PR #97, opened 2012-12-11
Implement SQS long polling

lsegal:
- Queue#poll now accept a :wait_time_seconds parameter which sets the
  maximum amount of time that SQS should long poll for.
  Defaults to 15 seconds. To use the default set for all messages
  on the queue, pass in nil.
- The :wait_time_seconds default does not apply to
  Queue#receive_message. For this method, you must pass the option
  directly, or set the global Queue#wait_time_seconds long polling
  value.
- Added Queue#wait_time_seconds attribute to set the queue-wide
  long poll timeout value. This setting applies to all calls to
  Queue#receive_message or Queue#poll
- Deprecates the Queue::DEFAULT_POLL_INTERVAL constant
- Queue#poll no longer uses :poll_interval; use :wait_time_seconds
  instead.

### DIFF
--- a/features/sqs/messages.feature
+++ b/features/sqs/messages.feature
@@ -64,6 +64,23 @@ Feature: SQS Messages
     And each message receipt handle should have been deleted
     And the forked process should have completed without errors
 
+  @long_polling @slow
+  Scenario: Long polling for SQS messages
+    Given I create a queue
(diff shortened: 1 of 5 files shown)


trevorrowe:
A replacement interface has been added to the version 2 AWS SDK for Ruby here: https://github.com/aws/aws-sdk-ruby/pull/740

PR #558, opened 2014-06-04
[s3] Do not prepend metadata keys if they start with x-amz

lukaszkorecki:
According to
http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUT.html

to enable redirect on an S3 object it needs to have
'x-amz-website-redirect' metadata key with new location value.

When creating an object like so:

``` ruby

bucket.object['foo'].write('', metadata: { "x-amz-website-redirect" =>
"http://example.com" }, acl: :public_read )
```

`x-amz-website-redirect` gets prepended with `x-amz-meta-` prefix thus
breaking the redirect feature.

This also affects any other meta data key which starts with 'x-amz'.

### DIFF
--- a/lib/aws/s3/client.rb
+++ b/lib/aws/s3/client.rb
@@ -139,7 +139,10 @@ module AWS
       def set_metadata request, options
         if metadata = options[:metadata]
           Array(metadata).each do |name, value|
-            request.headers["x-amz-meta-#{name}"] = value
+            # Do not add x-amz-meta- prefix to
+            # special metadata keys (like x-amz-webiste-redirect)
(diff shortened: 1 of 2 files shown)


maletor:
Does this conflict with how other headers should be stored?

At least according to [the Android SDK](https://aws.github.io/aws-sdk-android/docs/reference//com/amazonaws/services/s3/package-summary.html), it would appear that those encryption type meta tags _should_ include the `x-amz-meta-` prefix.

theincognitocoder:
The `a-amz-meta-` prefix is for user data that should be persisted with the object. All of the other x-amz- prefixed headers should be exposed as parameters to the request.

PR #598, opened 2014-08-08
Mention global stubbing in README

michaelbaudino:
It looks to me like something that should be in the README...


duane:
Is this no longer an option?

Doug-AWS:
We've documented stubbing in the developer guide at http://docs.aws.amazon.com/sdk-for-ruby/v2/developer-guide/stubbing.html.

PR #634, opened 2014-09-26
Add a configurable number of retries to the EC2 credential provider

### LINKED ISSUE #587: occasional AWS::Errors::MissingCredentialsError
We are using aws-sdk-ruby on ec2 (specifically writing to s3) in a script, it is running on about 100-200 hosts. This script is running multiple times every 5 minutes on those hosts and we are occasionally getting AWS::Errors::MissingCredentialsError. We rely on the EC2 credential provider. We initially thought that sometimes the failure was happening because the call to get the info was failing so we put a retry block with something like s3client.config.credential_provider.refresh but we are still getting the error. Any ideas?

menno:
In our production environment which uses IAM instance roles exclusively we have configured the EC2 provider as the only valid provider of credentials. However we experience sometimes intermittent timeouts on the metadata-service. In order to avoid the dreaded credential error, we are now able to retry the request for a number of times.

Should fix #587 as well.

### DIFF
--- a/lib/aws/core/credential_providers.rb
+++ b/lib/aws/core/credential_providers.rb
@@ -354,6 +354,8 @@ module AWS
         # @param [Hash] options
         # @option options [String] :ip_address ('169.254.169.254')
         # @option options [Integer] :port (80)
+        # @option options [Integer] :retries (0) Number of times to
+        #   retry retrieving credentials.
         # @option options [Float] :http_open_timeout (1)
         # @option options [Float] :http_read_timeout (1)
         # @option options [Object] :http_debug_output (nil) HTTP wire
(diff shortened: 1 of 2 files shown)


rb2k:
Just a question after the fact: Since the SDK initializes the object without any params itself over at: https://github.com/aws/aws-sdk-ruby/blob/v1.63.0/lib/aws/core/credential_providers.rb#L367

Would it be useful to just use AWS.configure for that?

trevorrowe:
@rb2k This seems reasonable. Open a new github issue and we can track that feature request.

rb2k:
Created one over here https://github.com/aws/aws-sdk-ruby/issues/717

Issue #654, opened 2014-10-29
Plus "+" character on S3Object key


aliismayilov:
It seems like there's some problem on `AWS::S3::S3Object#public_url`. `+` is being encoded. In fact the file location is:
* https://s3.amazonaws.com/bucket_name/files/document+copy.pdf
* NOT https://s3.amazonaws.com/bucket_name/files/document%2Bcopy.pdf

```ruby
[8] pry(main)> bucket = AWS::S3.new.buckets['bucket_name']
=> #<AWS::S3::Bucket:core_development>
[9] pry(main)> s3_object = bucket.objects['files/document+copy.pdf']
=> <AWS::S3::S3Object:bucket_name/files/document+copy.pdf>
[10] pry(main)> s3_object.public_url
=> #<URI::HTTPS:0x007feed60cbaf8 URL:https://s3.amazonaws.com/bucket_name/files/document%2Bcopy.pdf>
[11] pry(main)> s3_object.public_url.path
=> "/bucket_name/files/document%2Bcopy.pdf"
```

trevorrowe:
This is correct behavior. You are specifying the key of the object as "document+copy.pdf".  To create a public_url that is equivalent they key should be "document copy.pdf" (no plus sign).

```ruby
bucket = AWS::S3.new.buckets['bucket_name']
s3_object = bucket.objetcs['files/document copy.pdf']
s3_object.public_url
#=> #<URI::HTTPS:0x007fa0b88f2040 URL:https://s3.amazonaws.com/bucket_name/files/document%20copy.pdf>
```

Notice the `%20`, not `%2B`. The 2B is a URL encoded plus sign character, whereas 20 is a space character.

PR #655, opened 2014-10-29
UriEscape::escape specs

aliismayilov:
Issue is described here: https://github.com/aws/aws-sdk-ruby/issues/654

Currently `+` character in S3 urls is being encoded to `%2B`, but it should be treated as a whitespace `%2A`.
A file uploaded to S3 with key `/foo+bar.pdf`, is also available under `/foo%2Abar.pdf`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,8 @@
 /vendor
 /samples/config.yml
 *.gem
+/.ruby-*
+/.rvm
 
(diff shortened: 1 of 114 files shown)


trevorrowe:
Sorry for the slow response on this. The url escaping is correct. The issue is that spaces need to be provided as space characters and not "+" in the key.

PR #693, opened 2015-02-01
fixed bug with incorrect host when endpoint nil

rudkovsky:
Problem was when you set endpoint as nil there's no host. And so it leads to the output as this

```2.2.0 (main):0 > s3.buckets['transfer.celluloidheaven.net'].objects['test'].url_for(:read, :endpoint => nil, :secure => false)
=> #<URI::HTTP http:/transfer.celluloidheaven.net/test?AWSAccessKeyId=XXXX&Expires=XXXX&Signature=XXXX>```

Fixed now.


trevorrowe:
This has been merged into aws-sdk-v1.

PR #703, opened 2015-02-13
Added region to Aws::SharedCredentials

jimcroft:
Shared credential ini style files can contain a region setting which is used as a default by tools like the AWS CLI. I'd like the Aws::SharedCredentials object to provide me with that optional region setting for use in my Ruby code.

### DIFF
--- a/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
@@ -1,11 +1,15 @@
 module Aws
   class SharedCredentials < Credentials
 
+    # @return [String, nil]
+    attr_reader :region
+
     # @api private
     KEY_MAP = {
       'aws_access_key_id' => 'access_key_id',
       'aws_secret_access_key' => 'secret_access_key',
       'aws_session_token' => 'session_token',
+      'region' => 'region'
     }
 
     # Constructs a new SharedCredentials object. This will load AWS access
--- a/aws-sdk-core/spec/aws/shared_credentials_spec.rb
+++ b/aws-sdk-core/spec/aws/shared_credentials_spec.rb
@@ -24,6 +24,7 @@ module Aws
       expect(creds.access_key_id).to eq('ACCESS_KEY_0')
       expect(creds.secret_access_key).to eq('SECRET_KEY_0')
       expect(creds.session_token).to eq('TOKEN_0')
+      expect(creds.region).to eq('REGION_0')
     end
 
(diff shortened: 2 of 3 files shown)


jimcroft:
Picking the discussion up over here... if I amend this PR so that Aws::SharedCredentials cred and region attributes are set following the same logic as the AWS CLI is that something that would likely be merged in?

trevorrowe:
> I can confirm that region is honoured if set in ~/.aws/credentials. The docs do look a little muddled here as they also say "The AWS CLI will also read credentials from the config file". It looks like credentials and config files can be used interchangeably with only a slight difference in format (eg. [name] vs. [profile name]).
>
> Based on this I think this PR is incomplete as ~/.aws/config should be inspected for a default region too.

Interesting that the CLI loads a region from this file. I suspect this is a side-effect of sharing code for loading credentials & config as they used to be in the same file.

I understand that it would be handy for the Ruby SDK to be able to share more configuration with the CLI, especially region. I am concerned that loading region from a credentials configuration file is mixing concerns. Credentials are not scoped to a single region. This also introduces some other possible complexity if we decide to adopt the CLI's `~/.aws/config` file format for loading the region. Which file has precedence for setting the region? 

I'm going to continue the discussion over here with some of the other SDK and CLI engineers and I'll share what comes of this disucssion.

jimcroft:
Ok thanks @trevorrowe, will wait to see what feedback comes from your discussions. FYI where the CLI docs say "the keys in the credentials file will take precendence", I've just tested and it also seems to mean a region set in credentials takes precedence over a region set in the config file.

jimcroft:
Hi @trevorrowe any update on this?

trevorrowe:
I spoke with one of the CLI engineers. He said it is not intended behavior for the CLI to load a region from the shared credentials file, but a side-effect of implementation. He also indicated that it might be subject to change in the future. Given the purpose of this file was to segregate AWS secrets from other configuration we are not inclined to support loading a region from this file in the Ruby SDK.

That said, I do understand the desire to share region configuration between the CLI and Ruby SDK. It would be worth investigating loading the region from the `~/.aws/config` file directly.

trevorrowe:
I'm going to close this pull-request, as the current plan is to not support loading a region from the credentials file. If you'd like to discuss loading the region from the `~/.aws/config` file, hop into the [Gitter channel](gitter.im/aws/aws-sdk-ruby) and shoot me a message.

ColinHebert:
Is it possible to reconsider the option of loading the region from the aws cli configuration?

Issue #702, opened 2015-02-12
shared credentials ini file at HOME/.aws/credentials not used


stan3:
Per http://docs.aws.amazon.com/sdkforruby/api/frames.html I'm expecting to be able to create an ini file with aws credentials to be able to connect to s3.

> Default credentials are searched for in the following locations:
> 
>     * Aws.config[:credentials]
>     * ENV['AWS_ACCESS_KEY_ID'] and ENV['AWS_SECRET_ACCESS_KEY']
>     * The shared credentials ini file at HOME/.aws/credentials
>     * From an instance profile when running on EC2

With no details of what the ini file should look like I came across http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/SharedIniFileCredentials.html which documents a structure.  However this is not picked up by the client and I'm having to resort to environment variables to authenticate successfully.

trevorrowe:
The file needs to be at `~/.aws/credentials` for Linux and Mac. For Windows it should be at `%USERPROFILE%\.aws\credentials `. It should be formatted like:

```
[default]
aws_access_key_id =YOURACCESSKEY
aws_secret_access_key=YOURSECRET

[alternate-profile]
aws_access_key_id=YOURACCESSKEY
aws_secret_access_key=YOURSECRET
```

By default, the SDK will use the credentials in the "default" profile. You can change the profile like so:

```ruby
# default for all services
Aws.config[:profile] = 'profile-name'

# or a per client profile
s3 = Aws::S3::Client.new(profile:'other-profile-name')
```

I haven't heard any other issues reported about this feature. I use it myself regularly. Can you verify the paths and format and let me know if you are still having issues?

stan3:
Turns out my credentials file was in dos format.

trevorrowe:
I'm glad you were able to resolve the issue.

PR #704, opened 2015-02-13
Use a lazy star match so whitespace is trimmed

stan3:
Fix for #702.  Not sure how to run your tests so the test change is guess work.  Also I think would ideally include a doc change.


trevorrowe:
Thanks! Test worked well locally for me.

To run the tests, git clone the repo, bundle install, and then run `rake` or `rake:test`.

PR #706, opened 2015-02-15
Document correct permission for InvokeAsync

awendt:
I was giving a user permission for the action `lambda:invokeAsync` and it denied access with the following message:

    User: $user_arn is not authorized to perform: lambda:InvokeFunction on resource: $function_arn (Aws::Lambda::Errors::AccessDeniedException)

so the permission needed is actually `lambda:invokeFunction`.


trevorrowe:
Thanks for the fix. I'll merge this change and send the patch to the upstream source as well.

PR #716, opened 2015-02-24
Adding ELB waiter for healthy instance

omockler:
Adding a waiter that takes an ELB name and list of instance_ids and returns success when at least one of them reports back as "InService".

### DIFF
--- a/aws-sdk-core/apis/ElasticLoadBalancing.waiters.json
+++ b/aws-sdk-core/apis/ElasticLoadBalancing.waiters.json
@@ -1,0 +1,18 @@
+{
+  "version": 2,
+  "waiters": {
+    "InstanceHealthy": {
+      "delay": 15,
+      "operation": "DescribeInstanceHealth",
+      "maxAttempts": 40,
+      "acceptors": [
+        {
+          "expected": "InService",
+          "matcher": "pathAny",
+          "state": "success",
+          "argument": "InstanceStates[].State"
+        }
+      ]
+    } 
+  }
+}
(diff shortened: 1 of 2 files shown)


omockler:
I was just hoping to wait for one of them to be in the success state, should I change then name? I could also add one for checking if they are all healthy.

trevorrowe:
The name is fine. It should match "all" of the requested instances are healthy. You control what all means by omitting the instance if list or providing one or more odd. Does that make sense?

omockler:
Yeah, it makes sense, but it doesn't exactly cover the use case I had in mind.

I'm creating a number of new instances to replace the ones currently attached to an ELB. One instance can handle all of the load coming through by itself, but I am starting additional instances for high availability purposes. I was hoping to start the process of deregistering and terminating the old instances once any number of the new instances become available through the load balancer.

Do you think that this use case could warrant having a waiter for ```pathAll``` and ```pathAny```?

trevorrowe:
Maybe my confusion is the intended use of the waiter. Do you intend to wait until any of the instances are "InService" and use this as a condition to indicate the health of the load balancer? If so, maybe its just naming confusion.

My suggestion:

1. Rename the first waiter from "InstanceHealthly" to "InstanceInService". This makes it very clear what state is being polled for.
2. Rename the second waiter to "AnyInstanceInService".

Example:

```ruby
# poll until EVERY instance is in service
elb.wait_until(:instance_in_service)

# polls until both instances are in service
elb.wait_until(:instance_in_service, instances: [{ instance_id: 'id' }, { instance_id: 'id2'}])

# poll until ANY instance is in service
elb.wait_until(:any_instance_in_service)

# polls until any of the two instances are in service
elb.wait_until(:any_instance_in_service, instances: [{ instance_id: 'id' }, { instance_id: 'id2'}])
```

Does that cover your use cases?

omockler:
Exactly what I was looking for, thanks for the naming suggestion!

trevorrowe:
Thanks for the contribution!

PR #723, opened 2015-03-01
Added creation_date attribute to image

guss77:
This attribute is important when looking at a list of image updates and selecting the latest (or earliest) one.

### DIFF
--- a/lib/aws/ec2/image.rb
+++ b/lib/aws/ec2/image.rb
@@ -86,6 +86,8 @@ module AWS
     # @attr_reader [Array<String>] product_codes Returns an array of
     #   product codes attached to this instance.
     #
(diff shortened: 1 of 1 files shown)


trevorrowe:
Thanks!

guss77:
Thanks for the merge. When do you plan to release an update to rubygems.org?

trevorrowe:
Since version 2 has been released, we've been averaging about once a month of the v1 SDK releases. This is not a strict rule, just an observation. Releases happen generally when there is a bug or security fix, or an important API update for a service supported in v1. The last release was about 2 weeks ago.

guss77:
OK, thanks

guss77:
Hi, any chance of shooting a gem release with this API change? 
I'm really dependany on this for my workflow and I would love to drop my private repo - it makes a mess with bundle installs.

Thanks.

Issue #712, opened 2015-02-21
CloudWatch logs get_log_events pager seems to loop indefinitley


Tim-B:
I'm having some trouble using get_log_events with the .each do |response| pager functionality.

Basically it seems to loop indefinitely.

Here's a basic example:

```
require 'aws-sdk'

cwl = Aws::CloudWatchLogs::Client.new(region: 'us-east-1')

resp = cwl.get_log_events(
    log_group_name: 'CloudTrail/DefaultLogGroup',
    log_stream_name: '284243076752_CloudTrail_us-east-1',
    start_time: (Time.now.to_i - 3 * 60 * 60 * 24) * 1000,
    end_time: (Time.now.to_i - 1 * 60 * 60 * 24) * 1000,
)

resp.each do |event_page|
  puts "Page returned with #{event_page[:events].length} items."
  puts "back #{event_page[:next_backward_token]} forward #{event_page[:next_forward_token]}"
  puts "Last #{event_page.last_page?}"
  puts "Next #{event_page.next_page?}"
end
```

If you replace log_stream_name with your own log stream name this should print all events from 1 day ago to 3 days ago.

For me that returns:
```
Page returned with 1013 items.
back b/31761921002636013633299509555220269662210192519471497216 forward f/31765609545891850598367319401354689355894894871384293376
Last false
Next true
Page returned with 0 items.
back b/31765609545891850598367319401354689355894894871384293376 forward f/31765609545891850598367319401354689355894894871384293376
Last false
Next true
Page returned with 0 items.
back b/31765609545891850598367319401354689355894894871384293376 forward f/31765609545891850598367319401354689355894894871384293376
Last false
Next true
Page returned with 0 items.
back b/31765609545891850598367319401354689355894894871384293376 forward f/31765609545891850598367319401354689355894894871384293376
Last false
Next true
Page returned with 0 items.
back b/31765609545891850598367319401354689355894894871384293376 forward f/31765609545891850598367319401354689355894894871384293376
Last false
Next true
Page returned with 0 items.
back b/31765609545891850598367319401354689355894894871384293376 forward f/31765609545891850598367319401354689355894894871384293376
Last false
Next true
Page returned with 0 items.
back b/31765609545891850598367319401354689355894894871384293376 forward f/31765609545891850598367319401354689355894894871384293376
Last false
Next true
Page returned with 0 items.
[...]

PR #730, opened 2015-03-06
Fix for infinite paging loop on CloudWatchLogs::Client#get_log_events.

trevorrowe:
See related issue #712


trevorrowe:
@awood45 Could you give this a look over?

Issue #738, opened 2015-03-14
Allow custom waiters


danielsiwiec:
It would be nice to have an option to wait for an arbitrary attribute value, rather than be limited to a fixed list of waiters. For instance:

```ruby
operation = Proc.new { |instance| instance.public_dns_name.nil? }
instance.wait_until(&operation)
```

trevorrowe:
The implementation for this method would be fairly straightforward.  Internally it would yield a loaded object, refreshing self until the block returned a true value. The biggest question is how to ensure it has a consistent interface for max attempts, delay, etc to prevent polling indefinitely.

danielsiwiec:
Hmm, that's a very good question. For my need I hand-rolled the mechanism, but it feels like I'm duplicating logic, that's already there:

```ruby
class Waiter
  def initialize(options)
    @options = options
  end

  def wait
    attempts = 0
    while @options[:poller].call && attempts < @options[:max_attempts]
      sleep(@options[:delay])
      attempts += 1
    end
  end
end

operation = Proc.new {
  instance = ec2.instance(instance.instance_id)
  instance.public_dns_name.empty?
}

waiter = Waiter.new(
  poller: operation,
  max_attempts: 20,
  delay: 1,
)

waiter.wait
```

My naive thinking makes me hope that if we opened up the Waiter and Poller classes and let people create them with a custom ```operation``` and pass it in, it would *magically* work.
Someone with a deeper understanding of the ```Poller``` would have to testify to that though.

danielsiwiec:
Trevor, I took a stab at implementing a custom waiter. I also tried to follow your testing convention. Please look at the commit above and let me know what you think.

Pros:
- reused the original waiting logic
- the interface is very easy to use ```instance.wait_until(&condition)```

Caveats:
- implemented only in EC2::Instance class (I'd need some guidance with how to implement the function across all resources, if needed)
- hardcoded waiter parameters

Let me know what you think about the approach. I haven't written any docs yet, but if this gets a green light, I'll definitely add some.

trevorrowe:
I like the general approach. I took your idea and refined it a bit. I was interested in a generic solution that would work for all resource types. I pushed a proof-of-concept branch. It lacks automated tests, but it passes all of the smoke tests I ran it through manually.

https://github.com/aws/aws-sdk-ruby/commit/f9e33b4c8b0a3e26e95c61bb84ae046bd9f47355

Some thoughts:

[...]

PR #744, opened 2015-03-20
Add `Resource#wait_until { |resource| condition }` custom waiters.

danielsiwiec:
The wait_until method doesn't alter the Resource's internal state. It operates on a copy. Added docs and specs. This addresses issue #738

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 Unreleased Changes
 ------------------
 
+* Feature - Aws::Resources::Resource - Added support for custom waiters
+  with `Aws::Resources::Resource#wait_until`.
+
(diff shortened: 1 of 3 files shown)


trevorrowe:
Looks good! I added one comment above about the documented namespace for the waiter errors. Feel free also to add an entry into the `CHANGELOG.md` at the repository root. Its important to keep the change log updated with new features. The chage log is consumed to produce release notes. Feel free also to pass on that, and I'll write the change log entry.

danielsiwiec:
Doing it right now.

trevorrowe:
Thanks so much for driving this! This should go out with the next release of the SDK.

danielsiwiec:
Thank you. My pleasure to contribute!

PR #747, opened 2015-03-24
Main README file should expose correct exmples

rlishtaba:
Please, review and update examples in the main README file


trevorrowe:
I'm inclined to not merge this pull request. The examples are correct as it. They are loading credentials from one of the default locations documented previously in the readme. It is probably worth reviewing the readme to ensure this is more clear.

PR #755, opened 2015-03-30
Update error_handler.rb to handle nil error_code

ajcantu:
We're occasionally seeing errors on this line:

```
NoMethodError: undefined method `sub' for nil:NilClass
File "/opt/rubies/ruby-2.1.5/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.0.33/lib/aws-sdk-core/xml/error_handler.rb" line 55 in remove_prefix
File "/opt/rubies/ruby-2.1.5/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.0.33/lib/aws-sdk-core/xml/error_handler.rb" line 50 in extract_error
File "/opt/rubies/ruby-2.1.5/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.0.33/lib/aws-sdk-core/xml/error_handler.rb" line 21 in error
File "/opt/rubies/ruby-2.1.5/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.0.33/lib/aws-sdk-core/xml/error_handler.rb" line 9 in block in call
File "/opt/rubies/ruby-2.1.5/lib/ruby/gems/2.1.0/gems/aws-sdk-core-2.0.33/lib/seahorse/client/response.rb" line 43 in call
```

### DIFF
--- a/aws-sdk-core/lib/aws-sdk-core/xml/error_handler.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/xml/error_handler.rb
@@ -51,7 +51,7 @@ module Aws
       end
 
       def remove_prefix(context, error_code)
-        if prefix = context.config.api.metadata('errorPrefix')
+        if error_code && prefix = context.config.api.metadata('errorPrefix')
           error_code.sub(/^#{prefix}/, '')
         else
           error_code


trevorrowe:
Have you tested this in your environment to see if it solves the problem? It looks like this will just move the error up to where it attempts to construct an error class from a nil error_code and raise there instead.

I'm very interested in the situation that can cause the response to have a body, but the SDK is unable to extract an error code. Would you be willing to run a patched version of the SDK that would log the http response body in one of these situations so we can see what is the root cause?

If you are, let me know, and I'll put together some code that would collect the needed information.

ajcantu:
Ah, yes. It looks like it will just error later at:
https://github.com/aws/aws-sdk-ruby/blob/master/aws-sdk-core/lib/aws-sdk-core/errors.rb#L100

Ok, we'll try to get more information on why this is failing for us. It's only happened once in the last 4 days, so I'm not sure when we'll see it again, but we'll create a new pull request at that time.

Thanks for taking a look!

trevorrowe:
I went ahead an made a small change that will at a minimum prevent errors while attempting to parse the error code and messages and will fall-back on a friendly http status code based error. This should allow us to identify when this happens.

ajcantu:
Sweet. Thanks, @trevorrowe!

Issue #696, opened 2015-02-06
Missing "exists" method?


gshutler:
Against the v1 gem I had some code like this:

```ruby
client = AWS::S3::Client.new
bucket = client.buckets[name]
client.buckets.create(name) unless bucket.exists?
# bucket now definitely exists
```

However, the `#exists?` method isn't there any more and I can't see an equivalent.

What's the recommended method of implementing the same behaviour? The alternatives I've been able to spot don't seem like they would be recommended:

 * Calling `Aws::S3::Client#head_bucket`
    * Raises an error if it does not exist, errors as flow control makes me think this isn't recommended
 * Calling `Aws::S3::Resource#create_bucket`
    * Raises an error if it does exist, errors as flow control makes me think this isn't recommended
 * Calling `Aws::S3::Resource#buckets`
    * Seems wasteful to get the whole list when I only care about one

Similarly there's not an `#exists?` on `Aws::S3::Object` so guidance in that case would likely be useful too (though it doesn't affect me).

protochron:
I'd also be interested in knowing the best way to handle this case. It's especially useful when looking at CloudFormation stacks since it means you don't have to iterate over the entire stack collection.

trevorrowe:
Adding `#exists?` methods is on a short list of features I'd like to address with resources shortly.

For the v1 SDK, the exists methods were all hand-coded and used different logic for each scenario. For example, with Amazon S3 buckets, the exists method would make a `#get_bucket_versioning` request. With a successful response, the bucket exists. If a NoSuchBucket error was raised (404), false was returned. If an AccessDenied error was raised, then true was returned (the bucket exists, you just can't get information about it).

For v2, I'm inclined to solve this using waiter methods. By adding a waiter for "BucketExists" and CloudFormation "StackExists" we can share the logic. The exists method would simply poll the waiter exactly once, and not keep waiting.

philostler:
I was very surprised when I went looking for the `#exists?` method on `Aws::S3::Object` and couldn't find it. I think this is a pretty major omission tbh. I'd be interested when there is a nice solution available... atm i'm rescuing `Aws::S3::Errors::NoSuchKey` :disappointed:

PR #757, opened 2015-03-30
Added `#exists?` methods to a handful of resource classes.

trevorrowe:
The `#exists?` method works by polling the exists wait for the resource exactly once. If the waiter is successful, then `true` is returned, if it fails, then `false` is returned.

To add additional `#exists?` methods, a waiter must be added to the resource class as `Exists` and that waiter must be defined in the *.waiters.json document for that service.

See #696

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,15 @@
 Unreleased Changes
 ------------------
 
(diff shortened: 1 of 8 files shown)


trevorrowe:
This has been merged into master.

PR #769, opened 2015-04-07
Empty set. How to be?

antohaby:
Hi everyone!

Is it possible to cast empty set as empty list? Or shall I handle empty sets in my client code? 
I propose to cast the empty sets as empty lists because this is a good way to follow simple rule: "What You See Is What You Get"
We see the empty set so this is the same as the empty list, isn't it?

Referring to php-sdk, php has no set class but only array and all operations with sets performed over arrays. So under php context empty sets and empty lists are meaning same things.

Thanks,
Anton

### DIFF
--- a/aws-sdk-core/lib/aws-sdk-core/dynamodb/attribute_value.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/dynamodb/attribute_value.rb
@@ -48,6 +48,7 @@ module Aws
         private
 
         def format_set(set)
+          return { l: [] } if set.empty?
           case set.first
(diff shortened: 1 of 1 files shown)


trevorrowe:
The DynamoDB API makes a distinction between lists and sets. DynamoDB does not accept empty sets. By accepting from the user an empty set and converting it to a list type we've changed the nature of the data structure. This prevents users from round-tripping an item (e.g. GET, modify, PUT) multiple times without possibly introducing bugs.

I'm open to discussion on how best to handle an empty set. Another possible option is to drop the set.

PR #792, opened 2015-04-23
Replace multi_json with standard library json

sferik:
As on of the authors of `multi_json`, I no longer use it in my own projects and no longer recommend its use to others, now that the `json` gem is part of the Ruby standard library. I have replaced the `multi_json` dependency with a `json` gem dependency, despite the fact that this is not strictly necessary, just to ensure that users of older versions of Ruby, which ship with an older version of `json` are using the latest gem version instead. This dependency could be removed.

I have opened a [similar pull request against `jmespath`](https://github.com/trevorrowe/jmespath.rb/pull/7) to remove the transitive dependency on `multi_json`.

I have also added Ruby 2.2 to the Travis CI matrix for testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ rvm:
   - 1.9.3
   - 2.0.0
   - 2.1
+  - 2.2
   - jruby
 
 sudo: false
--- a/aws-sdk-core/aws-sdk-core.gemspec
+++ b/aws-sdk-core/aws-sdk-core.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |spec|
   spec.bindir = 'bin'
   spec.executables << 'aws.rb'
 
-  spec.add_dependency('multi_json', '~> 1.0')
+  spec.add_dependency('json', '~> 1.8')
   spec.add_dependency('builder', '~> 3.0')
   spec.add_dependency('jmespath', '~> 1.0')
 
--- a/aws-sdk-core/features/env.rb
+++ b/aws-sdk-core/features/env.rb
@@ -1,15 +1,15 @@
 $LOAD_PATH << File.join(File.dirname(__FILE__), '..', 'lib')
 
-require 'simplecov'
 require 'aws-sdk-core'
-require 'multi_json'
+require 'json'
+require 'simplecov'
 
 SimpleCov.command_name('test:integration:aws-sdk-core')
 
 cfg = './integration-test-config.json'
 
 if File.exist?(cfg)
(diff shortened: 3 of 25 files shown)


sferik:
@coveralls This is not helpful at all.

lsegal:
@sferik unhelpful bot is unhelpful :(

sferik:
@lsegal I was kind of hoping you wouldn’t delete all those spam messages so I could report them for abuse. Now that [GitHub supports multiple pull request status](https://github.com/blog/1935-see-results-from-all-pull-request-status-checks) checks, @coveralls bot should :skull:.

lsegal:
@sferik I didn't delete anything, maybe @trevorrowe or @awood45 did?

sferik:
I threaded to report them, so maybe @coveralls did.

trevorrowe:
@sferik @lsegal I deleted them as they weren't adding any value.

trevorrowe:
Continuing the discussion from: https://github.com/trevorrowe/jmespath.rb/issues/6#issuecomment-95875818

> > Given the small surface area required, I would like to explore removing the MultiJson dependency, while preserving the faster parsing options until at such time as Oj becomes standard.

> If you’re able to achieve this, I’d encourage you to contribute this code to MultiJson.

What I am considering is the following:

* Add a pair of non-public utility method to the SDK. Name this method very un-originally something like `Aws.load_json` and `Aws.dump_json`. 
* Update the SDK to use these when marshaling and unmarhsaling JSON payloads and when loading the large API definitions.

A simple implementation of this module could be:

```ruby
module Aws
  # @api private
  module Util
    class << self

      def load_json(json)
        ENGINE.load(json)
      end

      def dump_json(obj)
        ENGINE.dump(obj)
      end

      def oj_engine
        require 'oj'
        Oj
      rescue LoadError
        nil
      end

      def json_engine
        require 'json'
        JSON
      end

    end

    ENGINE = oj_engine || json_engine

  end
end
```

This is simplified version of what I do in place of using multi_xml: https://github.com/aws/aws-sdk-ruby/blob/master/aws-sdk-core/lib/aws-sdk-core/xml/parser.rb#L74.

trevorrowe:
@sferik I've addressed this in a branch that is being worked for `v2.1.0` of the SDK. You can see my relevant changes here: https://github.com/aws/aws-sdk-ruby/blob/version-2.1-refactor/aws-sdk-core/lib/aws-sdk-core/json.rb#L13-L56

This provides the shim we need to allow users to leverage Oj for performance while maintaining only the minimum interfaces.

PR #799, opened 2015-04-29
Add JSON::ParserError as a retry-able failure type.

pdrakeweb:


trevorrowe:
@pdrakeweb Thanks for the contribution. I'd like to understand a bit more about the situations that cause this error. I'm currently investigating an issue with Net::HTTP which I suspect may be the root cause of these errors. Do you have any more information about when these happen and perhaps a stack trace or any logs?

pdrakeweb:
Here is a partial stack trace (the remainder of the stack trace is inside custom code so it wouldn't help to include): https://gist.github.com/pdrakeweb/f548c5b7cb4a83409a74

rb2k:
Yeah, seen this happening too.
Would appreciate a merge :-/

trevorrowe:
I spent a non-trivial amount of time attempting to reproduce this issue and I can across an issue in how Net::HTTP handles networking errors and automatic retries. The commit above should resolve this issue.

glennpratt:
Interesting. Are you planning to create an upstream (Ruby) bug?

trevorrowe:
Yes. I have a write-up with suggestions, possible workarounds. I'm currently going through approval internally so that I can contribute these. There are related issues that affect the payload of a PUT request, stemming from the same issue. Lastly, there is a long-standing Expect 100-continue bug that I will be submitting a patch for. Three Net::HTTP issues in total.

glennpratt:
Nice, thanks for sharing!

tdg5:
Hey @trevorrowe, did you ever create an issue with Ruby Core? I couldn't find one so [I created this one](https://bugs.ruby-lang.org/issues/11526). Let me know if you created one previously so I can remove the dup. Otherwise, feel free to contribute if there's anything you want to add or disagree with.

I also made a GitHub PR here: https://github.com/ruby/ruby/pull/1019

For now, we've also gone the way of clearing `IDEMPOTENT_METHODS_`, but this seems like overkill so it'd be nice to get a fix going upstream.

trevorrowe:
@tdg5 If I recall correctly, the documentation on submitting issues indicated that I could do so via the issue tracker or via GitHub. I opted to go the GitHub route so that I could submit with suggested code-changes.  It did not seem to have received any attention and then it fell off my radar.

I'll go ahead an ping those issues and if I don't get any response I'll open issues on the Ruby issue tracker.

tdg5:
Oh duh, yeah I didn't even think to check GitHub. I found https://github.com/ruby/ruby/pull/951, thanks!

PR #820, opened 2015-05-19
Simple type fixes and method change

alkesh26:
1. aws.rb
    Instead of !array.include?() we can use exclude more readable. Benchmarked these methods. The result is same almost.
2. customizer.rb
    Since shape_name is not used we can use each_value.
3. parse.rb
    This can be initialised to empty array and hash.

    Rest all files are simple typo fixes.

### DIFF
--- a/aws-sdk-core/bin/aws.rb
+++ b/aws-sdk-core/bin/aws.rb
@@ -9,7 +9,7 @@ require 'logger'
 
 def env_bool key, default
   if ENV.key?(key)
-    !['0', 'false', 'no', 'off'].include?(ENV[key].downcase)
+    ['0', 'false', 'no', 'off'].exclude?(ENV[key].downcase)
   else
     default
   end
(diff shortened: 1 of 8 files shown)


trevorrowe:
I responded with a pull request to your fork here: https://github.com/alkesh26/aws-sdk-ruby/pull/1 - This addresses the parser errors. The root cause is that a `nil` value is being passed into the `#list` and `#map` methods preventing the defaults to setup from being used.  Instead, nil values are used which don't respond to `#[]=` or `#<<`.

If you merge those changes, I'll be happy to these these.

alkesh26:
@trevorrowe I am very thankful for your response. Can you please help me more in understanding the error. May be with an example or some link. I am unable to identify the root cause. 

It will be of great help. And thanks a lot once again. Your comment made my day. :)

trevorrowe:
Sure thing. When defining a Ruby method with a default value, that value is only used when the agument is not given.  For example:

```ruby
def example(value = "default-value")
  value
end

example
#=> "example-value"

example(nil)
#=> nil
```

The JSON parser was always passing the target parameter. This prevented the default values of `[]` and `{}` from ever being used. The patch I sent to your branch modified the parser to not send `nil` values. The fix above also resolves this.

trevorrowe:
Thanks for the contribution!

Issue #831, opened 2015-05-29
DynamoDB: Unable to put binary data


der-flo:
When trying to put binary data to DynamoDB, I get an exception regarding the usage of an IO (eg. `StringIO`) type. I'm using the actual gem (2.0.47).

Simple example to reproduce:
``` ruby
require 'aws-sdk'
client = Aws::DynamoDB::Client.new(region: 'eu-west-1')

item = {
  'id' => 'bar',
  'foo' => StringIO.new('baz')
}
client.put_item(table_name: 'foobar', item: item)
```

Exception and backtrace:
```
[...]/aws-sdk-core-2.0.47/lib/aws-sdk-core/json/builder.rb:54:in `format': undefined method `read' for "baz":String (NoMethodError)
from [...]/aws-sdk-core-2.0.47/lib/aws-sdk-core/json/builder.rb:24:in `block in structure'
from [...]/aws-sdk-core-2.0.47/lib/aws-sdk-core/json/builder.rb:20:in `each'
from [...]/aws-sdk-core-2.0.47/lib/aws-sdk-core/json/builder.rb:20:in `with_object'
from [...]/aws-sdk-core-2.0.47/lib/aws-sdk-core/json/builder.rb:20:in `structure'
from [...]/aws-sdk-core-2.0.47/lib/aws-sdk-core/json/builder.rb:50:in `format'
from [...]/aws-sdk-core-2.0.47/lib/aws-sdk-core/json/builder.rb:41:in `block in map'
from [...]/aws-sdk-core-2.0.47/lib/aws-sdk-core/json/builder.rb:40:in `each'
from [...]/aws-sdk-core-2.0.47/lib/aws-sdk-core/json/builder.rb:40:in `with_object'
from [...]/aws-sdk-core-2.0.47/lib/aws-sdk-core/json/builder.rb:40:in `map'
from [...]/aws-sdk-core-2.0.47/lib/aws-sdk-core/json/builder.rb:52:in `format'
from [...]/aws-sdk-core-2.0.47/lib/aws-sdk-core/json/builder.rb:24:in `block in structure'
from [...]/aws-sdk-core-2.0.47/lib/aws-sdk-core/json/builder.rb:20:in `each'
from [...]/aws-sdk-core-2.0.47/lib/aws-sdk-core/json/builder.rb:20:in `with_object'
from [...]/aws-sdk-core-2.0.47/lib/aws-sdk-core/json/builder.rb:20:in `structure'
[...]

PR #832, opened 2015-06-02
Don't Read IO Objects in Attribute Value Marshal

awood45:
Resolves Issue #831

Running the following code exercised the problem, and now works with
this change:

```
dynamodb.put_item(
  table_name: "foo",
  item: {
    id: 1,
    contents: StringIO.new("bar")
  }
)
```

The issue appeared to be that #read was called twice on the StringIO or
IO object, and the second call would fail as String#read does not
exist.

One thing I'd like to review further before merging is the fact that maps and collections do this same #read call but do not have this problem when tested.

### DIFF
--- a/aws-sdk-core/lib/aws-sdk-core/dynamodb/attribute_value.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/dynamodb/attribute_value.rb
@@ -34,7 +34,7 @@ module Aws
             end
           when String then { s: obj }
           when Numeric then { n: obj.to_s }
-          when StringIO, IO then { b: obj.read }
+          when StringIO, IO then { b: obj }
           when Set then format_set(obj)
           when true, false then { bool: obj }
           when nil then { null: true }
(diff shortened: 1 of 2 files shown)


trevorrowe:
The change looks good. Maps and lists definitely are effected, but I suspect what you meant to say is that they are not tested?

```ruby
dynamodb.put_item(
  table_name: "foo",
  item: {
    id: 1,
    contents: { # map
      data: StringIO.new("bar")
    }
  }
)
#=> raises NoMethodError: undefined method `read' for "bar":String
```

```ruby
dynamodb.put_item(
  table_name: "foo",
  item: {
    id: 1,
    contents: [ # list
      StringIO.new("bar"),
    ]
  }
)
#=> raises NoMethodError: undefined method `read' for "bar":String
```

awood45:
Per our earlier discussion, Hashes and Arrays worked fine from the first change, the second use of #read was isolated to Set types only. This change fixes the appropriate test and code.

PR #846, opened 2015-06-18
Fix presigned urls for copying encrypted snapshots

### LINKED ISSUE #845: Aws::Plugins::EC2CopyEncryptedSnapshot#presigned_url has incorrect argument list
Trying to copy an EBS snapshop from one region to another requires a presigned url, but the plugin that automates this uses an incorrect argument list:

`Signers::V4.new('ec2', config.credentials, config.region)`

should instead be:

`Signers::V4.new(config.credentials, 'ec2', config.region)`

In addition to this, I discovered that the inclusion of the Content-Type header in the signed request was causing the presigned url to be invalid.

david-mccullars:
Fixes #845


trevorrowe:
Thanks for the patch. I'm going to work on adding an integration test to ensure this does not regress.

PR #855, opened 2015-06-26
Fix Resource#wait_until(max_attempts: nil)

ktheory:
Happy Friday! :wave: 

I noticed that passing `max_attempts: nil` to a resource didn't work like [these docs](https://github.com/aws/aws-sdk-ruby/blob/17508581a6887fb170e608242165b1ee18ccd9a9/aws-sdk-resources/lib/aws-sdk-resources/resource.rb#L65-L66) said. It only tried 10 times before raising `Aws::Waiters::Errors::TooManyAttemptsError`. :bug: :rotating_light: 

So I changed the code match the docs. :tada: :relieved: 

### Is there even a test???

You betcha. Of course it's weird to test that we actually attempt something indefinitely. So I just test that `wait_until(max_attempts: nil)` attempts more than 10 times.

### Could there be another way???

Instead of changing the code, we could change the docs to recommend passing `max_attempts: -1` or something instead. That can be a workaround until this is released. IMHO, using `nil` instead of `-1` to indicate "unlimited" seems more conventional for Rubyists.

### DIFF
--- a/aws-sdk-resources/lib/aws-sdk-resources/resource.rb
+++ b/aws-sdk-resources/lib/aws-sdk-resources/resource.rb
@@ -101,7 +101,7 @@ module Aws
       def wait_until(options = {}, &block)
         resource_copy = self.dup
         attempts = 0
-        options[:max_attempts] ||= 10
+        options[:max_attempts] = 10 unless options.key?(:max_attempts)
         options[:delay] ||= 10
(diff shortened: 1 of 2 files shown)


awood45:
Looks good to me, nice catch! 👍

I would be inclined to agree that this is the best way rather than special casing `-1`, but I'm open to changing my mind.

PR #861, opened 2015-07-03
Ensure S3 presigned default expires time is not changing

j15e:
It occurred to us that signatures end up invalid randomly because the default signature time is computed twice and we can end up with a policy that was signed for a policy with an expiration time 1 second earlier.

To be specific, the policy is computed twice inside the `fields` method which uses the `formation_expiration` twice too, which in turn computes `Time.now` at two different times.

https://github.com/didacte/aws-sdk-ruby/blob/bugfix/s3-presigned-default-expires/lib/aws/s3/presigned_post.rb#L345-L351

@see https://github.com/marcel/aws-s3/pull/54 (similar issue)


trevorrowe:
Good catch! Thanks for the patch.

PR #862, opened 2015-07-04
Support unsigned_request for cognito-identity

DianthuDia:


trevorrowe:
I'm not very familiar with the Cognito Identity API, but I looked at the API reference documentation for a few of these operations and their documented examples use an authorization header. Are these strictly unsigned operations, or could/should they be conditionally signed?

DianthuDia:
Yes, documented examples use an authorization header. But please check beginning of the description.

http://docs.aws.amazon.com/cognitoidentity/latest/APIReference/API_GetId.html

> This is a public API. You do not need any credentials to call this API.

trevorrowe:
This should go out with the next release shortly.

DianthuDia:
Thank you for merging and adding test :yellow_heart:

PR #866, opened 2015-07-07
Create two additional waiters AppAvailable and InstanceProfileAvailable

djcp:
This creates 2 additional waiters:

- OpsWorks AppAvailable, to ensure an app is available before taking subsequent actions
- IAM InstanceProfileAvailable, to let you wait for an instance profile to be available

### DIFF
--- a/aws-sdk-core/apis/iam/2010-05-08/waiters-2.json
+++ b/aws-sdk-core/apis/iam/2010-05-08/waiters-2.json
@@ -1,6 +1,23 @@
 {
   "version": 2,
   "waiters": {
+    "InstanceProfileAvailable": {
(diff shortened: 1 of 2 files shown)


trevorrowe:
I pulled these changes locally, and made a few minor changes. Most notably, I renamed the waiters to `InstanceProfileExists` and `AppExists` to be consistent with other waiter names. I also added a resource exist waiter for `Aws::IAM::InstanceProfile#wait_until_exists` and `Aws::IAM::InstanceProfile#exists?`. The latter is automatically created when there is an exists waiter.

Thanks for the contribution! This should be released shortly.

djcp:
Thanks!

PR #871, opened 2015-07-13
Shared Examples

DanielNi:
Adds functionality and tests for generating shared documentation examples from the language-agnostic JSON format. These new examples will live alongside the JSON apis.

The code in this pull request hooks into the #example_tags method by adding a new #shared_examples method that returns an array of tags holding the YARD documentation examples. The SharedExample class simultaneously walks the JSON and Seahorse model to generate the example code.

RSpec tests pull JSON examples from disk (scraped from the AWS CLI examples) and use ParamValidators to check the correctness of the generated examples.


trevorrowe:
I haven't had a chance to fully review this yet.  One quick comment though, please feel free to remove the c2j prefixes. As an internal tool, this isn't meaningful in the public SDK. I would generally favor something more descriptive such as "SharedExample" instead of "C2jExample".

PR #889, opened 2015-08-05
Fix method name at queue_poller example

aleandros:
Small correction for SQS [QueuePoller](http://docs.aws.amazon.com/sdkforruby/api/Aws/SQS/QueuePoller.html) documentation, at the example titled **Throw :stop_polling**.

It calls `stats.receive_message_count` instead of `stats.received_message_count`.


trevorrowe:
Thanks!

PR #902, opened 2015-08-19
Update get_object examples

marcgarreau:
#### What does it do?
Updates example code for `get_object` API.
#### Any background context you want to provide?
When looking for how to save off a download to a specific file path, I searched the repo via Github for `get_object`. The examples provided a `response_target`, but when that didn't fly, I went googling and found Trevor's helpful post [here](https://ruby.awsblog.com/post/Tx354Y6VTZ421PJ/Downloading-Objects-from-Amazon-S3-using-the-AWS-SDK-for-Ruby), that points you to `target` instead. Then, magical success. Updating so future folks can skip a step.

### DIFF
--- a/doc-src/examples/s3/client/get_object/01_download_an_object_to_disk.rb
+++ b/doc-src/examples/s3/client/get_object/01_download_an_object_to_disk.rb
@@ -1,6 +1,6 @@
 # stream object directly to disk
 resp = s3.get_object(
-  response_target: '/path/to/file',
+  target: '/path/to/file',
   bucket: 'bucket-name',
   key: 'object-key')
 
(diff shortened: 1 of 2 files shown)


trevorrowe:
I'm curious what issue you had with `:response_target`. The following should work just fine:

```ruby
s3.get_object(bucket:'aws-sdk', key:'foo', response_target: '/path/to/file')
```

The following also works, but please note, the target option must be passed in a different hash. The first is the request parameters hash, the second is the request options hash:

```ruby
s3.get_object({ bucket:'aws-sdk', key:'foo' }, target: '/path/to/file')
```

As far as I can tell, the docs should be correct. Can you share the error you are having with `:response_target`?

trevorrowe:
Thanks!

trevorrowe:
Oops. I mean to merge a different pull request. Sorry, I'm going to revert this shortly.

PR #911, opened 2015-08-24
Fix Aws::PageableResponse to correctly answer `respond_to?` with methods different than #count.

fertapric:
`Aws::PageableResponse::UnsafeEnumerableMethods` is not forwarding `respond_to?` to its ancestors (using `super`) and it's always retuning `false` (except for #count). It breaks `Searhorse::Client::Response#respond_to?`:

```ruby
response # Aws::DynamoDB::Types::QueryOutput
response.respond_to?(:items) # => false
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 Unreleased Changes
 ------------------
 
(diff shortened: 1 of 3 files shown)


trevorrowe:
Thank you for the well formed pull-request!

trevorrowe:
Also, this should go out in tomorrows release.

PR #925, opened 2015-09-13
Fix storage class setting on PresignedPost

julik:
Previously storage_class would be passed as a vanilla form field
which would be ignored by S3. The field should be passed as an
"x-amz-storage-class" which requires a minimal change in the
class descriptor. Once the field is set, the storage class is
properly set on the POSTed object.


trevorrowe:
Thanks for the fix. This should go with the next release, possibly today.

julik:
:heart: my pleasure

PR #927, opened 2015-09-17
Don't enforce 1 week limit on presigned urls.

dziemid:
We have run into this restriction while updating from v1 version of the gem.

### DIFF
--- a/aws-sdk-core/lib/aws-sdk-core/s3/presigner.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/s3/presigner.rb
@@ -11,9 +11,6 @@ module Aws
     class Presigner
 
       # @api private
-      ONE_WEEK = 60 * 60 * 24 * 7
-
-      # @api private
       FIFTEEN_MINUTES = 60 * 15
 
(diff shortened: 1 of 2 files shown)


trevorrowe:
I'm afraid this is not technically possible. Removing the 1-week limit expiration will only permit the SDK to return invalid pre-signed URLs. Signatures extending beyond one week will not be honored by Amazon S3. This is a limitation of signature version 4 pre-signed URLs.

Version 2 of the SDK intentionally moved from v2 signatures to v4 signatures for security reasons. Additionally, v2 signatures simply do not work in an increasing number of scenarios, such as new Amazon S3 regions, operations involving encrypted objects and more.

dziemid:
Thank you for quick answer.

Is 1-week limitation enforced across all the regions? We are currently using 60 days in us-east-1.

trevorrowe:
I assume with the 60 days you are using the classic signature which does not have a limit on the expiration.

dziemid:
Thank you, make sense.

In that case, happy to close this one.

PR #937, opened 2015-09-29
Accept #to_str attributes as Strings in dynamodb

soulcutter:
Unlike #to_s, #to_str is only implemented by classes that act like
strings. This change allows user-defined classes which provide that
conversion to be used directly by dynamodb clients and be treated as
String attributes.

For example, users may define a class which is a composite of two or
more values which are stored as a single String in dynamodb, and pass
instances of that class into dynamodb clients without need for explicit
conversion.


trevorrowe:
Looks good, thanks!

PR #940, opened 2015-09-30
Fixes #938

### LINKED ISSUE #938: SignatureDoesNotMatch when using virtual_host option
I'm trying to use 2.1.23 to generate a presigned url for an item in an s3 bucket with a custom domain name.
Let's say the bucket's name is stuff.example.com and that there is a cname configured so that if you go to http://stuff.example.com/some-uuid/index.html you get a stack website.
This is all working, however I would like to turn on some security via the presigned_url feature.

My code is as follows:

```
  def get_object_url(bucket_name, object_name)
    res = Aws::S3::Resource::new
    res.bucket(bucket_name)
        .object(object_name)
        .presigned_url(:get, {
                               expires_in: 1.day,
                               virtual_host: true
                           })
  end
```

The result is: http://stuff.example.com/some-uuid/indexAPI.html?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=MYKEY%2F20150929%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20150929T192323Z&X-Amz-Expires=86400&X-Amz-SignedHeaders=host&X-Amz-Signature=718d7d44e3286266cc0cfd7b303f60c63df7425bea4952632816ee4491b6452a

And when I try it in a browser I get this:

SignatureDoesNotMatch --
The request signature we calculated does not match the signature you provided. Check your key and signing method.

denodster:
Fixes issue where a non https url is given https.

### DIFF
--- a/aws-sdk-core/lib/aws-sdk-core/s3/presigner.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/s3/presigner.rb
@@ -78,7 +78,9 @@ module Aws
         req.handle do |context|
           uri = context.http_request.endpoint
           uri.host = context.params[:bucket]
-          uri.path = uri.path.sub("/#{context.params[:bucket]}", '')
+          uri.path.sub!("/#{context.params[:bucket]}", '')
+          uri.path.sub!('https', 'http')
+          uri.port = 80
           @handler.call(context)
         end
       end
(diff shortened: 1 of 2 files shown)


denodster:
okay, what did I do wrong? I changed 2 lines. surely, that can't be breaking EVERYTHING...

denodster:
Looks like the regex in the test was broken by the addition of :80 to the end of the domain in the url, also it was kind of too permissive in that it would allow:
`http://myabucketacom/` but not `http://my.bucket.com:80/`

trevorrowe:
Thanks!

PR #954, opened 2015-10-12
add support for VPC peering connections

guss77:
Add the classes VPCPeeringConnection and VPCPeeringConnectionCollection
Add support for getting existing peering connections and creating new connections to VPC
Add support to RouteTable to target a VPC peering connection

### DIFF
--- a/lib/aws/ec2/route_table.rb
+++ b/lib/aws/ec2/route_table.rb
@@ -196,6 +196,11 @@ module AWS
           client_opts[:network_interface_id] = interface
         end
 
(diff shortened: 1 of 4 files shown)


trevorrowe:
Thanks for the contribution. I've merged this, but I'm not sure when our next scheduled release of version 1 will be. The version 1 SDK is primarily in maintenance mode at this point.

guss77:
Thanks. That's OK - currently I'm using a local patch to get my VPC peering connection fix. If there's a release every month or two, I'm fine with that and I'll just port features from my local branch to the official v1 as time permits.

PR #964, opened 2015-10-19
Is "Service Side Encryption" a typo?

igrep:
Seems [the document](http://docs.aws.amazon.com/AmazonS3/latest/dev/serv-side-encryption.html) calls it "Server-Side Encryption".


trevorrowe:
Thanks.

PR #976, opened 2015-10-29
Add support for prefix lists.

dtbartle:


dtbartle:
@trevorrowe ping?

awood45:
Thank you for your pull request. We're not taking new features for V1 of the SDK anymore, as it is on the deprecation path and we're doing bug fixes and security patches only.

If you'd like to add features to V1, I would recommend forking and vendoring the package.

I would also recommend looking at V2 of the SDK, which you can use in the same project as V1 with the gem structure we have created. See the README for more information.

Issue #949, opened 2015-10-09
Aws::S3::MultipartUploadError when sending large files AND using KMS (works in v 2.1.28)


danlherman:
Got the following error:

aws-sdk-resources-2.1.26/lib/aws-sdk-resources/services/s3/multipart_file_uploader.rb:72:in `abort_upload': multipart upload failed: undefined method `match' for nil:NilClass (Aws::S3::MultipartUploadError)

The method works with smaller files, or on the same large file (5.9gb) when sent without KMS. This issue does NOT happen using v 2.1.28, but does happen on 2.1.29 and 2.1.26.

Used the following code:

#!/usr/bin/env ruby
require 'aws-sdk'
kms_key = 'xxxxxxxx'
s3 = Aws::S3::Resource.new(
  credentials: Aws::Credentials.new('xxxxxxx', 'xxxxxxxx'),
  region: 'us-east-1'
)
obj = s3.bucket('bucket-name').object('path/to/file.zip')
obj.upload_file('file.zip',  server_side_encryption: 'aws:kms', ssekms_key_id: kms_key)

danlherman:
Sorry, it seems to happen on 2.1.28 with some larger files as well. Just tried sending a 48gb file and after a while got this error:

aws-sdk-resources-2.1.28/lib/aws-sdk-resources/services/s3/multipart_file_uploader.rb:72:in `abort_upload': multipart upload failed: undefined method `match' for nil:NilClass (Aws::S3::MultipartUploadError)

trevorrowe:
Thank you for reporting this issue. We are looking into it now.

trevorrowe:
Do you know the region of the target bucket? Also, would it be possible for you to rescue the `Aws::S3::MultipartUploadError` for some additional debug information. The upload works by uploading parts using multiple threads. If any of those threads generate errors, then they are aggregated and raised. You can access the original errors from the MPU error:

```ruby
begin
  # ...
rescue Aws::S3::MultipartUploadError => mpu_error
  mpu_error.errors.each do |error|
    puts error.message
    puts error.backtrace
  end
end
```

I think the issue is related signing the upload part requests. When using server side encryption with KMS the request must be signed with signature version 4. There is code in the SDK to rescue the error that indicates this and then attempts to resign and send the request. Its a bit tricky because the SDK must determine the region of the target bucket to generate the version 4 signature.

[...]

Issue #965, opened 2015-10-19
"The request signature we calculated does not match the signature you provided." when uploading with `Aws::S3::Object#upload_file`


BanzaiMan:
This is possibly a variation of https://github.com/aws/aws-sdk-ruby/issues/923. With 2.1.30, a user reports that uploading file results in signature mismatch. The back traces are almost identical:

Before:

```
/home/travis/.rvm/gems/ruby-1.9.3-p551/gems/aws-sdk-core-2.1.20/lib/seahorse/client/plugins/raise_response_errors.rb:15:in `call': The request signature we calculated does not match the signature you provided. Check your key and signing method. (Aws::S3::Errors::SignatureDoesNotMatch)
	from /home/travis/.rvm/gems/ruby-1.9.3-p551/gems/aws-sdk-core-2.1.20/lib/aws-sdk-core/plugins/s3_sse_cpk.rb:18:in `call'
	from /home/travis/.rvm/gems/ruby-1.9.3-p551/gems/aws-sdk-core-2.1.20/lib/aws-sdk-core/plugins/param_converter.rb:21:in `call'
	from /home/travis/.rvm/gems/ruby-1.9.3-p551/gems/aws-sdk-core-2.1.20/lib/seahorse/client/plugins/response_target.rb:21:in `call'
	from /home/travis/.rvm/gems/ruby-1.9.3-p551/gems/aws-sdk-core-2.1.20/lib/seahorse/client/request.rb:70:in `send_request'
	from /home/travis/.rvm/gems/ruby-1.9.3-p551/gems/aws-sdk-core-2.1.20/lib/seahorse/client/base.rb:207:in `block (2 levels) in define_operation_methods'
	from /home/travis/.rvm/gems/ruby-1.9.3-p551/gems/aws-sdk-resources-2.1.20/lib/aws-sdk-resources/services/s3/file_uploader.rb:42:in `block in put_object'
	from /home/travis/.rvm/gems/ruby-1.9.3-p551/gems/aws-sdk-resources-2.1.20/lib/aws-sdk-resources/services/s3/file_uploader.rb:49:in `open_file'
	from /home/travis/.rvm/gems/ruby-1.9.3-p551/gems/aws-sdk-resources-2.1.20/lib/aws-sdk-resources/services/s3/file_uploader.rb:41:in `put_object'
[...]

PR #979, opened 2015-10-30
SigV4 as Default for All Amazon S3 Calls

awood45:
Remove hacks that supported 'S3 Signer' default for many common code
paths. Now, SigV4 is always the default and special fallback logic is
removed.

You can still use the legacy signer, without special fallbacks, by
specifying the signature version in your client:
`client = Aws::S3::Client.new(signature_version: 's3')`

Resolves issue #949 and may be related to issue #965

### DIFF
--- a/aws-sdk-core/features/s3/step_definitions.rb
+++ b/aws-sdk-core/features/s3/step_definitions.rb
@@ -14,7 +14,9 @@ After("@s3") do
       break if objects.empty?
       @client.delete_objects(bucket: bucket, delete: { objects: objects })
     end
+    puts "MADE IT TO DELETING BUCKET"
     @client.delete_bucket(bucket: bucket)
+    puts "DELETED BUCKET"
   end
 end
 
--- a/aws-sdk-core/features/s3/step_definitions.rb
+++ b/aws-sdk-core/features/s3/step_definitions.rb
@@ -14,9 +14,7 @@ After("@s3") do
       break if objects.empty?
       @client.delete_objects(bucket: bucket, delete: { objects: objects })
     end
-    puts "MADE IT TO DELETING BUCKET"
     @client.delete_bucket(bucket: bucket)
-    puts "DELETED BUCKET"
   end
 end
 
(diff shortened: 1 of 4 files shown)


trevorrowe:
Some notes:

* Add entries into the CHANGELOG and possibly the UPGRADING guides.
* The `CachedBucketRegionHandler` in the S3 request signer plugin was previously forcing known `us-east-1` buckets to use the s3-external-1 endpoint. We should consider if this is the correct endpoint or if the classic endpoint is to be used.
* The comment block above the `BucketSigningErrorHandler` is not longer relevant. This handler now only deals with wrong region errors.

Look for more code comments.

aws-sdk-core/features/s3/step_definitions.rb
```diff
@@ -14,7 +14,9 @@
       break if objects.empty?
       @client.delete_objects(bucket: bucket, delete: { objects: objects })
     end
+    puts "MADE IT TO DELETING BUCKET"
```
trevorrowe: Be sure to remove the pair of debug statements here.

aws-sdk-core/lib/aws-sdk-core/plugins/s3_request_signer.rb
```diff
@@ -6,28 +6,15 @@ module Plugins
     # @api private
     class S3RequestSigner < Seahorse::Client::Plugin
 
-      class SigningHandler < RequestSigner::Handler
+      option :signature_version, 'v4'
```
trevorrowe: Stylistically, I prefer the method invocation with parenthesis, i.e. `option(:signature_version, 'v4')`

aws-sdk-core/lib/aws-sdk-core/plugins/s3_request_signer.rb
```diff
@@ -6,28 +6,15 @@ module Plugins
...
         def call(context)
           require_credentials(context)
-          version = signature_version(context)
-          case version
-          when /v4/ then apply_v4_signature(context)
-          when /s3/ then apply_v2_signature(context)
+          case context.config.signature_version
+          when 'v4' then apply_v4_signature(context)
+          when 's3' then apply_s3_legacy_signature(context)
           else raise "unsupported signature version #{version.inspect}"
```
trevorrowe: This is a good opportunity to improve this existing error message. Add into the message the valid signature versions.

PR #985, opened 2015-11-04
Fix InstanceExists Waiter

joaogbcravo:
Instead of checking just the status of DescribeInstances operation from aws, we should check if the reservations list have length > 0.

### DIFF
--- a/aws-sdk-core/apis/ec2/2015-10-01/waiters-2.json
+++ b/aws-sdk-core/apis/ec2/2015-10-01/waiters-2.json
@@ -7,8 +7,9 @@
       "operation": "DescribeInstances",
       "acceptors": [
         {
-          "matcher": "status",
-          "expected": 200,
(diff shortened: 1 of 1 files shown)


joaogbcravo:
To have more context about this change:
I created a machine, then I destroyed it, and waited to not be seen anymore on my dashboard.
When I do describe_instances({instance_ids: [DESTROYED_INSTANCE_ID]}) it returns a good response (200 OK) with an empty reservation list, instead of InvalidInstanceIDNotFound. 

This  makes the Aws::Resources::Resource exists? method to be a "liar" method.

trevorrowe:
Sorry for the slow merge. The fix looks good. We may be able to simplify the path expression and replace it with a pathAny matcher. Semantically it would be the same as the pathAny and pathAll matchers may not return empty lists.

This should go out with the next release.

PR #994, opened 2015-11-10
add retry to S3::Bucket delete!

tmornini:
We have regularly seen S3.Bucket delete! return Errors::BucketNotEmpty.

This should reduce that a great deal. :-)


tmornini:
The tests were green locally and it appears the failures are all gem install failures and not this PR's changes...

aws-sdk-resources/lib/aws-sdk-resources/services/s3/bucket.rb
```diff
@@ -24,8 +24,21 @@ def clear!
...
+
+        attempts = 0
+
+        begin
+          clear!
+          delete
+        rescue Errors::BucketNotEmpty => e
+          Kernel.sleep 1.3 ** attempts
+
+          fail e if attempts > 20
```
awood45: This is a fairly lengthy retry period, I'd like to see this start with a tighter default (like 3), with an option that can be passed in to set your own retry period.

awood45:
Generally looks good, I'd like to make the change around default attempt counts before we can merge. While you're adding that option, could just as easily make the wait period configurable as well.

tmornini:
@awood45 Options added

tmornini:
@awood45 Good call, sorry for being lazy. I actually had the semantics of max_attempts wrong...

tmornini:
Rebased on current master

aws-sdk-resources/spec/services/s3/bucket/delete_spec.rb
```diff
@@ -34,13 +34,119 @@ module S3
...
+              .with(bucket: 'bucket-name')
+              .and_raise(bucket_not_empty_exception).ordered
+
+            expect(Kernel).to receive(:sleep).with(1.3).ordered
+
+            expect(client).to receive(:delete_bucket)
+              .with(bucket: 'bucket-name')
+              .and_raise(bucket_not_empty_exception).ordered
+
+            expect(Kernel).to receive(:sleep).with(1.6900000000000002).ordered
```
awood45: Add a comment around this to explain the magic number, at least. That way, it's a less confusing failure if the defaults are changed later. Or, move out to variables that are grouped together.

awood45:
I think we're about set to go here. Last comment above, then I'll pull in to test and merge. Thanks!

tmornini:
@awood45 Done. After the rebase, specs now log:

```DEPRECATION WARNING: called deprecated batch method `#delete` on a batch of `Aws::S3::ObjectVersion` objects; use `#batch_delete!` instead```

A few quick pokes make me 80% certain that is unrelated to this work.

PR #1007, opened 2015-11-18
Remove deprecation warning in S3::Bucket#clear!

janko-m:
Since `#delete` has been renamed to `#batch_delete!`, we should also update `#clear!` to not issue deprecation warnings.


awood45:
Thanks!

PR #1015, opened 2015-11-24
Allow region to be specified in shared credentials file

jkanywhere:
Pass options to all the credential providers so the user can choose which
profile they want from their shared credentials file.

Allos that shared credential file to also specify the region associated with
each profile.

Update AWS client configuration to read region from the credential provider in
addition to all its normal places.

### DIFF
--- a/lib/aws/core/configuration.rb
+++ b/lib/aws/core/configuration.rb
@@ -467,12 +467,16 @@ module AWS
 
       add_option :session_token
 
+      add_option :path
+
+      add_option :profile_name
+
       add_option :region do |cfg,region|
         region || cfg.credential_provider.credentials[:region] || ENV['AWS_REGION'] || ENV['AMAZON_REGION'] || ENV['AWS_DEFAULT_REGION'] || 'us-east-1'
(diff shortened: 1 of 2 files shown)


awood45:
Thank you for your pull request. We're not taking new features for V1 of the SDK anymore, as it is on the deprecation path and we're doing bug fixes and security patches only.

If you'd like to add features to V1, I would recommend forking and vendoring the package.

I would also recommend looking at V2 of the SDK, which you can use in the same project as V1 with the gem structure we have created. See the README for more information.

felix-okta:
info for op: https://boto3.readthedocs.org/en/latest/reference/core/session.html#boto3.session.Session for how python does it

felix-okta:
```
a=boto3.session.Session(profile_name='aew1d')
b=a.client('ec2') # for ec2
c=a.client('s3')
```

jkanywhere:
I'm not asking for this to be re-opened.
Just noting this topic has been considered before: https://github.com/aws/aws-sdk-ruby/pull/703

PR #1019, opened 2015-12-04
fixes typos on readme.

common-nighthawk:
removes erroneous comma after last key/value pair in ruby hash.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ You can configure default credentials and region via `Aws.config`. **In version 
 ```ruby
 Aws.config.update({
   region: 'us-west-2',
-  credentials: Aws::Credentials.new('akid', 'secret'),
+  credentials: Aws::Credentials.new('akid', 'secret')
 })
 ```
 


awood45:
I wouldn't call this erroneous, in the sense that the code works just fine (it would be invalid JSON, but not invalid Ruby). Stylistically though this is a cleaner look.

awood45:
I'll go with the best practice on this one. Thanks!

common-nighthawk:
Thanks for the merge, Alex.

I'm embarrassed to say that I thought `{ 'hello' => 'world', 'foo' => 'bar', }` was invalid Ruby code.  I should have run it in IRB first.

However, I did confirm that this syntax is consistent with accepted best practices--
https://github.com/bbatsov/ruby-style-guide#no-trailing-array-commas

Which is all to say... sorry for using 'erroneous' :smile:

(I recently watched your talk from RailsConf15.  Thanks for everything re: AWS + RB.)

awood45:
Correct, it's a common misconception due to the fact that JSON *will* break, but it's a best practice to exclude the comma.

Also, I appreciate the thanks, hope the workshop was useful!

Issue #1012, opened 2015-11-23
Endpoint isn't used in presigned post


hakunin:
Hi,
I am using the `fake-s3` gem and it advises to use `endpoint` config to point `aws-sdk-ruby` to localhost.

What I found though is that the presigned post class isn't aware of endpoints and seems to always return s3 url no matter what.

awood45:
How are you trying to do this with the `fake-s3` gem? The presigned post method is primarily constructing headers, and does have a `url` parameter you can set.

trevorrowe:
@hakunin You should be able to simply Ignore the presigned-post `#url` and use `localhost` in your form. Is this not working?

hakunin:
I am using `shrine` and request the url with ajax. Shrine which asks the bucket to pre-sign a post, which results in the bucket url not using the endpoint.

Since the idea is to use `fakes3` not if statements to let the code go completely different path based on the environment, it would be good if setting the endpoint in `aws-sdk-ruby` would simply return the endpoint path.

andretanguy:
Experiencing a potentially related issue with the ruby aws-sdk (v2) whereby we're using a Ceph endpoint rather than S3 itself (ala similar setup to fake-s3) and when we include 'endpoint' in our config (alongside force_path_style), the gem always returns an error:

Aws::S3::Errors::InvalidArgument

awood45:
@andretanguy Could you please open a separate issue for that? That looks like a possible separate bug.

awood45:
@hakunin - I just merged pull request #1020 which plumbs endpoints through presigned posts created via S3 Objects as well as S3 Buckets within the resource interface.

If you're using our S3 resource abstraction, this should be all you need. If you're creating presigned post object directly, you should use the `:url` parameter provided.

Going to close this now, feel free to reach out if this isn't clear or doesn't fully cover what you need to do.

@andretanguy I'd still like to dive in to what you're seeing as a separate issue.

hakunin:
Thanks alex!

andretanguy:
@awood45 all done for you: #1021

PR #1020, opened 2015-12-05
Apply :endpoint option to Bucket#presigned_post

janko-m:
`Object#presigned_post` properly passes the bucket's url (which is affected by the `:endpoint` option), so to keep consistency we make `Bucket#presigned_post` pass it as well.


awood45:
Related to #1012

awood45:
Looks good with a first pass.

awood45:
Integration tests pass locally as well, and the behavior seems to match. Thanks!

janko-m:
Thanks!

Cool, I didn't know I've fixed an open issue. And I'm also the maintainer of Shrine which the opener mentions, it's cool how it's all connected.

PR #1026, opened 2015-12-10
Header Blacklist

awood45:
A proposed list of headers that we will never attempt to sign in SDK requests.


trevorrowe:
I took the change and made the minor change myself. We should see about getting a maintenance release out with this fix so the affected `Aws::CloudSearchDomain::Client` users can pickup the work-around for the Content-Length bug.

PR #1027, opened 2015-12-11
Fix Presigner not respecting :endpoint URL scheme

janko-m:
When presigned URL is generated for a client which has `:endpoint`, the URL scheme was always "https", even when the `:endpoint` had "http" scheme.

### DIFF
--- a/aws-sdk-core/lib/aws-sdk-core/s3/presigner.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/s3/presigner.rb
@@ -54,7 +54,11 @@ module Aws
       private
 
       def http_scheme(params, virtual_host)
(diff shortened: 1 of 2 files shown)


awood45:
There's a failing test here, essentially because the client call within expects the `:secure` parameter to not be present.

janko-m:
Updated.

awood45:
I've pulled down and confirmed that this doesn't break our integration tests around presigning.

awood45:
I like the idea, but I think this can be changed to be more generally applicable. I'm going to pull this down, modify it, and push.

trevorrowe:
I pulled this locally and made a small change to ensure this worked for all URIs by default. Your changes have been merged into mainline. Thanks for the PR!

PR #1043, opened 2015-12-30
Adding support for AES/GCM/NoPadding

DanielRedOak:
Adds support for AES/GCM/NoPadding and uses the auth_tag from the end of the ciphertext to authenticate the complete ciphertext when `cipher.final` is called in the code.  Should also add additional flexibility to add further formats in the future if necessary.

Presence of an auth tag is detected via `x-amz-meta-x-amz-tag-len` being present in the response headers.  Unencrypted data is then written to the IO Object (vs unencrypting on .write without auth tag) to obtain the auth tag from the last `x-amz-meta-x-amz-tag-len` bits of the ciphertext. The IO Object is then truncated to remove the auth tag and then re-processed and decrypted after setting the `cipher.auth_tag` and `cipher.auth_data=''` on the cipher (order of operations is critical).  When `cipher.final` is called, OpenSSL authenticates the ciphertext and returns.

Currently lacks tests, but was locally tested with and without a target_response to test the different IO classes that are used.  Would like to get tests written, but wanted to get this in front of people first.

**Simple Example Usage**
```ruby
# pass in the kms_key_id, bucket, object_key, and target
s3encrypt = Aws::S3::Encryption::Client.new(kms_key_id: kms_key_id)
s3encrypt.get_object(bucket:bucket, key:object_key, response_target: target)
```

### DIFF
--- a/aws-sdk-resources/lib/aws-sdk-resources/services/s3/encryption/decrypt_handler.rb
+++ b/aws-sdk-resources/lib/aws-sdk-resources/services/s3/encryption/decrypt_handler.rb
@@ -38,11 +38,35 @@ module Aws
           context.http_response.on_headers(200) do
             cipher = decryption_cipher(context)
             body = context.http_response.body
-            context.http_response.body = IODecrypter.new(cipher, body)
+            if context.http_response.headers['x-amz-meta-x-amz-tag-len']
+              # we'll need to get the entire body before decrypting
+              context.http_response.body = IODecrypter.new(cipher, body, contains_tag = true)
+            else
+              context.http_response.body = IODecrypter.new(cipher, body, contains_tag = false)
+            end
+
           end
 
           context.http_response.on_success(200) do
             decrypter = context.http_response.body
+
+            # if present, we'll need to authenticate the ciphertext
+            if context.http_response.headers['x-amz-meta-x-amz-tag-len']
+              auth_tag_len = context.http_response.headers['x-amz-meta-x-amz-tag-len'].to_i
+              decrypter.io.seek((decrypter.io.size - (auth_tag_len/8)))
+              auth_tag = decrypter.io.read
+              decrypter.cipher.auth_tag = auth_tag
+              decrypter.cipher.auth_data = ''
+              decrypter.io.truncate((decrypter.io.size - (auth_tag_len/8)))
+              decrypter.io.rewind if decrypter.io.respond_to?(:rewind)
+              # read back in the encrypted data
+              content = decrypter.io.read
+              decrypter.io.truncate(0)
+              decrypter.io.rewind if decrypter.io.respond_to?(:rewind)
+              # reprocess, decrypting data and authenticating ciphertext
+              decrypter.contains_tag = false
+              decrypter.write(content)
+            end
             decrypter.finalize
             decrypter.io.rewind if decrypter.io.respond_to?(:rewind)
             context.http_response.body = decrypter.io
--- a/aws-sdk-resources/lib/aws-sdk-resources/services/s3/encryption/decrypt_handler.rb
+++ b/aws-sdk-resources/lib/aws-sdk-resources/services/s3/encryption/decrypt_handler.rb
@@ -22,6 +22,11 @@ module Aws
 
         POSSIBLE_ENVELOPE_KEYS = (V1_ENVELOPE_KEYS + V2_ENVELOPE_KEYS).uniq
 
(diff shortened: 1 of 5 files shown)


awood45:
Thanks for doing this, looks like a good start. I'm going to review through this after the new year, and I'm happy to help with getting this ready to merge in.

DanielRedOak:
@awood45 need anything from me to get things moving?  We're using SES + encryption in S3 so we're using my branch for decryption atm.

awood45:
I'm going to take a deeper look at this now.

awood45:
I can confirm this is passing all tests for us. Reviewing code again now.

aws-sdk-resources/lib/aws-sdk-resources/services/s3/encryption/kms_cipher_provider.rb
```diff
@@ -25,7 +25,7 @@ def encryption_cipher
           envelope = {
             'x-amz-key-v2' => encode64(key_data.ciphertext_blob),
             'x-amz-iv' => encode64(cipher.iv = cipher.random_iv),
-            'x-amz-cek-alg' => 'AES/CBC/PKCS5Padding',
+            'x-amz-cek-alg' => 'AES/CBC/PKCS5Padding', #TODO allow selection
```
awood45: Let's take out any TODO comments, as a style matter - I agree we should add this (although that isn't a blocker to me for accepting this PR, given the bug related to this), but we can handle that via other means.

awood45:
This looks good to me and passes our existing tests - I would like to see unit tests though. Let me know if you need any guidance on this.

DanielRedOak:
Just glancing over things it looks like tests for this change won't be trivial since kms is hardcoded to use AES/CBC/PKCS5Padding but I'll see about matching things up when I get some time

DanielRedOak:
Added a test similar to the CBC test to verify decryption and renamed the CBC tests to be more descriptive now that there are multiple ciphers

DanielRedOak:
@awood45 anything else we need to get this merged?

awood45:
Very sorry for the delay - Trevor had code review comments he wanted to add.

DanielRedOak:
@awood45 any update?

DanielRedOak:
@awood45 just wondering about a timeline on this.  I have a ticket on our end I'd like to close when this gets merged in.  Thanks!

aws-sdk-resources/lib/aws-sdk-resources/services/s3/encryption/decrypt_handler.rb
```diff
@@ -33,11 +38,35 @@ def attach_http_event_listeners(context)
...
+              decrypter.io.truncate((decrypter.io.size - (auth_tag_len/8)))
+              decrypter.io.rewind if decrypter.io.respond_to?(:rewind)
+              # read back in the encrypted data
+              content = decrypter.io.read
+              decrypter.io.truncate(0)
+              decrypter.io.rewind if decrypter.io.respond_to?(:rewind)
+              # reprocess, decrypting data and authenticating ciphertext
+              decrypter.has_tag = false
+              decrypter.write(content)
+            end
```
trevorrowe: I'd like to see if we can improve the implementation to not buffer, seek, truncate and rewind the target IO object. I think this can be improved by making a 2nd GET request to the object with a range header that will retrieve only the trailing authentication tag. This can be used to initialize the cipher that we give to the IO decrypter which should eliminate the need for the extra reads.

I've pulled this down locally, and I'm going to try this out.

trevorrowe:
I've added a code comment. I'd like to avoid buffering the object to get the auth tag from the end of the body by making a GET Object with range request. This should allow us to initialize the cipher with the auth tag straightway.

aws-sdk-resources/lib/aws-sdk-resources/services/s3/encryption/decrypt_handler.rb
DanielRedOak: My thought was to minimize calls to the API, but that seems reasonable as well as long as we aren't forced into download the full object more than once.  I might be able to peek next week too.  Let me know how things go.

trevorrowe:
I've added https://github.com/aws/aws-sdk-ruby/pull/1204 which has an implementation my suggestion above.

trevorrowe:
I've merged these commits here: c3376f02c0490273674a37879a688438326bf84a

Thanks for the excellent contribution!

PR #1044, opened 2016-01-03
Test on Ruby 2.3

optimisticanshul:


awood45:
Thanks for this - I agree that starting Ruby 2.3 test coverage is a good idea.

It does look like the Travis build is failing, however, so I'm going to look in to that before I can accept the PR.

trevorrowe:
Closing in favor of 6acaac487c1a659e49da68e5f4694ada85fd7f33. Travis does not yet support using `2.3` as a Ruby version, but rather you must specify the explicit `2.3.0` version. This is an issue with an upstream dependency of theirs and they are waiting for it be resolved.

PR #1046, opened 2016-01-06
Include additional errors in the list of retryable network errors.

pdrakeweb:
Both host unreachable and connection refused are possible transient network errors.


awood45:
Thanks for this!

Issue #1049, opened 2016-01-08
JSON Engine issue


kronos:
While switching to aws sdk v2 I had a couple of experiments in irb to understand how to use the new version of the gem and, thanks to the authors, everything was quite straightforward. After putting new gem into my project I encounter a strange issue:
```
Aws::S3::Resource.new(region: S3_REGION, credentials: s3_credentials)
NoMethodError: undefined method `each' for nil:NilClass
	from /apps/singing/source/vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.2.9/lib/aws-sdk-core/waiters/provider.rb:9:in `initialize'
	from /apps/singing/source/vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.2.9/lib/aws-sdk-core/client_waiters.rb:13:in `new'
	from /apps/singing/source/vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.2.9/lib/aws-sdk-core/client_waiters.rb:13:in `set_waiters'
	from /apps/singing/source/vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.2.9/lib/aws-sdk-core/client.rb:43:in `define'
	from /apps/singing/source/vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.2.9/lib/aws-sdk-core.rb:378:in `block in <module:Aws>'
	from /apps/singing/source/vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.2.9/lib/aws-sdk-core.rb:368:in `call'
	from /apps/singing/source/vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.2.9/lib/aws-sdk-core.rb:368:in `block in add_service'
	from /apps/singing/source/vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.2.9/lib/aws-sdk-core.rb:367:in `each'
	from /apps/singing/source/vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.2.9/lib/aws-sdk-core.rb:367:in `add_service'
	from /apps/singing/source/vendor/bundle/ruby/2.2.0/gems/aws-sdk-core-2.2.9/lib/aws-sdk-core/s3.rb:1:in `<top (required)>'
[...]

PR #1054, opened 2016-01-11
Ensure that Oj does not Symbolize JSON

awood45:
Resolves GitHub Issue #1049

### DIFF
--- a/aws-sdk-core/lib/aws-sdk-core/json.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/json.rb
@@ -23,7 +23,7 @@ module Aws
     class << self
 
       def load(json)
-        ENGINE.load(json)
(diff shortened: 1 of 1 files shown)


trevorrowe:
Looks good. We should add a unit test that uses a before hook to set the Oj default option and then disables it in an after hook (to ensure we don't leak configuration).

Also worth noting, the `mode: :compat` is necessary for both load and dump. Previously we only had this on dump. Without compat mode on, Oj attempts to convert identifiers that begin with ":" into symbols which we do not want.

awood45:
Let's talk about a unit test strategy later today - most of the rest of our tests try to be unit test agnostic, and as you say, the global config change is more or less inevitable (an attempt to not leak configuration would likely just result in assuming the default behavior is always `:symbolize_names` being false, although that's a likely safe assumption.

PR #1055, opened 2016-01-12
addressed #1053, us-west-1 was hardcoded

omargallob:
i had an issue with yesterdays fix. 

the option(:sigv4_region) was returning us-west-1 and this hardcoded fix didn't work for me. i have pushed a fix that allows different regions (depending on the region included in the endpoint)

### DIFF
--- a/aws-sdk-core/lib/aws-sdk-core/plugins/request_signer.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/plugins/request_signer.rb
@@ -49,8 +49,11 @@ module Aws
 
       option(:sigv4_region) do |cfg|
         prefix = cfg.api.metadata['endpointPrefix']
+
         if cfg.endpoint.to_s.match(/#{prefix}\.amazonaws\.com/)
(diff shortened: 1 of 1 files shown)


awood45:
Can you tell me a bit more about your issue? I'm happy to roll back the second phase of yesterday's fix if there is a regression in behavior we missed, but this wouldn't be the fix.

Please open an issue so we can pin down what regressed.

awood45:
Any further information on this? I'm very curious about your use case because that regex should not be triggered for any endpoints that contain a region, only "global" endpoints.

omargallob:
sorry alex, i just saw this email. 

i was getting invalid region 'iot.eu-west-1', after the fix i was getting invalid region 'us-west-1'. i forked the gem and changed the hard coded us-west-1 to fit my region.

should of notified before attempting to make a fix.

awood45:
So for IOT, you should be correctly getting your actual region, because setting the region is required in the client configuration. Can you open an issue with a failing code example?

PR #1059, opened 2016-01-13
Updated to the latest versions of the protocol tests.

trevorrowe:
Bulk updated the protocol tests. There were some minor library updates necessary to support these.

### DIFF
--- a/aws-sdk-core/lib/aws-sdk-core/rest/request/endpoint.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/rest/request/endpoint.rb
@@ -77,7 +77,8 @@ module Aws
                 if StringShape === member.shape.member.shape
                   parts += list_of_strings(member.location_name, params[member_name])
                 else
-                  msg = "Only list of strings supported"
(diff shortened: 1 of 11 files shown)


aws-sdk-core/lib/aws-sdk-core/rest/request/endpoint.rb
```diff
@@ -73,6 +73,14 @@ def apply_querystring_params(uri, params)
                   raise NotImplementedError, msg
                 end
 
+              when ListShape
+                if StringShape === member.shape.member.shape
+                  parts += list_of_strings(member.location_name, params[member_name])
+                else
+                  msg = "Only list of strings supported"
```
awood45: Optional: Could we output the actual type we received in this case?
```ruby
msg = "Only list of strings supported, was: #{member.shape.member.shape}"
```

awood45:
LGTM

trevorrowe:
I've added more information about the given shape in the unsupported case.

trevorrowe:
I merged this locally, it has been pushed to master already.

PR #1060, opened 2016-01-14
Resolved an issue with EC2 request ids.

### LINKED ISSUE #1038: Cannot get an EC2 request ID
This code cannot get an EC2 request ID.
```ruby
region = "us-east-1"
client = Aws::EC2::Client.new(region: region)
response = client.describe_instances
response.context[:request_id]

=> nil
```

`Aws::Plugins::Protocols::EC2#parse_xml` method passes a block to `Aws::Xml::Parser#parse` method.
https://github.com/aws/aws-sdk-ruby/blob/master/aws-sdk-core/lib/aws-sdk-core/plugins/protocols/ec2.rb#L12-L21
```ruby
          def parse_xml(context)
            if rules = context.operation.output
              data = Xml::Parser.new(rules).parse(xml(context)) do |h|
                context.metadata[:request_id] = h['requestId']
              end
              data
            else
              EmptyStructure.new
            end
          end
```

But `Aws::Xml::Parser#parse` method does not treat a block.
https://github.com/aws/aws-sdk-ruby/blob/master/aws-sdk-core/lib/aws-sdk-core/xml/parser.rb#L23-L31
```ruby
      # @param [String] xml An XML document string to parse.
      # @param [Structure] target (nil)
      # @return [Structure]
      def parse(xml, target = nil)
        xml = '<xml/>' if xml.nil? or xml.empty?
        stack = Stack.new(@rules, target)
        @engine.new(stack).parse(xml.to_s)
        stack.result
      end
```

https://github.com/aws/aws-sdk-ruby/commit/52f3efadcd29e92d271da3ec15e661529438b115#diff-91359190dca8b9e40db446ea5c1a5654
```diff
 -      # @param [String<xml>] xml
 -      # @param [Hash, nil] target
 +      # @param [String] xml An XML document string to parse.
 +      # @param [Structure] target (nil)
        # @return [Structure]
 -      def parse(xml, target = nil, &block)
 -        xml = MultiXml.parse(xml).values.first || {}
 -        yield(xml) if block_given?
 -        structure(@shape, xml, target)
 +      def parse(xml, target = nil)
 +        xml = '<xml/>' if xml.nil? or xml.empty?
 +        stack = Stack.new(@shape, target)
 +        @engine.new(stack).parse(xml)
 +        stack.result
        end
```

trevorrowe:
This is a regression from when we switched to stream parsing the XML. Older versions of the SDK parsed the entire XML document into a DOM, and then walked that DOM to parse the values. Because the entire document was no longer available in memory, the parser could no longer yield it as a hash. This caused the EC2 protocol handler to silently fail, no longer able to populate the request ID.

This update adds a callback to the XML parser that yields when un-modeled/un-handled XML values are present in the response. The EC2 handler registers a callback and checks the context/path of the unhandled value and captures it if it is the request id.

To prevent a regression in the future, a unit test was added to ensure the EC2 client correctly extracts the request id from the HTTP response body and stores it in the response context.

Fixes #1038

### DIFF
--- a/aws-sdk-core/lib/aws-sdk-core/xml/parser.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/xml/parser.rb
@@ -20,6 +20,21 @@ module Aws
         @engine = options[:engine] || self.class.engine
       end
 
+      # Parses the XML document, returning a parsed structure.
+      #
+      # If you pass a block, this will yield for XML
+      # elements that are not modeled in the rules given
+      # to the constructor.
+      #
+      #   parser.parse(xml) do |path, value|
+      #     puts "uhandled: #{path.join('/')} - #{value}"
+      #   end
+      #
+      # The purpose of the unhandled callback block is to
+      # allow callers to access values such as the EC2
+      # request ID that are part of the XML body but not
+      # part of the operation result.
+      #
       # @param [String] xml An XML document string to parse.
       # @param [Structure] target (nil)
       # @return [Structure]
--- a/aws-sdk-core/lib/aws-sdk-core/xml/parser.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/xml/parser.rb
@@ -23,9 +23,9 @@ module Aws
       # @param [String] xml An XML document string to parse.
       # @param [Structure] target (nil)
       # @return [Structure]
-      def parse(xml, target = nil)
+      def parse(xml, target = nil, &unhandled_callback)
         xml = '<xml/>' if xml.nil? or xml.empty?
-        stack = Stack.new(@rules, target)
+        stack = Stack.new(@rules, target, &unhandled_callback)
         @engine.new(stack).parse(xml.to_s)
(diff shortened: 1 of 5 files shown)


aws-sdk-core/lib/aws-sdk-core/xml/parser.rb
```diff
@@ -23,9 +23,9 @@ def initialize(rules, options = {})
       # @param [String] xml An XML document string to parse.
       # @param [Structure] target (nil)
       # @return [Structure]
-      def parse(xml, target = nil)
+      def parse(xml, target = nil, &unhandled_callback)
```
awood45: I understand what `&unhandled_callback` is for because of the full context of the change, but it could be hard to remember or not intuitive to someone reading. Can we add this to the docs above?

awood45:
LGTM, one comment on documentation.

trevorrowe:
Additional code documentation added.

PR #1062, opened 2016-01-18
iot-data api has incorrect prefix, hence does not work

SStorm:
There is no data.iot url, and and as far as I'm aware, the data plane endpoint should use the same iot prefix. This seemingly never worked. 

The unit tests failed, and it looks like because of my change, but I'm a bit baffled as to why.


awood45:
This is an auto-generated file, so we can't take changes to it. In any case, this change wouldn't change any behavior.

You are right that `data.iot` isn't a real endpoint, and that each endpoint actually contains an identifier. Because of this, our Aws::IoTDataPlane clients require that you specify an endpoint explicitly.

If you've run into issues with that in our client, please open an issue. Thanks!

Issue #1070, opened 2016-01-22
Aws::IoTDataPlane Clients using "us-east-1" for Signing Region


awood45:
Related to #1052, possibly a regression introduced by #1053. No matter which region the client is instantiated to, the sigv4 signing region is set to us-east-1.

awood45:
The bug appears to be the fact that, for Aws::IoTDataPlane in particular (or any service where we require specification of an endpoint by the user in general, so possibly Aws::CloudSearchDomain as well), the "prefix" variable in `Aws::Plugins::RequestSigner` is nil, which means the "us-east-1" hardcoded endpoint is matched against `/\.amazonaws\.com/`. That would match just about any endpoint.

awood45:
Apologies for the regression - I've tested this fix against that failing case and others, and it will go out with the next release.

Issue #1052, opened 2016-01-11
Aws::IoTDataPlane::Errors::ForbiddenException


lsilvs:
I'm trying to use the ruby sdk in order to get the shadow of my Thing but I'm getting the following error:

Code used:
```ruby
client = Aws::IoTDataPlane::Client.new(
  endpoint: 'https://data.iot.eu-west-1.amazonaws.com',
  region: 'eu-west-1',
  credentials: credentials,
)
```

Error returned:
```ruby
Credential should be scoped to a valid region, not 'iot.eu-west-1'
```

Is the endpoint I'm using correct?

trevorrowe:
I think this may be an issue with how the service's regional signing name is extracted from the endpoint:
https://github.com/aws/aws-sdk-ruby/blob/master/aws-sdk-core/lib/aws-sdk-core/plugins/request_signer.rb#L53

To demonstrate:

```ruby
client = Aws::IoTDataPlane::Client.new(
  endpoint: 'https://data.iot.eu-west-1.amazonaws.com',
  region: 'eu-west-1',
)
cilent.config.sigv4_region
#=> "io.eu-west-1" -- oops, should be "eu-west-1"
```

This shouldn't be to difficult to fix. Thank you for reporting the issue.

awood45:
Pushed the fix, will go out with the next release.

lsilvs:
Tested and approved.

Cheers,

lsilvs:
Hi @awood45 
seems that your commit '[Fix SigV4 Signing Region for Aws::IoTDataPlane](https://github.com/aws/aws-sdk-ruby/commit/804d0f8d43a67132ae149a3ad0690be4c52e6022)' solved the problem but the commit '[Simplify Request Signer Region Assignment](https://github.com/aws/aws-sdk-ruby/commit/b0cb55845802955617bace5718219d6c0df83167)' brought it back.

Can you check it out, pls

awood45:
I can't reproduce an error, here is what I'm seeing with the latest version (using `aws.rb`):
```ruby
Aws> client = Aws::IoTDataPlane::Client.new(region: "us-east-1", endpoint: "https://#{iot.describe_endpoint.endpoint_address}")
[Aws::IoT::Client 200 0.67704 0 retries] describe_endpoint()
=> #<Aws::IoTDataPlane::Client>
Aws> client.config.sigv4_region
=> "us-east-1"
Aws> client.publish(topic: "foo")
[...]

PR #1072, opened 2016-01-22
Get Correct sigv4_region When No endpointPrefix

awood45:
Aws::IoTDataPlane::Client intentionally does not define `endpointPrefix`
in the configured API metadata as a mechanic to force the user to supply
an `:endpoint` value on client construction.

The previous attempted fix failed to note that, in absence of
`endpontPrefix`, all endpoints will match with the first selector,
hard-coded to "us-east-1". This change will check for this case, and use
your client configured region when it occurs.

* Resolves GitHub Issue #1070
* Related to GitHub Issue #1052
* Fixes Regression In GitHub Pull Request #1053


trevorrowe:
The change looks good, but I would be more inclined to see a functional test for a specific service to exercise this. Try adding a test to the IoT data plane client spec that does this.

Issue #1076, opened 2016-01-29
Indefinite wait while uploading_archive and file read pointer is at eof


dhunnapotha:
**How to reproduce**

```
glacier  = ::Aws::Glacier::Client.new :region => 'us-east-1'
fp = File.open '/tmp/archive.txt'
fp.read 1000

glacier.upload_archive ({
        vault_name: 'test',
        account_id: '-',
        archive_description: 'sample description',
        body: fp
      })
```

/tmp/archive.txt is a small text file with 1 word in it
fp.read 1000 puts the read pointer of fp at the end of file

The upload_archive hangs indefinitely.

Please note that this might not be a valid usecase, but in my case, this happened, as fp has been moved to eof by some other library, which was not entirely obvious

awood45:
Thanks for the report, will take a look at the pull request.

awood45:
Working on a fix within the Glacier Checksum plugin that rewinds the body before attempting calculation. That would prevent this code path from being triggered.

awood45:
Thanks for the report!

PR #1077, opened 2016-01-29
Raise error when treehash.digest called without any updates.

dhunnapotha:
This addresses the #1076 issue.

### DIFF
--- a/aws-sdk-core/lib/aws-sdk-core/tree_hash.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/tree_hash.rb
@@ -48,6 +48,8 @@ module Aws
     end
 
     def digest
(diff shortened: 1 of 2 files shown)


awood45:
Can you provide a failing test for this? Would like to have the issue covered by our test suite, especially for regression testing purposes.

dhunnapotha:
Hey 
- Dint find any specs related to tree_hash. So, I created a file tree_hash_spec and added a test case.
- Without the fix, the code goes into infinte loop. I can't think of a proper way to have a testcase to test for infinte loop. 

Can you go through the changes and let me know if you have any further inputs?

awood45:
Thank you for the code contribution. Looking at the Glacier Checksum Plugin, we're going to take a different approach that will prevent the invalid call. I'll push that change up shortly and close the issue.

PR #1082, opened 2016-02-02
Update README.md

bogdanvlviv:


trevorrowe:
The copy is correct. The intended word is `ini`, not init. This is describing the file format, which is ini. Thank you for the PR though! We do appreciate the help.

PR #1086, opened 2016-02-08
Fix Ruby fatal error: "uninitialized constant Aws::Client::Errors".

chupakabr:
Hello,

Recently, trying to utilise Firehose V2 client, I ended up receiving Ruby fatal errors, basically it was always the same error:
```
"uninitialized constant Aws::Client::Errors"
```

In the end I found out that this error happens every time I receive an error response from Firehose endpoint. After debugging I noticed that mentioned error is being generated because of invalid "svc" variable initialization in xml/error_handler.rb error() method which tried to get it like this:

```
svc = context.client.class.name.split('::')[1]
```

though, client class is always Module::Client here, on the other hand, we keep service name identificator in the context.client.class.identifier, so the idea of the fix is to just capitalize identificator so "svc" variable is always created using underlying module, in my case Aws::Firehose::Errors::... instead of Aws::Client::Errors::..., like this:

```
svc = context.client.class.identifier.to_s.capitalize
```

After I applied this fix I started receiving proper errors in my code like this, which I assume is an expected behaviour:
```
Firehose my-test-stream not found under account NNNNNNNNNNNNN.
```

I would appreciate if you review my fix and sort out this problem in the next release of aws-sdk-ruby. Feel free to contact me if you have any questions.

With best,
Valera

### DIFF
--- a/aws-sdk-core/lib/aws-sdk-core/xml/error_handler.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/xml/error_handler.rb
@@ -21,7 +21,7 @@ module Aws
         else
           code, message = extract_error(body, context)
         end
-        svc = context.client.class.name.split('::')[1]
+		svc = context.client.class.identifier.to_s.capitalize
         errors_module = Aws.const_get(svc).const_get(:Errors)
         errors_module.error_class(code).new(context, message)
       end


awood45:
The "capitalize" portion is causing multiple regressions in the test suite. May want to consider another approach. Will try to take a look at this on my end as well.

chupakabr:
Today I noticed that this is happening when I run the plugin from RSpec like this:
```
  let(:output) { LogStash::Outputs::Firehose.new({"codec" => "plain"}) }
```
However,  proper exception is being thrown/handled when I install the plugin into Logstash and run Logstash from the command line.

So might be a minor one as it is related to tests only or I just don't know how to properly use Ruby and RSpec :smile:

trevorrowe:
@chupakabr Are you by chance testing your code in JRuby?

trevorrowe:
Capitalizing the identifier is not sufficient to convert it into a class Name. Some services have multiple words. For example `:dynamodb` capitalizes to `Dynamodb` but would need to be `DynamoDB`.

My comment above was also highlighting that this is only an issue if your module does not correctly report its name, which is a feature of Ruby modules. There was a regression in JRuby that caused this not to work. I believe this has been corrected.

Lastly, I'm not happy about how this code works today. If I had a do-over, I would instead not generate the exception classes at runtime at all. Unfortunately this behavior can not change without a major version bump of the SDK, which is not currently planned.

chupakabr:
Hey Trevor,

Gotcha. I was using JRuby for testing. I'm not JRuby or Ruby expert, so maybe my configuration JRuby is messed up somewhere. Anyway, the code itself works fine for me, cannot say the same about tests though, but not a big deal for me now. :)

With best,
Valera

djpate:
Having the same issue with jRuby 9.0.5.0

PR #1089, opened 2016-02-09
Remove one of two indentical spec examples

tawan:


awood45:
Looks good, thanks for catching this!

Issue #1085, opened 2016-02-07
MD5 digest verification for SQS message attributes


tawan:
I implemented the calculation MD5 digests for message attributes in a [gem of mine](https://github.com/tawan/active-elastic-job/blob/v1.3.0/lib/active_elastic_job/md5_message_digest_calculation.rb)

Now I wonder if this is actually already done in this client, but I couldn't find it, or if it's still necessary to check the md5 digests at all.

Cheers,

tawan

awood45:
It's not currently done, but we would certainly take a pull request to do it, or add it as a feature request.

If you want to take a crack at the PR, I'd recommend writing such a validation as a plugin, similar to our S3 MD5 checker.

tawan:
Ok, I'll have a look at the S3 MD5 checker and wrap it in a PR. In case of questions, I'll drop them here.

PR #1093, opened 2016-02-11
Verify MD5 checksums returned by SQS

tawan:
Verify MD5 checksums returned by SQS for received messages.
 * Verify checksum of message body.
 * Verify checksum of message attributes if present.

 Refer to #1085

### DIFF
--- a/aws-sdk-core/lib/aws-sdk-core/plugins/sqs_md5s.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/plugins/sqs_md5s.rb
@@ -1,0 +1,148 @@
+require 'openssl'
+
+module Aws
+  module Plugins
+
+    # @seahorse.client.option [Boolean] :verify_checksums (true)
+    #   When `true` MD5 checksums will be computed for messages sent to
+    #   an SQS queue and matched against MD5 checksums returned by Amazon SQS.
+    #   `Aws::Errors::Checksum` errors are raised for cases where checksums do
+    #   not match.
+    class SQSMd5s < Seahorse::Client::Plugin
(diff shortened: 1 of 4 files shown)


trevorrowe:
I took a moment to review this. The code looks good. I had one piece of feedback above we need to address and then I'm happy to merge this in. Good work!

aws-sdk-core/lib/aws-sdk-core/plugins/sqs_md5s.rb
```diff
@@ -0,0 +1,148 @@
...
+        end
+
+        def validate_send_message_batch(context, response)
+          context.params[:entries].each do |entry|
+            id = entry[:id]
+            body = entry[:message_body]
+            attributes = entry[:message_attributes]
+            message_response = response.successful.select { |r| r.id == id }[0]
+            unless message_response.nil?
+              validate_single_message(body, attributes, message_response)
```
tawan: @trevorrowe This is the line you mentioned in the previous commit: I simple added a `nil` check. Only messages which are which are placed in the `successful` array are verified.

tawan:
@trevorrowe Thanks for your feedback. I incorporated your suggestion. I amended my changes and rebased, therefore your inline comment seems to be gone. I commented the line in question again for some context.

trevorrowe:
Thanks for the contribution! This should go out with our next release. 

I'm going to do a small amount of refactoring on the tests. We've been moving away from unit testing the handlers in isolation and instead functionally test the client class they are applied to. This has the benefit of ensuring the plugin is applied correctly. Your tests are organized well enough that moving this should be pretty straightforward.

PR #1104, opened 2016-02-25
Region-to-region multipart copy fix

### LINKED ISSUE #1083: Aws::S3::Object.copy_from fails cross-regions when multipart_copy: true
We're attempting to copy objects from a bucket in one region to a bucket in another region with the multipart_copy option set to true. We've found that the operation succeeds when `multipart_copy` is not specified in the `Object.copy_from` call, but as soon as we add it we instead receive an `Aws::S3::Errors:Http301Error`.

Digging in with the aws.rb client, we additionally found that when invoking `copy_from` with `multipart_copy: true` the `HEAD` call erroneously constructs an endpoint using the *target* region endpoint but the *source* region bucket, e.g.,

```
Aws> result = target_object.copy_from(source_object, multipart_copy: true)
opening connection to esp-dev-singapore.s3-us-west-2.amazonaws.com:443...
...
[Aws::S3::Client 301 0.474308 0 retries]
```

As opposed to doing the same call without `multipart_copy: true`...

```
Aws> result = target_object.copy_from(source_object)
opening connection to esp-dev-oregon.s3-us-west-2.amazonaws.com:443...
...
[Aws::S3::Client 200 3.406666 0 retries]
```

Is is a known restriction that you cannot copy using multipart across regions?

whazzmaster:
Fixes #1083 

:deciduous_tree: 

Ensures that the appropriate client is queried to find the source object's size when copying from region-to-region. 

I also noticed that, though the tests appear to explicitly state that passing content_length to `S3::Object#copy_from` will bypass calling HEAD on the source object, the documentation did not mention this option. I updated the `copy_from` documentation to include the content_length option as that's the workaround we're currently using.

### DIFF
--- a/aws-sdk-resources/lib/aws-sdk-resources/services/s3/object.rb
+++ b/aws-sdk-resources/lib/aws-sdk-resources/services/s3/object.rb
@@ -20,6 +20,10 @@ module Aws
       #   performance improvements on large objects. Amazon S3 does
       #   not accept multipart copies for objects smaller than 5MB.
       #
+      # @options options [Integer] :content_length When included
+      #   the source object will not be queried for its size before
+      #   starting the copy.
+      #
       # @example Basic object copy
       #
       #   bucket = Aws::S3::Bucket.new('target-bucket')
--- a/aws-sdk-resources/lib/aws-sdk-resources/services/s3/object.rb
+++ b/aws-sdk-resources/lib/aws-sdk-resources/services/s3/object.rb
@@ -5,7 +5,7 @@ module Aws
       alias size content_length
 
       # Copies another object to this object. Use `multipart_copy: true`
-      # for large objects. This is required for objects that exceed 5GB.'
+      # for large objects. This is required for objects that exceed 5GB.
       #
       # @param [S3::Object, String, Hash] source Where to copy object
       #   data from. `source` must be one of the following:
--- a/aws-sdk-resources/lib/aws-sdk-resources/services/s3/object.rb
+++ b/aws-sdk-resources/lib/aws-sdk-resources/services/s3/object.rb
@@ -20,9 +20,9 @@ module Aws
       #   performance improvements on large objects. Amazon S3 does
       #   not accept multipart copies for objects smaller than 5MB.
       #
-      # @options options [Integer] :content_length When included
-      #   the source object will not be queried for its size before
-      #   starting the copy.
+      # @options options [Integer] :content_length Used only when
+      # multipart_copy is TRUE.  When included the source object will not be
+      # queried for its size before starting the copy.
       #
       # @example Basic object copy
       #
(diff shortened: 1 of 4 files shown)


whazzmaster:
Hmm, looks like jruby fails one spec. Taking a look now.

whazzmaster:
Soooooo, after debugging this all afternoon, it appears that the rpec mocks are not thread safe and so the multipart upload stubbing of `client.upload_part_copy` may fail due to an `ArrayIndexOutofBoundsException`.

See `aws-sdk-resources/spec/services/s3/object/multipart_copy_spec:189` for the block where 60 mocks get setup. If, to test it out, you comment out that block and replace `aws-sdk-resources/lib/aws-sdk-resources/services/s3/object_multipart_copier.rb#copy_part` where the etag is retrieved from the stubbed call with some hard-coded string, you will never see the test fail. 

There seems to be an issue inside the rspec-mocks that if mocks are being 'consumed' in a multi-threaded context, at some point there is an index OOB exception. I just wonder why this only happens in jruby...

aws-sdk-resources/lib/aws-sdk-resources/services/s3/object.rb
```diff
@@ -20,6 +20,10 @@ class Object
       #   performance improvements on large objects. Amazon S3 does
       #   not accept multipart copies for objects smaller than 5MB.
       #
+      # @options options [Integer] :content_length When included
+      #   the source object will not be queried for its size before
+      #   starting the copy.
+      #
```
trevorrowe: We should reword this documentation to indicate the `:content_length` is only needed when `:multipart_copy` is `true`.  The content_length is used for calculating the part size and nothing else.

aws-sdk-resources/lib/aws-sdk-resources/services/s3/object_copier.rb
```diff
@@ -12,6 +12,7 @@ def initialize(object, options = {})
       end
 
       def copy_from(source, options = {})
+        options[:copy_source_client] = source.client if source.is_a?(S3::Object)
```
trevorrowe: This could also be expressed as:

```ruby
options[:copy_source_client] ||= source.client if source.respond_to?(:client)
```

The `||=` is less important, the `respond_to?` check makes this a little more future proof. Currently we have not added support for `Aws::S3::ObjectVersion`, but this is likely to happen. No need to change this unless you feel inclined. If you don't, we will make the change when adding object version support.

whazzmaster: Sounds good- I'll try to collect the feedback and push an update this afternoon.

trevorrowe:
This looks good. I think the only, non-necessary prior to merging, additions we could make it to improve this are:

* Add a `:copy_source_region` option where users can specify the region to allow the SDK to construct a client and HEAD the source object when needed.
* Document `:copy_source_client`

whazzmaster:
I think I got everything in there- take another look and let me know if I missed anything.

trevorrowe:
Thanks for the contributions! I've pulled branch locally and added to it support for object version copies.

trevorrowe:
Sorry, it looks like master changed between when I pulled your branch so this pull request didn't cleanly merge. That said, it does contain all of your commits.

whazzmaster:
:+1: Thanks!

Issue #1103, opened 2016-02-25
v2.2.19 nil to integer TypeError in md5_of_message_attributes(message_attributes)


zwang80:
aws-sdk-core v2.2.19

Aws::SQS::Client.send_message returns exception
sending message using option: {:queue_url=>"https://sqs.us-west-2.amazonaws.com/1234567890/aws-sqs-test", :message_body=>"{\"mirror\":true}", :message_attributes=>{"att_string"=>{:data_type=>"String.test", :string_value=>"String value"}}}
name: "att_string", attribute: {:data_type=>"String.test", :string_value=>"String value"}

in md5_of_message_attributes(message_attributes)
[TRANSPORT_TYPE_ENCODINGS[attribute[:data_type]]] turns out to be [nil]

TypeError: no implicit conversion of nil into Integer
    /home/dummydir/vendor/bundle/ruby/2.1.0/gems/aws-sdk-core-2.2.19/lib/aws-sdk-core/plugins/sqs_md5s.rb:111:in `pack'
    /home/dummydir/vendor/bundle/ruby/2.1.0/gems/aws-sdk-core-2.2.19/lib/aws-sdk-core/plugins/sqs_md5s.rb:111:in `block in md5_of_message_attributes'
    /home/dummydir/vendor/bundle/ruby/2.1.0/gems/aws-sdk-core-2.2.19/lib/aws-sdk-core/plugins/sqs_md5s.rb:92:in `each'
    /home/dummydir/vendor/bundle/ruby/2.1.0/gems/aws-sdk-core-2.2.19/lib/aws-sdk-core/plugins/sqs_md5s.rb:92:in `md5_of_message_attributes'
    /home/dummydir/vendor/bundle/ruby/2.1.0/gems/aws-sdk-core-2.2.19/lib/aws-sdk-core/plugins/sqs_md5s.rb:74:in `validate_attributes'
    /home/dummydir/vendor/bundle/ruby/2.1.0/gems/aws-sdk-core-2.2.19/lib/aws-sdk-core/plugins/sqs_md5s.rb:57:in `validate_single_message'
    /home/dummydir/vendor/bundle/ruby/2.1.0/gems/aws-sdk-core-2.2.19/lib/aws-sdk-core/plugins/sqs_md5s.rb:40:in `validate_send_message'
[...]

PR #1105, opened 2016-02-25
Fix TypeError raised by SQS MD5 verification

tawan:
Fix TypeError which is raised during verification of  MD5 checksum of SQS
message attributes. This error is caused by the usage of data type with custom
labels. Transport encodings for data types with custom labels weren't obtainable,
thus packing a `nil` value resulted in a TypeError.

Refer #1103


trevorrowe:
Thanks for the bug fix!

Issue #1106, opened 2016-02-27
standard_ia storage class not honored on multi-part upload


jaypatrickhoward:
Using aws-sdk 2.2.0 with ruby 2.1.5  on CentOS6.

The following code:

```
object = bucket.object("#{prefix}/#{object_name}")
object.upload_file("#{object_name}", storage_class: "STANDARD_IA")

```
...creates objects with storage class "STANDARD_IA" for files smaller than 15 MB but storage class "STANDARD" for files 15 MB or larger, i.e. the default threshold for triggering a multi-part upload.

Is that by design?

jaypatrickhoward:
Verified that "storage_class" isn't being passed on the "create_multipart_upload", "upload_part" or "complete_multipart_upload" calls when those are provoked by a call to "upload_file".

mirceal:
+1

mirceal:
:+1:

awood45:
This would not be by design for any reason I can think of, thanks for the report.

jaypatrickhoward:
Happy to oblige.  Let me know if there's anything else I can do to expedite the process of this getting into an official release.  Update the test cases, etc.  (Will require me to figure out how the test cases work, though...)

awood45:
Certainly happy to see a PR for this. The unit tests would likely just confirm that the relevant headers are being lifted to the request (which would, at this point, be a failing test).

trevorrowe:
Thanks for the PR.

PR #1108, opened 2016-02-27
use correct set of params when calling create_multipart_upload

jaypatrickhoward:
See issue #1106 on aws-sdk-ruby.  The lines:

```
      CREATE_OPTIONS =
        Set.new(Client.api.operation(:upload_part).input.shape.member_names)
```
were most likely a cut/paste error from the next couple lines:

```
      UPLOAD_PART_OPTIONS =
        Set.new(Client.api.operation(:upload_part).input.shape.member_names)
```

Changing the operation for CREATE_OPTIONS from ":upload_part" to ":create_multipart_upload" results in the "storage_class" parameter correctly being passed.  Otherwise it is omitted since it's not a valid parameter for the "upload_part" call.


trevorrowe:
Good catch. Thanks for the PR!

PR #1121, opened 2016-03-03
Fixed exception on Aws::AutoScaling::Instance#load

jamesiarmes:
Calling `load`/`reload` on an instance of `Aws::AutoScaling::Instance` causes `Uncaught exception: unexpected value at params[:instances_ids]`. This can be reproduced by the following:

```ruby
require 'aws-sdk'

name = 'ASGName'
asg = Aws::AutoScaling::AutoScalingGroup.new(
  name: name
)
instance = asg.instances.first
instance.reload
```

I was able to identify a typo in the parameters for the operation.


trevorrowe:
Thanks for the fix!

PR #1126, opened 2016-03-11
Presigned POST requests got error response, when using SSE-C.

takeyuweb:
The encryption key must be  base64-encoding.

See at http://docs.aws.amazon.com/AmazonS3/latest/dev/ServerSideEncryptionCustomerKeys.html


trevorrowe:
Thanks for the fix!

PR #1128, opened 2016-03-11
Include 302 redirect code

fgouteroux:
The W3 spec seems to indicate that they're both used for temporary redirects.

### DIFF
--- a/aws-sdk-core/lib/aws-sdk-core/plugins/s3_redirects.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/plugins/s3_redirects.rb
@@ -13,7 +13,8 @@ module Aws
 
         def call(context)
           response = @handler.call(context)
-          if context.http_response.status_code == 307 or 302
+          status_code = context.http_response.status_code
+          if status_code == 307 || status_code == 302
             endpoint = context.http_response.headers['location']
             context.http_request.endpoint = endpoint
             context.http_response.body.truncate(0)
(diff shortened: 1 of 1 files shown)


awood45:
This code is breaking a large number of tests.

aws-sdk-core/lib/aws-sdk-core/plugins/s3_redirects.rb
```diff
@@ -13,7 +13,7 @@ class Handler < Seahorse::Client::Handler
 
         def call(context)
           response = @handler.call(context)
-          if context.http_response.status_code == 307
+          if context.http_response.status_code == 307 or 302
```
awood45: This if statement will *always* return true. You would instead write something similar to:

```ruby
status_code = context.http_response.status_code
if status_code == 307 || status_code == 303
  # do things
end
```

awood45:
I'm going to test this somewhat carefully to avoid any unexpected behavior change. What will help is, are you addressing a particular piece of missing functionality, or is change just being made "in case"?

fgouteroux:
It's not a missing functionality from the sdk. We use an s3 compatible storage with Geo-IP, and our load-balancer send a 302 redirect. I'm not sure that this PR is right.

awood45:
Given that there isn't anything that hits this, going to close this PR. If you see a case where this would be hit that I can reproduce, feel free to reopen!

PR #1136, opened 2016-03-17
Support New Regions and Endpoints File

awood45:


aws-sdk-core/lib/aws-sdk-core/endpoint_provider.rb
```diff
@@ -2,30 +2,52 @@ module Aws
...
+          '{dnsSuffix}', partition["dnsSuffix"]
+        )
       end
 
-      def expand(pattern, region, service)
-        'https://' + pattern.sub('{region}', region).sub('{service}', service)
+      def get_partition(region)
+        RULES["partitions"].find do |p|
+          region.match(p["regionRegex"])
+        end
```
trevorrowe: The #get_partition method could own the logic of returning 'aws' as the default partion if the region does not match the regex of any known partition. This removes the need for the nil check above.

mtdowling:
Did you consider also utilizing the endpoint specific metadata for influencing the connection? For example, utilizing signaturerVersions, signing names, etc?

aws-sdk-core/lib/aws-sdk-core/partitions.rb
```diff
@@ -0,0 +1,153 @@
...
+  # To get a partition by name:
+  #
+  #     aws = Aws.partition('aws')
+  #
+  # You can also enumerate all partitions:
+  #
+  #     Aws.partitions.each do |partition|
+  #       puts partition.name
+  #     end
+  #
```
awood45: I'd love the "all everything in everywhere" example to be present, since that's a use case some have. For example:

```ruby
Aws.partitions.each do |partition|
  partition.regions.each do |region|
    region.services.each do |service|
      # Do anything you'd want to do for all services in all regions.
      puts "#{service} is in #{region.name} in #{partition.name}"
    end
  end
end
```

PR #1137, opened 2016-03-19
Fix error message on unsupported S3 signature version

tyok:
When an unsupported signature version is passed to the S3 client, it will raise a confusing error message:

```
NameError: undefined local variable or method `version' for #<Aws::Plugins::S3RequestSigner::SigningHandler:0x007f896f5273a8>
```

This change will make it raises the intended error message:

```
RuntimeError: unsupported signature version "v2", valid options: 'v4' (default), 's3'
```


awood45:
Good catch, thanks!

PR #1141, opened 2016-03-23
Remove Incorrect Resource Waiter Documentation

awood45:
The resource waiter documentation uses the underlying client call
parameter data, which is not necessary and is in fact incorrect. This
removes that part of the generated documentation.

Also removes unused code around plumbing extra params through to the
waiter, which was not supported anyways.


awood45:
Reviewed with Trevor offline.

Issue #1143, opened 2016-03-29
copy_from and copy_to not working when a object file with a space is used and multipart_copy enabled


johan-smits:
When I copy a object from a source bucket to a new bucket and the source file has a space, it fails with the multipart => true.

The source_object.key is for example: "foo bar.baz"

Code:
```ruby
source_object.copy_to("#{AWS_S3_GLACIER_DIR}/#{target_object.key}",
                                     {multipart_copy: (source_object.size > 5.megabytes)})
```

Back trace:
```
/var/lib/gems/2.1.0/gems/aws-sdk-core-2.2.30/lib/seahorse/client/plugins/raise_response_errors.rb:15:in `call': Aws::S3::Errors::NotFound
	from /var/lib/gems/2.1.0/gems/aws-sdk-core-2.2.30/lib/aws-sdk-core/plugins/s3_sse_cpk.rb:18:in `call'
	from /var/lib/gems/2.1.0/gems/aws-sdk-core-2.2.30/lib/aws-sdk-core/plugins/param_converter.rb:20:in `call'
	from /var/lib/gems/2.1.0/gems/aws-sdk-core-2.2.30/lib/seahorse/client/plugins/response_target.rb:21:in `call'
	from /var/lib/gems/2.1.0/gems/aws-sdk-core-2.2.30/lib/seahorse/client/request.rb:70:in `send_request'
	from /var/lib/gems/2.1.0/gems/aws-sdk-core-2.2.30/lib/seahorse/client/base.rb:207:in `block (2 levels) in define_operation_methods'
	from /var/lib/gems/2.1.0/gems/aws-sdk-resources-2.2.30/lib/aws-sdk-resources/services/s3/object_multipart_copier.rb:124:in `source_size'
	from /var/lib/gems/2.1.0/gems/aws-sdk-resources-2.2.30/lib/aws-sdk-resources/services/s3/object_multipart_copier.rb:31:in `copy'
	from /var/lib/gems/2.1.0/gems/aws-sdk-resources-2.2.30/lib/aws-sdk-resources/services/s3/object_copier.rb:31:in `copy_object'
	from /var/lib/gems/2.1.0/gems/aws-sdk-resources-2.2.30/lib/aws-sdk-resources/services/s3/object_copier.rb:15:in `copy_from'
	from /var/lib/gems/2.1.0/gems/aws-sdk-resources-2.2.30/lib/aws-sdk-resources/services/s3/object.rb:66:in `copy_from'
[...]

PR #1144, opened 2016-03-29
Avoid Double URL Encoding S3 Keys

awood45:
When performing a copy operation with the S3 Object resource, the object
key would be URL encoded before being sent on to the S3 client, which
URL encodes the key a second time. This was causing a 404 to
short-circuit the copy operation.

Tests were checking the inputs into the client call, which would not
catch the second step of the double encoding.

Resolves issue #1143


awood45:
This turns out to be the wrong approach, so I'm abandoning this PR, to be replaced by a new one later.

PR #1146, opened 2016-03-31
Add a couple examples of how to require the library properly for new users

rgooler:
Helping newbies avoid the trap I just fell in.


trevorrowe:
Thanks!

PR #1159, opened 2016-04-14
FIX: describe_launch_configurations's max_records default is 50

kenjiskywalker:
Hi.

I tried to this code.

```rb
client = Aws::AutoScaling::Client.new(
  region: REGION_NAME,
  access_key_id: ACCESS_KEY,
  secret_access_key: SECRET_KEY
)

launch_configs = {}
client.describe_launch_configurations.launch_configurations.each do | lc |
  p lc.launch_configuration_name
end
```

This code return `50` items.

But, [This Official Document](http://docs.aws.amazon.com/sdkforruby/api/Aws/AutoScaling/Client.html#describe_launch_configurations-instance_method) says
`The maximum number of items to return with this call. The default is 100.`

which one is correct?
make sure, I send this PR.

Please check it. thank you.


awood45:
I'll check this with the team and doc writers, but we can't take pull requests to an auto-generated file. Thanks for the heads up!

kenjiskywalker:
OK, thanks. 😉

PR #1163, opened 2016-04-19
Added `:use_accelerate_endpoint` option to Aws::S3::Client.

trevorrowe:
You can now pass a configuration option to accelerate `Aws::S3::Client` operations. You can construct a client with `use_accelerate_endpoint: true` to enable this feature.

```ruby
s3 = Aws::S3::Client.new(use_accelerate_endpoint: true)
s3.put_object(bucket: 'bucket-name', key:'key')
#=> uses https://bucket-name.s3-accelerate.amazonaws.com
```

You can pass `:use_accelerate_endpoint` to client operations to override the client default.

```ruby
s3 = Aws::S3::Client.new(use_accelerate_endpoint: true)
s3.put_object(bucket: 'bucket-name', key:'key')
#=> uses https://bucket-name.s3-accelerate.amazonaws.com
```

You can pass `:use_accelerate_endpoint` to client operations to override the client default.

```ruby
# non-accelerated client
s3 = Aws::S3::Client.new

# non-accelerated
s3.put_object(bucket: 'bucket-name', key:'key')

# accelerated
s3.put_object(bucket: 'bucket-name', key:'key', use_accelerate_endpoint: true)
```

[See the Amazon S3 documentation for more information](http://docs.aws.amazon.com/AmazonS3/latest/dev/transfer-acceleration.html).

Not supported for the following operations:

* `#create_bucket`
* `#list_buckets`
* `#delete_bucket`

### DIFF
--- a/aws-sdk-core/lib/aws-sdk-core/api/docs/param_formatter.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/api/docs/param_formatter.rb
@@ -44,7 +44,7 @@ module Aws
             end
           when ListShape then list(ref, i, visited)
           when MapShape then map(ref, i, visited)
-          when BooleanShape then "true"
+          when BooleanShape then "false"
           when IntegerShape then '1'
           when FloatShape then '1.0'
           when StringShape then string(ref)
--- a/aws-sdk-core/lib/aws-sdk-core/plugins/s3_accelerate.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/plugins/s3_accelerate.rb
@@ -1,0 +1,73 @@
+module Aws
+  module Plugins
+
+    # Provides support for using `Aws::S3::Client` with Amazon S3 Transfer
+    # Acceleration.
+    #
+    # Go here for more information about transfer acceleration:
+    # [http://docs.aws.amazon.com/AmazonS3/latest/dev/transfer-acceleration.html](http://docs.aws.amazon.com/AmazonS3/latest/dev/transfer-acceleration.html)
+    #
+    # @seahorse.client.option [Boolean] :use_accelerate_endpoint (false)
+    #   When set to `true`, accelerated bucket endpoints will be used
+    #   for all object operations. You must first enable
+    #   accelerate for each bucket. [Go here for more information](http://docs.aws.amazon.com/AmazonS3/latest/dev/transfer-acceleration.html).
+    #
+    class S3Accelerate < Seahorse::Client::Plugin
+
+      option(:use_accelerate_endpoint, false)
+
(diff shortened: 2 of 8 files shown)


aws-sdk-core/lib/aws-sdk-core/api/docs/param_formatter.rb
```diff
@@ -44,7 +44,7 @@ def ref_value(ref, i, visited)
             end
           when ListShape then list(ref, i, visited)
           when MapShape then map(ref, i, visited)
-          when BooleanShape then "true"
+          when BooleanShape then "false"
```
awood45: This is just for doc examples, right? And I assume this is to avoid the default presumption of examples being "accelerate on"?

aws-sdk-core/lib/aws-sdk-core/plugins/s3_accelerate.rb
```diff
@@ -0,0 +1,73 @@
...
+    #   for all object operations. You must first enable
+    #   accelerate for each bucket. [Go here for more information](http://docs.aws.amazon.com/AmazonS3/latest/dev/transfer-acceleration.html).
+    #
+    class S3Accelerate < Seahorse::Client::Plugin
+
+      option(:use_accelerate_endpoint, false)
+
+      def add_handlers(handlers, config)
+        operations = config.api.operation_names - [
+          :create_bucket, :list_buckets, :delete_bucket,
```
awood45: Long term nervous about hard naming these. Fine for now, should consider alternatives down the road.

awood45:
LGTM, assuming my assumption within was correct.

aws-sdk-core/lib/aws-sdk-core/api/docs/param_formatter.rb
trevorrowe: Yes, that is only for the documented request syntax examples. Another option would be to do something like `"true || false"`. I would prefer to keep it valid Ruby syntax. The important part of this is to ensure the example shows that a boolean is provided.

aws-sdk-core/lib/aws-sdk-core/plugins/s3_accelerate.rb
trevorrowe: Completely understood, not sure of a valid alternative. `#create_bucket` and `#list_buckets` are simply not supported. The reason is the s3-accelerate endpoint must have an accelerated bucket as the endpoint prefix. Create is creating a new bucket that does not previously exist and list buckets is outside the context of a single bucket. `#delete_bucket` was on request from the S3 team. There is no deterministic characteristic of these that makes it possible to not hardcode.

PR #1164, opened 2016-04-19
Added stricter resource collection arg validation.

trevorrowe:
Resource has many operation accept a hash argument only. Updated collection methods to now raise on argument errors.

```ruby
bucket = Aws::S3::Bucket.new('bucket-name')

bucket.objects('key')
#=> now raises ArgumentError, expecting Hash, got String

bucket.objects(1,2,3)
#=> now raises ArgumentError, expecting 0..1 args, got 2
```

### DIFF
--- a/aws-sdk-resources/lib/aws-sdk-resources/operations.rb
+++ b/aws-sdk-resources/lib/aws-sdk-resources/operations.rb
@@ -108,6 +108,7 @@ module Aws
         # @option (see Base#call)
         # @return [Collection]
         def call(options)
+          validate_args!(options)
           Collection.new(self, options)
         end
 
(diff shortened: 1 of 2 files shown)


awood45:
Sanity Check: Is there any collection, anywhere, that behaves differently? Are we going to be able to ensure that all collections going forward behave in this manner?

awood45:
Change itself looks fine w/r/t stated intent.

trevorrowe:
All resource collection / has many associations are wrappers around paginated API calls. The only accepted arguments are `:limit` and pass-through arguments to the client operation. This is strictly always a hash. This is per the resources spec.

PR #1177, opened 2016-05-03
Use SVG badges over PNG

olivierlacan:
Looks all crisp and stuff. ;-)


awood45:
That is, indeed, very crisp. I approve!

PR #1178, opened 2016-05-06
Shared Configuration Support

awood45:
This change adds the `Aws::SharedConfig` file, which parses both the
Shared Credentials file and the Shared Config file. From that, functions
are provided to support:

* Credentials from shared credentials or shared config.
* Default region selection from shared config.
* Assume role credentials from shared config.

Shared config support is toggled behind the `AWS_SDK_LOAD_CONFIG`
environment variable.

Assume role credentials support is working, but not yet wired in to any
default credential provider chain. I'd also still like to test edge
cases, as CLI configuration supports many more values than the shared
credential file.


Gemfile
```diff
@@ -51,4 +51,5 @@ end
 
 group :repl do
   gem 'pry'
+  gem 'byebug'
```
trevorrowe: You will want to remove byebug from the Gemfile and gitignore.

aws-sdk-core/lib/aws-sdk-core.rb
```diff
@@ -266,6 +268,9 @@ module Xml
     autoload :Parser, 'aws-sdk-core/xml/parser'
   end
 
+  # @api private
+  SHARED_CONFIG = SharedConfig.new
+
```
trevorrowe: I would suggest replacing the constant with a method that memoizes. We had to make a similar change because of how autoload works. It is possible for a second threat to get a no such constant error under concurrency. The code below mitigates this. It is possible that multiple threads hit the method, but a single object will be memoized for later use.

```ruby
def shared_config
  @shared_config ||= SharedConfig.new
end
``

File: aws-sdk-core/lib/aws-sdk-core/ini_parser.rb [hunk 3]
```diff
@@ -0,0 +1,77 @@
+module Aws
+  class IniParser
+
```
trevorrowe: Given there are three class level methods, I would move them inside the eigenclass:

```ruby
class << self
  def ini_parse(raw)
  end
  # ...
end
```

aws-sdk-core/lib/aws-sdk-core/plugins/regional_endpoint.rb
```diff
@@ -33,6 +34,15 @@ def after_initialize(client)
         end
       end
 
+      private
+      def self.resolve_region(cfg)
+        keys = %w(AWS_REGION AMAZON_REGION AWS_DEFAULT_REGION)
+        env_region = ENV.values_at(*keys).compact.first
+        env_region = nil if env_region == ''
```
trevorrowe: Good catch on the empty ENV region string. I don't believe we check for this previously.

aws-sdk-core/lib/aws-sdk-core/shared_config.rb
```diff
@@ -0,0 +1,179 @@
...
+    attr_reader :profile_name
+
+    # Constructs a new SharedConfig provider object. This will load the shared
+    # credentials file, and optionally the shared configuration file, as ini
+    # files which support profiles.
+    #
+    # By default, only the shared credential file (the default path for which is
+    # `~/.aws/credentials`) is loaded. However, if you set the
+    # `ENV['AWS_SDK_LOAD_CONFIG']` environment variable, the shared config file
+    # will also be loaded (the default path for which is `~/.aws/config`).
```
trevorrowe: I would suggest moving the `ENV['AWS_SDK_LOAD_CONFIG']` completely outside the shared config class. Loading from the shared config should be the default behavior of this class. It could instead accept a boolean at construction time to disable this feature. The code above that constructs the default shared config object could do the environment check. This ensures defaults are preserved and eliminates the need for a process wide environment variable in applications that construct this directly.

aws-sdk-core/lib/aws-sdk-core/shared_config.rb
```diff
@@ -0,0 +1,179 @@
...
+
+    def credentials(opts = {})
+      p = opts[:profile] || @profile_name
+      validate_profile_exists(p) if @parsed_credentials || @parsed_config
+      if credentials = credentials_from_shared(p, opts)
+        credentials
+      elsif credentials = credentials_from_config(p, opts)
+        credentials
+      else
+        nil
```
trevorrowe: Rather than using assignments in the conditionals, I would suggest:

```ruby
credentials = credentials_from_shared(p, opts)
credentials ||= credentials_from_config(p, opts)
credentials
```

trevorrowe: Also, I haven't tested this, but I suspect this will be problematic if the credential keys exist in the document, but they have no values. The `credentials` object will exist, but it will be empty, and we should continue to the next check.

aws-sdk-core/lib/aws-sdk-core/shared_config.rb
```diff
@@ -0,0 +1,179 @@
...
+      end
+    end
+
+    # Currently assumes config is present.
+    def assume_role_credentials_from_config(opts = {})
```
trevorrowe: I understand this may still be in flux, but it does not appear that this is hooked up anywhere. It should probably return an instance of `AssumeRoleCredentials` automatically when the assume role profile is set.

trevorrowe:
This doesn't appear to support any of the other possible shared configuration values from `~/.aws/config` besides region and credentials. These should be released together as we don't want the opt-in flag to initially only add partial support and then later pull additional configuration.

Gemfile
awood45: I've removed it from the Gemfile, I still think it's fine in the `.gitignore` file unless we would intend on tracking some sort of byebug history in the repo at any point.

aws-sdk-core/lib/aws-sdk-core.rb
awood45: Moved to a method, updated tests and usage to reflect this.

aws-sdk-core/lib/aws-sdk-core/ini_parser.rb
awood45: Done. Also removed an unused method from the class (development artifact I neglected to remove).

aws-sdk-core/lib/aws-sdk-core/shared_config.rb
awood45: You're correct, this hasn't yet been hooked up. Next change will hook it up (conditional to the env variable being set) to the default credential provider chain.

aws-sdk-core/lib/aws-sdk-core/shared_config.rb
awood45: This is done with the most recent change.

aws-sdk-core/lib/aws-sdk-core/shared_config.rb
awood45: Last diff doesn't wire this up yet since we're still resolving behavior questions. Will resolve that before merging.

aws-sdk-core/lib/aws-sdk-core/shared_config.rb
awood45: I think that may also be the current behavior. I'm going to add some tests for this specific case so we can handle it properly.

awood45: I've added some specific tests and configuration cases around this behavior. It will now explicitly fall through shared credentials, then to shared config, then to returning `nil` if credentials are incomplete as defined by the `Aws::Credentials#set?` method.

temujin9:
Is there an ETA on this getting finished and merged? We have a large environment (including several ruby tools) which is moving to role assumption for our core architecture.

awood45:
@temujin9 This last commit is the end of the planning/development cycle. Once we finish reviewing and ensuring that the test harness is solid (since this touches a lot of code that's run by just about every single client construction) we're intending to release.

temujin9:
@awood45 Thanks for the follow-up. Can I ask you ping here when it does release?

awood45:
When we release this, you'll definitely see a note here and it will be a minor version bump, for e.g. to `2.4.0`.

aws-sdk-core/lib/aws-sdk-core.rb
```diff
@@ -268,6 +270,12 @@ module Xml
 
   class << self
 
+    # @api private
+    def shared_config
+      enabled = ENV["AWS_SDK_LOAD_CONFIG"] ? true : false
+      @shared_config ||= SharedConfig.new(config_enabled: true)
+    end
+
```
trevorrowe: Minor nitpick. I try to limit the top-level utility functions. A simple change would be to keep this function inside the `Aws::SharedConfig` module. You could name the method something like `.default_config`, e.g `Aws::SharedConfig.default_config`.

aws-sdk-core/lib/aws-sdk-core/credential_provider_chain.rb
```diff
@@ -67,9 +68,28 @@ def shared_credentials(options)
       nil
     end
 
+    def assume_role_credentials(options)
+      if Aws.shared_config.config_enabled?
+        if options[:config]
+          assume_role_with_profile(options[:config].profile, options[:config].region)
+        else
+          assume_role_with_profile(nil)
```
trevorrowe: This method call will fail with an incorrect number of arguments error.  I would suggest something like:

```ruby
def assume_role_credentials(options)
  if Aws.shared_config.config_enabled?
    profile = nil
    region = nil
    if options[:config]
      profile = options[:config].profile
      region = options[:config].region
    end
    assume_role_with_profile(profile, region)
  else
    nil
  end
end
```

trevorrowe:
I finished going though this today. I have some additional feedback from simply a code-organization perspective that should affect functionality. I can go over this with you in person. If you add `@api private` to the `Aws::SharedConfig` class, then I would say we are good to ship (given the feedback above has been addressed).

aws-sdk-core/lib/aws-sdk-core/credential_provider_chain.rb
awood45: This was fixed in the most recent commit.

awood45:
I've marked the `Aws::SharedConfig` class as API private.

trevorrowe:
Ship it!

temujin9:
Shipment received. Thanks, guys.

Issue #700, opened 2015-02-10
[Question] Is there intent to support generation of signed CloudFront URLs?


dandunckelman:
Currently, we use this tool:  https://github.com/dylanvaughn/aws_cf_signer

But I was just curious if this operation is something the SDK would do at some point?

Thanks

trevorrowe:
I haven't given it much consideration to this point. I would be generally in favor of supporting this feature, but I am unable to comment on timeline. I'd have to see where this fits into our priorities.

dandunckelman:
@trevorrowe sounds good!

PR #1194, opened 2016-05-27
Add support for Signed CloudFront URLs

cjyclaire:
This PR adds support for signed URL option for AWS CloudFront. 
[issue](#700), AWS [Doc](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-signed-urls.html).

cc: @trevorrowe @awood45


aws-sdk-core/lib/aws-sdk-core/cloudfront/url_signer.rb
```diff
@@ -0,0 +1,136 @@
...
+        # Example Use:
+        # signer = Aws::CloudFront::UrlSigner.new
+        # url = signer.signed_url(
+        #	"CF_KEYPAIR_ID",
+        #	"/path/to/private/key/cf_private_key.pem"
+        # )
+        #
+        class UrlSigner
+        	# @key_pair_id [String] :ID of the key pair
+            # @private_key_path [String] :Path to the private key used for signing
```
trevorrowe: The yard comments should be formatted like:

```
@param [String] key_pair_id Then documentation goes at the end.
```

aws-sdk-core/lib/aws-sdk-core/cloudfront/url_signer.rb
```diff
@@ -0,0 +1,136 @@
...
+            def initialize(key_pair_id, private_key_path)
+                @key_pair_id = key_pair_id
+                @private_key = File.open(private_key_path, 'rb') { |f| f.read }
+            end
+
+            # create a signed Amazon CloudFront URL
+            # @param [String] URL to sign
+            # @option param [String | Integer] UTC Unix timestamp
+            # @option param [String] JSON Policy
+            def signed_url(url, expires = nil, policy = nil)
```
trevorrowe: I would suggest:

```ruby
# @param [String] url
# @option options [Time, DateTime, Date, String, Integer<timestamp>] :expires 
# @option options [String<JSON>] :policy A JSON policy document.
def signed_url(url, options = {})
  expires = options[:expires]
  policy = options[:policy]
end

File: aws-sdk-core/lib/aws-sdk-core/cloudfront/url_signer.rb [hunk 3]
```diff
@@ -0,0 +1,136 @@
...
+                url_sections = url.split('://')
+                if url_sections.length < 2
+                    raise ArgumentError, "Invaild URL:#{url}"
+                end
+
+                scheme = url_sections[0].gsub('*', '')
```
trevorrowe: We can probably use `URI.parse(string)` here on the url, instead of parsing it ourselves.

aws-sdk-core/lib/aws-sdk-core/cloudfront/url_signer.rb
```diff
@@ -0,0 +1,136 @@
...
+                start_flag = (uri.include? '?') ? '&' : '?'
+
+                uri = "#{uri}#{start_flag}#{signature}"
+                if scheme == 'rtmp'
+                    rtmp_url(URI(uri))
+        		else
+        			uri
+                end
+            end
+
```
trevorrowe: You should add a `private` statement here to keep these methods from being documented as part of the public interface.

aws-sdk-core/lib/aws-sdk-core/cloudfront/url_signer.rb
```diff
@@ -0,0 +1,136 @@
...
+            #    signing with a canned policy. Not required when passing
+            #    a custom policy.
+            # @option param [String] JSON policy. Use this option when
+            #    creating a signature for a custom policy.
+            def signature(resource = nil, expires = nil, policy = nil)
+                signature_content = []
+                if !policy.nil?
+                    policy = policy.gsub('/\s/s', '')
+                    signature_content << "Policy=#{encode(policy)}"
+                elsif !resource.nil? && !expires.nil?
```
trevorrowe: In Ruby, only `nil` and `false` evaluate to `false` in a boolean context. You can remove the not nil checks here.

```ruby
elsif resource && expires
```

aws-sdk-core/spec/aws/cloud_front/url_signer_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+        describe UrlSigner do
+
+            let(:key_pair_id) { "CF_KEYPAIR_ID" }
+            let(:private_key_path) { "#{File.dirname(__FILE__)}/cf_private_key.pem" }
+            let(:signer) { UrlSigner.new(key_pair_id, private_key_path) }
+        	let(:expires) { 1357034400 } # January 1, 2013 10:00 am UTC (Unix timestamp)
+
+            describe '#initialize' do
+                it 'takes keypair id and privatekey path' do
+                    expect(signer.class).to eq(UrlSigner)
```
trevorrowe: Inline your major assertion.

```ruby
signer = UrlSigner.new(key_pair_id, private_key_path)
expect(signer).to ...
```

Try to not hide the main code under test in a let expression.

aws-sdk-core/lib/aws-sdk-core/cloudfront/url_signer.rb
```diff
@@ -0,0 +1,136 @@
...
+        # signer = Aws::CloudFront::UrlSigner.new
+        # url = signer.signed_url(
+        #	"CF_KEYPAIR_ID",
+        #	"/path/to/private/key/cf_private_key.pem"
+        # )
+        #
+        class UrlSigner
+        	# @key_pair_id [String] :ID of the key pair
+            # @private_key_path [String] :Path to the private key used for signing
+            def initialize(key_pair_id, private_key_path)
```
trevorrowe: We should add options here:

```ruby
# @param [String] private_key_id
# @option options [String] :private_key_pair
# @option options [String] :private_key_path
def intilaize(key_pair_id, options = {})
  private_key = options[:private_key]
  # if both :private_key and :private_key_path are unset we should raise an argument error
  private_key ||= File.open(options[:private_key_path], 'rb') { |f| f.read }
end
```

cjyclaire:
Fix updated. 
Note: URI.parse() cannot be applied because CloudFront URLs need to handle wildcard characters`*` while URI module doesn't.

Issue #991, opened 2015-11-10
Param checker could be friendlier


vemv:
What it prints:

`expected params[:pipeline_objects][1][:fields][5][:string_value] to be a string`

What it should print:

`expected params[:pipeline_objects][1][:fields][5][:string_value] to be a string, got instead the value 2 (class: Fixnum)`

Cheers - Victor

awood45:
That sounds like a good idea, I'll see if there's any reason not to do that. If you'd like, feel free to open a pull request for this as well.

awood45:
I've added this to the feature request list, let me know if you have any other questions or ideas!

PR #1199, opened 2016-06-01
Add object value & class in param checker

cjyclaire:
[Add to Param Validation Error Message](https://github.com/aws/aws-sdk-ruby/blob/master/FEATURE_REQUESTS.md#add-to-param-validation-error-message), feature request from customer issue #991 

cc: @awood45 @trevorrowe

### DIFF
--- a/aws-sdk-core/lib/aws-sdk-core/param_validator.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/param_validator.rb
@@ -60,7 +60,7 @@ module Aws
     def list(ref, values, errors, context)
       # ensure the value is an array
       unless values.is_a?(Array)
-        errors << "expected #{context} to be an array"
+        errors << "expected #{context} to be an array, got value #{values} (class: #{values.class}) instead."
         return
       end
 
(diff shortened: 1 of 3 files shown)


aws-sdk-core/lib/aws-sdk-core/param_validator.rb
```diff
@@ -60,7 +60,7 @@ def structure(ref, values, errors, context)
     def list(ref, values, errors, context)
       # ensure the value is an array
       unless values.is_a?(Array)
-        errors << "expected #{context} to be an array"
+        errors << "expected #{context} to be an array, got value #{values} (class: #{values.class}) instead."
```
trevorrowe: I would suggest DRY'ing up the code by having a single re-usable string for the got value (class) instead messaging. You can create a const string at the top of the class like:

```ruby
GOT_INSTEAD = "got value %s (class: %s) instead."
```

Then you can simply append the following to each error message

```ruby
GOT_INSTEAD % [values.inspect, values.class.name]
```

trevorrowe: Also note, we should use the `#inspect` of the value for each error message instead of the default `#to_s` that results from passing the raw value through string interpolation. This makes for a better error message.

trevorrowe:
Great addition! I added a few small comments.

cjyclaire:
Good point, make sense!

PR #1204, opened 2016-06-08
Adding support for AES/GCM/NoPadding - take 2

trevorrowe:
This pull request merges in https://github.com/aws/aws-sdk-ruby/pull/1043 and changes how the encrypted stream is handled on download. When an auth tag is appended to the body cipher text as a trailer, it is now fetched via a second GET Object with/range header request. This is used to initialize the decryption cipher upfront. The IO Decrypter is then wrapped so the body stream stops reading bytes off the body when it gets to the trailing auth tag in the initial response.

See related pull request:
https://github.com/aws/aws-sdk-ruby/pull/1043


CHANGELOG.md
```diff
@@ -1,6 +1,30 @@
...
+  encrypted with AES/GCM/NoPadding with a trailing authentication tag.
+  This makes it possible to objects that have been encrypted from
+  Amazon SES inbound.
+
+  ```ruby
+  # you must use a KMS client that is the same region as the kms key id
+  kms = Aws::KMS::Client.new(region: 'us-west-2')
+
+  s3_enc = Aws::S3::Encryption::Client.new(
+    kms_key_id: 'arn:aws:kms:us-west-2:469596866844:alias/aws/ses',
```
awood45: Nitpick: Personally prefer non-realistic looking account IDs, such as `123456789012`.

aws-sdk-resources/lib/aws-sdk-resources/services/s3/encryption/decrypt_handler.rb
```diff
@@ -31,9 +36,41 @@ def call(context)
...
-            context.http_response.body = IODecrypter.new(cipher, body)
+            cipher = decryption_cipher(context)
+
+            if http_resp.headers['x-amz-meta-x-amz-tag-len']
+
+              content_length = http_resp.headers['content-length'].to_i
+              auth_tag_length = http_resp.headers['x-amz-meta-x-amz-tag-len']
+              auth_tag_length = auth_tag_length.to_i / 8
+
+              auth_tag = context.client.get_object(
```
awood45: I know currently that we're doing this ranged get call due to the nature of GCM decryption, but a comment explaining why we're doing this (and why we're doing this instead of double-reading the object) would be useful for 1-2 years down the road.

aws-sdk-resources/lib/aws-sdk-resources/services/s3/encryption/io_auth_decrypter.rb
```diff
@@ -0,0 +1,52 @@
...
+      # @api private
+      class IOAuthDecrypter
+
+        # @option options [required, IO#write] :io
+        #   An IO-like object that responds to {#write}.
+        # @option options [required, Integer] :encrypted_content_length
+        #   The number of bytes to decrypt from the `:io` object.
+        #   This should be the total size of `:io` minus the length of
+        #   the cipher auth tag.
+        # @option options [required, OpenSSL::Cipher] :cipher
```
awood45: Can we document these two options, even if very succinctly? Very helpful down the line when debugging and needed to reload context.

aws-sdk-resources/lib/aws-sdk-resources/services/s3/encryption/io_decrypter.rb
```diff
@@ -4,19 +4,20 @@ module Encryption
...
-        # @param [OpenSSL::Cipher] cipher
-        # @param [#write] io An IO-like object that responds to {#write}.
-        def initialize(cipher, io)
-          @orig_cipher = cipher.clone
-          @cipher = cipher.clone
-          @io = io
+        # @option options [required, #write] :io
+        #   An IO-like object that responds to {#write}.
+        # @option options [required, OpenSSL::Cipher] :cipher
+        def initialize(options = {})
```
awood45: By previous convention, we may want to consider releasing this feature as `2.4.0`. Arguably minor enough to not bother, but it's technically a breaking change of a private API.

aws-sdk-resources/lib/aws-sdk-resources/services/s3/encryption/io_decrypter.rb
```diff
@@ -31,6 +32,53 @@ def reset_cipher
         end
 
       end
+
+      # @api private
+      class IOAuthDecrypter
```
awood45: Why is this defined twice? Copy error?

aws-sdk-resources/lib/aws-sdk-resources/services/s3/encryption/kms_cipher_provider.rb
```diff
@@ -41,7 +41,18 @@ def decryption_cipher(envelope)
             encryption_context: encryption_context,
           ).plaintext
           iv = decode64(envelope['x-amz-iv'])
-          Utils.aes_decryption_cipher(:CBC, key, iv)
+          type = envelope['x-amz-cek-alg'] || false
+          block_mode = :CBC #default to CBC
```
awood45: Is it possible for this to end up undefined, given the case switch below?

awood45:
Mostly minor comments, plus the in-progress resolution of test failure causes.

aws-sdk-resources/lib/aws-sdk-resources/services/s3/encryption/io_decrypter.rb
trevorrowe: Yup. I wrote the code original in the one file for convience and failed to remove it when I added it to its own file. I'll clean that up.

trevorrowe:
Merged into master here: c3376f02c0490273674a37879a688438326bf84a

Issue #990, opened 2015-11-09
How to download a big file from s3 ? 


ahmad-alkheat:
Currently ```s3.get_object``` is freezing and not returning anything because the file is too big, is there a way to go around this? 

Thanks

awood45:
You can work around this by making ranged requests to the large object and reassembling the parts.

Here's one example:

```ruby
client = Aws::S3::Client.new
part = "bytes=0-4999"
resp = client.get_object(bucket: bucket, key: key, range: part)
```

This will get 5000 bytes of the object, and you could do this (with larger blocks) sequentially for the full size of the object.

That's the workaround, I'm also working on putting something like this together as an SDK feature that compliments multipart uploads.

ahmad-alkheat:
@awood45  Thanks a bunch, that makes sense.

Issue #987, opened 2015-11-05
Unable to download entire rds log file with download_db_log_file_portion


wr0ngway:
I'm using aws-sdk 2.1.35 to try and download a postgresql rds log file

When using either the block based pagination, or doing it manually, I'm unable to download the entire rds log file.  The entire log file is 260MB, but either method only gets 100MB.  The final paged request shows additional_data_pending=false, marker ="17:105920837" for either method.  I am able to download the entire file using the web console.  The code I'm using is shown below.

 Is there some other way to download the entire log file?  That is, a way to get a handle on the signed url the console generates to download the entire thing directly would be great.

          manual_pagination = true
          open(path, 'wb+') do |f|

            if manual_pagination
              opts = {
                db_instance_identifier: db_id,
                log_file_name: file,
                marker: "0"
              }
              additional_data_pending = true
              while additional_data_pending do
                print "."
                out = rds.download_db_log_file_portion(opts)
                f.write(out[:log_file_data])
                opts[:marker] = out[:marker]
                additional_data_pending = out[:additional_data_pending]
              end
              puts "."
            else
              rds.download_db_log_file_portion(db_instance_identifier: db_id,
                                               log_file_name: file, marker: "0").each do |r|
                print "."
                f.write r[:log_file_data]
              end
              puts "."
            end

wr0ngway:
If I manually start a new request with the marker returned from the last request, the download proceeds, but eventually fails again in a similar way.  I am thus able to get all the way to the end of the file with some retry logic.  I noticed that the size of the returned data is 1048576 bytes when additional_data_pending=false, but 1048611 when true, so I'm guessing something about the data or how it's processed is causing the abort.

wr0ngway:
Here is a work around that works:

          open(path, 'wb+') do |f|

            opts = {
              db_instance_identifier: db_id,
              log_file_name: file,
              marker: "0"
            }

            while true do
              print "."
              out = rds.download_db_log_file_portion(opts)
              f.write(out[:log_file_data])

              break if opts[:marker] == out[:marker]
              opts[:marker] = out[:marker]
            end
            puts "."

          end

awood45:
That makes sense, this looks essentially like multipart download functionality.

Looking at this as a feature request now.

awood45:
Closing this issue, I've added the feature request and it looks like we have a workaround in the meantime (the feature would just be a high level wrapper around this same workaround).

Feel free to reach out if you have any questions or further issues.

PR #1213, opened 2016-06-14
Add Multipart Download Helper

cjyclaire:
Feature [request](https://github.com/aws/aws-sdk-ruby/blob/master/FEATURE_REQUESTS.md#add-multipart-download-helpers), Github issue #987 #990 

cc:\ @trevorrowe @awood45


cjyclaire:
Since this feature has been unblocked, PR is updated and ready for another review.
@trevorrowe @awood45 

Noted that the build failure is irrelevant to this PR.

ms-ati:
Any progress on this since last year?

eredi93:
+1

aws-sdk-resources/lib/aws-sdk-resources/services/s3/file_downloader.rb
```diff
@@ -0,0 +1,134 @@
...
+          cnt.times do
+            thread = Thread.new(chunks.shift) do |chunk|
+              @client.get_object(bucket: @bucket, key: @key, range: chunk)
+            end
+            threads << thread.value
+          end
+          count += cnt
+        end
+
+        File.open(@path, 'wb') do |f|
```
awood45: Curious: Why not have each thread's `#get_object` call write directly to a file? It seems like we're going to be keeping every single chunk in memory, which could be problematic.

aws-sdk-resources/lib/aws-sdk-resources/services/s3/file_downloader.rb
```diff
@@ -0,0 +1,134 @@
...
+        parts.times do |i|
+        # partNumber starts from 1
+          part = i + 1
+          thread = Thread.new(part) do
+            @client.get_object(bucket: @bucket, key: @key, part_number: part)
+          end
+          threads << thread.value
+        end
+
+        File.open(@path, 'wb') do |f|
```
awood45: Same comment here - parallel writing seems to be the best approach, no?

cjyclaire:
@awood45 Good suggestions, buffering everything into memory is definitely bad. I made changes and clean up the code, ready for a second review now : ).

aws-sdk-resources/lib/aws-sdk-resources/services/s3/file_downloader.rb
```diff
@@ -0,0 +1,135 @@
...
+          remains_count = chunks.size - count
+          cnt = @thread_count < remains_count ? @thread_count : remains_count
+          cnt.times do
+            thread = Thread.new(chunks.shift) do |chunk|
+              resp = @client.get_object(bucket: @bucket, key: @key, range: chunk)
+              File.open(@path, 'ab') do |f|
+                f.write(resp.body.read)
+              end
+            end
+            threads << thread.join
```
mayanks: Will this not write the chunks out of order?

cjyclaire:
@awood45 I updated this PR with multithreading fix, ready for another review : )

aws-sdk-resources/features/s3/download_file.feature
```diff
@@ -0,0 +1,57 @@
...
+    Then 0 head_object request should have been made
+    And 1 get_object request should have been made
+    And the downloaded file should match the uploaded file
+
+  # It is possible for the user to control the chunk size of each
+  # part downloaded, by specifing a mode "get-range" and also
+  # an option to define how large each chunk is.
+  @mode
+  Scenario: Download a large object using part numbers
+    Given I have a 16M file
```
awood45: Now that we are creating large files, I'm a bit concerned that we might be leaking tempfiles with how our tests are written. We should add explicit cleanup steps and a shared tag for these tests like `@large-file` or similar.

aws-sdk-resources/features/s3/download_file.feature
```diff
@@ -0,0 +1,57 @@
...
+
+  # "larger" objects (exceeding 15MB) are downloaded in chunks,
+  # using RANGE requests, even if the it was not uploaded in parts
+  @get-range
+  Scenario: Download a large object using ranges
+    Given I have a 16M file
+    And I upload the file using put_object
+    When I download the file with mode "auto"
+    Then 4 get_object requests should have been made
+    And the downloaded file should match the uploaded file
```
awood45: Could we also integration test our cleanup of temp files? That would be a bad regression if it ever happened.

aws-sdk-resources/lib/aws-sdk-resources/services/s3/file_downloader.rb
```diff
@@ -0,0 +1,163 @@
...
+
+      def sort_files(files)
+        # sort file by start range count or part number
+        files.sort do |a, b|
+          a[/([^\=]+)$/].split('-')[0].to_i <=> b[/([^\=]+)$/].split('-')[0].to_i
+        end
+      end
+
+      def concatenate_files(files)
+        File.open(@path, 'w', encoding: 'UTF-8') do |obj|
```
awood45: We could do a bit of refactoring here for clarity. Let's rename `obj` in the block to make it more clear it's the output file, something like `output_file`. Similarly, the inner `file` could become `part_file`. This will make it more clear what we are doing.

aws-sdk-resources/lib/aws-sdk-resources/services/s3/file_downloader.rb
```diff
@@ -0,0 +1,163 @@
...
+          sort_files(files).each do |file|
+            File.open(file, 'r', encoding: 'UTF-8') do |f|
+              obj.write(f.read)
+            end
+            File.unlink(file)
```
awood45: I'm concerned about error handling in this case. We may want to put this method's logic in a private inner method, and wrap it in a begin/ensure loop. For example:

```ruby
def concatenate_files(fileparts)
  begin
    concatenate_parts(fileparts) # put the existing method logic in a private method here
  ensure
    fileparts.each do |part|
      File.unlink(part) if File.exist?(part)
    end
  end
end
```

awood45: The thought here is that if we have some sort of File issue partway through the write, we're going to leak temp files. The `ensure` step performs a cleanup of remaining part files.

aws-sdk-resources/lib/aws-sdk-resources/services/s3/file_downloader.rb
```diff
@@ -0,0 +1,163 @@
...
+          end
+        end
+      end
+
+      def get_range(chunks)
+        count = 0
+        threads = []
+        file_chunks = []
+        dir = Dir.tmpdir
+        while count < chunks.size
```
awood45: I'd like to consider refactoring this, as the logic feels a bit fragile. Could we, for example, create the chunks and sort the chunks into thread batches ahead of time, then process the thread batches one at a time? I'm worried that a later logic mistake makes it really easy to hit an infinite loop here.

awood45: Happy to pair on this as well.

aws-sdk-resources/lib/aws-sdk-resources/services/s3/file_downloader.rb
```diff
@@ -0,0 +1,163 @@
...
+      def get_part(parts)
+        threads = []
+        file_parts = []
+        dir = Dir.tmpdir
+        parts.times do |i|
+        # partNumber starts from 1
+          part = i + 1
+          threads << Thread.new(part) do
+            resp = @client.get_object(bucket: @bucket, key: @key, part_number: part)
+            file_parts << file = File.join(dir, 'part_number=' + part.to_s)
```
awood45: The `=` character is considered something to avoid in a filename. Let's change our naming scheme (possibly even in a private method) to something like: `targetfilename.txt_part_0013` and so on. The other advantage is this is string sortable (reducing complexity elsewhere) when done right. We can also pair on this.

aws-sdk-resources/features/s3/download_file.feature
cjyclaire: That's fair, tests added : )

aws-sdk-resources/features/s3/download_file.feature
cjyclaire: Added.

aws-sdk-resources/lib/aws-sdk-resources/services/s3/file_downloader.rb
cjyclaire: Make sense, renaming done.

aws-sdk-resources/lib/aws-sdk-resources/services/s3/file_downloader.rb
cjyclaire: Fixed and refactored in the latest commit.

aws-sdk-resources/lib/aws-sdk-resources/services/s3/file_downloader.rb
cjyclaire: Sounds good, fixed and refactored in the latest commit. Happy for offline discussion if I still miss something  : )

aws-sdk-resources/lib/aws-sdk-resources/services/s3/file_downloader.rb
cjyclaire: Good to know, Initially, I use `=` because for ranges, it will look like `bytes=12345-56789`, thus I use `part_number=` prefix to echo that, sounds like it might not be a good option there. Happy to discuss a better naming strategy : )

P.S. `sort_files` is introduced because I noticed that original Ruby string compare is having trouble in comparing strings like "12345-56789" in a expected numerical way.

aws-sdk-resources/lib/aws-sdk-resources/services/s3/file_downloader.rb
```diff
@@ -0,0 +1,175 @@
...
+        batches.each do |batch|
+          threads = []
+          batch.each do |chunk, file|
+            threads << Thread.new do
+              resp = @client.get_object(
+                :bucket => @bucket,
+                :key => @key,
+                param.to_sym => chunk
+              )
+              File.open(file, 'w', encoding: 'UTF-8') {|f| f.write(resp.body.read)}
```
janko-m: Here you can use the `:response_target` option with a path, which will automatically download the file in chunks, and also use the (correct) binary format.

```rb
@client.get_object(
  :bucket => @bucket,
  :key => @key,
  param.to_sym => chunk,
  :response_target => file
)
```

cjyclaire: That's correct, thanks for catching! Will do that :)

aws-sdk-resources/lib/aws-sdk-resources/services/s3/file_downloader.rb
```diff
@@ -0,0 +1,173 @@
...
+      def map_files(slice, dir, mode)
+        case mode
+        when 'range'
+          slice.inject({}) {|h, chunk| h[chunk] = File.join(dir, chunk); h}
+        when 'part_number'
+          slice.inject({}) {|h, part| h[part] = File.join(dir, "part_number=#{part}"); h}
+        end
+      end
+
+      def get_range(chunks)
```
awood45: I would consider renaming `#get_range` and `#get_part` as these function names are confusing elsewhere in code given what they do. Perhaps:

* `#multithreaded_get_by_ranges`
* `#multithreaded_get_by_parts`

That way, when you run in to these method names, you understand what's going on, especially with `#get_part`

aws-sdk-resources/lib/aws-sdk-resources/services/s3/file_downloader.rb
```diff
@@ -0,0 +1,173 @@
...
+        thread_batches(chunks, 'range')
+      end
+
+      def get_part(parts)
+        thread_batches(parts, 'part_number')
+      end
+
+      def thread_batches(chunks, param)
```
awood45: This needs to have some safety checks added. As it appears to operate now, if you get an exception during processing, you risk having orphaned files because the exception from the client will bubble up as soon as we join a failed thread.

Since this method creates the intermediate files, it should be responsible for cleaning them all up before completing.

aws-sdk-resources/lib/aws-sdk-resources/services/s3/file_downloader.rb
```diff
@@ -135,31 +133,44 @@ def map_files(slice, dir, mode)
...
+                  :key => @key,
+                  param.to_sym => chunk,
+                  :response_target => file
+                )
+              rescue => error
+                # clear file parts once failed
+                clean_up_parts(parts)
+                raise error
+              end
+              thread.abort_on_exception = true
```
janko-m: Good catch on setting `Thread#abort_on_exception`, I think this `ObjectMultipartCopier` and `MultipartFileUploader` should also be updated to set `Thread#abort_on_exception` (otherwise the thread won't propagate the error, and nothing will be cleaned up).

Btw, you probably need to set `Thread#abort_on_exception` *before* the exception is raised (at the top of the thread definition), otherwise that part of code will never get executed in case an exception is raised. While here, instead of using a `thread` local variable, you could also use `Thread.current` to access the current thread:

```rb
threads << Thread.new do
  Thread.current.abort_on_exception = true

  begin
    # execute
  rescue
    # clean-up
  end
end
```

aws-sdk-resources/lib/aws-sdk-resources/services/s3/file_downloader.rb
```diff
@@ -135,31 +133,44 @@ def map_files(slice, dir, mode)
...
       end
 
-      def get_part(parts)
+      def multithreaded_get_by_parts(parts)
         thread_batches(parts, 'part_number')
       end
 
       def thread_batches(chunks, param)
         batches = file_batches(chunks, param)
+        parts = batches.inject([]) {|a, batch| a.push(*batch.values); a}
```
janko-m: `Array#push` returns the modified receiver, so this could be just

```rb
parts = batches.inject([]) {|a, batch| a.push(*batch.values) }
```

Even better, you could use `Array#flat_map`:

```rb
parts = batch.flat_map(&:values)
```

aws-sdk-resources/lib/aws-sdk-resources/services/s3/file_downloader.rb
```diff
@@ -0,0 +1,173 @@
...
+          msg = "Invalid mode #{@mode} provided, "\
+            "mode should be :single_request, :get_range or :auto"
+          raise ArgumentError, msg
+        end
+      end
+      
+      private
+
+      def multipart_download
+        resp = @client.head_object(bucket: @bucket, key: @key, part_number: 1)
```
cjyclaire: @janko-m Great catch! I did some simple test and you are right about this. I started this PR before the partNumber is released, looks like the actually API differs a bit, anyways, thank you so much for catching this! I'll  fix this and add another cucumber test to make sure that scenario is taken care of properly : )

aws-sdk-resources/lib/aws-sdk-resources/services/s3/file_downloader.rb
cjyclaire: Make sense, thanks for the advice, will address those : )

aws-sdk-resources/lib/aws-sdk-resources/services/s3/file_downloader.rb
cjyclaire: Ah that's right, I'll fix that

aws-sdk-resources/lib/aws-sdk-resources/services/s3/file_downloader.rb
```diff
@@ -135,31 +133,43 @@ def map_files(slice, dir, mode)
...
+                resp = @client.get_object(
+                  :bucket => @bucket,
+                  :key => @key,
+                  param.to_sym => chunk,
+                  :response_target => file
+                )
+              end
             end
+            threads.each(&:join)
+          rescue => error
```
awood45: Given that we're rescuing all of StandardError, perhaps we could simplify this to be an `ensure` block, which means we only ever need to call `#clean_up_parts` here whether the method succeeds or fails. It also eliminates the need to catch only to re-raise the same exception.

aws-sdk-resources/lib/aws-sdk-resources/services/s3/file_downloader.rb
```diff
@@ -135,31 +133,43 @@ def map_files(slice, dir, mode)
...
+            # clear file parts once failed
+            clean_up_parts(parts)
+            raise error
           end
-          threads.each(&:join)
         end
-        parts = batches.inject([]) {|a, batch| a.push(*batch.values); a}
         concatenate_files(parts)
```
awood45: Lift this in to the begin/ensure block.

aws-sdk-resources/lib/aws-sdk-resources/services/s3/file_downloader.rb
```diff
@@ -135,31 +133,43 @@ def map_files(slice, dir, mode)
         end
       end
 
-      def get_range(chunks)
+      def clean_up_parts(parts)
```
awood45: Make sure this can handle a `nil` value for `parts`, since an ensure block is going to call this method.

aws-sdk-resources/lib/aws-sdk-resources/services/s3/file_downloader.rb
```diff
@@ -104,9 +104,7 @@ def concatenate_files(fileparts)
         begin
           concatenate_parts(fileparts)
         ensure
-          fileparts.each do |part|
-            File.unlink(part) if File.exists?(part)
-          end
+          clean_up_parts(fileparts)
```
awood45: Not needed here if this is run within the outer begin/ensure block which handles cleanup.

cjyclaire:
Updated, ready for another review ; )

PR #1214, opened 2016-06-14
Fix class organization headings in docs

kddeisz:
The SDK is doing some organization to the API docs page (http://docs.aws.amazon.com/sdkforruby/api/) class list. Clicking on the "Non-Service Classes" or "Services" headings results in being redirected to an "undefined" page because YARD is adding links to the lis programatically. This PR adds a small JS file to remove the linking from the heading lis.


trevorrowe:
Thanks for the fix!

Issue #1184, opened 2016-05-12
Aws::S3 unable to parse "Expires" header


hosh:
When I tried doing a `head_object`, I get the following error:

```
irb(main):003:0* s3 = Aws::S3::Client.new(region: 'us-east-1')
=> #<Aws::S3::Client>
irb(main):004:0> s3.head_object(bucket: bucket, key: key)
ArgumentError: no time information in "5d"
        from /home/deployer/.rbenv/versions/2.2.4/lib/ruby/2.2.0/time.rb:252:in `make_time'
        from /home/deployer/.rbenv/versions/2.2.4/lib/ruby/2.2.0/time.rb:364:in `parse'
        from /home/deployer/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/aws-sdk-core-2.3.3/lib/aws-sdk-core/rest/response/headers.rb:42:in `cast_value'
        from /home/deployer/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/aws-sdk-core-2.3.3/lib/aws-sdk-core/rest/response/headers.rb:27:in `extract_header_value'
```

Just to make sure, I added a debug line:

```
DEBUG: #<Seahorse::Model::Shapes::ShapeRef:0x007fce41fc1b80 @metadata={}, @required=false, @deprecated=false, @shape=#<Seahorse::Model::Shapes::TimestampShape:0x007fce42212dd0 @metadata={"type"=>"timestam
p"}, @name="Expires", @documentation=nil>, @location="header", @location_name="Expires", @documentation=nil> "5d"
```

Looking at the code, https://github.com/aws/aws-sdk-ruby/blob/master/aws-sdk-core/lib/aws-sdk-core/rest/response/headers.rb#L41 it looks like it is expecting a well-formatted timestamp.

Just to be sure, I tried this with the `aws-cli` tool:

```
$ aws s3api head-object --bucket $BUCKET --key $KEY
{
    "AcceptRanges": "bytes",
    "ContentType": "image/png",
    "LastModified": "Fri, 04 Mar 2016 02:06:23 GMT",
    "ContentLength": 78616,
    "Expires": "5d",
    "ETag": "\"0a937f9c6deab66e91a87e0741cb42df\"",
    "StorageClass": "REDUCED_REDUNDANCY",
    "CacheControl": "public",
    "Metadata": {}
}
```

The aws cli handles this fine. The V1 handles this fine.

[...]

PR #1216, opened 2016-06-16
Fix ArgumentError when S3 responds with invalid Expires Header

trevorrowe:
Resolved an issue with `Aws::S3::Client#head_object` and `#get_object` where an `ArgumentError` was raised if Amazon S3 responded with an Expires header that contained an unparsable string.

The `#head_object` and `#get_object` response now return `nil` when the Expires header contains an invalid value. You can now access the raw string value of the Expires header with `#expires_string`.
```ruby
# If Amazon S3 responds with `Expires: abc` as a header
resp = s3.head_object(bucket:'bucket', key:'key')
resp.expires #=> nil
resp.expires_string #=> "abc"
```

See related [GitHub issue #1184](https://github.com/aws/aws-sdk-ruby/issues/1184).

### DIFF
--- a/aws-sdk-core/lib/aws-sdk-core/api/customizations.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/api/customizations.rb
@@ -157,6 +157,24 @@ module Aws
             end
           end
         end
+
+        api['shapes']['ExpiresString'] = { 'type' => 'string' }
+        %w(HeadObjectOutput GetObjectOutput).each do |shape|
+          members = api['shapes'][shape]['members']
+          # inject this into sort order so this appears directly
+          # after #expires
+          api['shapes'][shape]['members'] = members.inject({}) do |h, (k,v)|
+            if k == 'Expires'
+              h['ExpiresString'] = {
+                'shape' => 'ExpiresString',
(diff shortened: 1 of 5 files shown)


aws-sdk-core/lib/aws-sdk-core/api/customizations.rb
```diff
@@ -157,6 +157,24 @@ def apply_plugins(client_class)
             end
           end
         end
+
+        api['shapes']['ExpiresString'] = { 'type' => 'string' }
+        %w(HeadObjectOutput GetObjectOutput).each do |shape|
+          members = api['shapes'][shape]['members']
+          # inject this into sort order so this appears directly
+          # after #expires
```
awood45: Technically this will make it show up BEFORE 'Expires' in my testing, not after.

awood45:
There's a mismatch in the comments about where `ExpiresString` goes in the hash order and where it is actually put. I'm not really sure why it matters though, at least in the context of the diff.

trevorrowe:
I've confirmed that the order does not matter, but I've pushed a commit that makes sure they line up.

awood45:
LGTM then

Issue #1051, opened 2016-01-09
Why is cache_control header blacklisted for generation of signed URLs?


toaster:
Hi there,

for #1026 a list of headers were excluded from signature for various reasons.
One of them (`cache_control`) is used by me in presigned URLs and I want to rely upon it to ensure that uploaded objects have the proper cache control settings.
For me this is a blocker for upgrading beyond 2.2.5. Of course I could monkey patch the signer but that is not really what I want to do.
If it is really necessary to have this header excluded from signature it may be a solution to make its exclusion optional.

awood45:
So, if I understand you correctly, you're relying on the signature to show that the `cache-control` header is not modified by your customer?

In any case, I'll re-check that it needs to be on the blacklist, although our direction is moving towards a whitelist of headers in the future, potentially.

I think there's a feature request here to whitelist headers for signing. I think it's fair for you to say: "Sign this header, I understand that I'm now responsible for ensuring it is not modified in transit to the server, and that there will be an error if it is."

toaster:
Yes, you understood me correctly. And yeah, that whitelist would be totally okay for me.

awood45:
Adding to feature request backlog. Will also take a PR for this if you want to take a crack at it.

toaster:
Hi there!

Due to this issue we are still stuck at SDK version 2.2.5. We now want to use the upload acceleration which is not supported by 2.2.5 AFAICS.
Is there any chance that #1228 will be merged any time soon?

awood45:
I've added review notes to that PR, and we'll work on getting that in.

toaster:
Hi again,

we worked around our upload acceleration issue with 2.2.5 by simply patching the URL.
But now we want to use signed CloudFront URLs. But, alas, the `Aws::CloudFront::UrlSigner` does not exist in 2.2.5.
This problem has now evolved from “annoying” to “blocker” for us.
When will the PR get merged?

PR #1228, opened 2016-06-27
Add option for :whitelist_headers in V4 signer

cjyclaire:
Feature [request](https://github.com/aws/aws-sdk-ruby/blob/master/FEATURE_REQUESTS.md#add-header-whitelist-for-request-signing)
Issue related: #1051 

cc:\ @awood45 @trevorrowe


aws-sdk-core/lib/aws-sdk-core/signers/v4.rb
```diff
@@ -39,10 +39,16 @@ def self.sign(context)
       #   the endpoint prefix.
       # @param [String] region The region (e.g. 'us-west-1') the request
       #   will be made to.
-      def initialize(credentials, service_name, region)
+      # @param [Array] whitelist headers provided to escape blacklist check
+      def initialize(credentials, service_name, region, whitelist_headers = [])
         @service_name = service_name
         @credentials = credentials.credentials
         @region = EndpointProvider.signing_region(region, service_name)
+        if whitelist_headers.any? && (BLACKLIST_HEADERS & whitelist_headers).any?
```
awood45: I might be missing something, but why not just remove this whole block and use:
```ruby
@blacklist = BLACKLIST_HEADERS - whitelist_headers
```
You already set whitelist headers to be an array by default, so once we get to this point, I'm not sure the if-else block is doing anything.

awood45:
The inner implementation appears to be reasonable, but how would an end user use this? I don't believe we have users creating signer classes directly, so we would need to wire this up to configuration in some way.

aws-sdk-core/lib/aws-sdk-core/signers/v4.rb
cjyclaire: That's true, I was thinking checking the intersection of  the 2 array before getting the result. It seems redundant, I'll remove that.

awood45:
Some other notes after a deeper dive:

- In the `Aws::Signers::V4` class, this will need to work in the `sign` method, and the whitelisted headers will have to plumb down to the `signed_headers` method.
- This does not necessarily involve the instantiation of a new signer instance, in fact, it usually does not.
- You may be changing how the internal interface works, so after we get this done we may want to look at if this is a minor version bump.

cjyclaire:
This PR is updated and ready for another review.
@trevorrowe , @awood45

trevorrowe:
I've done a quick review of this. I think we should discuss the method in which white-listed headers are provided. I'm of the opinion that this should not be done as client configuration options. Given the client will never consume these options, they increase the surface area of a client just to act as a convenient bag for grouping options. I suspect this was suggested because a `Client` instance is passed to the pre-signer.  Retrospectively, the `Aws::S3::Presigner` should not have ever accepted a client instance. Instead it should have accepted a credential provider and a region. We can not change that interface now, but we can deprecate it. Lets do a more in-depth review in person and we can update the pull-request.

cjyclaire:
This PR is updated and ready for another review @awood45 @trevorrowe

cjyclaire:
@awood45 Sounds good, make sense. I'll add that : )

cjyclaire:
PR #1477 just opened includes this feature, we'll focus on that one : )
Closing, feel free to chime in and add comments in that PR : )

PR #1251, opened 2016-07-21
OpenSSL::Digest::MD5.file cannot be used on filehandles whose links have been removed from the filesystem.

### LINKED ISSUE #1232: jRuby StackOverflowError in versions above 2.3.14
I have an integration test that started throwing a StackOverflowError in jRuby when I tried to upgrade from 2.3.14 to 2.3.19. I tested the versions between and discovered that it’s caused by a change included in 2.3.15.

Because the StackOverflowError blows away the actual calling stack, it’s hard for me to speculate on what might be causing it, but I am happy to dig into it if needed.

If you have trouble reproducing it, I could probably git bisect and get it down to a particular commit.

I’m running:
jruby-9.1.2.0
rspec-3.5.0
(and Rails-5.0, but I don’t believe that’s relevant)
I can provide the Gemfile.lock if that would help.

The (slightly simplified) spec looks like this:

```ruby
RSpec.describe MyController, type: :controller do
  let(:s3) { Aws::S3::Resource.new }
  let(:s3path) { 'a/random/s3/test_file.csv' }

  before do
    Aws.config[:s3] = {
      stub_responses: {
        list_objects: { contents: [{ key: s3path }] },
        get_object: { body: File.read(Rails.root.join('spec/fixtures/signature_data.txt')) },
        # TODO: Figure out if there is a way to do this
        put_object: { }
      }
    }

    allow(RedshiftExporter::CONNECTION_POOL).to receive(:with)

    # TODO: Figure out the prescribed way to mock put_object and test its body
    allow(Aws::S3::Resource).to receive(:new).and_return(s3)
    s3client = s3.instance_variable_get('@client')
    # Call the original method
    original_put_object = s3client.method(:put_object)
    allow(s3client).to receive(:put_object) do |obj, &block|
      @csv_file = StringIO.new(obj[:body].read, 'r:UTF-16le')
      obj[:body].rewind # probably unnecessary
      original_put_object.call(obj, &block)
    end
  end

  subject do
    @csv_file.read(3) # Remove BOM
    CSV.new(@csv_file, col_sep: "\t")
  end

   it 'should have headers' do
    Sidekiq::Testing.inline! do
      post :create, params: params, body: body.to_json
    end

    expect(subject.shift).to eq headers
  end

end
```

erikogan:
Instead fallback to reading data from the handle.

Fixes #1232.

There were no existing tests for these methods. I could add some if that would help.

### DIFF
--- a/aws-sdk-core/lib/aws-sdk-core/checksums.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/checksums.rb
@@ -12,7 +12,7 @@ module Aws
       # @param [File, Tempfile, IO#read, String] value
       # @return [String<SHA256 Hexdigest>]
       def sha256_hexdigest(value)
-        if File === value || Tempfile === value
+        if (File === value || Tempfile === value) && !value.path.nil? && File.exist?(value.path)
           OpenSSL::Digest::SHA256.file(value).hexdigest
         elsif value.respond_to?(:read)
           sha256 = OpenSSL::Digest::SHA256.new
@@ -26,7 +26,7 @@ module Aws
       # @param [File, Tempfile, IO#read, String] value
       # @return [String<MD5>]
       def md5(value)
-        if File === value || Tempfile === value
+        if (File === value || Tempfile === value) && !value.path.nil? && File.exist?(value.path)
           Base64.encode64(OpenSSL::Digest::MD5.file(value).digest).strip
         elsif value.respond_to?(:read)
           md5 = OpenSSL::Digest::MD5.new


trevorrowe:
We do not have unit tests for this code, as it is not a public interface. It is tested functionally and in our integration tests.

What is the expected behavior here if the File has been unlinked and then removed from the file system. It appears its going to move down one step to the next condition (respond to #read) and then fail there. Is that indented?

erikogan:
Sorry, I should have been clearer in my description. 

Yes, that is exactly as intended, but it won’t fail the next step (at least not on a POSIX-based system (Linux/OS X/BSD, etc), because the kernel will keep the file available to the open file descriptor wrapped by the Ruby File object. File syscalls wrapped by Ruby File methods will continue to work, so it is still possible to `read(2)` data from the file object and then `rewind(3)` the fd back to the beginning.

erikogan:
Incidentally, I tested this fix before posting the PR and it solves the issue I was seeing in #1232.

trevorrowe:
I've merged this and added a CHANGELOG entry and a test to cover the behavior.

erikogan:
Great! Thanks for adding the test. I should have looked harder for where to add one. I feel better knowing there is one.

PR #1254, opened 2016-07-25
Resolved  issue with AutoScaling `:group_in_service` waiter.

### LINKED ISSUE #1252: 2.4.1 Aws::AutoScaling::AutoScaleGroup.wait_until_in_service not functional
How to replicate:

Create Launch Configuration
Create Autoscale group (ELB attached or not)
Poll Autoscale group with .wait_until_in_service or .wait_until(:autoscale_group_in_service {})

Regardless of the health or the instances in the ASG or ELB (if associated) the method will never return success.

Within a .before_attempt block, the yielded method 'response' returns an empty response.

Example code:
```
          Aws::AutoScaling::AutoScalingGroup.new("example-asg").wait_until_in_service do |w|
            w.before_attempt do |attempts, response|
              puts "#{attempts} attempts made"
              puts response
            end
          end
```

trevorrowe:
When converting the JMESPath expressions from the waiter definition into a Ruby SDK compatible path expression with snake_cased identifiers, the string literals for service values were being incorrectly converted. String literals will no longer be inflected during translation.

In issue #1252, the path is:

```
contains(AutoScalingGroups[].[length(Instances[?LifecycleState=='InService']) >= MinSize][], `false`)
```

This was resulting in `'InService'` being inflected to `'in_service'`. In JMESPath, single quotes are JSON string literals, and back-ticks are JSON value literals. These should not be inflected, only the JMESPath double-quoted strings (which are identifiers) should be translated. I've added a unit test to ensure this behavior does not regress.

Fixes #1252


awood45:
LGTM - I think I see some possible edge cases, though I don't think they're likely to ever happen in the wild. We can discuss that offline (and I may be wrong), but that won't block this change.

PR #1268, opened 2016-08-07
Fix instance variable typo

malept:
Fairly self-explanatory.


awood45:
There's a failing test here, so I'll take a look when I get in the office on Monday - I want to get this fixed for the next release.

malept:
Fixed :smile: 

It looks like that private `fresh()` method isn't tested, that probably would have caught that error earlier.

awood45:
The root cause of the test failure was elsewhere - I've made a fix that I'll be pushing up shortly. Thanks for pointing this out!

PR #1272, opened 2016-08-17
s3 markers do not get included in the list

thomasdziedzic:
I was confused by the documentation which made it seem inclusive, but
looking at the s3 api docs at http://docs.aws.amazon.com/AmazonS3/latest/API/v2-RESTBucketGET.html
I see that there is a parameter named start-after.


awood45:
This file is automatically generated, so a PR would just get clobbered with the next release. I'll raise this with the S3 doc team. Thanks!

PR #1274, opened 2016-08-23
Allow setting config/credential files via environment variables

malept:
Adds support for setting the paths to the config and credentials files via environment variables, named the same as the ones used by botocore/awscli.
- config file: `AWS_CONFIG_FILE`
- credentials file: `AWS_SHARED_CREDENTIALS_FILE`

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core/shared_config.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/shared_config.rb
@@ -20,7 +20,11 @@ module Aws
     # `~/.aws/credentials`) and the shared config file (the default path for
     # which is `~/.aws/config`) are loaded. However, if you set the
     # `ENV['AWS_SDK_CONFIG_OPT_OUT']` environment variable, only the shared
-    # credential file will be loaded.
+    # credential file will be loaded. You can specify the shared credential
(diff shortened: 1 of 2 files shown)


awood45:
Considering if this would be a breaking change, but looks good at a first pass.

awood45:
I've resolved to include this for `aws-sdk-core` 3.0, as there are already some semi-breaking changes involved there. Now considering merging it in sooner as the surface area for issues seems to be low.

malept:
Out of curiosity, now that v3 is out, is this now bumped to v4, or could it go in sooner?

awood45:
We will review this against V3. The upgrade from 2 to 3 promises no code breaking changes, that still holds, so adding this is fine.

awood45:
We will need to open this PR again against V3 code, however, since the structure is a bit different.

malept:
That's fine. I should be able to do that within the next week.

malept:
I've rebased this PR on master.

haines:
@awood45 sorry to be "that guy" but is there any chance of getting this in soon? It seems to be supported by the majority of the other SDKs, and we are relying on this feature (I've had to monkey-patch it in until this PR lands), so would really appreciate having it in the released gem.

Thanks ❤️

PR #1278, opened 2016-08-26
fixes spelling errors in several files. "encyrption" => "encryption"

runongirlrunon:
"Encryption" was spelled as "encyrption" in several files. This commit fixes all those errors.


trevorrowe:
Thanks for the fixes!

PR #1293, opened 2016-09-19
Fix InstanceProfileCredentials warning

kddeisz:
Previously it was rescuing an error but then doing nothing with it, which results in warnings being generated. Also fixes a typo in the tests.

```
/Users/kdeisz/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/aws-sdk-core-2.4.3/lib/aws-sdk-core/instance_profile_credentials.rb:122: warning: assigned but unused variable - error
```


trevorrowe:
Thanks for the contribution!

PR #1297, opened 2016-09-20
Return key length rather than arbitrary number

smithyuk:
Current fixed number might make it difficult to debug issues


awood45:
Looks good, thanks for the contribution!

Issue #1265, opened 2016-08-03
Redundant calls to "describe_stacks" from Aws::CloudFormation::Stack#data


mdub:
Most resource-collection methods appear to fully "load" with a single API call:

```
$ AWS_DEBUG=true aws.rb
Aws> Aws.ec2.resource.instances.first.data; nil
[Aws::EC2::Client 200 1.41273 0 retries] describe_instances()  
=> nil
Aws> Aws.ec2.resource.subnets.first.data; nil
[Aws::EC2::Client 200 0.250955 0 retries] describe_subnets()  
=> nil
Aws> Aws.ec2.resource.security_groups.first.data; nil
[Aws::EC2::Client 200 0.760407 0 retries] describe_security_groups()  
=> nil
Aws> Aws.ec2.resource.volumes.first.data; nil
[Aws::EC2::Client 200 1.107822 0 retries] describe_volumes()  
=> nil
```

but calling `Aws::CloudFormation::Stack#data` appears to trigger another call to `describe_stacks`, which I imagine to be redundant:

```
Aws> Aws.cloudformation.resource.stacks.first.data; nil
[Aws::CloudFormation::Client 200 0.949843 0 retries] describe_stacks()  
[Aws::CloudFormation::Client 200 0.05216 0 retries] describe_stacks(stack_name:"woollyams-test")  
=> nil
```

This makes dealing with large numbers of stacks inefficient:

```
Aws> Aws.cloudformation.resource.stacks.map(&:outputs); nil
[Aws::CloudFormation::Client 200 0.886849 0 retries] describe_stacks()  
[Aws::CloudFormation::Client 200 0.11976 0 retries] describe_stacks(stack_name:"woollyams-test")  
[Aws::CloudFormation::Client 200 0.061462 0 retries] describe_stacks(stack_name:"tlap-woollyams")  
[Aws::CloudFormation::Client 200 0.063231 0 retries] describe_stacks(stack_name:"cp-agentadmin-in-e2e-shipper-internal")  
[Aws::CloudFormation::Client 200 0.061777 0 retries] describe_stacks(stack_name:"cp-agentadmin-in-e2e-shipper-public")  
[Aws::CloudFormation::Client 200 0.061863 0 retries] describe_stacks(stack_name:"property-updates-notifier-dev-mwang")  
[Aws::CloudFormation::Client 200 0.062491 0 retries] describe_stacks(stack_name:"inbox-resources")  
[...]

PR #1311, opened 2016-10-08
Add missing "path" for CloudFormation "Stacks" collection

mdub:
The adds a "path" to for the "Stacks" resource collection, allowing data to be loaded into Stack objects by the "DescribeStacks" call.

Addresses aws/aws-sdk-ruby#1265.

As far as I could see, "Stacks" is the only "hasMany" collection that is missing a "path".

### DIFF
--- a/aws-sdk-core/apis/cloudformation/2010-05-15/resources-1.json
+++ b/aws-sdk-core/apis/cloudformation/2010-05-15/resources-1.json
@@ -36,7 +36,8 @@
           "type": "Stack",
           "identifiers": [
             { "target": "Name", "source": "response", "path": "Stacks[].StackName" }
-          ]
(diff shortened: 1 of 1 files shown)


mdub:
Hmm. Travis CI build appears to have been failing (one test) since 9ee5ac9. @awood45?

awood45:
I'm looking at this, and it seems to still make the extra calls when inspecting stack data. It doesn't look wrong, but does this address the linked issue for you?

awood45:
Okay, I take that back, it looks like when I pulled down this branch to test, it wasn't actually taking in the diff.

awood45:
In any case, that looks good then. I'll merge this now, and push it upstream. Thanks!

PR #1312, opened 2016-10-10
Fix EC2 version bump in unit test

cjyclaire:
Fix unit test.

@awood45


cjyclaire:
Looks like the fix is already merged :)

PR #1315, opened 2016-10-12
Add Support for S3 Accelerate with Dualstack

awood45:
Allows for use of S3 Accelerate and IPv6 together. Added to the S3 Accelerate plugin due to the specific endpoint.


awood45:
I've integration tested this against an accelerate-enabled bucket that I own. It is hitting the correct endpoint and returning responses successfully.

PR #1318, opened 2016-10-18
Recognise Kinesis throttling errors.

ags:
Kinesis can raise a `LimitExceededException` that isn't retried like other throttling errors as the name doesn't include 'throttl' and it's not included in the whitelist of other throttling errors.

This adds the exception to the whitelist so it can be retried.


ags:
Is there any chance of this being looked at?

awood45:
Thanks for this, merging in.

ags:
Thanks!

PR #1326, opened 2016-10-24
remove unnecessary documentation note

mziwisky:
in my testing, Aws::S3::Object#copy_to and Aws::S3::Object#copy_from
both work equally well for cross-region copies.  if that isn't the case,
then i think the note this commit removes needs more detail about which
scenarios #copy_to cannot handle that #copy_from can.

### DIFF
--- a/aws-sdk-resources/lib/aws-sdk-resources/services/s3/object.rb
+++ b/aws-sdk-resources/lib/aws-sdk-resources/services/s3/object.rb
@@ -70,9 +70,6 @@ module Aws
       # Copies this object to another object. Use `multipart_copy: true`
       # for large objects. This is required for objects that exceed 5GB.
       #
-      # @note If you need to copy to a bucket in a different region, use
-      #   {#copy_from}.
-      #
       # @param [S3::Object, String, Hash] target Where to copy the object
       #   data to. `target` must be one of the following:
       #


awood45:
I think it may be correct that both work now that more robust bucket redirect logic is in place, but we'd like to do some verification before accepting this PR. I do agree that if we find anything, that note should be updated, and removed if we don't find anything.

cjyclaire:
@mziwisky Interestingly, when I try to reproduce a cross region `#copy_to` call, it fails with
```
 Aws::S3::Errors::PermanentRedirect The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.
Aws::S3::Errors::PermanentRedirect: The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.
```
It's expected because I was trying to copy my object to another region.

I'm curious, could you explain a bit how you find `copy_to` works cross region for you?

mziwisky:
@cjyclaire and `#copy_from` works without raising that PermanentRedirect?  I thought I inspected the web request created by each of those method calls and found them to be the same, but I can't be sure anymore.  I wish now I'd have included my test procedure with this issue in the first place.  Alas, I'll have to investigate again.  Will report back after I've done so.

cjyclaire:
@mziwisky Yeah, `#copy_from` works without problem. No worries, appreciate the feedback, happy to see if you have some interesting findings, then we could make our documentation better : )

cjyclaire:
Closing due to ages, yet feel free to reopen anytime once you have further details and interesting findings ; )

mziwisky:
The closing reminded me of this!  So I dug in a bit.  I found that using `copy_to` to go from us_east_1 to us_west_2 works fine because aws-sdk handles the redirect itself, and then caches the region for the bucket in us-west-2 and uses that for future requests.  So you get a warning the first time you try that, but not on subsequent tries.  However the same can't be said for `copy_to` from us_west_2 to us_east_1.

So the note is true for certain pairs of regions, and false for others.  In that case it's certainly better to keep it around than to delete it, but maybe all regions could gain this auto-redirecting behavior?

You can view the test cases and an annotated run output for them at https://gist.github.com/mziwisky/4e9e71c1bdb6c70145acc785057aa545

Issue #1342, opened 2016-11-11
Aws::S3::Object#exists? doesn't work with encrypted objects


XmmmiR:
```Aws::S3::Object#exists?``` relays on client`s ```head_object```, which uses same encryption options as ```get_object```, but doesn't provide any ways to pass options

```ruby
key = OpenSSL::Cipher::AES256.new(:ECB).random_key
encryption_options = {
  sse_customer_algorithm: "AES256",
  sse_customer_key: key,
  sse_customer_key_md5: Digest::MD5.digest(key)
}

resource = Aws::S3::Resource.new(access_key_id: "", secret_access_key: "", region: "us-east-1")
bucket = resource.bucket(bucket_name)
object = bucket.object("test_obj")

object.exists? #=> false

# Standard object
object.put body: "value"
object.exists? #=> true

# Object with encryption
object.put({ body: "value" }.merge!(encryption_options))
object.exists? # Aws::S3::Errors::BadRequest:
object.get(encryption_options).body.read #=> "value"
```

I can still find out if object exists(error appears only for encrypted objects), yet this definitely isn't desired behaviour.

XmmmiR:
With proposed changes, options passed to ```#exists```, will be used in ```Aws::S3::Client#head_object```, but it gives same ability to any classes inherited from ```Aws::Resources::Resource```

trevorrowe:
I added the missing tests to the pull requests and merged it. Thanks for raising the issue!

julik:
It would seem that either there was a regression, or the fix in #1343 did not quite do the trick. I can't seem to be able to pass the encryption headers to `exists?`. I can bypass it using the well-known `0-0` byte range trick, so there is a workaround.

PR #1343, opened 2016-11-11
Allow passing extra option to Aws::Resources::Resource#exists?

XmmmiR:
Adds ability to pass additional waiter params to ```Aws::Resources::Resource#exists?``` and ```Aws::Resources::Operations::WaiterOperation```

Related issue #1342

### DIFF
--- a/aws-sdk-resources/lib/aws-sdk-resources/operations.rb
+++ b/aws-sdk-resources/lib/aws-sdk-resources/operations.rb
@@ -221,7 +221,7 @@ module Aws
 
           resource = options[:resource]
 
-          params_hash = {}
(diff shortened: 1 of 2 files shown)


baburdick:
+1

albertosaurus:
No tests?

trevorrowe:
Thanks for the contribution. I've merge the change and added tests that cover this behavior.

julik:
It would seem that either there was a regression, or the fix in #1343 did not quite do the trick. I can't seem to be able to pass the encryption headers to `exists?`. I can bypass it using the well-known `0-0` byte range trick, so there is a workaround.

PR #1355, opened 2016-12-05
allow DynamoDB::Client.update_item to PUT empty sets

makeready:
Added a few lines to ensure that blank sets can be saved in dynamoDB. We're using sets and having to delete the attribute entirely just to blank them out feels counterintuitive.


makeready:
Current master build is failing CI due to a gem that doesn't compile in ruby < 2, I'll update the pull request once master is fixed.

awood45:
Can you tell me a bit more about the problem you're trying to solve, maybe with a reproducing example? Tests are breaking, but I can review this a bit better with more context.

makeready:
Essentially the SDK currently prevents us from performing a PUT action for an attribute when the new value is an empty set. This is because it can't be marshalled - there is a case statement that checks the datatype of the first member of the set and incorrectly raised a type error if the set was empty.

BTW, note that this fork doesn't actually break any of the tests, but Travis CI is flagging the build for not being compatible with ruby 1.9 - this incompatibility was forked from master and was introduced to the main build on November 29.

makeready:
Example of update_item that this PR fixes:

```
client.update_item({
  table_name: "myTable",
  key: { 
    "id" => "1"
  },
  attribute_updates: {
    "mySet" => {
      value: Set.new
      action: "PUT"
    }
  }
})
```

PR #1358, opened 2016-12-09
Run test with ruby 2.3.3

ojiry:
#### Purpose
Test with latest ruby version.

#### Changes
Modify ruby version of .travis.yml.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ rvm:
   - 2.0.0
   - 2.1
   - 2.2
-  - 2.3.0
+  - 2.3.3
   - jruby
 
 sudo: false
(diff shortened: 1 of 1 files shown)


awood45:
Any particular reason we want to change the version here? I'm more inclined to add Ruby 2.4 as it comes out rather than extend into additional point changes unless there's a particular change to concern ourselves with.

ojiry:
> Any particular reason we want to change the version here?

Because we using the latest ruby, we want to verify by Travis CI.

>  I'm more inclined  to add Ruby 2.4~

I'm sorry. My English was poor and I couldn't understand the meaning of the second half. 🙇

awood45:
I'm going to decline this PR, but I will make a change to try and remove the `2.3.0` point and test `2.3` instead. That should solve the problem you're mentioning and would be more consistent with how we test other Ruby versions. Thanks for bringing this to our attention!

ojiry:
I agree with it. Thanks.

PR #1359, opened 2016-12-09
Support Ruby 2.4

ojiry:
This PR made the following changes in preparation Ruby 2.4.
* Fix deprecated warnings caused by NIL and Fixnum
* Add ruby-head version to travis.yml


ojiry:
This PR close once because it takes time to fix the tests.

Issue #1301, opened 2016-09-23
Aws::Errors::MissingCredentialsError: unable to sign request without credentials set


rayway30419:
Hi

I use IAM-role to assign policy to my EC2 instance, but sometimes it raise missing credential error.

`Aws::Errors::MissingCredentialsError: unable to sign request without credentials set`

Does anyone have same problem?

awood45:
So you're saying that on an EC2 instance, you're seeing this periodically? How is your environment set up?

rayway30419:
Hi @awood45 

It happens randomly, 

What kind of enviroment setting?
--

Ruby 2.3.0
Rails 5.0.0.1
aws-sdk 2.6.3
Ubuntu 14.04 Trusty (3.13.0-63-generic)

awood45:
Is the Ubuntu version you listed your development environment, or an EC2 AMI? How are you configuring your clients in code?

rayway30419:
It's on the EC2 AMI.
```
# region in configuration file
client = Aws::DynamoDB::Client.new(
  region: Aws.config[:region]
)
# do sth.
client.put_item params
```

rayway30419:
BTW, I found a clue.....it's running on my sidekiq worker, it may happen to mult-ithreading?

rayway30419:
Another error message when initialize DynamoDB client
`NameError: uninitialized constant Aws::CredentialProviderChain::InstanceProfileCredentials`

awood45:
Okay, it looks like this may be related to how Sidekiq initializes - you might need to do eager_autoload within your Sidekiq workers, rather than only in the Rails initialization. Can you try that?

awood45:
For what it's worth, we're working on an overhaul of the SDK which should reduce or eliminate these autoload issues - message definitely received that they cause problems when threading.

rayway30419:
@awood45 
With my understand, the sidekiq process will execute the procedures within rails initializer, too.

I am glad to hear this new.

glennpratt:
https://bugs.ruby-lang.org/issues/12688

trevorrowe:
Just chiming in here. The preview release of the modularized SDK is coming shortly. With modularization, we have eliminated the use of `autoload`.

trevorrowe:
[...]

PR #1360, opened 2016-12-09
Fixes empty credential hash never getting refreshed

GeorgFleig:
This fix forces set? to also recheck for new credentials in case the credentials hash is empty and not only when it is nil.

In case of an EC2 instance using an instance profile the error AWS::Errors::MissingCredentialsError occurred from time to time (this also seems to occur in v2: https://github.com/aws/aws-sdk-ruby/issues/1301). It turned out to be a connection problem between the instance and the meta data service running on 169.254.169.254. However, even when the service is reachable again, the cached credentials (an empty hash) are never refreshed and clients keep ending up with the MissingCredentialsError.

Either each and every tool using calls which depend on AWS credentials could catch this error and force a AWS.config.credential_provider.refresh or the problem is fixed at it's root, as proposed by this pull request.

### DIFF
--- a/lib/aws/core/credential_providers.rb
+++ b/lib/aws/core/credential_providers.rb
@@ -47,9 +47,11 @@ module AWS
         #
         def set?
           @cache_mutex ||= Mutex.new
-          unless @cached_credentials
+          if @cached_credentials.nil? || @cached_credentials.empty?
             @cache_mutex.synchronize do
-              @cached_credentials ||= get_credentials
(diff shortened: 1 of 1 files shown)


awood45:
Unfortunately, V1 of the SDK is deprecated and we are only addressing security vulnerabilities going forward. However, I would look at a PR for the issue in V2 of the SDK, if the issue presents the same way (I'll take a look at a repro test for that).

PR #1370, opened 2016-12-22
Parse EAI_NONAME SocketError from GNU libc

chtitux:
The GNU libc returns a different error message than MacOS.
The glibc error message for EAI_NONAME is defined in https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/posix/gai_strerror-strs.h;h=19040a51381b83127c947233fee8cf56e185f92f;hb=HEAD#l8 ,
and should be considered as a DNS resolution error.

### DIFF
--- a/aws-sdk-core/lib/seahorse/client/net_http/handler.rb
+++ b/aws-sdk-core/lib/seahorse/client/net_http/handler.rb
@@ -30,7 +30,10 @@ module Seahorse
         end
 
         # @api private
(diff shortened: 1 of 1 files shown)


mehlah:
Curious to know btw why it's not enough to rely only on `SocketError` without adding the condition on `DNS_ERROR_MESSAGES`

awood45:
Socket errors are generic to a number of root causes, this error message is generally intended to help with misconfigured clients hitting a non-existent endpoint or region.

In any case, this PR looks good. Thanks!

PR #1373, opened 2016-12-27
removed Fixnum reference

jamespdo:
Since the Ruby-2.4.0 release, Fixnum and Bignum are combined into Integer class. This fix is to remove all Fixnum reference.

### DIFF
--- a/aws-sdk-core/spec/aws/param_validator_spec.rb
+++ b/aws-sdk-core/spec/aws/param_validator_spec.rb
@@ -114,14 +114,24 @@ module Aws
 
       it 'validates map keys' do
         validate({ string_map: { 'abc' => 'mno' }})
-        validate({ string_map: { 123 => 'xyz' }},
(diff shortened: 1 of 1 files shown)


awood45:
This change has numerous failing tests - I'll take a look at the issue w/r/t Ruby 2.4's release, but I can't merge this PR as-is.

jamespdo:
Yes I understand. This is a breaking change to any ruby version that <=2.4.0.
I have updated the code to handle and rolled them under this same PR. Thanks!

awood45:
I don't think I want tests to be detecting Ruby versions in this manner. The tests may only be passing because this code isn't hit, besides.

I'm taking a look at Ruby 2.4 compatibility concerns and I'll keep an eye on this PR for guidance as I do so.

awood45:
Thank you for pointing this out - I'm addressing this in a separate branch/PR to solve this (and any other issues I can find) for Ruby 2.4.

PR #1374, opened 2016-12-30
Re-enforce path addressing for S3 GetBucketLocation requests

benesch:
Signature v4 broke GetBucketLocation requests that use virtual host
addressing of buckets; GetBucketLocation requests must now use path
addressing. This was addressed in 68a6cde, but this fix was later broken
by refactoring related to f19e4d8. This commit fixes 68a6cde and adds an
integration test for the behavior.

### DIFF
--- a/aws-sdk-core/features/s3/buckets.feature
+++ b/aws-sdk-core/features/s3/buckets.feature
@@ -54,3 +54,12 @@ Feature: S3 Buckets
     When I create a bucket with the location constraint "us-west-2"
     When I put a large object
     Then the object should exist
+
+  Scenario: Get location of DNS-compatible bucket in nonstandard region
+    Given I am using the S3 "us-west-2" region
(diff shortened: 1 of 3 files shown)


benesch:
I believe this is only failing because of the issue in #11!

awood45:
Looks good at a first pass - I'll run the integration tests and give a final review when I get back to work on Tuesday. Thanks!

benesch:
Sounds good. Thanks, @awood45!

I should note that I never tracked down exactly what caused `context.operation_name` to switch from a string to a symbol; `git bisect` only narrowed it down to the fallout from f19e4d8.

awood45:
I'll look in to that question since I'm now wondering if we have similar issues hiding elsewhere in the code.

For now, looks good and merging. Thanks!

benesch:
Awesome! Thanks so much, @awood45.

PR #1378, opened 2017-01-03
Update Travis Config For Ruby 2.3 and 2.4

awood45:
- Make Ruby 2.3 unbound to a specific point release.
- Add Ruby 2.4

Successor to PR #1358


awood45:
Scrapping as this has gone off original intent. Will replace with a refactor to handle 2.4

PR #1383, opened 2017-01-05
Add VPC Peer deletion waiter.

pdrakeweb:


cjyclaire:
@pdrakeweb Thanks for the contribution! 

Yet this file is automatically generated, I'll make those changes internally from root to address the need. Will be tagged in the following release after it merged. I'll ping there when the fix is done.
 
Thanks : )

cjyclaire:
A tagged release containing the fix is out, soft ping.

Issue #1386, opened 2017-01-06
Aws::CloudFront::UrlSigner fails with more than one query parameter


dobrite:
Using the following code I am able to successfully access a file from Cloudfront pointed at S3. All query parameters are forwarded in the Cloudfront distribution.

```ruby
signer = Aws::CloudFront::UrlSigner.new(key_pair_id: '...', private_key: '...')
signer.signed_url("https://abc.cloudfront.net/test.txt?one=red", expires: 5.minutes.from_now)
```

However, when using 2 query parameters I get the standard XML "Access Denied":

```ruby
signer.signed_url("https://abc.cloudfront.net/test.txt?one=red&two=blue", expires: 5.minutes.from_now)
```

I poked at the code, but didn't turn up anything obvious.

Thank you for any insight on this.

dobrite:
I was able to get it to work with the aws cli using `aws cloudfront sign`. This points to a bug in this library. I'll work on providing a failing spec.

dobrite:
I was able to generate valid signatures using the test case, but not in my development environment. This points to an issue with my development environment (open ssl?), and not this library. I'll reopen this issue if I find something otherwise, but for now I don't want you to waste any (more) time on this.

awood45:
Please do also update us if you find something that fixes your local environment, in case this comes up again. Thanks!

dobrite:
Ok, I figured it out!

Rails has a [default](http://stackoverflow.com/questions/27379432/prevent-rails-from-encoding-the-ampersands-in-a-url-when-outputting-json/27460529#27460529) setting that encodes `&` as `\\u0026` which explains why the signatures didn't match when an ampersand was involved. It also explains why the test case produced valid signatures and my dev environment didn't (Rails was loaded).

As a workaround `JSON.dump(json_hash)` can be used [here](https://github.com/aws/aws-sdk-ruby/blob/master/aws-sdk-core/lib/aws-sdk-core/cloudfront/url_signer.rb#L141). This will fix cloud front url signer for vanilla Rails apps when an ampersand is involved.

Does that sound like a reasonable fix?

```
# Rails irb
[35] pry(main)> JSON.generate(json_hash)
=> "{\"Statement\":[{\"Resource\":\"http://www.google.com?q=blah&a=yo\",\"Condition\":{\"DateLessThan\":{\"AWS:EpochTime\":191981981981}}}]}"

# Ruby irb
[36] pry(main)> json_hash.to_json
[...]

PR #1387, opened 2017-01-07
AWS Cloudfront UrlSigner: failing spec signing url with 2 query parameters

dobrite:
Failing spec for issue #1386 

Signature generated with:

```bash
$ aws cloudfront sign --url "https://abc.cloudfront.net/images/image.jpg?color=red&size=large" --key-pair-id "CF_KEYPAIR_ID" --private-key file://aws-sdk-core/spec/aws/cloud_front/cf_private_key.pem --date-less-than 1357034400
```

### DIFF
--- a/aws-sdk-core/spec/aws/cloud_front/url_signer_spec.rb
+++ b/aws-sdk-core/spec/aws/cloud_front/url_signer_spec.rb
@@ -74,6 +74,19 @@ module Aws
             "&Key-Pair-Id=CF_KEYPAIR_ID"
             expect(url).to eq(expected_url)
         end
+        it 'can generate signed urls with canned policy with 2 query parameters' do
+          url = signer.signed_url(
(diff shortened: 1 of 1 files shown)


dobrite:
`expected_url` is incorrect. This spec passes.

I did re-verified that 1 query param works with `aws-sdk-ruby`, 2 does not, but 2 _does_ work with the aws cli using same credentials, distribution, etc.

awood45:
I'll definitely take a look at this. It's a bit more involved to set up and repro than the usual issue, so it's taken a bit longer to set aside the time. Happy to take a PR in the meantime, of course, but I agree there should not be an inconsistency.

dobrite:
I understand completely. I too am having issues with setup and repro. I was hoping to provide a failing test case, but narrowing it down is providing difficult. In any case, thank you to taking a look!

PR #1390, opened 2017-01-09
Fix Cloudfront url_signer for more than one query parameter in a Rails environment

dobrite:
Vanilla rails escapes html entities in json by default using `to_json`

Use `JSON.dump` to preserve html entities (e.g. `&` for more than one query param)

Fixes issue #1386


awood45:
This looks good, thanks! Will add this to the next release.

dobrite:
Thank you for your work!

PR #1395, opened 2017-01-25
Add presign-url support for create_db_instance_read_replica

cjyclaire:


cjyclaire:
Note that failure is jruby and irrelevant with this change.

Issue #1336, opened 2016-11-07
Add some randomness to the exponential backoff of max_retries


keymon:
The [retry logic configured by `max_retries`](http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/Core/Configuration.html#max_retries-instance_method), for the case of Rate limit related errors, does a exponential backoff [following this function](https://github.com/aws/aws-sdk-ruby/blob/395e0072c731815c2d4a8bc002d516d89916b11f/lib/aws/plugins/retry_errors.rb#L9): `2.0 ** retries * 0.3`

That is great, but there is a problem: It is exactly the same backoff time for every execution. This is problematic when you are doing lots of operations from multiple threads. For instance, creating a environment with multiple VMs using 32 threads. All the threads will execute the same API calls at the same time, and if several threads fail at the same time because the Rate limit from AWS, they will retry again after the same amount of seconds.

What I want to propose is add a little bit of randomness to that backoff. We can use this function:

```
def backoff(r)
    b = 2.0 ** r.to_f * 0.3
    v = 1.25 ** r.to_f * 0.3 * (Random.rand() - 0.5)
    return b+v
end
```

With this code we can test it a little bit:

``` 
(1..5) each { |run| # Repeat the experiment
    puts "---- Execution: #{run} ----"
    (1..10).each { |r| 
        b = backoff(r)
        puts "Retries #{r}: #{b}s == #{b.floor()}m #{b % 60}s"
    }
}
``` 

```
---- Execution: 1 ----
Retries 1: 0.4723107748709313s == 0m 0.4723107748709313s
Retries 2: 1.4083360795375834s == 1m 1.4083360795375834s
Retries 3: 2.271092661113177s == 2m 2.271092661113177s
Retries 4: 4.916736747923815s == 4m 4.916736747923815s
Retries 5: 10.000144551307137s == 10m 10.000144551307137s
Retries 6: 18.839097183818104s == 18m 18.839097183818104s
Retries 7: 38.480158713168855s == 38m 38.480158713168855s
Retries 8: 77.45500289555731s == 77m 17.455002895557314s
Retries 9: 152.7204691297255s == 152m 32.7204691297255s
Retries 10: 307.6726209158677s == 307m 7.672620915867697s
---- Execution: 2 ----
Retries 1: 0.6050370631541719s == 0m 0.6050370631541719s
Retries 2: 1.345491076957416s == 1m 1.345491076957416s
[...]

PR #1397, opened 2017-01-26
Additional backoff strategies with cap + jitter  #1336

lwoggardner:
Just checking if this may be on the right track.

Not entirely sure that exposing constants to represent the various strategies is a good option or if the default backoff strategy should/could be changed to something with jitter + max delay cap.


lwoggardner:
Have decided to try an alternate approach

PR #1398, opened 2017-01-26
Fix S3 bucket dns for IP endpoint

cmarodrigues:
In case we have IP endpoint, the function move_dns_compat_bucket_to_subdomain change the endpoint path and doesn't allow to connect to the S3 API.

This is problematic when we have S3 endpoint of Dell/EMC Centera S3 API.

### DIFF
--- a/aws-sdk-core/lib/aws-sdk-core/plugins/s3_bucket_dns.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/plugins/s3_bucket_dns.rb
@@ -39,6 +39,7 @@ module Aws
           if
             bucket_name &&
             S3BucketDns.dns_compatible?(bucket_name, https?(endpoint)) &&
+            endpoint.host !~ /(\d+\.){3}\d+/ &&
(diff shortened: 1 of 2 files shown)


cjyclaire:
Thanks for the contribution! Could you also add some rspec test for the change as well?

cmarodrigues:
After read more about it, this pull request can make no sense and could be fixed with flag force_path_style = true. Although however, i submit rspec for tests of s3_bucket_dns, if you think this could be helpful, you can merge it.

cjyclaire:
@cmarodrigues , That's true, taking a second look, we already have similar checks at [`dns_compatible`](https://github.com/aws/aws-sdk-ruby/blob/master/aws-sdk-core/lib/aws-sdk-core/plugins/s3_bucket_dns.rb#L79). And S3 DNS bucket [style](http://docs.aws.amazon.com/AmazonS3/latest/dev/VirtualHosting.html) do have stricter requirement. I'd suggest using path style (which is recommended) there.

Closing, same great thanks for your thoughts in contributions : )

Issue #1401, opened 2017-01-28
Explicit port :80 on presigned_url with virtual_host set to true causes problems


kjwierenga:
**Problem statement**
`Presigner#presigned_url` with `virtual_host` set to `true` returns a URL where the port (for http) is explicitly set to `:80`. Some HTTP clients (in my case `ffprobe`) pass on this port in the `Host` header (`Host: files.mydomain.dom:80`) which invalidates the signature resulting in a `403 Forbidden` response.

**TL;DR**
It would be nice if `Presigner#presigned_url` didn't have `:80` in the URL so all HTTP clients can use the URL without problems.

**Full explanation**

First we show that `:80` is indeed included in the presigned URL.
```ruby
> params = { bucket: 'files.mydomain.com', key: 'song.mp3', virtual_host: true }
 => {:bucket=>"files.mydomain.com", :key=>"song.mp3", :virtual_host=>true} 
> Aws::S3::Presigner.new.presigned_url(:get_object, params)
[Aws::S3::Client 0 0.000621 0 retries] get_object(bucket:"files.mydomain.com",key:"song.mp3")  

 => "http://files.mydomain.com:80/song.mp3?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=XXXXXXXXXXXXXXXXXXXX%2F20170127%2Feu-west-1%2Fs3%2Faws4_request&X-Amz-Date=20170127T165250Z&X-Amz-Expires=900&X-Amz-SignedHeaders=host&X-Amz-Signature=5ce3411e5b4016a670c159c62e72563be1a4aa256d332185a9d7570b866c8b66" 
```

If we use this URL with curl everything we get a `HTTP/1.1 200 OK`.
```bash
$ curl -s -v "http://files.mydomain.com/song.mp3?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=XXXXXXXXXXXXXXXXXXXX%2F20170128%2Feu-west-1%2Fs3%2Faws4_request&X-Amz-Date=20170128T104954Z&X-Amz-Expires=10800&X-Amz-SignedHeaders=host&X-Amz-Signature=c620db0d8e3615eb6e5fa6bc5027c1ccee5472c00b11b505de7af801328b75d4" -o /dev/null
*   Trying 52.218.64.40...
* TCP_NODELAY set
* Connected to files.mydomain.com (52.218.64.40) port 80 (#0)
[...]

PR #1402, opened 2017-01-28
Don't set default port 80 explicitly. Fixes #1401

kjwierenga:
No need to set port 80 explicitly on virtual_host presigned_url. Not setting it prevents problems for some HTTP clients. Fixes #1401

PR #1403, opened 2017-01-28
Re-parse the virtual_host URI in presigned_url after changing the scheme. Fixes #1400, #1401

### LINKED ISSUE #1400: Presigner#presigned_url mutates params hash
The `params` hash passed to `Presigner#presigned_url` is mutated in the call (`virtual_host` attribute is removed). This is unexpected and can cause problems for callers.

```ruby
> params = { bucket: 'files.mydomain.com', key: 'test', virtual_host: true }
 => {:bucket=>"files.mydomain.com", :key=>"test", :virtual_host=>true} 
> Aws::S3::Presigner.new.presigned_url(:get_object, params)
[Aws::S3::Client 0 0.000683 0 retries] get_object(bucket:"files.mydomain.com",key:"test")  

 => "http://files.mydomain.com:80/test?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=XXXXXXXXXXXXXXXXXXXX%2F20170127%2Feu-west-1%2Fs3%2Faws4_request&X-Amz-Date=20170127T163857Z&X-Amz-Expires=900&X-Amz-SignedHeaders=host&X-Amz-Signature=21d2a78d1dda9485469d48a57b2e4e3f173d795c74f15ed5fff3319b7b3df716" 
> params
 => {:bucket=>"files.mydomain.com", :key=>"test"} 
> Aws::S3::Presigner.new.presigned_url(:get_object, params)
[Aws::S3::Client 0 0.000572 0 retries] get_object(bucket:"files.mydomain.com",key:"test")  

 => "https://s3-eu-west-1.amazonaws.com/files.mydomain.com/test?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=XXXXXXXXXXXXXXXXXXXX%2F20170127%2Feu-west-1%2Fs3%2Faws4_request&X-Amz-Date=20170127T163857Z&X-Amz-Expires=900&X-Amz-SignedHeaders=host&X-Amz-Signature=3c2635c9dec08209a5eeb9f010e8cff69bea56c7cbd7e63438356b13ddbb0724" 
```

First call returns a https virtual_host URL, while the second call returns a https URL.

I think `params` should be cloned inside `Presigner::presigned_url`.

kjwierenga:
`Presigner#presigned_url` should not mutate params hash when it processes paramaters such as `virtual_host` and `secure`. This is fixed by duplicating the hash. Fixes #1400

### DIFF
--- a/aws-sdk-core/lib/aws-sdk-core/s3/presigner.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/s3/presigner.rb
@@ -76,11 +76,12 @@ module Aws
       def use_bucket_as_hostname(req)
         req.handlers.remove(Plugins::S3BucketDns::Handler)
         req.handle do |context|
-          uri = context.http_request.endpoint
+          uri = context.http_request.endpoint.dup
           uri.host = context.params[:bucket]
           uri.path.sub!("/#{context.params[:bucket]}", '')
           uri.scheme = 'http'
           uri.port = 80
+          context.http_request.endpoint = URI.parse(uri.to_s)
           @handler.call(context)
         end
       end
--- a/aws-sdk-core/lib/aws-sdk-core/s3/presigner.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/s3/presigner.rb
@@ -42,6 +42,7 @@ module Aws
         if params[:key].nil? or params[:key] == ''
           raise ArgumentError, ":key must not be blank"
         end
+        params = params.dup
         virtual_host = !!params.delete(:virtual_host)
         scheme = http_scheme(params, virtual_host)
 
--- a/aws-sdk-core/spec/aws/s3/presigner_spec.rb
+++ b/aws-sdk-core/spec/aws/s3/presigner_spec.rb
@@ -109,6 +109,16 @@ module Aws
           expect(url).to match(/^http:/)
         end
 
+        it 'can generate virtual hosted url' do
+          signer = Presigner.new(client: client)
+          url = signer.presigned_url(:get_object,
+            bucket:'virtual.hosted.com',
+            key:'foo',
+            virtual_host: true
+          )
+          expect(url).to match(/^http:\/\/virtual.hosted.com\/foo/)
+        end
+
       end
     end
   end
--- a/aws-sdk-core/spec/aws/s3/presigner_spec.rb
+++ b/aws-sdk-core/spec/aws/s3/presigner_spec.rb
@@ -119,6 +119,18 @@ module Aws
           expect(url).to match(/^http:\/\/virtual.hosted.com\/foo/)
         end
 
+        it 'returns same url when called twice' do
+          signer = Presigner.new(client: client)
+          params = {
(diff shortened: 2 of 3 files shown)


kjwierenga:
This PR is dependent on PR #1402

awood45:
Update on this - I'm trying to get an environment set up to properly test this. I believe that the port 80 explicit mention may have been for a specific bug, and I would want to avoid a regression of the same. We only have an integration test around this currently.

awood45:
Okay, there does appear to be a problem which makes this issue a bit more complicated than the included PR. Consider the original purpose of this code snippet, which is to switch the URI to http. For your use case of a non-standard port, your PR is fine:

```ruby
uri = URI("https://foo.bar:3000") #=> #<URI::HTTPS https://foo.bar:3000>
uri.port #=> 3000
uri.scheme = "http"
uri #=> #<URI::HTTPS http://foo.bar:3000>
uri.port #=> 3000
```

However, it doesn't work for the standard HTTP/HTTPS ports - that explicit setter is needed:

```ruby
uri = URI("https://foo.bar") #=> #<URI::HTTPS https://foo.bar>
uri.port #=> 443
uri.scheme = "http"
uri #=> #<URI::HTTPS http://foo.bar>
uri.port #=> 443 - wrong
```

That doesn't mean we can't fix this issue - I'd like to support your ability to set ports, but we need to do a bit more. I'll copy this comment to the PR.

kjwierenga:
You can change the scheme of a URI object, but that doesn't change the class of the object. This is what is causing the default port to be appended.

```ruby
uri = URI("https://foo.bar") #=> #<URI::HTTPS https://foo.bar> 
uri.to_s # => "https://foo.bar" 
uri.scheme='http'
uri.port=80
uri.to_s # => "http://foo.bar:80" 
URI.parse(uri.to_s).to_s
# Fix like this
URI(uri.to_s) # => #<URI::HTTP http://foo.bar> 
```

Same goes the other way around:

```ruby
uri=URI("http://foo.bar") # => #<URI::HTTP http://foo.bar> 
uri.to_s # => "http://foo.bar" 
uri.scheme='https'
uri.port=443
uri.to_s # => "https://foo.bar:443" 
# Fix like this
URI(uri.to_s) # => #<URI::HTTPS https://foo.bar> 
```

Note how the URI subclass is determined at first parse of the input URL.

Fix is to re-parse the URI after changing the scheme to a URI object with the appropriate subclass gets created.

kjwierenga:
This is a better way of fixing this issue I think. The solution was there all along in `sign_but_dont_send` which applies exactly the same technique.

awood45:
Indeed, a topic of discussion in our team right now is to potentially move a lot of this logic into a single class to avoid exactly these kinds of inconsistencies. Let's change the PR to use this kind of approach and then we should be good to go.

cjyclaire:
PR #1477 has opened including support for this PR, closing, tracking in #1477 instead.
The spec test has been included as well.
Feel free to chime in and add comment in that PR, thanks : )

PR #1412, opened 2017-02-07
Add Doc cross linking support for Code Gen

cjyclaire:
Adding doc cross linking support for Code Gen.
@awood45 @trevorrowe


cjyclaire:
I have tested this locally and the `cross_link` is basically same with what we currently have.

PR #1416, opened 2017-02-09
Avoid errors due to mutations to options argument in the upload_file method

markmercedes:
I was getting this error while trying to upload objects to S3 using the SDK

```
Can't modify a frozen hash
```

This was happening because the `upload_file` method was deleting a key from the hash with the uploading options provided to the method (mutating the hash)

```Ruby
PUBLIC_PERMISSIONS = { acl: "acl: "public-read"}.freeze

object.upload_file('path/to/file.ext', PUBLIC_PERMISSIONS)
```

A code like the one shown above will generate the error mentioned before, this PR has the fix for it

Hope this helps

### DIFF
--- a/aws-sdk-resources/lib/aws-sdk-resources/services/s3/object.rb
+++ b/aws-sdk-resources/lib/aws-sdk-resources/services/s3/object.rb
@@ -245,10 +245,11 @@ module Aws
       #   without any errors.
       #
       def upload_file(source, options = {})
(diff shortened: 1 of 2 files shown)


cjyclaire:
Thank you for the contribution! I can reproduce the issue and your PR looks good to me.
I'll merge it, it will be tagged in our next release.

Issue #1419, opened 2017-02-11
How do I get a DBClusterSnapshot using just the db_cluster_snapshot_identifier?


lwoggardner:
For non-Aurora RDS snapshots I can do this..

```ruby
r = Aws::RDS::Resource.new(**client_args)
r.db_snapshots(db_snapshot_identifier: db_snapshot_identifier).first
```

but there seems no equivalent for Aurora cluster snapshots.

The only method to get a DBClusterSnapshot requires knowing the cluster_id, which for instance is not embedded DBClusterSnapshot.db_cluster_snapshot_arn

Note that ideally there would also be a method directly on RDS::Resource

```ruby
r = Aws::RDS::Resource.new(**client_args)
aurora_snapshot = r.db_cluster_snapshot(db_cluster_snapshot_identifier)
db_snapshot = r.db_snapshot(db_snapshot_identifier)
```

awood45:
You're right that we don't have the top level method for this in the resource abstractions yet, as far as I can see. We're happy to take a PR for it, otherwise can take on as a feature request to add.

Interestingly, there may be a bug in the `Aws::RDS::DBClusterSnapshot` object, as it requires the cluster ID itself. I'll look further in to that as well.

awood45:
I can pretty easily add a top-level `Aws::RDS::Resource#db_cluster_snapshots` method which accepts a snapshot ID to filter down to a collection of one, but it could be a breaking change to add the direct resource creation with only a single parameter. I'll make a branch for the WIP Resource PR.

mullermp:
Going to close this issue for now. Please reopen if this is still any concern. This is a breaking change.

PR #1420, opened 2017-02-13
Add DBClusterSnapshot to Aws::RDS::Resource

awood45:
Related to issue #1419

Still need to look at options for direct creation of a snapshot object.


stormsilver:
This would be really great to have. Any reason not to update this and merge it?

We are in the process of moving all of our environments from RDS Postgres -> RDS Postgres Aurora, and API parity is quite helpful!

mullermp:
This is a breaking change. We will not be merging this now.

Issue #1371, opened 2016-12-24
How can I use stub_responses with Aws::S3::Resource instead of Aws::S3::Client?


apurvis:
I have in the past been able to make very good use of `stub_responses` with `Aws::S3::Client`; however when it comes to `Aws::S3::Resource` I am at a bit of a loss as to how to actually stub anything.

As an example, I would like to stub a particular key as a response to `Aws::S3::Bucket.objects`. I have managed to get things to work in the sense of not exceptioning out with something like
```ruby
s3 = Aws::S3::Resource.new(stub_responses: true)
bucket = Aws::S3::Bucket.new(name: 'bucket', stub_responses: true)
```

But this doesn't work:
```ruby
bucket.stub_responses(objects: ['file'])

NoMethodError: undefined method `stub_responses' for #<Aws::S3::Bucket name="x">
	from (irb):140
	from /data/warehouse/shared/bundle/ruby/2.3.0/gems/railties-4.0.13/lib/rails/commands/console.rb:90:in `start'
	from /data/warehouse/shared/bundle/ruby/2.3.0/gems/railties-4.0.13/lib/rails/commands/console.rb:9:in `start'
	from /data/warehouse/shared/bundle/ruby/2.3.0/gems/railties-4.0.13/lib/rails/commands.rb:62:in `<top (required)>'
	from script/rails:6:in `require'
	from script/rails:6:in `<main>'
```

and neither does 
```
s3_stub.stub_responses(bucket: bucket_stub)
```

i tried to create a stub for `Aws::S3::Client` ahead of time with stubbed responses and pass that into the `Resource` constructor but that also didn't work:

```ruby
client_stub = Aws::S3::Client.new(region: 'region', stub_responses: true)
client_stub.stub_responses(:list_buckets, {
  buckets: [{ name: 'my-bucket' }]
})
resource_stub = Aws::S3::Resource.new(client_stub)
resource_stub.buckets
=> #<Aws::Resources::Collection type="Aws::S3::Bucket" limit=nil params={}>
```

and when i tried
```
resource_stub.list_buckets
``` 
i didn't actually get a stubbed response - i got a real response.

awood45:
I was able to get your example to work with a slight change:

```ruby
client_stub = Aws::S3::Client.new(stub_responses: true)
client_stub.stub_responses(:list_buckets, {
  buckets: [{ name: 'my-bucket' }]
})
resource_stub = Aws::S3::Resource.new(client: client_stub)
[...]

PR #1422, opened 2017-02-14
Add example of client_stubs with a Resource class

Iambchop:
Related to #1371


cjyclaire:
Looks good to me, thank you for contributing examples that could be helpful for other developers : )
I'll merge it.

Issue #1346, opened 2016-11-23
Aws::S3::Client.stub_responses doesn't stub metadata anymore?


porbas:
I'm in course of upgrading aws-sdk gems in my projects. Unfortunately my tests goes red :disappointed: 

In aws-sdk v 2.2.37 I could stub metadata:
```
piotrek@t530:~$ aws.rb _2.2.37_
Aws> s3 = Aws::S3::Client.new(stub_responses: true)
=> #<Aws::S3::Client>
Aws> s3.stub_responses :head_object, metadata: {'custom_key' => 'abc'}
=> [{:http=>
   #<Seahorse::Client::Http::Response:0x00000006269700
    @body=#<StringIO:0x00000006248398>,
    @complete=false,
    @done=nil,
    @error=nil,
    @headers={"x-amzn-requestid"=>"stubbed-request-id", "x-amz-meta-custom_key"=>"abc"},
    @listeners={},
    @status_code=200>}]
Aws> s3.head_object bucket: 'bucket', key: 'key'
[Aws::S3::Client 200 0.242652 0 retries] head_object(bucket:"bucket",key:"key")  
=> #<struct Aws::S3::Types::HeadObjectOutput
 delete_marker=nil,
 accept_ranges=nil,
 expiration=nil,
 restore=nil,
 last_modified=nil,
 content_length=nil,
 etag=nil,
 missing_meta=nil,
 version_id=nil,
 cache_control=nil,
 content_disposition=nil,
 content_encoding=nil,
 content_language=nil,
 content_type=nil,
 expires=nil,
 website_redirect_location=nil,
 server_side_encryption=nil,
 metadata={"custom_key"=>"abc"},
 sse_customer_algorithm=nil,
 sse_customer_key_md5=nil,
 ssekms_key_id=nil,
 storage_class=nil,
 request_charged=nil,
 replication_status=nil>
```

In current 2.6.28:
```
piotrek@t530:~$ aws.rb _2.6.28_
Aws> s3 = Aws::S3::Client.new(stub_responses: true)
=> #<Aws::S3::Client>
Aws> s3.stub_responses :head_object, metadata: {'custom_key' => 'abc'}
=> [{:http=>
   #<Seahorse::Client::Http::Response:0x00000006069ce8
    @body=#<StringIO:0x000000060465e0>,
    @complete=false,
    @done=nil,
    @error=nil,
[...]

PR #1425, opened 2017-02-16
No Response Body for Certain Stubbed Methods

awood45:
Resolves surface issue of #1346


awood45:
There's some room to ask why XML bodies are smashing the header value, when the API model clearly says metadata comes from the headers. This list could also be extracted or generated via other means, potentially.

PR #1427, opened 2017-02-17
Add :phone_number to SNS::Client#publish

maximebedard:
This PR patches SNS::Client#publish to also allow sending a PhoneNumber. This was already possible in v2, but backport the feature in v1 of the SDK.

This is documented in v2 here: http://docs.aws.amazon.com/sdkforruby/api/Aws/SNS/Client.html#publish-instance_method and on the original AWS documentation here http://docs.aws.amazon.com/sns/latest/api/API_Publish.html.

### DIFF
--- a/lib/aws/api_config/SNS-2010-03-31.yml
+++ b/lib/aws/api_config/SNS-2010-03-31.yml
@@ -325,6 +325,8 @@
     - :string
     MessageStructure:
     - :string
+    PhoneNumber:
+    - :string
   :outputs:
     :children:
       PublishResult:


awood45:
V1 of the SDK is end of life, so unfortunately we cannot accept further API updates to it as we won't be doing releases unless there is a security issue.

Keep in mind, you can use V1 and V2 in the same project, and upgrade piecemeal for calls like this. Let me know if you have any questions about how to do that.

maximebedard:
Got it, thanks for the quick response. I'll investigate toward upgrading the client to v2 👍

PR #1429, opened 2017-02-20
[Code Gen Doc] Fix yard template bug

cjyclaire:
Fix the grouping for service gems and register templates.
Removing the duplicate search context in layout.

@awood45 @trevorrowe


cjyclaire:
Update: 
    Looks like this fix works per gems only, still doesn't work when run `yardoc` from root, working on the fix.

cjyclaire:
Merging temple fixing first, will addressing yard tag grouping issue in a separate PR :)

Issue #1438, opened 2017-03-01
Add public API for clearing Seahorse::Client::NetHttp::ConnectionPool


iamatypeofwalrus:
**Metadata**
SDK Version: 2.3.x
Platforms: RHEL5, Amazon Linux

**Issue:**
When forking on linux based systems race conditions can arise between a parent process, child processes, and long-lived connections in `Seahorse::Client::NetHttp::ConnectionPool`. If a connection exists in the pool before a parent forks the child will also have access to this file descriptor and can possibly attempt to send a request on the connection at the same time another process is listening for a response.

SQS Example:

1. Parent process long polls `Aws::SQS::Client#receive_messages` and receives a batch of messages
2. Parent forks a child process for each message
3a. Parent process long polls `Aws::SQS::Client#receive_messages`
3b. Child process finishes processing and calls `Aws::SQS::Message.delete`

It is possible for 3a and 3b to utilize the same open file descriptor available in the `Seahorse::Client::NetHttp::ConnectionPool` and cause some unrelated error to be thrown.

Creating new instances of a client in the child process will not clear the ConnectionPool [since all of the pools are stored in a class variable.
](https://github.com/aws/aws-sdk-ruby/blob/master/aws-sdk-core/lib/seahorse/client/net_http/connection_pool.rb#L16)
 
I created an [example script](https://gist.github.com/iamatypeofwalrus/8105e291760c6235025a63fe65ed7705) that showcases this issue and it is based on a similar application that my team uses in production to process async jobs.

I was able to fix this by adding the following code after any fork:

```ruby
Seahorse::Client::NetHttp::ConnectionPool.pools.each do |pool|
  pool.empty!
end
```

As far as I can tell from the public documentation consumers are discouraged from relying on implementation details inside of Seahorse.

**Feature Request:**
Add a public API for clearing connection pools (`Aws.empty_connection_pools!`?) 

**Documentation Request:**
Add documentation around `fork`-ing best practices with respect to the Aws SDK

awood45:
Thanks for raising this. I think these requests make sense and I'm happy to either take a PR, and/or treat this as a relatively high priority feature request.

iamatypeofwalrus:
I'd like to take a stab at a pull request. Should land sometime this week.

awood45:
This will go out with the next release.

PR #1451, opened 2017-03-11
Add Aws.empty_connection_pools! API

iamatypeofwalrus:
This pull request is in response to #1438 and adds an empty connection pool API to the top level `Aws` name space.

### DIFF
--- a/aws-sdk-core/lib/aws-sdk-core.rb
+++ b/aws-sdk-core/lib/aws-sdk-core.rb
@@ -411,6 +411,9 @@ module Aws
     # there are no race conditions between the parent process and its children
     # for the pooled TCP connections.
     #
+    # Child processes that make multi-threaded calls to the SDK should block on
+    # this call before beginning work.
+    #
     # @return [nil]
     def empty_connection_pools!
       Seahorse::Client::NetHttp::ConnectionPool.pools.each do |pool|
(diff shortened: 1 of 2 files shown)


iamatypeofwalrus:
I wasn't quite sure how to test the pool without being too prescriptive of the underlying implementation. The spec file for ConnectionPool didn't have any tests for guidance :-(

iamatypeofwalrus:
Specs passed locally but not on CI. Taking a look.

aws-sdk-core/lib/aws-sdk-core.rb
```diff
@@ -403,6 +403,21 @@ def use_bundled_cert!
       ))
     end
 
+    # Close any long-lived connections maintained by the SDK's internal
+    # connection pool.
+    #
+    # Applications that rely heavily on the `fork()` system call on POSIX systems
+    # should call this method in the child process directly after fork to ensure
+    # there are no race conditions between the parent process and its children
+    # for the pooled TCP connections.
```
awood45: Should expand this to note that the method should return before making client calls in the child process. In theory, a threaded system might call this, start API calls before we clear, and have issues.

iamatypeofwalrus: Updated documentation in latest commit

Issue #1464, opened 2017-03-22
:http_response_body in log formatter errors with BlockIO


nate:
If the log formatter string includes `:http_response_body` and a logger is set a `NoMethodError` is raised.

Code to recreate the error:

```ruby
require "rubygems"
require "logger"
require "aws-sdk"

access_key_id = ""
secret_access_key = ""
region = ""
bucket = ""
key = ""

credentials = Aws::Credentials.new(access_key_id, secret_access_key)
log_formatter = Aws::Log::Formatter.new(":http_response_body")
logger = Logger.new(STDOUT)

client = Aws::S3::Client.new(region: region, credentials: credentials, log_formatter: log_formatter, logger: logger)

client.get_object(bucket: bucket, key: key){|_chunk| }
```

Backtrace:

```
/Users/nate/.rvm/gems/ruby-2.2.2@test-gemset/gems/aws-sdk-core-2.8.9/lib/seahorse/client/http/response.rb:45:in `body_contents': undefined method `rewind' for #<Seahorse::Client::BlockIO:0x007fb46a6e8068> (NoMethodError)
from /Users/nate/.rvm/gems/ruby-2.2.2@test-gemset/gems/aws-sdk-core-2.8.9/lib/aws-sdk-core/log/formatter.rb:174:in `_http_response_body'
from /Users/nate/.rvm/gems/ruby-2.2.2@test-gemset/gems/aws-sdk-core-2.8.9/lib/aws-sdk-core/log/formatter.rb:100:in `block in format'
from /Users/nate/.rvm/gems/ruby-2.2.2@test-gemset/gems/aws-sdk-core-2.8.9/lib/aws-sdk-core/log/formatter.rb:100:in `gsub'
from /Users/nate/.rvm/gems/ruby-2.2.2@test-gemset/gems/aws-sdk-core-2.8.9/lib/aws-sdk-core/log/formatter.rb:100:in `format'
from /Users/nate/.rvm/gems/ruby-2.2.2@test-gemset/gems/aws-sdk-core-2.8.9/lib/aws-sdk-core/plugins/logging.rb:58:in `format'
from /Users/nate/.rvm/gems/ruby-2.2.2@test-gemset/gems/aws-sdk-core-2.8.9/lib/aws-sdk-core/plugins/logging.rb:51:in `log'
[...]

PR #1469, opened 2017-03-24
Adding specific error message for response body that cannot be rewind

cjyclaire:
Addressing issue #1464 

This PR helps making debugging experience better with #1464 situation.
Additional thoughts: I choose to raise error at `#body_content` only because it could potentially break customers depending on `#body` if error is raised there as well.

@awood45


cjyclaire:
Closing, will open a new one instead

PR #1483, opened 2017-04-11
Fix Gitter badge not displaying properly

kriskhaira:
* Also update to latest format shown in the Gitter Room's settings page
* See http://blog.gitter.im/2014/01/27/badgers/


cjyclaire:
Thanks for the fix, I'll merge it : )

PR #1487, opened 2017-04-18
Require only cgi/util for escaping and unescaping

janko-m:
For CGI escaping it's not needed to require the whole cgi.rb standard library, it's enough to require only cgi/util.rb, which is almost 5x smaller.

While here we also remove superfluous cgi.rb requires, and add cgi/util.rb requires that were missing.


awood45:
This looks reasonable, thanks!

PR #1488, opened 2017-04-18
Set Thread#abort_on_exception on spawned threads

janko-m:
By default threads won't propagate exceptions that occur inside them to the main thread, unless we call `Thread#value` on a thread that exited with an exception, in which case its exception will be raised in the main thread. Because some folks might not know that last bit, for readability we explicitly set `Thread#abort_on_exception` on spawned threads.

While we're here, we also refactor `MultipartFileUploader` to be like `ObjectMultipartCopier`, so that we can make the same `Thread#abort_on_exception` update.


janko-m:
I will close this, as I don't feel too strongly about setting `Thread#abort_on_exception` to `true` anymore (I actually like the exception propagating only when we call `Thread#value`).

However, I think the first commit is nice to have because it brings consistency between `MultipartFileUploader` and `ObjectMultipartCopier` implementations. I will open a separate PR for that.

PR #1489, opened 2017-04-18
Allow string keys for required options

johnathanludwig:
This fix updates the parameter check to also see if the values are present under string keys in addition to symbols. With my testing the API calls still work correctly when string keys are used.

In some cases we have some symbols that end up getting converted to strings before being passed to the SDK calls. These work fine on all methods except ones with required options.

### DIFF
--- a/aws-sdk-core/lib/aws-sdk-core/param_validator.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/param_validator.rb
@@ -39,7 +39,7 @@ module Aws
       # ensure required members are present
       if @validate_required
         shape.required.each do |member_name|
-          if values[member_name].nil?
+          if values[member_name].nil? && values[member_name.to_s].nil?
             param = "#{context}[#{member_name.inspect}]"
             errors << "missing required parameter #{param}"
           end
--- a/aws-sdk-core/spec/aws/param_validator_spec.rb
+++ b/aws-sdk-core/spec/aws/param_validator_spec.rb
@@ -52,7 +52,7 @@ module Aws
         validate({})
       end
 
-      it 'raises an error when a required paramter is missing' do
+      it 'raises an error when a required parameter is missing' do
         shapes['StructureShape']['required'] = %w(String)
         validate({}, 'missing required parameter params[:string]')
       end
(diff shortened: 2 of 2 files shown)


awood45:
Can you elaborate more on the case which "works fine on all methods except ones with required options"? Would like to document/test that if taking this PR.

johnathanludwig:
Yeah, sorry. So this change fixes a call like this with DynamoDB for example:

```ruby
Aws::DynamoDB::Client.new.describe_table('table_name' => 'test')
# => ArgumentError: missing required parameter params[:table_name]
```

However optional params like with EC2 DescribeInstances already work:

```ruby
Aws::EC2::Client.new.describe_instances('filters' => [{ name: 'tag-key', values: ['key1'] }])
#<Aws::Structure:Aws::EC2::Types::DescribeInstancesResult
```

cjyclaire:
Thanks for the info, while from the example you provided, this PR does make sense. However I do have some concerns for allowing both styles at same time, it could potentials causing some weirdness around parameters values.
For example: 
if use one style solo ( as you provided ): 
This works fine 
```ruby
Aws::EC2::Client.new.describe_instances('filters' => [{ name: 'tag-key', values: ['key1'] }])
#<Aws::Structure:Aws::EC2::Types::DescribeInstancesResult
```
However
``` ruby
s3.put_object('bucket' => 'a-fancy-bucket', 'key' => 'wow', 'body' => 'hello')
# => Aws::S3::Errors::InvalidBucketName: The specified bucket is not valid.

# Yet
s3.put_object(bucket: 'a-fancy-bucket', key: 'wow', body: 'hello') 
# => #<struct Aws::S3::Types::PutObjectOutput
```
And if mixing those 2 styles together
``` ruby
s3.put_object(bucket: 'a-fancy-bucket', key: 'wow', 'body' => 'hello') 
# => succeed yet the `body` value is ignored
```
I think the behavior there need dig deeper, thoughts?

johnathanludwig:
Interesting. Is S3 just doing something different with the keys then? I haven't dug into all the code, but when I tested with DynamoDB it seemed to search correctly with the string key once I passed the required check that is in this PR.

Ultimate it is not a big deal if we can't accept this. On our end we have added a call to `symbolize_keys` on the arguments before we make the SDK call. It just seemed appropriate since these are hashes and both styles are common with them.

cjyclaire:
Thanks for the info. appreciate that, closing this right now, yet please feel free to reopen anytime if this becomes pain point for you or further questions exists : )

PR #1494, opened 2017-04-26
Adding support for Rds Auth Login

cjyclaire:
@awood45


awood45:
I'm sufficiently confident that this build failure is not related to this change, given that master has flipped from passing to failing independent of any change. While we pin down that issue, we can continue with this.

Issue #1497, opened 2017-04-27
CloudFormation::Client missing client_request_token support for create_stack/update_stack/delete_stack operations


bobziuchkovski:
The upstream CloudFormation API supports the `ClientRequestToken` param for the [CreateStack](http://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_CreateStack.html), [UpdateStack](http://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_UpdateStack.html), and [DeleteStack](http://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_DeleteStack.html) operations.

However, the latest published version of the aws-sdk for ruby (2.9.12) doesn't support the param for any of these operations.  Oddly enough, it _does_ support `client_token` for the [create_change_set](http://docs.aws.amazon.com/sdkforruby/api/Aws/CloudFormation/Client.html#create_change_set-instance_method) operation, though.

Supporting `client_request_token` for `create_stack`, `update_stack`, and `delete_stack` would be incredibly helpful for us.  We use the CloudFormation NotificationArn support to log all of the stack operations that occur on our CloudFormation stacks.  These notification SNS messages include the corresponding `ClientRequestToken` value when it is specified.  If we could pass it from the ruby sdk, we would be able to correlate our API requests with the corresponding stack events.

bobziuchkovski:
Hmm...it looks like I can make it work by editing `aws-sdk-core/apis/cloudformation/2010-05-15/api-2.json` and adding the `ClientRequestToken` shape and member definitions for those operations.  Is there anything else I'd need to update, or could I just submit those changes via a pull request?

cjyclaire:
@bobziuchkovski Thanks for reporting and appreciate that! This API update is definitely in our radar and will be delivered through all SDKs soon : ) Stay tuned : )

cjyclaire:
The API update has been included in the latest release, closing : ) Thank you for the patience.

bobziuchkovski:
@cjyclaire Thank you very much for updating that so quickly!!

PR #1499, opened 2017-04-27
Add ClientRequestToken support to CloudFormation::Client operations

bobziuchkovski:
Per https://github.com/aws/aws-sdk-ruby/issues/1497, the `ClientRequestToken` param isn't supported in the ruby sdk for any of the CloudFormation operations that support it.

I've taken a stab at adding support for this param.  I know @cjyclaire mentioned support would be coming soon, but I thought perhaps I could expedite "soon".  😄 

As far as I can tell, this shouldn't cause any breakages / backwards incompatibility, though I noticed the "We may choose not to accept pull requests that change service descriptions" note in `CONTRIBUTING.md`.  It looks harmless to me, but I'm not familiar with the code base at all.

### DIFF
--- a/aws-sdk-core/apis/cloudformation/2010-05-15/api-2.json
+++ b/aws-sdk-core/apis/cloudformation/2010-05-15/api-2.json
@@ -366,7 +366,8 @@
       "type":"structure",
       "required":["StackName"],
       "members":{
-        "StackName":{"shape":"StackName"}
+        "StackName":{"shape":"StackName"},
+        "ClientRequestToken":{"shape":"ClientRequestToken"}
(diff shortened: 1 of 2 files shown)


awood45:
The reasoning for this is that we have a central source of these JSON files, and need to keep them in sync with service teams. In the case of `api-2.json` and `docs-2.json`, we cannot accept pull requests. Instead, we would reach out to the relevant owners for updates. That's what we've done here. I do appreciate your efforts and I'll push to get this supported ASAP.

bobziuchkovski:
Okay, thank you for the clarification.

Issue #1510, opened 2017-05-06
Aws::EC2::RouteTable#routes has nil entry when IPv6 route exists


cunnie:
Behavior: Aws::EC2::RouteTable#routes has nil entry instead of IPv6 route. IPv4 route is present & accounted for.

Output from `pry-byebug` session:
```
[9] pry(#<Bosh::AwsCloud::Instance>)> table
=> #<Aws::EC2::RouteTable id="rtb-52235b36">
[10] pry(#<Bosh::AwsCloud::Instance>)> table.routes
=> #<Aws::Resources::Batch resources=[#<Aws::EC2::Route route_table_id="rtb-52235b36", destination_cidr_block="10.0.0.0/16">, nil]>
```
Screenshot of Route Table via GUI

<img width="614" alt="routes" src="https://cloud.githubusercontent.com/assets/1020675/25768517/594765de-31ba-11e7-8a92-65f4ac0cde6e.png">

Expected value of `table.routes`:
```
=> #<Aws::Resources::Batch resources=[#<Aws::EC2::Route route_table_id="rtb-52235b36", destination_cidr_block="10.0.0.0/16">, #<Aws::EC2::Route route_table_id="rtb-52235b36", destination_ipv_6_cidr_block="2600:1f1c:591:f000::/56"]>
```
Version of SDK **2.9.15**, from `Gemfile.lock`:
```
    aws-sdk (2.9.15)
      aws-sdk-resources (= 2.9.15)
    aws-sdk-core (2.9.15)
      aws-sigv4 (~> 1.0)
      jmespath (~> 1.0)
    aws-sdk-resources (2.9.15)
      aws-sdk-core (= 2.9.15)
    aws-sigv4 (1.0.0)
```

awood45:
Can you provide a wire trace of your call? When you create a client or resource, just add a new parameter:

```ruby
client = Aws::S3::Client.new(http_wire_trace: true)
resource = Aws::S3::Resource.new(http_wire_trace: true)
```

That will help us diagnose the discrepancy.

cunnie:
Hi @awood45 ,

I've posted a gist of the trace [here](https://gist.github.com/cunnie/1aa747751aa68aaab97932df38883677#file-route_table_trace-txt-L409). Line 409 seems to be the point of interest, and it seems to have the proper information.

[...]

PR #1513, opened 2017-05-08
Update Ec2 resource model identifier

cjyclaire:
Fix issue #1510 

Once EC2 VPC added support for IPv6 with `destination_ipv_6_cidr_block`, `destination_cidr_block` is not the identifier for route of a routing table anymore. For one route of a routing table, either of these 2 parameter will be nil.

Tested with:
``` ruby
table = Aws::EC2::RouteTable.new(id: "rtb-1234abcd")
res = table.routes
res.each do |r|
   puts "Ipv4: "
   puts r.destination_cidr_block
   puts "Ipv6: "
   puts r.destination_ipv_6_cidr_block
end
```

@awood45

### DIFF
--- a/aws-sdk-core/apis/ec2/2016-11-15/resources-1.json
+++ b/aws-sdk-core/apis/ec2/2016-11-15/resources-1.json
@@ -1444,11 +1444,7 @@
     },
     "Route": {
       "identifiers": [
-        { "name": "RouteTableId" },
-        {
-          "name": "DestinationCidrBlock",
-          "memberName": "DestinationCidrBlock"
-        }
+        { "name": "RouteTableId" }
       ],
       "shape": "Route",
       "actions": {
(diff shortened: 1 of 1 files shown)


cjyclaire:
I'm hesitate to merge this because now `destination_cidr_block ` is no longer required, thus `Routes` resource doesn't have an suitable identifier per `Route` anymore, which even though pass V2 resource structure yet looks like:
```
=> #<Aws::Resources::Batch resources=[#<Aws::EC2::Route route_table_id="rtb-123">, #<Aws::EC2::Route route_table_id="rtb-123">]>
```
And this will cause problem for the code-gen version, and it makes sense since there is a resource without identifier.

cunnie:
I defer to your wisdom as far as whether to merge the pull request. We've inserted a nil-check to fix our problem, so we're okay on our end.

PR #1523, opened 2017-05-21
S3: Refactor MultipartFileUploader to be like ObjectMultipartCopier

janko:
This aim of this PR is to bring consistency between the two implementations.


aws-sdk-resources/lib/aws-sdk-resources/services/s3/multipart_file_uploader.rb
```diff
@@ -158,7 +159,7 @@ def part_size(total_size, part_size, offset)
       end
 
       # @api private
-      class PartList
+      class PartQueue
```
janko-m: Only because the `MultipartObjecCopier` uses  the name `PartQueue`. I reverted it back to `PartList`.

janko:
I don't feel strongly about this anymore, both kinds of implementations are good 👍

PR #1529, opened 2017-05-31
Remove unnecessary newline escapes.

gddc:


awood45:
I'm actually inclined to accept this. Normally we don't because we source this from shared models, but:

* Fixing it is on the backlog.
* It's just a documentation change, so no harm if it reverts.

PR #1530, opened 2017-06-07
Add support for s3 host id

cjyclaire:
Adding support for all request ids for s3: http://docs.aws.amazon.com/AmazonS3/latest/dev/troubleshooting.html

@awood45

### DIFF
--- a/aws-sdk-core/lib/aws-sdk-core/plugins/s3_host_id.rb
+++ b/aws-sdk-core/lib/aws-sdk-core/plugins/s3_host_id.rb
@@ -1,0 +1,26 @@
+module Aws
+  module Plugins
+
+    # Support S3 host id, more information, see:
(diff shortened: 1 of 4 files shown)


aws-sdk-core/lib/aws-sdk-core/plugins/s3_host_id.rb
```diff
@@ -0,0 +1,26 @@
...
+    # This plugin adds :host_id for s3 responses when available
+    # @api private
+    class S3HostId < Seahorse::Client::Plugin
+
+      class Handler < Seahorse::Client::Handler
+
+        def call(context)
+          response = @handler.call(context)
+          h = context.http_response.headers
+          context[:host_id] = h['x-amz-id-2']
```
awood45: Given that this field is S3-specific (especially in case other such fields come up for other services in the future), perhaps a name like `:s3_host_id` would be appropriate.

cjyclaire:
Updated with `:s3_host_id`, will merge once CI test finished

PR #1533, opened 2017-06-15
[Version 3 - Preview] Switch the order of `aws-sdk` and `aws-sdk-resources` gems

cjyclaire:
As we discussed off-line. Now `aws-sdk` gem is a wrapper of `aws-sdk-resources` gem.

@awood45 

Once this gets merged, I'll update README.md etc. accordingly.


gems/aws-sdk-resources/CHANGELOG.md
```diff
@@ -1,13 +1,85 @@
 Unreleased Changes
```
awood45: This is probably something to revisit - we want this changelog at the top level, despite the disconnect in the dependency chain.

I believe this is also related to the test failure this PR creates.

gems/aws-sdk-resources/VERSION
```diff
@@ -1 +1 @@
-3.0.0.rc1
+3.0.0.rc8
```
awood45: This isn't really right - we should reverse the code, but the release count is still the release count.

cjyclaire:
@awood45 Changelog is  fixed, README.md is updated and build task is updated as well : )

PR #1539, opened 2017-06-20
Document dynamic S3 stubbing

janko-m:
I found out about this feature when I was looking at the test suite for aws-sdk itself, and it helped me immensely when I was rewriting test suite for [Shrine](https://github.com/janko-m/shrine/blob/90ea7693e6ac32c0bf361b041d9e9cdfaf8c24d0/test/s3_test.rb) and [tus-ruby-server](https://github.com/janko-m/tus-ruby-server/blob/ff3890b5d05b7a05f904488a880c1365de46d034/test/s3_test.rb) to use stubs, because I was able to test the parameters which the operations were being called with, and dynamically create new stubs based those parameters.


cjyclaire:
@janko-m This is cool, looks great, thanks for the contribution!

PR #1545, opened 2017-06-23
Reduce memory allocation when generating checksums and signatures

janko-m:
When IO is #read without a buffer string, it allocates a new string for each read chunk. Even though by reading the file in chunks we keep memory usage low, we still give GC work to clean up all those strings.

However, if we pass an additional buffer string to #read, instead of allocating new string for each chunk, each chunk will be read into the existing buffer string (replacing any previous value). That means that the buffer string is the only string that's allocated.

The IO body already has to accept the additional buffer argument, because Net::HTTP uses IO.copy_stream to stream the IO to the TCP socket, and IO.copy_stream will throw an exception if the IO body only accepts the length argument.

### DIFF
--- a/gems/aws-sdk-s3/lib/aws-sdk-s3/file_part.rb
+++ b/gems/aws-sdk-s3/lib/aws-sdk-s3/file_part.rb
@@ -56,14 +56,12 @@ module Aws
       end
 
       def read_from_file(bytes, output_buffer)
-        if bytes
-          data = @file.read([remaining_bytes, bytes].min)
-          data = nil if data == ''
-        else
-          data = @file.read(remaining_bytes)
-        end
+        length = [remaining_bytes, *bytes].min
+        data   = @file.read(length, output_buffer)
+
         @position += data ? data.bytesize : 0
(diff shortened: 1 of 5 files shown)


sorah:
Note: #1545 contains a same change

janko-m:
As @sorah has already discovered in https://github.com/aws/aws-sdk-ruby/pull/1516, the `FilePart#read` method had incorrect behaviour, which doesn't match Ruby's `IO#read`. This is the relevant bit of `IO#read` behaviour:

* when length argument is not given and file is at EOF, return `""`
* when length argument is given and file is at EOF, return `nil`

`FilePart#read` respects this behaviour when output buffer is *not* used, but it doesn't when it is used; in that case it returns `""` on EOF even if length argument was given. This would cause an endless loop when generating checksums.

Could this be merged now?

awood45:
This looks good, but before I can merge, a quick piece of administrivia:

Can you confirm: "By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license."

janko-m:
Sure thing.

PR #1546, opened 2017-06-23
Use Digest#base64digest method

janko:
Other than being shorter, it has two qualities over `Base64.encode64` that are desirable in this case. One is that it doesn't include a trailing newline at the end of the result. Other one is that it doesn't wrap the result in 60 columns, meaning it doesn't add a newline after each 60 characters.

A general alternative to `Base64.encode64` is `Base64.strict_encode64`, which should have the same desirable behaviour, but calling `#base64digest` is shorter.

### DIFF
--- a/gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/md5s.rb
+++ b/gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/md5s.rb
@@ -1,5 +1,4 @@
 require 'openssl'
-require 'base64'
 
 module Aws
   module S3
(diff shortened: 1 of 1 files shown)


awood45:
Same administrivia on this one, while I rerun the failed test case can you confirm: "By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license."

janko-m:

I'm not sure what's making this change fail on JRuby. Maybe we should try upgrading to a newer version.

mullermp:
Hey @janko - it looks like this change was never shipped. I'm sorry about that. I would like to merge this in.

My one question is: Will the `encode64` method be breaking for consumers who are expecting line feeds? My first thought is that, for true backwards compatibility, we should prefer `strict_encode64`.

janko:
I think people shouldn't generally be expecting line feeds in request headers (which this MD5 digest is being used for). I would argue that having line feeds in request headers could already be problematic.

`Base64.strict_encode64` and `MD5#base64digest` should IMO have the same output, but I guess it's safer to use `Base64.strict_encode64`. In that case the PR can be closed, as the point of `MD5#base64digest` was to simplify code and remove a string allocation.

mullermp:
That makes sense to me. I like the code simplification, however JRuby seems to fail on the CI - some values come back nil on tests. I can try to make these changes myself (and close this PR) or would you like to try?

janko:
Feel free to make the changes and close this PR 👍

mullermp:
Sounds good. I will keep this open until then. I am new around here - I see you are (or were) a regular contributor :). Let me know if you need anything.

mullermp:
JRuby is giving me trouble on these. When trying to reproduce, even without the changes, it fails anyway. I think that they're already an allowed failure in master for travis, and the build doesn't even get past the bundle step. I am going to merge these regardless.

PR #1548, opened 2017-06-23
[Version 3 Preview] Add/Port Over Shared Examples support for V3

cjyclaire:
Have tested against all current example generation : )

@awood45


cjyclaire:
The failure is caused by forgetting add changelog entry for this change, adding changelog right now ...

cjyclaire:
Merging this right now, this only involves documentation changes and I have verified that the generated examples are the same with current v2 examples.

PR #1561, opened 2017-07-28
Ensure SpotInstanceRequestFulfilled waiter retries on request not found

BRMatt:
The EC2 spot instance API appears to be eventually consistent.

If you make a `DescribeSpotInstanceRequests` API call immediately after
a `RequestSpotInstances` call then you may get a
`InvalidSpotinstanceRequestID.NotFound` error.

The waiter should continue waiting if it encounters this error.

It seems there are no tests for the contents of this file, so I presume this is ok?

See issue for more details.

fixes aws/aws-sdk-ruby#1559

### DIFF
--- a/aws-sdk-core/apis/ec2/2016-11-15/waiters-2.json
+++ b/aws-sdk-core/apis/ec2/2016-11-15/waiters-2.json
@@ -413,6 +413,11 @@
           "matcher": "pathAny",
           "argument": "SpotInstanceRequests[].Status.Code",
           "expected": "system-error"
+        },
+        {
+          "state": "retry",
+          "matcher": "error",
+          "expected": "InvalidSpotInstanceRequestID.NotFound"
         }
       ]
     },


BRMatt:
I tried out this patch on our system, it seems to handle the error correctly and waits until the instance is available:

image 2017-07-31 at 12 18 13 pm

image 2017-07-31 at 12 21 19 pm

cjyclaire:
Thanks for the contribution! Will pull this upstream to share with all SDKs and CLI and release from there : )

BRMatt:
Oh so the json files are auto generated?

awood45:
The ones you've changed have a "single source of truth" outside this repo, to ensure the SDKs are in sync.

BRMatt:
Gotcha, makes sense. Any idea on when that might happen? Just so I know when we can stop using our fork.

cjyclaire:
@BRMatt Should be today ; )

cjyclaire:
Release contains the change is out, resolving : )

PR #1573, opened 2017-08-16
add a space to address binary operator warning

fallwith:
add a space to address warnings such as:

```
aws-sdk-resources-2.7.14/lib/aws-sdk-resources/documenter/has_many_operation_documenter.rb:7: warning: `+' after local variable or literal is interpreted as binary operator
aws-sdk-resources-2.7.14/lib/aws-sdk-resources/documenter/has_many_operation_documenter.rb:7: warning: even though it seems like unary operator
```

these warnings are thrown at require time by every JRuby app that requires the gem


cjyclaire:
Looks good, thank you for the contribution! Will be tagged in the next release

PR #1576, opened 2017-08-21
Make travis wait for running jruby test

cjyclaire:
Travis jruby suit for code-gen is failing due to timeout, the PR proposes a fix.

Reference: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,8 @@ matrix:
       env: OLD_OJ=1
     - rvm: 2.3.0
       env: OLD_OX=1
+  allow_failures:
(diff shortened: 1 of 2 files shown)


cjyclaire:
Hmm it does look like `wait` is not the proper fix there, I tried same version jruby test locally, it all passed instantly.

cjyclaire:
Finally I have located the cause ... it has to do with jruby parsing very long strings hanging ...
Closing this one, will open another one against master

Issue #1583, opened 2017-08-30
No Changelog for version 3


greysteil:
Version 3.0.0 was released on [Rubygems](https://rubygems.org/gems/aws-sdk) yesterday, but there's no changelog entry for it and `master` on this repo still points to [v2.10.35](https://github.com/aws/aws-sdk-ruby/blob/master/VERSION). Can we get an update on what's going on, and where the code for version 3.0.0 is?

menno:
After checking some branches it seems to be located in this branch: https://github.com/aws/aws-sdk-ruby/tree/code-generation

awood45:
We're moving V3 to the master branch soon, we have to coordinate that change with some internal tooling. For now V3 lives in code-generation, correct.

jamtur01:
No release tag either.

greysteil:
Thanks for the update. Would be awesome if you could add the above details to the changelog, or a GitHub release - at the moment anyone doing a bundle update without a pinned version will be getting V3 but won't be able to find details of what's changed / currently happening easily.

(My particular interest is that I run [Dependabot](https://dependabot.com) which automatically creates PRs to bump dependencies - we've had a few people ask if there's been an error on our side because they haven't been able to find any details of the new version in this repo!)

awood45:
Release tags are a more interesting issue now that we have modularized into 100+ gems. `aws-sdk` will probably be on `3.0.0` for a long time, it's the underlying service gems which will be updating. I'm open to ideas about how to add dozens of new GitHub releases/tags to this repo without it getting overwhelming.

The CHANGELOGs in the `code-generation` branch are current and will soon be in `master`. Similarly, the V2 CHANGELOG will continue to update once it has moved to a branch.

greysteil:
Interesting. Release notes aren't going to work for you with the above setup, but you should be able to get Rubygems to point to the right changelogs, which will help a lot - I'll create a PR for that now.

awood45:
The use case you have will be interesting with V3. We actually will be having hundreds of CHANGELOGs, and users will be migrating to only using a subset of service gems relevant to their use cases. It's unfortunately impractical for us to create repos for each gem, so a "single" CHANGELOG doesn't fit well going forward. Happy to discuss what we do have, or if there are ways to make our multiple changelogs easier to consume.

greysteil:
Got a solution for you on that one: https://github.com/aws/aws-sdk-ruby/pull/1585

awood45:
[...]

PR #1584, opened 2017-08-30
Add details of version 3.0.0 to changelog

greysteil:
Follow-up from https://github.com/aws/aws-sdk-ruby/issues/1583.

Whilst V2 is on master, it makes sense to have a link in the changelog to the code for V3 (for anyone who has just installed V3 and comes looking for it. If/when V3 becomes master, this changelog entry could be deleted (from the v2 branch).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@ Unreleased Changes
 
 * Issue - Aws::CloudHSM - This has been deprecated, remove smoke test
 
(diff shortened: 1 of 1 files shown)


greysteil:
@awood45 - sensible compromise?

awood45:
I'm fine with this - I agree that we will delete this after we switch branches, but happy to bridge things for convenience. Let's keep the conversation going in the other thread as well.

PR #1585, opened 2017-08-30
Tell Rubygems correct sourcecode and changelog links

greysteil:
Add metadata to each gem's `gemspec` which tells Rubygems where to find its source code and changelog.


awood45:
This is awesome, thanks! We are going to look at adding this into our generation process, as you stated, so this PR may be on hold for a bit but I'm strongly inclined to do this in general.

cjyclaire:
Thanks for the contribution! Just a soft heads-up, I'll take the work from there and do a bit refractor before merge after we switch to master shortly.

greysteil:
👌

cjyclaire:
Closing, PR #1586 is under reviewed, tracking with #1586 instead.

Issue #1588, opened 2017-08-31
Regression: S3 object streaming is not working anymore


janko-m:
### Issue description

In aws-sdk V2 you could stream an object from S3 by passing a block:

```rb
object.get { |chunk| puts chunk }
# or
client.get_object({...}) { |chunk| puts chunk }
```

In aws-sdk-s3 this doesn't work anymore, even though it's still [documented](https://github.com/aws/aws-sdk-ruby/blob/dad6433b6c4940442b2eeda431c15f3cc576e94f/gems/aws-sdk-s3/lib/aws-sdk-s3/client.rb#L2526-L2532). It doesn't work because [`Aws::S3::Object#get`](https://github.com/aws/aws-sdk-ruby/blob/dad6433b6c4940442b2eeda431c15f3cc576e94f/gems/aws-sdk-s3/lib/aws-sdk-s3/object.rb#L557-L564) and [`Aws::S3::Client#get_object`](https://github.com/aws/aws-sdk-ruby/blob/dad6433b6c4940442b2eeda431c15f3cc576e94f/gems/aws-sdk-s3/lib/aws-sdk-s3/client.rb#L2595-L2598) don't handle a given block anymore.

### Gem name

The problem is occurring on `aws-sdk-s3` 1.0.0.

### Version of Ruby, OS environment

```
$ ruby -v
ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-darwin16]
$ uname -a
Darwin Jankos-MacBook-Pro-2.local 16.7.0 Darwin Kernel Version 16.7.0: Thu Jun 15 17:36:27 PDT 2017; root:xnu-3789.70.16~2/RELEASE_X86_64 x86_64
```

### Code snippets / steps to reproduce

```rb
require "aws-sdk-s3"

resource = Aws::S3::Resource.new(
  access_key_id:     "...",
  secret_access_key: "...",
  region:            "...",
)
client = resource.client
bucket = resource.bucket("my-bucket")

object = bucket.object("foo")
object.put(body: "content")

# Both of these versions yield chunks in aws-sdk V2, but not in aws-sdk-s3
object.get { |chunk| puts chunk }
client.get_object(bucket: "my-bucket", key: "foo") { |chunk| puts chunk }
```

awood45:
I am able to reproduce this error. Taking a look.

cjyclaire:
[...]

PR #1591, opened 2017-08-31
Update streaming operations with block streaming ability

cjyclaire:
Address #1588 

Tested gem generation and manual tested behavior locally.


cjyclaire:
@awood45 Ready for another review!

I added spec test for a mock service with operations that's similar to s3, glacier and polly. The mock `api.json` also cover the the corner cases I met when locating correct streaming shape. I can add another service smoke test if you recommed ; )

Issue #1592, opened 2017-09-01
Aws::VERSION is not up


k1LoW:
### Issue description

Aws::VERSION is not up

### Gem name ('aws-sdk', 'aws-sdk-resources' or service gems like 'aws-sdk-s3') and its version

- aws-sdk-core

### Version of Ruby, OS environment

```sh
$ ruby -v
ruby 2.3.3p222 (2016-11-21 revision 56859) [x86_64-darwin15]
```

### Code snippets / steps to reproduce

https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/lib/aws-sdk-core/version.rb#L2

cjyclaire:
@k1LoW Thanks for catching, working on fix this

cjyclaire:
A side note is that for core version, it's tracking [VERSION](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/VERSION) file instead, which has the correct version, I'll remove the redundant one.

cjyclaire:
PR has been merged, closing : )

JaganNarayanan:
Sorry, newbie here. I get this error 'uninitialized constant Aws::VERSION' when loading my page after using the gem 'aws-sdk'. How do I remove the version file? Will it fix this? Thank you in advance!

cjyclaire:
@JaganNarayanan Could you provide more context? Which version of the gem are you on? I tried install the `aws-sdk` gem from scratch and tried several API calls, I cannot reproduce the error here.

JaganNarayanan:
Gemfile 
`gem 'aws-sdk'
`
config/environments/development.rb 
`  config.paperclip_defaults = {
      storage: :s3,
      s3_region: ENV["aws_bucket_region"],
      s3_credentials: {
        # s3_host_name: ENV["aws_bucket_name"],
        bucket: ENV["aws_bucket_name"],
        access_key_id: ENV["amazon_key"],
        secret_access_key: ENV["amazon_secret_access_key"]
        }
      }`

config/environment/production.rb
`  config.paperclip_defaults = {
  :storage => :s3,
  :preserve_files => true,
  :bucket => ENV['aws_bucket_name']
}`

That was my code. I also tried using older versions of the gem, 1 & 2 but I get the same version error. Any help would be much appreciated, thanks!

easybills-admin:
Possible side effect: https://github.com/thoughtbot/paperclip/issues/2484

PR #1595, opened 2017-09-01
Remove redundant version file at core gem

cjyclaire:
As described, address #1592 

The core gem version is [here](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/VERSION), and extracted [here](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/lib/aws-sdk-core.rb#L71)

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core/version.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/version.rb
@@ -1,3 +1,0 @@
-module Aws
-  VERSION = '2.8.2'
-end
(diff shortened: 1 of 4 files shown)


cjyclaire:
Added changelog entry, the only failure is the jruby hanging problem, which has been fixed and merged.
Merging this right now.

gems/aws-sdk-core/lib/aws-sdk-core/version.rb
```diff
@@ -1,3 +0,0 @@
-module Aws
-  VERSION = '2.8.2'
-end
```
easybills-admin: Possible side effect: https://github.com/thoughtbot/paperclip/issues/2484

Issue #1603, opened 2017-09-07
NameError: undefined local variable or method 'v' for Aws::Util:Module


tarcieri:
Occurring here:

https://github.com/aws/aws-sdk-ruby/blob/734a23dfb940c4dfa8e2fb8fb6056d947bfef20a/gems/aws-sdk-core/lib/aws-sdk-core/util.rb#L34

It's likely `v` should be `obj` (or perhaps `obj.dup`/`obj.clone`?). I can send a PR...

cjyclaire:
@tarcieri Ah thanks for catching! exactly, happy to review the PR if possible, if you don't have time, let me know and I can take this too.

tarcieri:
If I were to do a PR, I guess my first question would be what is the correct behavior here re: `dup`/`clone`?

cjyclaire:
@tarcieri Taking a second look, this `:deep_copy` is actually not used else where in the sdk, you can refer the `:deep_copy` we have at code generation [here](https://github.com/aws/aws-sdk-ruby/blob/734a23dfb940c4dfa8e2fb8fb6056d947bfef20a/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/helper.rb#L152).

tarcieri:
I encountered this error attempting to invoke `batch_create_tags` on an `Aws::EC2::Instance`:

```
NameError: undefined local variable or method `v' for Aws::Util:Module
  /home/tony/.gem/ruby/2.4.0/gems/aws-sdk-core-3.0.0/lib/aws-sdk-core/util.rb:34:in `deep_copy'
  /home/tony/.gem/ruby/2.4.0/gems/aws-sdk-core-3.0.0/lib/aws-sdk-core/util.rb:28:in `block in deep_copy'
  /home/tony/.gem/ruby/2.4.0/gems/aws-sdk-core-3.0.0/lib/aws-sdk-core/util.rb:27:in `each'
  /home/tony/.gem/ruby/2.4.0/gems/aws-sdk-core-3.0.0/lib/aws-sdk-core/util.rb:27:in `inject'
  /home/tony/.gem/ruby/2.4.0/gems/aws-sdk-core-3.0.0/lib/aws-sdk-core/util.rb:27:in `deep_copy'
  /home/tony/.gem/ruby/2.4.0/gems/aws-sdk-core-3.0.0/lib/aws-sdk-core/util.rb:32:in `block in deep_copy'
[...]

PR #1604, opened 2017-09-08
Fix batch operation behavior

cjyclaire:
Address #1603 

Also add a fix to batch operation, add related tests.


cjyclaire:
Have manually tested EC2 resource batch operations and Autoscaling batch operation, will merge once CI completed

cjyclaire:
The only failure is caused by ruby 1.9.3 kramdown dependency issue, which is addressed at #1605

Issue #1607, opened 2017-09-11
NoMethodError - undefined method `last_modified' for #<Hash:0x1e8b16d2>


PetrKaleta:
`/.rbenv/versions/jruby-9.1.13.0/lib/ruby/gems/shared/gems/aws-sdk-s3-1.2.0/lib/aws-sdk-s3/object_summary.rb:43:in `last_modified'`

Please have a look in `ObjectSummary` class...

```ruby
    # @return [Time]
    def last_modified
      data.last_modified
    end

    # @return [String]
    def etag
      data.etag
    end

    # @return [Integer]
    def size
      data.size
    end
```

but `data` is a hash...

cjyclaire:
Taking a look

cjyclaire:
@PetrKaleta Hmm, this sounds like a bug, could you help provide some code snippet to help me reproduce the issue?
I tried to reproduce in the simplest way:
```ruby
obj = Aws::S3::Resource.new.bucket('a-fancy-bucket').object('hello')
obj.last_modified
# => 2017-08-31 16:37:14 +000
```
And this works fine for me.

PetrKaleta:
This code was working without any issues until V3 was introduced

``` ruby
      def structure_at(bucket:, path: nil)
        bucket.client.list_objects_v2 prefix: path,
                                      bucket: bucket.name,
                                      delimiter: '/'
      end

      def files_at(bucket:, path: nil)
        structure = structure_at bucket: bucket, path: path

        structure
          .contents
          .map do |object|
            next if object.key.end_with? '/'

            Aws::S3::ObjectSummary.new bucket_name: bucket.name,
                                       key: object.key,
                                       data: object.to_h
          end
          .compact
      end
```

So it looks like the weak point is data attribute in `ObjectSummary` constructor...

cjyclaire:
@PetrKaleta Ah sorry about the confusion, I should try with object summary instead. I can reproduce this, looks like this will work if remove `.to_h` for the object in `:data` field. Double checking v2 behavior now before adding back compatiblity

cjyclaire:
@PetrKaleta A quick follow up question, may I ask for the documentation source that lead you to use `.to_h` in v2? Just trying to make sure I covered all in my fix PR.

Adding notes here: at V2, passing an object or hash both works.

PetrKaleta:
@cjyclaire Hmmmmm, I think I found it directly in the source code....

cjyclaire:
[...]

PR #1608, opened 2017-09-11
Allow resource constructor taking data hash

cjyclaire:
Addressing #1607 

To keep v2 compatibility, allow resource object taking data hash, spec test is added as well.

Some sample generated diff, e.g s3:
image

### DIFF
--- a/build_tools/aws-sdk-code-generator/spec/interfaces/resources_spec.rb
+++ b/build_tools/aws-sdk-code-generator/spec/interfaces/resources_spec.rb
@@ -309,6 +309,16 @@ describe 'Interfaces' do
           expect(ticket.purchase_price).to be(10)
         end
 
+        it 'allows data to be provided as hash' do
+          data = client.stub_data(:get_ticket, ticket:{purchase_price:10}).ticket
+          expect(client).not_to receive(:get_ticket)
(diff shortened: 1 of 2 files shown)


cjyclaire:
Marking this as WIP, trying to simply the diff

cjyclaire:
Soft update ... I'm working on switching struct style usage everywhere in resource since changing data attribute along won't work if customer depending on this behavior for all resource functionalities ...

Adding some thoughts here, an alternative fix is per every resources class, convert hash data when initializing like:
```ruby
data = options.delete(:data)
@data = data.is_a?(Hash) ? Struct.new(*data.keys).new(*data.values) : data
```
We don't need to consider nested hashes because in V2 [code](https://github.com/aws/aws-sdk-ruby/blob/version-2/aws-sdk-resources/lib/aws-sdk-resources/resource.rb#L222), same hash initializing usage will only work one layer as well. Pros for this approach is that we keep the consistency in using structs and cut resources diffs. The only drawback is that if customer is passing `:data` a hash directly and call `.data` later, struct will be returned instead of the original hash.

cjyclaire:
Observing some weird git issue locally, will open a new one to address this.

Issue #1610, opened 2017-09-12
Undefined method `wait_until' in v3 SDK


ajkerr:
Please fill out the sections below to help us address your issue

### Issue description

We have uncovered a backwards-compatibility problem after attempting to upgrade from the v2 SDK to the v3 version.

Methods such as wait_until no longer exist on classes such as Volume. This issue appears to be related to the fact that classes such as Volume extended Resource in v2, and now extend Object in v3.

### Gem name ('aws-sdk', 'aws-sdk-resources' or service gems like 'aws-sdk-s3') and its version

aws-sdk 3.0.1

### Version of Ruby, OS environment

ruby 2.4.1p111 on macOS (via Homebrew)

### Code snippets / steps to reproduce

This code works in v2, but fails in v3 with the following message:

```
ruby-sdk-3-test.rb:13:in `<main>': undefined method `wait_until' for #<Aws::EC2::Volume:0x007f96f834a378> (NoMethodError)
```

```ruby
require "aws-sdk"

ec2 = Aws::EC2::Resource.new(region: 'us-west-2')

volume = ec2.create_volume({
  size: 1,
  availability_zone: 'us-west-2a',
  volume_type: 'gp2'
})
puts "Created new volume: #{volume.id}"

volume.wait_until() {|volume| volume.state == 'available' }
puts "Volume #{volume.id} is now available."
```

cjyclaire:
@ajkerr Thanks for reporting, taking a look

cjyclaire:
Soft update, I can reproduce this, working on fixing the regression, in the meantime, you can use ec2 client waiters as work around like:
```ruby
client.wait_until(:volume_available, volume_ids:['vol-0b123323123123'])
```

cjyclaire:
The fix PR has been merged, will ping the thread once release is out : )

cjyclaire:
The release is [out](https://rubygems.org/gems/aws-sdk-ec2/versions/1.6.0), closing : )

PR #1614, opened 2017-09-13
Add Resource#wait_until regression with deprecate warnings

cjyclaire:
Address #1610 

Some background, in V2 those resource [objects](http://docs.aws.amazon.com/sdkforruby/api/Aws/EC2/Volume.html) all inherits [`Aws::Resources::Resource`](http://docs.aws.amazon.com/sdkforruby/api/Aws/Resources/Resource.html). This [`#wait_until`](https://github.com/aws/aws-sdk-ruby/blob/version-2/aws-sdk-resources/lib/aws-sdk-resources/resource.rb#L101) method looks like the only that make sense and need extra port over to V3. This PR ported this over to all resource objects that has waiters for their service, avoid depending on `resource.json` waiter check because `resource.json` usually quite outdated. Tested manually.

Other alternative approaches considered yet not followed:
I have considered port this over to` resource.rb`s solo, and let those objects' `#wait_until` just call the one of service resource. However this won't work for block usage [here](https://github.com/aws/aws-sdk-ruby/blob/version-2/aws-sdk-resources/lib/aws-sdk-resources/resource.rb#L102).

Notes, this usage behavior looks like should have be been [deprecated](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-ec2/lib/aws-sdk-ec2/client.rb#L21323), so I have added @deprecated tag as well

Sample diff almost the same with the mustache diff : )

### DIFF
--- a/gems/aws-sdk-ec2/features/step_definitions.rb
+++ b/gems/aws-sdk-ec2/features/step_definitions.rb
@@ -1,7 +1,6 @@
 Before("@ec2") do
   @service = Aws::EC2::Resource.new
   @client = @service.client
-  @volume_ids = []
 end
 
 After("@ec2") do
--- a/gems/aws-sdk-ec2/features/step_definitions.rb
+++ b/gems/aws-sdk-ec2/features/step_definitions.rb
@@ -1,8 +1,28 @@
 Before("@ec2") do
   @service = Aws::EC2::Resource.new
   @client = @service.client
+  @volume_ids = []
 end
 
 After("@ec2") do
-  # shared cleanup logic
+end
+
+Given(/^I create a volume$/) do
+  @volume = @service.create_volume(size: 1, availability_zone: 'us-west-2a', volume_type: 'gp2')
+  @volume_id = @volume.id
+end
+
+When(/^I use \#wait_until to wait until volume is available$/) do
+  expect {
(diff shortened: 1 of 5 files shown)


cjyclaire:
Ah will follow up with tests

awood45:
Looks good on a first pass, will wait for tests.

cjyclaire:
Thought the same. I also thought S3 test might be cleaner, however S3 waiters might not be a good place to test states from block, so I mock the same EC2 issue here. I also make sure the volume is cleaned up. Now ready for another review : )

gems/aws-sdk-ec2/features/step_definitions.rb
```diff
@@ -4,5 +4,24 @@
...
+  @volume_id = @volume.id
+end
+
+When(/^I use \#wait_until to wait until volume is available$/) do
+  expect {
+    @resp = @volume.wait_until {|v| v.state == 'available'}
+  }.not_to raise_error
+end
+
+Then(/^Waiter works and volume is cleaned up$/) do
```
awood45: Use the cleanup logic you stubbed above. If the test fails you won't clean up the resource as written now.

cjyclaire: Make sense, fixed in the latest commit.

PR #1615, opened 2017-09-15
Allow DynamoDB to write empty set attribute values, part 2

makeready:
Hello,

First off, thanks for merging in my original PR for this feature. However, now that it's usable in the wild, we're trying to save empty sets and getting errors from param_validator.rb. After tracing through the error it became clear that I forgot to create the new AttributeValue definitions to support empty sets. This PR is to fix those omissions.

BTW, I wasn't clear on how to test this change - param_validator_spec uses a fake list of shapes, and client_api.rb doesn't have a corresponding test file, which means the definitions 
in client_api aren't actually tested anywhere. If you'd like me to write an integration test to cover it I can, but I assume this feature is too minor to deserve an integration test.

Thanks


cjyclaire:
@makeready Thanks for the contribution, those files are auto generated from shapes defined at .json file [here](https://github.com/aws/aws-sdk-ruby/blob/master/apis/dynamodb/2012-08-10/api-2.json), so, add those files directly won't work. Could you guide me to your original PR that I can see how can I add this to our generator? : )

makeready:
Here's the original PR: https://github.com/aws/aws-sdk-ruby/pull/1355

It occurs to me that there doesn't need to be a new shape created - an empty set could be typed as a BooleanSetAttribute or some other existing set attribute type instead. Depends on whether adding a new shape is more disruptive than allowing a type to exist with a misleading name. 

Let me know which way you'd like to go with it and I can amend the PR, either by adding the new shape to the jsons, or by removing references to empty set and typing it using an existing shape instead.

Thanks!

Issue #1535, opened 2017-06-16
Aws::S3::Client#get_object can use range requests to resume interrupted downloads


janko-m:
This [Amazon article](https://aws.amazon.com/blogs/developer/downloading-objects-from-amazon-s3-using-the-aws-sdk-for-ruby/) describes various ways in which you can download objects from S3. The Seahorse client currently retries any failed requests, which includes interrupted downloads. It retries them like any other request, by re-issuing the same request. That way it's currently not possible for streaming downloads, i.e. `Aws::S3::Client#get_object` with a block, to be resilient to network errors, because when `:response_target` is an IO object it can be truncated before retrying the request, while with a block (`BlockIO`) you cannot do that, the chunks were already yielded.

I propose that for `Aws::S3::Client#get_object` requests, instead of retrying the entire download request from the beginning, aws-sdk uses `Range` header to **resume downloading** from the last downloaded byte. That way retrying interrupted downloads will be faster, and it will also work for streaming downloads (where a block is passed to `Aws::S3::Client#get_object`).

cjyclaire:
Thanks for bring up the idea, appreciate that. I'd happy to dig more to see how do we make it clear as a feature request or enhance the user experience there. For the download, we release [`multipart downloader`](https://github.com/aws/aws-sdk-ruby/blob/master/CHANGELOG.md#2912-2017-04-26) in April, when it's making requests by `:range` or `:part`, when a single request fails, it ensures cleaning up completed chunks. For those requests, if retry happens, it's actually retrying the same request with `:range` or `:part`.

From my understanding of your proposal (feel free to correct me if I'm missing something : )), you are suggesting special handling for "basic"(no `:range` or `:part` in parameters) `:get_object` operation when making retries to make it more robustic. The problem is, if `:range` or `:part_number` is not specified in the `#get_object` api call, the response will contains `nil` value for `content_range` and `parts_count`, which will be useless if we want to resume the range.

Personally, I feel `multipart downloader` works better for this case if the file is big, for single `#get_object` request, I don't see a clear way for making recover less error prone. Thoughts?

janko-m:
Thanks for a thorough answer!

[...]

PR #1617, opened 2017-09-19
Don't retry streaming requests with blocks after data has been recieved

janko:
When downloading S3 objects with a block, aws-sdk-s3 internally creates a `BlockIO` to be the writable response target.

```rb
object.get do |chunk|
  # streaming
end
```

Normally aws-sdk-s3 retries any failed downloads in case of network errors, but not for streaming downloads. This is because `BlockIO` is not truncatable, unlike a response target like `StringIO` or `Tempfile`, and aws-sdk-s3 needs it to be because it retries the whole download from the start.

In [Shrine](https://github.com/janko-m/shrine) and [tus-ruby-server](https://github.com/janko-m/tus-ruby-server) I'm using aws-sdk-s3's streaming download feature, mainly for streaming the S3 object through a web application. It would be great if these downloads were to be retried in case of network errors in this case as well, especially for tus-ruby-server where the S3 objects will typically be very large.

This PR adds this functionality. It does so by remembering how many bytes of the content was "written" to the response target so far; when the request is retried, part of the response body that has already been written is simply skipped, until it reaches the part it hasn't written yet.

After this PR I plan to add support for [range requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests), so that the requests are retried from the last byte offset. But for start I wanted to add this because I didn't know whether all S3 endpoints support range requests, so I thought we would still need this as a safety net.


janko-m:
Ready for review.

janko-m:
Note that this addresses the feature request discussed in https://github.com/aws/aws-sdk-ruby/issues/1535 (that is, it will be fully addressed once `Range` header is added on retries).

gems/aws-sdk-core/lib/seahorse/client/net_http/handler.rb
```diff
@@ -79,8 +79,16 @@ def transmit(config, req, resp)
               bytes_received = 0
               resp.signal_headers(status_code, headers)
               net_resp.read_body do |chunk|
+                if bytes_received < resp.body.size
```
awood45: Can you talk a bit about the motivation of this part of the change? I can follow what you're doing, but want to check about any possible unintended consequences.

awood45:
Additionally, administrivia, can you confirm: "By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license."

gems/aws-sdk-core/lib/seahorse/client/net_http/handler.rb
janko-m: Imagine we're downloading an S3 object to a file. Seahorse will append each chunk to the file object as it is being downloaded. However, if a timeout error occurs, currently Seahorse will truncate the incomplete file object, retry the request, and start writing to the file from the beginning. This is not efficient, because there is nothing wrong with what we've already downloaded.

This change detects when the request has been retried and remembers how many bytes have already been written, and when it starts downloading again, instead of truncating the file and writing from the beginning, it will simply skip the data that was already written and start sending only new chunks. As a side effect, this allows streaming downloads (where block is provided as a target) to be retried in case of timeouts, where previously they were not retried, as a block is not "truncatable".

Ideally when we retry the request, we should make a `Range` request to ask S3 only for the remainder of the data that we haven't managed to download. I plan to make that change next, but I thought it's good to have this safety net for when that particular AWS endpoint doesn't support `Range` requests (since Seahorse is used throughout all the AWS services, if I understood correctly).

janko-m:

mullermp:
@janko Is this PR still desirable? If so I will do a review and get this in.

mullermp:
Thanks for the info. The code looks good to me. I would gladly merge it in.

There is however a [broken test](https://travis-ci.org/aws/aws-sdk-ruby/jobs/476041704):

`Aws::S3::Encryption::Client encryption methods #get_object decrypts the object with response target under retry`

Could you update this PR to fix that?

mullermp:
Attempted to fix this. When printing out a `prev_chunk`, I expected next + prev = body size

```
received: 0 body size: 8
next_chunk 5
prev_chunk 8
      populates part of response body that has not been written yet
received: 0 body size: 13
next_chunk 0
prev_chunk 13
      skips part of response body that has already been written
```

But for the failed test, the body size doesn't match the chunks.

received: 0 body size: 48
next_chunk 16
prev_chunk 48
      decrypts the object with response target under retry (FAILED - 1)

mullermp:
I attempted this but I hit a road block as well. It seems the chunk size is 64 while the response body is 48. This could be from cipher padding?

```
next_chunk 16 # in if condition
prev_chunk 48 # in if condition
chunk size: 64 # before signal data
final receive: 16 # before complete_response
      decrypts the object with response target under retry (FAILED - 1)
```

janko:
Thanks a lot for working on this 😃  Yes, the client-side encryption is definitely difficult to handle, I spent a lot of time trying to figure it out before. I think I would skip the retry mechanism when client-side encryption is used, if that's possible to detect.

mullermp:
No problem. I wanted to waste my afternoon debugging this! (just kidding of course...)

I found that it can certainly be skipped with:

```
                if Aws::S3::Encryption::IODecrypter != resp.body.class &&
                  bytes_received < resp.body.size
```

We can do this as a last resort but where's the fun in that? If the chunk size is 64, but the response body size is 48, what are the missing 16 bytes? I feel like we're closer now.

alextwoods:
I've been looking into fixes for #2311 and this seems related.

I've updated this PR with a fix for the S3 encryption client.  The issue is that the underlying body that the IODecrypter wraps is NOT reset between retries (The IODecrypter is recreated with the previous StringIO on the new request).

There are a few additional issues I think that I've tried to solve:
1.  When a request is retried, it would get a new BlockIO created and hence updating size and then checking it wouldn't be meaningful.  On 2XX headers, I changed it to only create a new target_response IO if the current body is a StringIO.
2.  When an error is encountered, the old code would set the body to StringIO.new (unless the body responded to :io).  This was first added [here](https://github.com/aws/aws-sdk-ruby/commit/ba932c10b5e51fa6eb10dc4fbb9fe997191b077a) and is intended to keep us from writing errors to either the block or the provided file and was modified later to add a check for the IODecrypter (which manages errors itself).  If we reset the body to a new StringIO object on retryable errors, the next retry will create a new BlockIO with size 0.  However, this change now opens us to the possibility of yielding errors to the passed block.

Additionally - this change will not work for the `S3::Encryption::Client` when used with a block because we need to create a new cipher on retry.

alextwoods:
There are a few issues with specs from the merge with master, but I think we can resolve those fairly easily (mostly the changes need to be applied into the adaptive retry logic as well).  

However, there are still 2 main issues I mentioned above.  Specifically:
1. This does not work for client side encryption when using a block (the current code results in incorrect decryption because the cipher is reset between retries).  
2. We would yield an error body to the provided block (a consequence of removing the setting of body to a new StringIO which is required so that the BlockIO w/ its stored size can be re-used in the next request when retired).  

@janko  - Given these I think we have a few options and I wanted to get your take.  

1.  We can move forward with this PR.  For client side encryption, we drop support for retry with a block for 200 ok responses (ie, we would still retry if we get say a throttling exception, but after we start getting data in the body and we're using a block as the response_target, if the request fails, we don't retry).    To fix the error body yielded to the block - we can revert my change, but store the previous body object (ie, the BlockIO) on the context and then in the on headers in the response_target plugin we just re-use that previous body object from the context if it exists.  
2.  We drop this PR and don't provide support for retires after 200ok, for generic streaming operations when a block is used.  However, we can then create a plugin specific to S3 get_object which is able to take advantage of the range parameter to avoid having to re-download/re-request data that the client has already received.  S3's get_object is the only streaming operation currently that supports the range parameter, so regardless we'd need specific behavior to support this.  

At this point I'd lean towards #2 - I'd prefer to reduce the complexity of the response_target and retry code for generic streaming operations and do provide the best support for retries of interrupted downloads in the get_object operations.  What do you think?

janko:
@alextwoods Thank you for picking this up, option 2️⃣ sounds good to me 👍

alextwoods:
I've created #2326 as a feature request for adding the S3 plugin for retry with range.  I've updated this PR to keep a few things that will be useful for that request + fix #2311.  If you're able to pitch in on the get_object specific retry w/ range (#2326 ) that would be great, otherwise, I'm hoping to have some time next week to pick it up.

alextwoods:
@janko  - I've created a draft PR for adding retry support using the range header for S3 get_objects: #2343  - Since get_objects is the only streaming operation that supports range its limited to that operation.  Additionally because of the way we do retries, the code gets a bit messy.  Lots of special cases.

PR #1619, opened 2017-09-22
Add refresh mechanism to SharedCredentials

SOYJUN:
Add the refresh mechanism to shared_credentials.rb
The default refresh cycle is one hour.

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
@@ -4,6 +4,7 @@ module Aws
   class SharedCredentials
 
     include CredentialProvider
+    include RefreshingCredentials
 
     # @api private
     KEY_MAP = {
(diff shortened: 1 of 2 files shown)


cjyclaire:
Making refreshing at default behavior might be breaking to current customer, could you make an optional flag for this? Also, may I ask for more background information about this feature request? Some use case perhaps? Any pain point there with the no refreshing behavior from file?

awood45:
I'm going to echo the fact that I'm not sure we want to do this. Is it not feasible to use existing refreshable credential sources?

SOYJUN:
Hi, I am from IDAS Turtle team, and this is our customer onboarding wiki: https://w.amazon.com/index.php/AWSAuth/Projects/HOSMEverywhere/Onboarding.
Our tracked SIM item: https://issues.amazon.com/issues/HOSM-1248
After the customers onboard to our service, we will deliver and rotate their credentials in host. 
And this is the 2017 Amazon security goal. So I assume that in the future, all the static creds in host should be able to be rotated.
It seems like it would be a better idea that I should set up another creds_provider.rb to implement this feature right now.  e.g. refresh_shared_credentials < shared_credentials ?

Issue #1624, opened 2017-10-03
Explicitly deleting tempfile after closing to save disk space


amalagaura:
### Issue description
We upload large encrypted files and have limited disk space. Tempfile are created to encrypt but they are not cleaned up after the file is uploaded.
  
### Gem name ('aws-sdk', 'aws-sdk-resources' or service gems like 'aws-sdk-s3') and its version
aws-sdk-s3

### Version of Ruby, OS environment
Ruby 2.3

### Code snippets / steps to reproduce
https://github.com/aws/aws-sdk-ruby/blob/9ea447b7e9e57e27bb7924aca40a5bf0c3265a94/gems/aws-sdk-s3/lib/aws-sdk-s3/encryption/io_encrypter.rb#L42 Closes the tempfile but does not `unlink` it. This leaves large files on disk which only get cleaned up when the process exits. This is not suitable for large running processes which upload large files. Can we add an `unlink` to this?

amalagaura:
I see there is a spec which expects the tempfile to not be deleted in order to enable caching of retries. 

Is the spec correct? I see a close operation here:
https://github.com/aws/aws-sdk-ruby/blob/9ea447b7e9e57e27bb7924aca40a5bf0c3265a94/gems/aws-sdk-s3/lib/aws-sdk-s3/encryption/encrypt_handler.rb#L42-L44

Doesn't the close happen on a succesful response? Is the IOEncryption handler instance really closed and then reused or is there a new IOEncryption instance on a retry?

amalagaura:
@cjyclaire I removed my PR

amalagaura:
Until this is done I am using the following code after an upload:

```
response = s3_client.put_object bucket: bucket, key: key, body: File.open(file_path)
encrypted = response.context.http_request.body.instance_variable_get(:@encrypted)
encrypted.close! if encrypted.respond_to?(:close!)
```

mullermp:
Going to close this issue for now. Please reopen if this is still any concern.

PR #1625, opened 2017-10-03
Close and unlink tempfiles when encryption is complete

amalagaura:
This adds a delete to the close of a tempfile. Otherwise tempfiles stick around until the ruby process is ended.

Issue #1624.


amalagaura:
I see there are specs which want the tempfile to not be deleted for caching purposes.

PR #1631, opened 2017-10-06
Add support for custom plugins

cjyclaire:
Adds support for adding custom plugins to generated service clients.


cjyclaire:
WIP, the error is that I need to create a foo plugin files separately can not use the existing one :(, which is easy to fix
Working on researching other approaches as well ...

PR #1638, opened 2017-10-12
Add serviceId for metadata field check

cjyclaire:
If service metadata now has serviceId, make sure it doesn't break build


cjyclaire:
The error is not related to the change and is 1.9.3 only, cannot reproduce in a local rbenv 1.9.3.
Merging fix changes right now, will work on a separate fix if the failing presists.

PR #1647, opened 2017-10-23
[Aws Partition] handler empty endpoint entry for service identifier

cjyclaire:
As the json file goes:

```
 "es" : { },
```

Tried to add spec test, yet if I check nil or empty, doesn't sounds like a sustainable test suit : )


cjyclaire:
I'm merging this right now, there are other SDKs exposing this already and this will not introducing breaking changes because "es" is the only one with empty entry.

Issue #1621, opened 2017-09-27
Shared credentials not being loaded from ~/.aws/credentials or ~/.aws/config


matt9949:
### Issue description

SharedCredentials are not being loaded from ~/.aws/config or ~/.aws/credentials at all. 

As per http://docs.aws.amazon.com/sdkforruby/api/Aws/EC2/Client.html#initialize-instance_method :

:credentials (required, Credentials) — Your AWS credentials. The following locations will be searched in order for credentials:
:access_key_id, :secret_access_key, and :session_token options
ENV['AWS_ACCESS_KEY_ID'], ENV['AWS_SECRET_ACCESS_KEY']
HOME/.aws/credentials shared credentials file
EC2 instance profile credentials See Plugins::RequestSigner for more details.
Tested with aws-sdk version: aws-sdk-core-2.10.53

When i set AWS_SECRET_ACCESS_KEY and AWS_ACCESS_KEY_ID as the 'services' creds, the instance is initialised correctly and i can authenticate.

However when using the shared credentials file, it doesn't pick up the creds.

### Gem name ('aws-sdk', 'aws-sdk-resources' or service gems like 'aws-sdk-s3') and its version
aws-sdk:2.10.53

### Version of Ruby, OS environment
OSX 10.12.3, Ruby 2.4.0

### Code snippets / steps to reproduce

~/.aws/credentials

[services]
aws_access_key_id=xyz
aws_secret_access_key=xyz

~/.aws/config

[profile dev]
role_arn = arn:aws:iam::<id>
source_profile = services
mfa_serial = arn:aws:iam::<id>
region = eu-west-1

[profile staging]
role_arn = arn:aws:iam::<id>
source_profile = services
mfa_serial = arn:aws:iam::<id>
region = eu-west-1

Using the SDK:

Without AWS_SECRET_ACCESS_KEY and AWS_ACCESS_KEY_ID set:

$ bundle exec irb
irb(main):001:0> require 'aws-sdk'
=> true
irb(main):002:0> client = Aws::EC2::Client.new(region: 'eu-west-1')
=> #<Aws::EC2::Client>
irb(main):003:0> client.run_instances(image_id: "ami-8e74b3f7", max_count: "1", min_count: "1")
Aws::Errors::MissingCredentialsError: unable to sign request without credentials set
[...]

PR #1657, opened 2017-11-13
Update README.md

rhyas:
Clarify credentials and profile options when fed to the SDK to be a little more clear. This is specifically a result of the #1621


awood45:
These changes look good, thanks!

PR #1660, opened 2017-11-18
Add support a protocol and custom build service

cjyclaire:
@alexwood45


build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/apig_endpoint_class.rb
```diff
@@ -0,0 +1,20 @@
...
+  module Views
+
+    class APIGEndpointClass < View
+
+      # @option options [required, String] :module_name
+      # @option options [required, String] :default_endpoint
+      def initialize(options)
+        @module_name = options.fetch(:module_name)
+        endpoint = options.fetch(:default_endpoint)
+        @default_endpoint = endpoint.start_with?('http') ? endpoint : "https://#{endpoint}"
```
awood45: This seems like it's prone to an unusual, but possible, edge case where somebody makes something called "httpservice" or similar which would then be an invalid endpoint. We should be checking for `http://` or `https://`.

cjyclaire: Ah that's fair, fixed this in the latest commit.

PR #1665, opened 2017-11-27
Remove old ox testing

cjyclaire:
Will wait for CI
Removing old ox testing suites

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -40,8 +40,12 @@ group :test do
       gem 'oj' unless RUBY_VERSION == '1.9.3'
     end
 
-    # As ox suggestion
-    gem 'ox', '~> 2.8.1'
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -40,12 +40,8 @@ group :test do
...
     end
 
-    # Ox after 2.4.12 has a change in default behavior. Test both.
-    if ENV['OLD_OX']
-      gem 'ox', '2.4.12'
-    else
-      gem 'ox'
-    end
+    # As ox suggestion
+    gem 'ox', '~> 2.8.1'
```
awood45: This is possibly too specific - if regressions appear in new 'ox' versions, we will miss it with this version lock. If we want, this could be the "new" `OLD_OX` and we keep the unbounded dependency for later tests.

PR #1679, opened 2017-12-19
Update APIG User-Agent to Prefix

cjyclaire:
As title

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core/plugins/api_key.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/plugins/api_key.rb
@@ -21,7 +21,7 @@ When provided, `x-api-key` header will be injected with the value provided.
       class OptionHandler < Seahorse::Client::Handler
         def call(context)
           # apply APIG user-agent by default
-          context.config.user_agent_suffix ||= 'aws-apig-ruby'
+          context.config.user_agent_prefix = 'aws-apig-ruby'
 
           if context.operation.require_apikey
             api_key = context.params.delete(:api_key)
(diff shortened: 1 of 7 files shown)


gems/aws-sdk-core/lib/aws-sdk-core/plugins/api_key.rb
```diff
@@ -21,7 +21,7 @@ def add_handlers(handlers, config)
       class OptionHandler < Seahorse::Client::Handler
         def call(context)
           # apply APIG user-agent by default
```
awood45: I think either here or in the near future we should pull this out of the API Key plugin - this wouldn't cause harm if the API Key plugin were removed, but it's turning this plugin into a catch-all for bringing along other configuration.

cjyclaire:
Separate another user agent plugin, ready for a second look : )

gems/aws-sdk-core/lib/aws-sdk-core/plugins/apig_user_agent.rb
```diff
@@ -0,0 +1,43 @@
...
+      class Handler < Seahorse::Client::Handler
+
+        def call(context)
+          set_user_agent(context)
+          @handler.call(context)
+        end
+
+        def set_user_agent(context)
+          ua = "aws-apig-ruby"
+          ua += "aws-sdk-ruby3/#{CORE_GEM_VERSION}"
```
awood45: This is going to create a difficult to read user agent. I'd go with this:

```ruby
ua = "aws-apig-ruby/#{CORE_GEM_VERSION}"
```

PR #1690, opened 2018-01-12
Assigned but ununsed variable

utilum:
Ruby 2.5 warns about this.


awood45:
`gems/aws-sdk-s3/lib/aws-sdk-s3/resource.rb` is a generated class, if you can take it out of the PR I can review (and add an item to improve the code generator). Good catch!

utilum:
Thank you for reviewing, @awood45 . I've removed it, and would be happy to improve the generator as well if you would provide some guidance, not sure where to start.

awood45:
Thanks!

PR #1691, opened 2018-01-12
Shadowing outer local variable

utilum:
Ruby 2.5 warns about this.


utilum:
Resolved by other commits.

Issue #1495, opened 2017-04-26
Build Failure in KITCHEN_SINK Test Harness


awood45:
The `KITCHEN_SINK` tests bring in third-party JSON/XML parsing gems to test compatibility. Those tests are now failing, and so this issue will be to determine:

1. What changed?
    * `oj`: `2.18.5` -> `3.0.1`
    * `ox`: `2.4.12` -> `2.4.13`
2. What's the fix?

awood45:
It would seem, despite the patch version bump, that `ox` is the most likely failure candidate due to the failures being in the XML tests.

awood45:
Local testing is failing now even with older versions. Expanding the investigation.

awood45:
I can confirm that, code wise, this issue now propagates all the way back to `2.9.0` and `2.8.0`. Going to need to keep searching for which dependency is now breaking, as older, passing builds don't have an obvious explanation in the gem set.

awood45:
Nevermind, after revising some local testing settings, discovered the regression is in `ox` `2.4.13`.

awood45:
Opened an issue in the [ox repo](https://github.com/ohler55/ox/issues/173).

PR #1695, opened 2018-01-17
Fixed ox parsing of xml documents containing significant whitespace

penntaylor:
...*semantically*-significant whitespace, that is.

The `:skip_white` flag being passed to Ox parser clobbers all whitespace during parsing. This strips out whitespace that may be semantically significant, such as in the case of sending YAML documents via SQS. The setting proposed in this pull request strips only trailing spaces from contained content, which should leave most real-world white-space significant documents semantically intact. It certainly fixes Ox for use with YAML over SQS.

This proposed change explicitly changes a prior fix made in #1496, related to issue #1495. ~~I was unable to locate the "kitchen_sink" test harness referenced there, so was unable to manually check this change against whatever the undesirable behavior for that issue was.~~ [Edit: I see now that KITCHEN_SINK is a Travis target.]

### DIFF
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Fixed ox parsing of xml documents containing significant whitespace
+
 3.14.0 (2018-01-15)
 ------------------
 
--- a/gems/aws-sdk-core/lib/aws-sdk-core/xml/parser/engines/ox.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/xml/parser/engines/ox.rb
@@ -14,7 +14,7 @@ module Aws
           Ox.sax_parse(
             @stack, StringIO.new(xml),
             :convert_special => true,
-            :skip => :skip_white
+            :skip => :skip_none
           )
         end
 


awood45:
Looks like the underlying issue may have been fixed in subsequent `ox` versions (per tests passing), so this may now be OK. Going to try and confirm this.

penntaylor:
The underlying `ox` problem still exists in most recent ox release (2.8.2); however, I discovered today that my patch **breaks some functionality** in `Aws::EC2::Instance`. Still working on tracking down the root cause. Will also try to come up with appropriate additions to the tests to catch both defects.

awood45:
Any updates on this? Would be especially interested in something long-term sustainable, as the `:skip_white` change itself was due to a patch version breakage.

mullermp:
Closing this for now. Reopen if there are any questions.

jdempster:
I'm also having this issue. After upgrading to AWS Elasticbeanstalk Amazon Linux 2, it would seem they've upgraded the ruby SDK version. In the worker tier, there is an aws-sqsd. The daemon reads from SQS and in turn makes an HTTP request.

Since the elastic beanstalk team upgraded the Ruby SDK, whitespace has been stripped from the SQS message. In PHP serialization this whitespace would be important to unserializing the data correctly, otherwise, it would end in a fatal.

I'm currently resolving this by patching the file during deployment
/opt/elasticbeanstalk/lib/ruby/lib/ruby/gems/2.6.0/gems/aws-sdk-core-2.11.546/lib/aws-sdk-core/xml/parser/engines/ox.rb

```
--- ox.rb       2021-04-09 20:20:20.862069509 +0000
+++ ox-new.rb   2021-04-09 20:21:05.181225573 +0000
@@ -14,7 +14,7 @@
           Ox.sax_parse(
             @stack, StringIO.new(xml),
             :convert_special => true,
-            :skip => :skip_white
+            :skip => :skip_none
           )
         end
```

But it does seem broken and makes sense to fix this upstream.

mullermp:
What version of Ox are you using?

jdempster:
From my understanding, it's ox-2.13.2.

I think the issue would arise no matter which version, as the issue is with regard to telling Ox to effectively strip whitespace.

mullermp:
In the SDK we have seen breaking behavioral changes from Ox, such as unintentional dropped support of older Ruby versions, changing of defaults, syntax, etc. In our tests, we have different tests for an older Ox version and a newer one.

Regarding this change, Ruby SDK V2 is deprecated and we do not do releases for it anymore. When you say that Elastic Beanstalk upgraded the Ruby SDK version, are you saying they upgraded to V3 and you are currently patching V2? Or did they upgrade from V1 to V2 and you are now patching V2? I do think it's worth looking at this issue and getting it fixed, but any fix would be to V3.

mullermp:
When using Elastic Beanstalk are you not able to select which gems and versions to install?

mullermp:
Alternatively, are you committed to using Ox? What happens if you use other engines such as Nokogiri?

jdempster:
Sorry, I'm not a ruby developer. Ideally, this is something the Elastic Beanstalk team need to be involved in fixing but I can't get in touch with anyone in the team.

They had upgraded from aws-sdk-core-2.3.0 to aws-sdk-2.11.546 which included the change to file mention above "aws-sdk-core/xml/parser/engines/ox.rb"

Ideally, they need to upgrade to v3 of the SDK but the problem also currently exists there.
From what I can tell there isn't any direct use of Ox, rather the usage comes from this SDK, which configures Ox to strip whitespace.

The message read back from SQS when using the Ruby SDK has been stripped of additional whitespace. Once there is a fix for whichever version then I can try and get the Elastic Beanstalk team to upgrade their use of the SDK.

mullermp:
No problem. My understanding of Elastic Beanstalk is that you can configure whatever applications you want, including the SDK version. The [docs say to use](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/ruby-development-environment.html#ruby-development-environment-sdk) `gem install aws-sdk` which would install V3. I'm not sure how it's being bundled in Elastic Beanstalk, if at all..

The way the engines work in the SDK is that we attempt to load certain ones with fallbacks. Ox is the first loaded and tried. If you're able to run `gem uninstall ox` for all versions of Ox on your instances, then a different engine will be used, such as rexml and nokogiri. We have confirmed that Ox is the only engine that will strip white space.

Another thing to try, if you are unable to uninstall Ox or upgrade the SDK version, you can base64 encode and decode your values when sending with SQS. For blobs of text, which may require formatting, this would be ideal.

jdempster:
Thank you for the advice.

I'm using a PHP worker tier Elastic Beanstalk, in this case, Ruby is bundled in to provide a daemon that reads from SQS and converts them into HTTP requests. This is known as aws-sqsd.

The solution I highlighted above changing the Ox configuration from skip_white to skip_none works well for me. Thought it would be worth raising the issue so it could get resovled.

mullermp:
I see - so you don't have access to a Ruby environment in your use case. My concerns for patching V2 is that 1) it's deprecated, 2) this is a change to core which can impact existing usage not just for your issue, and 3) assuming we patch it, I don't know if there a guarantee that the new version would be picked up by Elastic Beanstalk, since you mentioned that they upgraded recently, and the version mentioned is not the latest V2 version.. If you're able to control the Ruby SDK version I can look to make a patch - otherwise patching it wouldn't do anything for your problem.

When sending your data to SQS, are you not able to use Base64 encoding/decoding (using PHP?) My understanding of your use case would be that you can Base64 encode your data prior to sending to SQS, then your `aws-sqsd` worker would pick that up, and forward it as an HTTP request somewhere, and then you can extract that Base64 encoded value and decode it into the formatted string.

PR #1701, opened 2018-01-31
Close file descriptor in case of exceptions in Object#put

janko-m:
When a block is passed to `File.open`, file descriptor will be automatically closed even if an exception occurs while executing the block (e.g. the `Object#put` call fails with a timeout error).


awood45:
This looks good, and per what's becoming routine, can you confirm: "By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license."

janko-m:

PR #1705, opened 2018-02-03
Handle unlinked (temporary) files in logging formatter

grddev:
Uploading a temporary file to S3 fails with "TypeError: no implicit conversion of nil into String", when running under JRuby (many versions), with logging enabled, and the file was unlinked after it was opened.

The stack trace ends in `summarize_file` that basically prints roughly the same as `tempfile.inspect`, but also includes the size of the file. To handle paths and files the same way, the code uses `File.size(tempfile.path)` instead of `tempfile.size`, but this doesn't work if the file has been unlinked (in which case `tempfile.path` is nil).

The problem only affects JRuby, as `File === Tempfile.new` is false in MRI (but true in JRuby), but this means that the code just prints `#<Tempfile:/some/path>` for MRI 2.x and `#<File:/some/path>` for MRI 1.9.3.

This PR changes the implementation to use `tempfile.size` instead, meaning that opened files will be handled correctly even after unlinked (or renamed), but also adds explicit handling of `Tempfile` so that the output contains `#<File:/some/path (X bytes)>` for temporary files on both MRI and JRuby.

I changed the existing test cases to make use of unlinked regular `File` objects, but I was uncertain how I could add another test case for a `Tempfile` without duplicating the whole formatting test.

### DIFF
--- a/gems/aws-sdk-core/lib/seahorse/client/logging/formatter.rb
+++ b/gems/aws-sdk-core/lib/seahorse/client/logging/formatter.rb
@@ -213,6 +213,7 @@ module Seahorse
           when String   then summarize_string(value)
           when Hash     then '{' + summarize_hash(value) + '}'
           when Array    then summarize_array(value)
+          when Tempfile then summarize_file_io(value)
           when File     then summarize_file_io(value)
           when Pathname then summarize_file(value)
           else value.inspect
--- a/gems/aws-sdk-core/lib/seahorse/client/logging/formatter.rb
+++ b/gems/aws-sdk-core/lib/seahorse/client/logging/formatter.rb
@@ -213,7 +213,7 @@ module Seahorse
           when String   then summarize_string(value)
           when Hash     then '{' + summarize_hash(value) + '}'
           when Array    then summarize_array(value)
-          when File     then summarize_file(value.path)
+          when File     then summarize_file_io(value)
           when Pathname then summarize_file(value)
           else value.inspect
           end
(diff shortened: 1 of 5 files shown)


gems/aws-sdk-core/lib/seahorse/client/logging/formatter.rb
```diff
@@ -230,6 +231,14 @@ def summarize_string str
           end
         end
 
+        # Given an open file, this method returns a summarized inspecton
+        # string that includes the file size.
+        # @param [IO] file
+        # @return [String]
+        def summarize_file_io io
+          "#<File:#{io.path} (#{io.size} bytes)>"
```
awood45: This looks like we took a bit of a shortcut, knowing we could list the type as `File` because of the switch condition. If we're going to also log `Tempfile`, which makes sense, we should have that display here. For example, we could accept the class as a param:

```ruby
when Tempfile then summarize_file_io(value, "Tempfile")
when File     then summarize_file_io(value, "File")
```

```ruby
def summarize_file_io(io, classname)
  "#<#{classname}:#{io.path} (#{io.size} bytes)>"
end
```

gems/aws-sdk-core/spec/aws/log/formatter_spec.rb
```diff
@@ -24,6 +24,10 @@ def format(pattern, options = {})
         end
 
         it 'provides a :request_params replacement' do
+          file = Tempfile.open('') do |tempfile|
```
awood45: I'd actually prefer if we test both paths, rather than just replacing with a tempfile. Better way to ensure we could catch regressions with either path. Though given that it is not critical path in this test, you could choose to change one of the two test files, and not the other and I think that would meet the requirement.

grddev:
As should be evident from the lack of activity, I don't have any time to pursue this. I confirm that my contribution is made under the terms of the Apache 2.0 license, so feel free to either integrate this into the code base or not.

PR #1717, opened 2018-02-22
Fix QuickStart example #1 warnings

rcrews:
Warnings appear when running the sample code under [QuickStart Guide to Using the AWS SDK for Ruby](https://docs.aws.amazon.com/sdk-for-ruby/v3/developer-guide/quick-start-guide.html) with warnings on:

```ruby
#!/usr/bin/env/ruby -w
require 'aws-sdk-s3'  # v2: require 'aws-sdk'

s3 = Aws::S3::Resource.new(region: 'us-west-2')

s3.buckets.limit(50).each do |b|
  puts "#{b.name}"
end
```
The warnings are:

* `aws-sdk-core-3.16.0/lib/aws-sdk-core/rest/request/endpoint.rb:52: warning: shadowing outer local variable - parts`
* `aws-sigv4-1.0.2/lib/aws-sigv4/signer.rb:441: warning: shadowing outer local variable - headers`
* `aws-sdk-core-3.16.0/lib/seahorse/model/shapes.rb:39: warning: instance variable @location not initialized`

These  updates silence the warnings.


rcrews:
The TravisCI problem is at [aws-sdk-s3/lib/aws-sdk-s3/object_multipart_copier.rb#L56](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-s3/lib/aws-sdk-s3/object_multipart_copier.rb#L56). However, the TravisCI test failure was on JRuby 9.1.5.0. I am not able to reproduce the error on JRuby 9.1.16.0. I wonder if the problem is with JRuby 9.1.5.0 (and has since been fixed in 9.1.16.0)? Note that [the spec says @location should default to nil](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/spec/seahorse/model/shapes_spec.rb#L17), which is why I used `nil` for the initialization.

Robert

rcrews:
I reverted the commit that initialized the location field. I'll work on solving the warning in another way.

awood45:
Looks good, thanks!

PR #1720, opened 2018-02-27
Allow `__` in shape names

cjyclaire:
Tested with generated SDK with API call, request succeed.
Below changes do look messy, the thing is even though I want to strip as many as invalid characters, I need to find the reference in the model the other way around. Thus I limit it to known limitations. Let me know if you have other ideas.

### DIFF
--- a/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/api.rb
+++ b/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/api.rb
@@ -7,9 +7,15 @@ module AwsSdkCodeGenerator
       # @return [String, nil]
       def docstring(shape_or_shape_ref, api)
         ref, shape = resolve(shape_or_shape_ref, api)
+        # APIG models, downcase shape name in origin or "__" prefix in origin
         if shape.nil?
-          # APIG models, downcase shape name in origin
           ref, shape = resolve(AwsSdkCodeGenerator::Helper.downcase_first(shape_or_shape_ref), api)
+          if shape.nil?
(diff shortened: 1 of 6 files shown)


cjyclaire:
Also run `rake build` and no diff is generated for AWS services

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/api.rb
```diff
@@ -7,9 +7,15 @@ class << self
       # @return [String, nil]
       def docstring(shape_or_shape_ref, api)
         ref, shape = resolve(shape_or_shape_ref, api)
+        # APIG models, downcase shape name in origin or "__" prefix in origin
         if shape.nil?
-          # APIG models, downcase shape name in origin
           ref, shape = resolve(AwsSdkCodeGenerator::Helper.downcase_first(shape_or_shape_ref), api)
+          if shape.nil?
```
awood45: This is probably a place where comments are worthwhile, to explain why we nest this twice.

cjyclaire:
Updated test model, without this diff, it would ran into
```ruby
FFFF

Failures:

  1) Client Interface: API Gateway white label SDK populates x-api-key header correctly
     Failure/Error: Kernel.module_eval(code)

     SyntaxError:
       (eval):904: class/module name must be CONSTANT
           class __AnonymousListElement < Struct.new(
                                         ^
     # ./spec/spec_helper.rb:120:in `module_eval'
     # ./spec/spec_helper.rb:120:in `generate_service'
     # ./spec/interfaces/client/api_gateway_spec.rb:7:in `block (3 levels) in <top (required)>'
```

Issue #1730, opened 2018-03-16
Aws::S3::Object#presigned_url(:get, reponse_expires: ... ) raise NotImplementedError


hiboma:
### Issue description

When I test the following code, `Aws::S3::Object#presigned_url(:get)` with `:response_expires`, it raises `NotImplementedError` instead of generating a pre-sign URL.

```ruby
#!/usr/bin/env ruby

require 'aws-sdk-s3'

s3 = Aws::S3::Resource.new(
  access_key_id: '***',
  secret_access_key: '***',
)

obj = s3.bucket('examplebucket').object('test.txt')
puts obj.presigned_url(:get, response_expires: Time.now)
```

While other `:response_*` parameters

 * `:response_content_type`
 * `:response_content_language`
 * `:response_cache_control`
 * `:response_content_disposition`
 * `:response_content_encoding`

 are specified as an argument, preceding code works fine. Only `:reponse_expires` raises exceptions.

### Gem name ('aws-sdk', 'aws-sdk-resources' or service gems like 'aws-sdk-s3') and its version

`aws-sdk-s3` `1.8.2`

##### Gemfile

```
$ cat Gemfile
source 'https://rubygems.org'

gem 'aws-sdk-s3'
```

```
$ cat Gemfile.lock
GEM
  remote: https://rubygems.org/
  specs:
    aws-partitions (1.70.0)
    aws-sdk-core (3.17.0)
      aws-partitions (~> 1.0)
      aws-sigv4 (~> 1.0)
      jmespath (~> 1.0)
    aws-sdk-kms (1.5.0)
      aws-sdk-core (~> 3)
      aws-sigv4 (~> 1.0)
    aws-sdk-s3 (1.8.2)
      aws-sdk-core (~> 3)
      aws-sdk-kms (~> 1)
      aws-sigv4 (~> 1.0)
    aws-sigv4 (1.0.2)
    jmespath (1.3.1)

PLATFORMS
  ruby

DEPENDENCIES
  aws-sdk-s3

BUNDLED WITH
   1.16.1
```

### Version of Ruby, OS environment

```sh
$ ruby -v
ruby 2.4.3p205 (2017-12-14 revision 61247) [x86_64-darwin17]
```

### Code snippets / steps to reproduce

1. Prepare a script code (Please see above)
2. Run the script.

backtrace will be like following.

```
[] /tmp $ bundle exec ruby presign.rb
[...]

PR #1731, opened 2018-03-19
support timestamp shape at querystring

cjyclaire:
Addressing #1730 
Timestamp shape param value is already converted based on API model definition, just need to serialized onto querystring

tested with #get_object and #presigned_url, updated protocol tests


cjyclaire:
Taking a look, the test was succeed for me locally

cjyclaire:
Referring S3 API Doc for [GetObject](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGET.html), the timestamp shape at querystring appears to be `.utc.httpdate`, however, still working on confirming escape pattern

PR #1735, opened 2018-03-21
Fix typo: s/otpional/optional/

hfm:


cjyclaire:
Thanks for the fix, merging

PR #1736, opened 2018-03-26
Add frozen_string_literal: true to freeze strings

oniofchaos:
This will prevent unnecessary allocations of strings. There were a few
files that I had to remove it from because they depend on being able to
mutate strings. Those files were:

gems/aws-sdk-core/lib/aws-sdk-core/xml/doc_builder.rb
gems/aws-sdk-core/spec/aws/xml/doc_builder_spec.rb
gems/aws-sdk-core/spec/seahorse/client/logging/handler_spec.rb
gems/aws-sdk-core/spec/seahorse/client/request_spec.rb

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 $REPO_ROOT = File.dirname(__FILE__)
 $GEMS_DIR = "#{$REPO_ROOT}/gems"
 $CORE_LIB = "#{$REPO_ROOT}/gems/aws-sdk-core/lib"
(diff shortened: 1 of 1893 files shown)


oniofchaos:
If someone wants to point me to the right place to update the code generators in order to get `frozen_string_literal: true` in newly generated files, I'd be happy to look into making that update as well.

oniofchaos:
I didn't change the line that is failing in Ruby 1.9.3, looks like it has to deal with an encoding issue. Any thoughts?

awood45:
Taking a look at that test failure, but yes, I can also look at this as a feature request to the code generators to put this into generated classes. That seems like the correct path forward (for example, I'm not too concerned with adding it to the code generator classes themselves, they aren't shipped with the gem.

mullermp:
Closing this for now as we had no response. Feel free to reopen and add relevant lines to the code generators.

Issue #1738, opened 2018-03-26
Getting undefined method 'each' for Enumerable::Enumerator


david-a:
Please fill out the sections below to help us address your issue

### Issue description
When trying to fetch a list of objects and do something with them - `each`, `first` etc. it gives the error `NoMethodError: undefined method 'each' for #<Enumerable::Enumerator:0x007fcc8f4f63d0>`
originating in `aws-sdk-core`'s `collection.rb`.

it looks like all occurences of `Enumerator.new` in this `collection.rb` are lacking since they refer to `Enumerable::Enumerator` (becuase of the `include Enumerable` in the file) which has NO `each` method indeed. Instead, the should refer to the regular `Enumerator`. simple as adding `::` to those `Enumerator.new`.

### Gem name: 
'aws-sdk-core' / aws-sdk-s3 v1.8.2

### Version of Ruby, OS environment
ruby 2.3.3p222, macOS 10.13.2

### Code snippets / steps to reproduce
`Aws::S3::Resource.new(amazon_credentials).bucket('some_bucket').objects({...}).each {}`

david-a:
Just want to update, I've found the cause for the problematic (empty) definition of `Enumerable::Enumerator` in my environment - for me it was the gem `CFPropertyList-rails` (v1.0.2) which uses an old version of `CFPropertyList` (v2.2.8).

PR #1739, opened 2018-03-26
Replacing Enumerable::Enumerator with Enumerator

david-a:
Fixes the error `undefined method each for Enumerable::Enumerator`

Issue: https://github.com/aws/aws-sdk-ruby/issues/1738

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core/resources/collection.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/resources/collection.rb
@@ -26,7 +26,7 @@ module Aws
       # @deprecated
       # @api private
       def batches
-        Enumerator.new do |y|
+        ::Enumerator.new do |y|
           batch_enum.each do |batch|
             y << self.class.new([batch], size: batch.size)
           end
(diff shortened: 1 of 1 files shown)


awood45:
Is it possible for you to add a (previously) failing test to the spec suite that reflects the error?

It otherwise looks good, and happy to help if you're not sure about the best way to do this.

david-a:
Sure, I'll look into it later today. 
Thanks for your quick response!

david-a:
@awood45 done :)
This issue occurs only in environments which have another `Enumerable` module defined, and this module has an `Enumerator` class defined without `each` and `next` methods. 
Somehow my environment has this module, and I know it may be an edge case, but it won't hurt to add those `::` as a precaution.

david-a:
Just want to update, I've found the cause for the problematic (empty) definition of `Enumerable::Enumerator` in my environment - for me it was the gem `CFPropertyList-rails` (v1.0.2) which uses an old version of `CFPropertyList` (v2.2.8).

cjyclaire:
Thanks! Could you also add a changelog entry [here](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/CHANGELOG.md), such that your changes can be included in the next release of `aws-sdk-core` gem?

david-a:
@cjyclaire sure, under what version number?

cjyclaire:
Under "unreleased changes" section is fine, just follow the format to add an entry. Our build tools will auto bump the version in the next release : )

david-a:
@cjyclaire done, Ihope it's ok now

PR #1751, opened 2018-04-05
The Cognito Identity Provider InitiateAuth action does not require a signed request

halostatue:
This is the API called by the JavaScript, iOS, and Android clients where the IAM keys may not be available for security reasons.

Verified with `httpie`:

```
http --debug POST https://cognito-idp.us-east-1.amazonaws.com/ X-Amz-Target:AWSCognitoIdentityProviderService.InitiateAuth ClientId=clientId AuthFlow=USER_PASSWORD_AUTH AuthParameters:='{"USERNAME":"username","PASSWORD":"password"}' Content-Type:application/x-amz-json-1.1
```

I suspect that `RespondToAuthChallenge` should also be `o['authtype'] = "none"`, but it is not needed by my particular use case (I am implementing a user-level client in Ruby; that is, the user is using the Ruby client the same way it would be using the JavaScript client).

In the interim, this behaviour can be monkey-patched with

```ruby
Aws::CognitoIdentityProvider::ClientApi::API.operation(:initiate_auth)['authtype'] = 'none'
```

as long as it is applied before `Aws::CognitoIdentifier::Client.new`.

### DIFF
--- a/gems/aws-sdk-cognitoidentityprovider/lib/aws-sdk-cognitoidentityprovider/client_api.rb
+++ b/gems/aws-sdk-cognitoidentityprovider/lib/aws-sdk-cognitoidentityprovider/client_api.rb
@@ -2739,6 +2739,7 @@ module Aws::CognitoIdentityProvider
         o.name = "InitiateAuth"
         o.http_method = "POST"
         o.http_request_uri = "/"
+        o['authtype'] = "none"
         o.input = Shapes::ShapeRef.new(shape: InitiateAuthRequest)
         o.output = Shapes::ShapeRef.new(shape: InitiateAuthResponse)
         o.errors << Shapes::ShapeRef.new(shape: ResourceNotFoundException)
--- a/gems/aws-sdk-cognitoidentityprovider/spec/client_spec.rb
+++ b/gems/aws-sdk-cognitoidentityprovider/spec/client_spec.rb
@@ -10,6 +10,16 @@ module Aws
         expect(resp.context.http_request.headers['Authorization']).to be(nil)
       end
 
+      it "does not sign calls to initiate_auth" do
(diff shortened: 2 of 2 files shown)


awood45:
This is actually a change to a generated class, so it's not going to work as formed. I can look at this as a +1 to the requests to support unsigned requests for operations which accept them broadly (there are others in other services as well, as you probably know).

halostatue:
I was briefly looking at the JSON, and it looks like all that’s needed on that is a change to add `"auth_type": "none"` to the `"InitiateAuth"` operation in the JSON service description. Except maybe the tests. But I‘ll leave that to people who know how the code is generated.

mullermp:
Is this still a problem/pain point?

halostatue:
@mullermp I am no longer working with the software that needed to use Cognito. However, `InitiateAuth` does not require a signature (and, as I understand it, _should not_ have a signature), so I think that this is still something that should have the signature not required.

IIRC, the generated Ruby code at the time that this was submitted differed from other generated code on what it used. But I’m not entirely sure. Either way, unless and until I need to use Cognito again, whatever happens with this PR doesn’t matter to me.

mullermp:
@halostatue Thanks for your response, sorry about this not being addressed in a timely manner - I was not on the project when this PR was made. I found a file responsible for customizations like this and made a change (https://github.com/aws/aws-sdk-ruby/pull/2335) if you'd like to take a look. It adds `o['authtype'] = "none"` to both InitiateAuth and RespondToAuthChallenge. I've been bugging the cognito team about looking into adding it to their API model so that this customization isn't needed in the future. I'm happy to add other unauthenticated APIs.

halostatue:
Your fix looks like the correct fix.

PR #1753, opened 2018-04-09
Adding support for S3 Select [eventstream]

cjyclaire:
**Note, the PR is still in progress (tests, documentation code gen)**
**Binary decode/encode gem will be in a separate gem `aws-eventstream` in a separate PR.**

This PR is aiming to support following usage for unidirectional eventstream operations over `http1.1` protocol

1. Streaming along the way
``` ruby
opts = {
  bucket: 'bkt',
  key: 'my_file',
  ...
}
s3.select_object_content(opts) do |stream|
  # process a records event when it arrives
  stream.on_records_event do |event|
    puts event.payload.read
    # => String, output payload member of records event
    puts "<--- caught a records event --->"
  end

  # process a stats event when it arrives
  stream.on_stats_event do |event|
    puts event.details.inspect
    # => <struct Aws::S3::Types::Stats bytes_scanned=23, bytes_processed=23>, event_type=:stats>
    puts "<--- caught a stats event --->"
  end

  # process an end event when it arrives
  stream.on_end_event do |event|
    puts event.inspect
    # => <struct Aws::S3::Types::EndEvent event_type=:end>
    puts "<--- caught an end event --->"
  end
end
```
2. Or passing over a callback handler, a callback handler can be a Proc object or a EventStream instance

a) EventStream object
```ruby
handler = Aws::S3::EventStreams::SelectObjectContentEventStream.new
handler.on_records_event do |e|
  puts e.payload.read
  puts "<---caught a records event-->"
end
handler.on_stats_event do |e|
  puts e.details.inspect
  puts "<---caught a stats event-->"
end
s3.select_object_content(.. opts ..., event_stream_handler: handler)
```
b) Proc object
```ruby
callbacks = Proc.new do |stream|
  stream.on_records_event do |event|
    puts event.payload.read
    puts "<--- caught a records event --->"
  end

  stream.on_stats_event do |event|
    puts event.details.inspect
    puts "<--- caught a stats event --->"
  end

  stream.on_end_event do |event|
    puts event.inspect
    puts "<--- caught an end event --->"
  end
end
s3.select_object_content(.. opts ..., event_stream_handler: callbacks)
```
2.5 Or hybrid pattern of above 2
```ruby
handler = Aws::S3::EventStreams::SelectObjectContentEventStream.new
handler.on_records_event do |event|
  File.write('/path/to/file', event.payload.read)
end
s3.select_object_content(... opts ..., event_stream_handler: handler) do |stream|
  stream.on_stats_event do |event|
     puts "I want statistics ...."
     puts event.details.inspect
  end
end
```
Noted that now handler variable has both callbacks registered

3. Or wait for full response
```ruby
resp = s3.select_object_content(opts)
# => #<struct Aws::S3::Types::SelectObjectContentOutput
# payload=#<Aws::S3::Types::SelectObjectContentEventStream: ...>>
stream = resp.payload
# => #<Aws::S3::Types::SelectObjectContentEventStream: ...>

stream.next
# => #<struct Aws::S3::Types::RecordsEvent
# payload=#<StringIO:0x007fd99912c5f8>,
# event_type=:records>

stream.next
# => #<struct Aws::S3::Types::StatsEvent
# details=#<struct Aws::S3::Types::Stats bytes_scanned=23, bytes_processed=23>,
# event_type=:stats>

stream.next
# => #<struct Aws::S3::Types::EndEvent event_type=:end>

stream.next
# StopIteration: iteration reached an end
```
For all usage, after response ends, all events object are still available at the enumerate (eventstream object)
Will add examples of generated  S3 gem diff below.


cjyclaire:
For `build_tools` changes, they are all for generated diffs like following pattern (only include some important diff)

1. Add `eventstream_handler` option
```ruby
...
add_plugin(Aws::Plugins::EventStreamConfiguration)
...
```

2. For S3 event stream operation
```ruby
def select_object_content(params = {}, options = {}, &block)
   req = build_request(:select_object_content, params)
   req.handlers.add(Aws::Binary::DecodeHandler, priority: 95)
   req.send_request(options, &block)
end
```

3. For event shapes
```ruby
class RecordsEvent < Struct.new(
  :payload,
  :event_type)
  include Aws::Structure
end
```

4. For event stream shapes
``` ruby
class SelectObjectContentEventStream < Enumerator

  def event_types
    [
    :records,
    :stats,
    :progress,
    :cont,
    :end
    ]
  end

end
```

5. Adding default eventstream wrapper for callbacks
```ruby
module Aws::S3
  module EventStream
    class SelectObjectContentEventStream

      def initialize(options)
        @data = options.fetch(:data)
      end

      def on_records_event(&block)
        on_event(:records, Proc.new)
      end

      def on_stats_event(&block)
        on_event(:stats, Proc.new)
      end

      def on_progress_event(&block)
        on_event(:progress, Proc.new)
      end

      def on_cont_event(&block)
        on_event(:cont, Proc.new)
      end

      def on_end_event(&block)
        on_event(:end, Proc.new)
      end

      private

      def on_event(type, callback)
        callback.call(@data) if @data.event_type == type
      end

    end

  end
end
```

cjyclaire:
Pending thoughts for `on_initial_response` trigger for fetching headers members values available besides stream data from payload body.

trevorrowe:
I would recommend that the `:event_stream_handler` option accept an instance of the EventStream class or a block that is used to configure the class.

gems/aws-sdk-core/lib/aws-sdk-core/binary/decode_handler.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      def attach_eventstream_listeners(context, rules)
+
+        context.http_response.on_headers(200) do
+          protocol = context.config.api.metadata['protocol']
+          context.http_response.body = IODecoder.new(
+            protocol,
+            rules,
+            context.http_response.body,
+            context.config.eventstream_handler)
```
trevorrowe: Try to remove this from config and instead use `context[:event_stream_handler]`

gems/aws-sdk-core/lib/aws-sdk-core/binary/event_parser.rb
```diff
@@ -0,0 +1,92 @@
...
+      def parse(raw_event)
+        message_type = raw_event.headers.delete(":message-type")
+        if message_type
+          case message_type.value
+          when 'error'
+            parse_error_event(raw_event)
+          when 'event'
+            parse_event(raw_event)
+          else
+            raise 'Unrecognized :message-type value for the event'
```
awood45: Raise an appropriate typed error.

gems/aws-sdk-core/lib/aws-sdk-core/binary/event_parser.rb
```diff
@@ -0,0 +1,92 @@
...
+          error_code ? error_code.value : error_code,
+          error_message ? error_message.value : error_message
+        )
+      end
+
+      def parse_event(raw_event)
+        event_type = raw_event.headers.delete(":event-type").value
+        # content_type = raw_event.headers.delete(":content-type").value
+
+        # TODO Pending
```
awood45: ?

gems/aws-sdk-core/lib/aws-sdk-core/binary/event_stream_decoder.rb
```diff
@@ -0,0 +1,44 @@
...
+        # wrap decoded parsed event with callbacks in a event stream
+        @decoder.decode(StringIO.new(chunk)) do |raw_event|
+          event = @event_parser.apply(raw_event)
+          @emitter.signal(event.event_type, event) unless @emitter.nil?
+        end
+      end
+
+      private
+
+      # TODO, more protocol
```
awood45: Assuming this is coming before merging - correct?

gems/aws-sdk-core/lib/aws-sdk-core/param_validator.rb
```diff
@@ -130,6 +141,7 @@ def correct_type?(ref, value, errors, context)
       case value
       when Hash then true
       when ref.shape.struct_class then true
+      when Enumerator then ref.eventstream
```
awood45: We should discuss this a bit further. This might be rather broad in what it accepts.

gems/aws-sdk-core/lib/aws-sdk-core/binary/event_parser.rb
cjyclaire: Fixed, will be `EventStreamParserError.new ...`

gems/aws-sdk-core/lib/aws-sdk-core/binary/event_parser.rb
cjyclaire: Yeah ... though current no real server response is available for testing this, I'll adding this

gems/aws-sdk-core/lib/aws-sdk-core/binary/event_stream_decoder.rb
cjyclaire: As talked offline, I'll just let it throw error there for other protocols to make sure we proper test those before release their support

gems/aws-sdk-core/lib/aws-sdk-core/param_validator.rb
cjyclaire: Nice catch, I'll definitely consider more careful check for this one

jclusso:
Hey, was wondering if this is still being worked on? Really excited to be able to use this feature with S3!

cjyclaire:
@jclusso Thanks for the ping! Just let you know that this has been actively working on, and it's our top priority, we would like to get this out ASAP.

Currently functionality has been completed, and it's in the progress of completing test suite and documentation :D

jclusso:
@cjyclaire oh that's great! I'm going to pull it in and see if I can get it working if that's the case!

jclusso:
@cjyclaire not sure what I'm doing wrong but is there some special way to bundle this gem using a branch. I get "Could not find gem 'aws-sdk' in git://github.com/aws/aws-sdk-ruby.git (at flow@65cd4c7).
The source does not contain any versions of 'aws-sdk'" when trying to use github to bundle it.

cjyclaire:
@jclusso It's because that gem is in a separate PR: https://github.com/aws/aws-sdk-ruby/pull/1754 ;)
I'm working on get that released soon as well

cjyclaire:
Attached s3 code gen changes in the last commit.

Ready for a final review @trevorrowe @awood45

cjyclaire:
CI is failing on region_spec which shouldn't be relevant, also received warning on initializing enumerator. working on fixing those

gems/aws-sdk-core/lib/aws-sdk-core/binary/event_parser.rb
```diff
@@ -53,8 +58,10 @@ def parse_event(raw_event)
         event_type = raw_event.headers.delete(":event-type").value
         # content_type = raw_event.headers.delete(":content-type").value
 
-        # TODO Pending
-        # if event_type == 'initial-response' end
+        # Pending
+        raise Aws::Errors::EventStreamParserError.new(
```
awood45: Style point: Don't use inline `if` when the code is not a single line. Prefer a wrapping `if` statement.

gems/aws-sdk-core/aws-sdk-core.gemspec
```diff
@@ -14,7 +14,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency('jmespath', '~> 1.0')
   spec.add_dependency('aws-partitions', '~> 1.0')
   spec.add_dependency('aws-sigv4', '~> 1.0') # necessary for making Aws::STS API calls
-  spec.add_dependency('aws-eventstream', '~> 0.0.0')
+  spec.add_dependency('aws-eventstream', '~> 1.0') # necessary for binary eventstream
```
awood45: This is insufficient for launch, as we would also need to ensure that service gems which require the presence of eventstream to have a minimum bound on their dependency on `aws-sdk-core`, otherwise bundler will accept invalid dependency graphs.

build_tools/services.rb
```diff
@@ -123,6 +124,22 @@ def model_path(model_name, models_dir)
...
+      api['shapes'].each do |_, ref|
+        return true if ref['eventstream'] || ref['event']
+      end
+      false
+    end
+
+    def core_version(api)
+      version_file = File.read("#{$GEMS_DIR}/aws-sdk-core/VERSION").rstrip
+      if eventstream?(api)
+        EVENTSTREAM_CORE_VERSION
```
awood45: This is not right, we're going to lock customers on an old version of core. We need a bounded dependency like:

"~> 3", ">= 3.21.0"

Reference: https://guides.rubygems.org/patterns/#declaring-dependencies

PR #1762, opened 2018-04-18
instance_profile_credentials.rb: add User-Agent to the metadata crede…

willbengtson:
Adds the sending of a User-Agent other than the default User-Agent used, 'Ruby'.  Similar to golang SDK
that starts with aws-sdk-go/1.8.19.

This will enable protection of AWS credentials from Server Side Request Forgery (SSRF) vectors by using a metadata proxy and rejecting User-Agents that do not meet a regex.

### DIFF
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,7 +1,7 @@
 Unreleased Changes
 ------------------
 
-* Feature - Aws::InstanceProfileCredentials - Add sending a User-Agent other than the default User-Agent in Ruby.  Make the User-Agent `aws-sdk-ruby3/<version>` to allow protection against Server Side Request Forgery (SSRF) credential theft vectors by use of a metadata proxy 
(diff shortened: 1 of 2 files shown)


cjyclaire:
Thanks for the contribution! Could you also add a changelog entry [here](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/CHANGELOG.md) such that your change can be tracked in the next release?
Just add an entry under Unreleased Changes section same pattern would be fine, don't need to worry about version number or date : )

willbengtson:
Thanks, add this to the CHANGELOG.md.  Let me know if you need the text changed.

cjyclaire:
A nit picking there : ) 
I'd like to see ` Add sending a User-Agent other than the default User-Agent in Ruby.  Make` changed to  `Adding`. Because for regular API request, Ruby SDK request stack does has a [plugin for adding user-agent](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/lib/aws-sdk-core/plugins/user_agent.rb) It's special that those instance credentials are making request directly  and avoid the request handler stack we have.

willbengtson:
Ok, made the wording change.  Let me know what you think now.

cjyclaire:
Thanks! merged, will be out in the next release

PR #1773, opened 2018-04-30
Fix documentation error

jkburges:
Retries defaults to 5 if not specified in initialiser (not 0).


awood45:
This is probably something where we could expand context. While you are correct about the initializer if created directly, the documentation is correct for the most common use case - the default credential provider chain: https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/lib/aws-sdk-core/credential_provider_chain.rb#L27

I'm happy to take this change if we draw that distinction, or I can take this as a task to update the documentation myself. Up to you!

jkburges:
Thanks @awood45 - if it's any help, I arrived at this code because just from reading the doc, I was confused as to what the default retries is for default credential provider chain, so this was my attempt to clarify, but I didn't understand things properly as you point out.

PR #1789, opened 2018-05-21
aws-sdk-dynamodb - Allow objects responding to #to_h to be marshaled as a Hash

alexagranov:


alexagranov:
If accepted, I can make a similar PR against `version2` branch.

alexagranov:
Updated spec to pass on 1.9.3 (was using OpenStruct but no `to_h` for 1.9.3 :-()

awood45:
This is kind of an interesting case, it sounds like we're expecting different behavior in 1.9.3 if adopting this change?

awood45:
Ahh, I'm understanding the purpose a bit more. I am concerned though that this could catch some objects which weren't intended to be treated as hashes. Granted, we survive this now because of ordering, but I'm now considering if it's important to explicitly ask you to USE your hash transformation ability before calling low-level clients.

alexagranov:
Hi, thanks for getting back to this.

I do see the merit in requiring clients of an SDK to explicitly provide proper objects to marshal - otherwise raise an exception.  

But then couldn't one make the similar argument for not automatically calling `to_s` on provided objects?

I think `to_h` on Ruby objects has reached sufficient ubiquity that automatically marshaling with it if found could be reasonable.  Up to you of course but made sense to me to offer.

awood45:
Your counterpoint is fair as well, though we'd have big problems if we implemented, for example, the `:to_s` check in the same way. Different because this is at the end of the switch stack, but just something I need to think about.

Not worried about the 1.9.3 point anymore, I was confusing myself on the OpenStruct emphasis. No worries, you're correct on that count.

alexagranov:
Sounds good, let me know what you decide.  Happy to tackle specs from a different perspective/direction if you have something in mind.

mullermp:
@alexagranov Sorry that this issue was dropped. We took a look and think it should be safe to merge.

PR #1801, opened 2018-06-11
Fix JRuby Issues with IO.copy_stream

awood45:
In short, JRuby had some major issues around IO.copy_stream until about
`9.1.15.0`, with even more fixes coming in `9.1.16.0`. This proposed
patch works around that issue and integration tests are passing in
JRuby.

Fixes #1711

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ rvm:
   - 2.3
   - 2.4
   - 2.5
-  - jruby-9.1.5.0
+  - jruby-9.1.16.0
 
 sudo: false
 
@@ -22,7 +22,7 @@ bundler_args: --without docs release repl
 
 matrix:
   exclude:
-    - rvm: jruby-9.1.5.0
+    - rvm: jruby-9.1.16.0
       env: KITCHEN_SINK=1
   include:
     - rvm: 2.3
--- a/gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
+++ b/gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
@@ -219,6 +219,9 @@ module Aws
       #     # Passed chunks automatically split into multiupart multipart upload parts
       #     # and the parts are uploaded in parallel. This allows for streaming uploads
       #     # that never touch the disk.
+      #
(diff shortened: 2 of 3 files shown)


awood45:
From the related issue #1711 - this issue is likely to increase memory usage by a decent amount, including outside of JRuby.

Fonsan:
There is a memory usage increase. It will use twice the memory for when the stream is in memory mode and it will act as if in memory mode for tempfile which defeats the purpose of the tempfile apporach

Fonsan:
There might be a convoluted way to make it backwards compatible by reading and writing small chunks but I think going down that road would prove some of the points I layed out in #1711

Fonsan:
@awood45 Would you consider adding a explicit exception that would be raised if this functionality was called from a incompatible version of jruby allowing for a better experience for the developer?

awood45:
That's definitely possible. There's a few ways to go about this that I'm comparing right now. Most ugly might be to have this approach as a fallback, or gating on language versions (the approach I most want to avoid).

I'm leaning towards the belief that punishing the vast majority of users with memory bloat just because JRuby's standard library was buggy is not the way to go. Question is how well we can support older JRuby versions.

Fonsan:
I looked into reimplementing the same semantics as IO.copy_stream and realised that rubinius must do the same thing. 

https://github.com/rubinius/rubinius/blob/15b12dbf61f91f44851282af2956587eed9bc128/core/io.rb#L1041
https://github.com/rubinius/rubinius/blob/15b12dbf61f91f44851282af2956587eed9bc128/core/io.rb#L1123

But I am not even sure that they get it right as they call seek on the IO object and not all IO objects support seek

Fonsan:
Nevermind the bit about seek, MRI makes the same assumption for the fourth parameter https://ruby-doc.org/core-2.3.1/IO.html#method-c-copy_stream

awood45:
Ok, verdict - I'm going to go with your implementation with a documentation note about the JRuby issues. If we get reports about it, my fallback plan (which I'd rather not do) is to fork the logic to the memory bloat approach if and only if the user is using an old version of JRuby by checking the Ruby version at runtime.

PR #1805, opened 2018-06-19
add acm CertificateValidated waiter

davidcpell:

This adds a `DescribeCertificate` waiter to `aws-sdk-acm`. I was working on a deploy pipeline that uses the SDK, and needed to wait for certificate validation before the pipeline continues.

First time contributing to the generated code in `apis/`. I read through the Contributing doc but please let me know if there's anything I need to add!

The one thing I know may need to be revised is the error matcher - not sure if I've specified the "expected" value correctly. Thanks!

### DIFF
--- a/apis/acm/2015-12-08/waiters-2.json
+++ b/apis/acm/2015-12-08/waiters-2.json
@@ -1,7 +1,7 @@
 {
   "version": 2,
   "waiters": {
-    "CertificateIssued": {
+    "CertificateValidated": {
       "delay": 60,
       "maxAttempts": 40,
       "operation": "DescribeCertificate",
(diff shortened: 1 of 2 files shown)


awood45:
We can't accept this directly, as these are files we consume from an outside source. However, I am going to work to push this upstream, which would update the waiters for ALL SDKs, rather than Ruby alone.

awood45:
This will go out with the next release after today. Thank you!

davidcpell:
Thanks, can you point me to where I should submit PRs like this in the future?

davidcpell:
Thanks @awood45. Can you point me to where I should submit PRs like this in the future?

Issue #1813, opened 2018-06-28
Aws::Route53::Client#change_resource_record_sets fails if passing instance of ChangeResourceRecordSetsRequest


AngelCantugr:
### Issue description
When calling Aws::Route53::Client#change_resource_record_sets passing an instance of Aws::Route53::Types::ChangeResourceRecordSetsRequest instead of a Hash, the following error is being thrown: `no member 'id' in struct`

The reason is that the line https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-route53/lib/aws-sdk-route53/plugins/id_fix.rb#L32 assumes that the method parameters are always going to be in a Hash.

### Gem name ('aws-sdk', 'aws-sdk-resources' or service gems like 'aws-sdk-s3') and its version
aws-sdk - 2.11.77
aws-sdk-route53 - 1.10.0 

### Version of Ruby, OS environment
Ruby 2.4
MacOS 10.13.5

### Code snippets / steps to reproduce
1. Initialize a client of Aws::Route53::Client
1. Call the instance method #change_resource_record_sets passing an instance of Aws::Route53::Types::ChangeResourceRecordSetsRequest instead of a Hash

cjyclaire:
Thanks for the contribution, PR has been merged, will be out in the next release!

PR #1814, opened 2018-06-28
Calling Aws::Route53::Client#change_resource_record_sets passing an instance of ChangeResourceRecordSetsRequest throws an error

AngelCantugr:
Bug fix for issue https://github.com/aws/aws-sdk-ruby/issues/1813

When calling Aws::Route53::Client#change_resource_record_sets passing an instance of Aws::Route53::Types::ChangeResourceRecordSetsRequest instead of a Hash, the following error is being thrown: no member 'id' in struct.

This fix makes sure to convert the Struct to hash by using the #to_hash method before trying to find the the given key exists in it. If a Hash is passed, the #.to_hash method will return self. This is to ensure compatibility between ruby versions.

### DIFF
--- a/gems/aws-sdk-route53/CHANGELOG.md
+++ b/gems/aws-sdk-route53/CHANGELOG.md
@@ -1,6 +1,11 @@
 Unreleased Changes
 ------------------
 
+1.11.0 (2018-06-26)
+------------------
+
+* Issue - Update `IdFix` plugin to work with Hash and Struct requests.
+
 1.10.0 (2018-06-26)
 ------------------
 
--- a/gems/aws-sdk-route53/CHANGELOG.md
+++ b/gems/aws-sdk-route53/CHANGELOG.md
@@ -1,9 +1,6 @@
 Unreleased Changes
 ------------------
 
-1.11.0 (2018-06-26)
-------------------
-
 * Issue - Update `IdFix` plugin to work with Hash and Struct requests.
 
 1.10.0 (2018-06-26)
--- a/gems/aws-sdk-route53/lib/aws-sdk-route53/plugins/id_fix.rb
+++ b/gems/aws-sdk-route53/lib/aws-sdk-route53/plugins/id_fix.rb
@@ -29,7 +29,7 @@ module Aws
             # Many operations accept of :id or :hosted_zone_id as a root-level
             # param, pruning prefixes from those.
             [:id, :hosted_zone_id, :delegation_set_id].each do |key|
-              params[key] = remove_prefix(params[key]) if params[key]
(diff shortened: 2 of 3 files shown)


gems/aws-sdk-route53/lib/aws-sdk-route53/plugins/id_fix.rb
```diff
@@ -29,7 +29,7 @@ def remove_id_prefixes(params)
             # Many operations accept of :id or :hosted_zone_id as a root-level
             # param, pruning prefixes from those.
             [:id, :hosted_zone_id, :delegation_set_id].each do |key|
-              params[key] = remove_prefix(params[key]) if params[key]
+              params[key] = remove_prefix(params[key]) if params.to_hash.key?(key)
```
cjyclaire: how about using `to_h` of [Aws::Structure](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/lib/aws-sdk-core/structure.rb#L29) : )

AngelCantugr: I had to rollback the change since it broke the build using Ruby 1.9.3 https://travis-ci.org/aws/aws-sdk-ruby/builds/398037219.

After doing some investigation in the Aws::Structure , I found out that the Aws::Structure#to_h has an alias https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/lib/aws-sdk-core/structure.rb#L46 . So both #to_hash and #to_h point to the same method in the Aws::Structure class.

cjyclaire:
Ah I see, ruby `1.9.3` hash object doesn't have `to_h` method at that time. And your test is providing a mixture of Struct and Hash (which is good as caught by testing!)

Could you also add a changelog entry [here](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-route53/CHANGELOG.md) such that we can track this PR in the next release? It can start with `Issue` pattern

gems/aws-sdk-route53/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 Unreleased Changes
 ------------------
 
+1.11.0 (2018-06-26)
```
cjyclaire: Ah, you don't need to provide the version stuff here, our release tool will bump the version automatically.
Sorry for the confusion, could you remove these 3 lines and just keep the issue entry with its following new line?
After this is cleaned up, I'll merge the PR

AngelCantugr:
@cjyclaire - CHANGELOG.md got updated

PR #1820, opened 2018-07-03
Extensible Credential Providers

YashoSharma:
- [x] Adds process based credential providers into credential resolution chain
- [x] Adds unit tests for credential provider process declarations in config files


awood45:
There are other benefits to it being a standalone credential provider - for example, it would allow users to invoke it directly, rather than as a shared configuration side effect.

gems/aws-sdk-core/lib/aws-sdk-core/process_credentials.rb
```diff
@@ -0,0 +1,54 @@
+module Aws
+  class ProcessCredentials
+
+    include CredentialProvider
+    include RefreshingCredentials
+
+    def initialize(process)
```
cjyclaire: could you add more documentation for `#initialize` like what kind of process variable it's expecting?

gems/aws-sdk-core/lib/aws-sdk-core/process_credentials.rb
```diff
@@ -0,0 +1,54 @@
...
+      @process = process
+      @credentials = credentials_from_process(@process)
+    end
+
+    # @return [Credentials]
```
cjyclaire: `Aws::Credentials`?

gems/aws-sdk-core/lib/aws-sdk-core/process_credentials.rb
```diff
@@ -0,0 +1,54 @@
+module Aws
+  class ProcessCredentials
+
+    include CredentialProvider
+    include RefreshingCredentials
+
```
cjyclaire: Could you also add some doc here like an example of how to use this?
e.g https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/AssumeRoleCredentials.html

gems/aws-sdk-core/spec/aws/process_credentials_spec.rb
```diff
@@ -0,0 +1,43 @@
+require_relative '../spec_helper'
+
```
cjyclaire: Test suites looks good, just checking does these covered all cli test suits for the process credentials as well?

gems/aws-sdk-core/lib/aws-sdk-core/errors.rb
```diff
@@ -102,6 +102,10 @@ def initialize(*args)
       end
     end
 
+    # Raised when a credentials provider process returns a JSON
```
cjyclaire: surprise! ;)  you need to update the region line here: https://github.com/aws/aws-sdk-ruby/blob/master/build_tools/spec/region_spec.rb#L12

That spec is for testing we don't hard code region names, since the error file has been modified, the line number need to be modified as well :D That' why you are seeing the travis error

gems/aws-sdk-core/lib/aws-sdk-core/process_credentials.rb
```diff
@@ -0,0 +1,80 @@
...
+    # credentials provider. 
+    def initialize(process)
+      @process = process
+      @credentials = credentials_from_process(@process)
+    end
+
+    private
+    def credentials_from_process(proc_invocation)
+      begin
+        raw_out = `#{proc_invocation}`
```
awood45: My last point is for us to take another pass on the process invocation choice. Backticks run this way are going to be unescaped, and while we're taking in the data from the credentials file, we may want to be cautious. Take a look at [Open3](http://ruby-doc.org/stdlib-2.3.0/libdoc/open3/rdoc/Open3.html) and let's see if `Open3.capture2` works equally well.

gems/aws-sdk-core/lib/aws-sdk-core/process_credentials.rb
```diff
@@ -0,0 +1,74 @@
...
+    include CredentialProvider
+    include RefreshingCredentials
+
+    # Creates a new ProcessCredentials object, which allows an
+    # external process to be used as a credential provider.
+    #
+    # @param [String] process Invocation string for process
+    # credentials provider. 
+    def initialize(process)
+      @process = process
```
awood45: We're close now, at this point we just need to make sure we're splitting and escaping the string in the same way Python/CLI does: https://github.com/boto/botocore/blob/25a6a4c7c8a8b1797fc78c6a34dc7bca3f0bfba1/botocore/credentials.py#L793-L796

awood45:
So the last thing we need before merging is to add credential provider chain tests to cover what should and should not happen (making sure this is the end of the credential stack, ensuring it's a terminal state).

gems/aws-sdk-core/spec/aws/credential_resolution_chain_spec.rb
```diff
@@ -69,6 +69,11 @@ module Aws
         expect(client.config.credentials.access_key_id).to eq("ACCESS_KEY_SC1")
       end
 
+      it 'prefers process credentials over metadata credentials' do
```
awood45: Last bit here, I would also add some tests with this same profile where we ensure that static profile credentials, for example, are pulled in first. It's a bit thorough, but important for regression.

PR #1824, opened 2018-07-07
Reduce memory usage of Aws::S3::Object#upload_stream

janko:
Once we've taken content from the pipe and uploaded it, we can safely deallocate it. This change greatly reduces memory usage of Aws::S3::Object#upload_stream; when I profiled with a 15MB file, the memory usage went from **17MB** to just **2MB**.

### DIFF
--- a/gems/aws-sdk-s3/CHANGELOG.md
+++ b/gems/aws-sdk-s3/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Reduce memory usage of `Aws::S3::Object#upload_stream` when `StringIO` is used
+
 1.36.0 (2019-03-27)
 ------------------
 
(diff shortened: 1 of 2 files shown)


yvancastilloux:
Hi, we've seen this issue on our side too. We upload a file that has 100MB+ and the memory use is around 100MB. This fix is definitely needed. What's preventing this small fix to be merged? I see that the tests are not passing...

yvancastilloux:
Awesome! Can't wait to try this out!

cjyclaire:
Apologies for the delay and appreciate the contribution!
Mind help add a changelog entry [here](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-s3/CHANGELOG.md) before I merge this? (thus it can be tracked in the next release)

janko:
@cjyclaire Added the changelog entry

yvancastilloux:
Can't wait to see this released! Thank you so much guys!

PR #1825, opened 2018-07-07
Reduce memory usage of IOEncryptor and IODecryptor

janko:
These changes reduce memory usage of `Aws::S3::Encryption::Client` when uploading and downloading. I profiled memory usage with a 15MB file, and I got the following results:

* `#put_object` – Before: 32MB, After: 5MB
* `#get_object` – Before: 34MB, After: 19MB

I chose to use binary buffers because from my experience an UTF-8 buffer was unsafe for `Cipher#update`.


janko-m:
It seems that in JRuby, `OpenSSL::Cipher#update` doesn't accept the buffer argument – https://github.com/jruby/jruby/issues/5242. I've tested that it's still behaviour in JRuby 9.2.0.0.

mullermp:
@janko Is this addressable now?

Seems that JRuby team had merged your change, and they released `jruby-openssl` 0.10.2 
 with it (https://github.com/jruby/jruby/pull/5639)

According to https://github.com/jruby/jruby-openssl#compatibility , 0.10.0 is compatible with JRuby 1.7.20 all the way to 9.2.x. That would imply it's dependent on the openssl version and not JRuby version.

mullermp:
I fixed the merge conflict in the travis config, let's see if it succeeds!

mullermp:
Aha. It succeeds. I think this might be safe to merge? @janko your thoughts?

janko:
@mullermp Great, I think it's ready to merge then 👍

mullermp:
Ahh.. It's missing a changelog. Are you around to add an entry real quick? If not I can see if I have permission to add a commit to this fork or add a follow up commit.

mullermp:
Yuck. JRuby-1.7 runs now, but still fails.

It seems rvm fails to get the gemset `Error running 'gemset_import_list jruby-openssl jruby-launcher gem-wrappers rubygems-bundler bundler rake rvm',`

mullermp:
It looks like `jruby-openssl` installed the correct version, unsure if the other errors are related. Any ideas? https://travis-ci.org/aws/aws-sdk-ruby/jobs/649613817#L581

mullermp:
Going to sit on this (again) .. I've added 2.7 and jruby1.7 (not jruby19-mode which appears to be not correct) to our build in master. We're trying to move customers from V2 to V3, and a large number of JRuby1.7 customers are on V2. We'd like to keep this open (for now).

mullermp:
We released a V2 encryption client that has been disabled for Ruby 1.9 (and consequently JRuby 1.7). We can merge these changes into it.

mullermp:
This might be a bust..

gems/aws-sdk-s3/lib/aws-sdk-s3/encryptionV2/io_encrypter.rb
```diff
@@ -52,8 +52,12 @@ def encrypt_to_stringio(cipher, plain_text)
         def encrypt_to_tempfile(cipher, io)
           encrypted = Tempfile.new(self.object_id.to_s)
           encrypted.binmode
-          while chunk = io.read(ONE_MEGABYTE)
-            encrypted.write(cipher.update(chunk))
+          while chunk = io.read(ONE_MEGABYTE, read_buffer ||= String.new)
+            if cipher.method(:update).arity == 1
```
mullermp, line 56: Yuck...

alextwoods: Agreed - I don't love this.  I don't think we can count on the versions of openssl a user has so we have three options:
1. Check openssl versions instead of arity.  This is less direct but maybe makes it more clear what why we're doing this.  
2.  Create a wrapper around cipher to normalize this interface difference (push this mess down).  
3. Drop this change -  >80% of our users would see improved performance from this.

I lean towards taking this change with the arity check (with better documentation in comments) since its a fairly significant improvement in performance for the bulk of our users. 

Side note: I created a seperate PR for updating the travis.yml: #2340

mullermp:
Thanks for your contribution and sorry this took way too long.

PR #1827, opened 2018-07-10
Logged API Requests interface

YashoSharma:

- [x] Adds an accessible `api_requests` to all clients. When a client is stubbed it will write all requests it makes to `api_requests` which can be accessed from the stubbed client.

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb
@@ -27,9 +27,8 @@ module Aws
       requests = @api_requests
       self.handle do |context|
         requests << {
-          metadata: context.metadata,
-          request: context.http_request,
-          response: context.http_response,
+          operation_name: context.operation_name,
+          params: context.params,
           context: context
         }
         @handler.call(context)
--- a/gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb
@@ -6,6 +6,9 @@ module Aws
   # return when a client is using stubbed responses. Pass
   # `:stub_responses => true` to a client constructor to enable this
   # behavior.
+  #
+  # Also allows you to see the requests made by the client by reading the
+  # api_requests instance variable
   module ClientStubs
 
     # @api private
@@ -17,6 +20,16 @@ module Aws
           apply_stubs(operation_name, Array === stubs ? stubs : [stubs])
         end
       end
+
+      # When a client is stubbed allow the user to access the requests made
+      @api_requests = []
+
+      requests = @api_requests
+      self.handle do |context|
+        requests << context
+        @handler.call(context)
+      end
+      self.class.attr_reader :api_requests
     end
 
     # Configures what data / errors should be returned from the named operation
--- a/gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb
@@ -29,7 +29,6 @@ module Aws
         requests << context
         @handler.call(context)
       end
-      self.class.attr_reader :api_requests
     end
 
     # Configures what data / errors should be returned from the named operation
@@ -175,6 +174,16 @@ module Aws
       else
         msg = 'stubbing is not enabled; enable stubbing in the constructor '
         msg << 'with `:stub_responses => true`'
+        raise msg
+      end
+    end
+
+    def api_requests
+      if config.stub_responses
+        @api_requests
+      else
+        msg = 'you must enable stubbing in order to client requests;'
+        msg << ' enable stubbing in the constructor with `:stub_responses => true`'
         raise msg
       end
     end
--- a/gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb
@@ -184,7 +184,8 @@ module Aws
 
     # Allows you to access all of the requests that the stubbed client has made
     #
-    # @return [Array] Returns an array of the api requests made
+    # @return [Array] Returns an array of the api requests made, each request object contains the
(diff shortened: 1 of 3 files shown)


gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb
```diff
@@ -179,6 +178,16 @@ def stub_responses(operation_name, *stubs)
       end
     end
 
+    def api_requests
+      if config.stub_responses
+        @api_requests
+      else
+        msg = 'you must enable stubbing in order to client requests;'
```
awood45: Maybe just say:

"This method is only implemented for stubbed clients, and is available when you enable stubbing in the constructor with `stub_responses: true`."

I think this wording is important because it can read like you're supposed to turn on stubbing, when really you're just using this if you know you need it, like for testing.

I'd also raise a typed exception, perhaps `NotImplementedError`.

gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb
```diff
@@ -17,6 +20,15 @@ def setup_stubbing
           apply_stubs(operation_name, Array === stubs ? stubs : [stubs])
         end
       end
+
+      # When a client is stubbed allow the user to access the requests made
+      @api_requests = []
+
+      requests = @api_requests
+      self.handle do |context|
```
awood45: I think as discussed before, we should have a more complex object here, so long as we're actually providing this functionality as a first class member. Maybe create a dedicated class for this, even potentially a dedicated handler class which is inserted when we stub the client.

But for a dedicated class, we could also record more data about each request. We can record the parameters, keep the context object in full, record the method name, and provide convenient helper methods for each.

gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb
YashoSharma: The `stub_responses` method raises a generic `RuntimeError`, should I leave the error handling the way it is for consistency or use a typed exception?

gems/aws-sdk-core/spec/aws/client_spec.rb
```diff
@@ -150,6 +150,22 @@ module Aws
         expect(client.example_operation.string).to eq('value')
       end
 
+      it 'allows api requests to be logged when stubbed' do 
+        client = client_class.new(stub_responses: {
+          example_operation: { string: 'value' }
+        })
+        expect(client.api_requests.length).to eq(0)
+        client.example_operation
+        expect(client.api_requests.length).to eq(1)
```
awood45: Let's actually verify the contents, and use an example operation with parameters to verify that users can do the same validations on a first-class basis as you can in the aws-record examples. This is especially important because if we make changes later that would break stubbing behavior, we want to exercise all of it (assertions on everything we expose) so that those changes would break the test. Right now, the test has no contract on what actually is in the api_requests object.

gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb
```diff
@@ -17,6 +20,20 @@ def setup_stubbing
...
+        requests << {
+          metadata: context.metadata,
+          request: context.http_request,
+          response: context.http_response,
+          context: context
+        }
```
awood45: I think it would make sense to explicitly pull out and provide the params. See: https://github.com/aws/aws-sdk-ruby-record/blob/master/spec/aws-record/record/item_operations_spec.rb#L45

Consider pulling out `context.params` as the member `:params`, `context.operation.name` as the member `:operation`, and then just dropping the `:context` object. I think that will focus on the most important parts, while still providing the full context.

gems/aws-sdk-core/spec/aws/client_spec.rb
```diff
@@ -150,6 +150,31 @@ module Aws
...
+
+        it 'allows api requests to be logged when stubbed' do 
+          expect(client.api_requests.empty?).to be(true)
+          client.create_bucket(bucket:'aws-sdk')
+          expect(client.api_requests.length).to eq(1)
+          
+          log_obj = client.api_requests[0]
+          expect(log_obj[:operation_name]).to eq(:create_bucket)
+          expect(log_obj[:params]).to eq({:bucket=>"aws-sdk"})
+          expect(log_obj[:context].metadata).to eq({:gem_name=>"aws-sdk-sampleapi2", :gem_version=>"1.0.0", :response_target=>nil, :original_params=>{:bucket=>"aws-sdk"}, :request_id=>"stubbed-request-id"})
```
cjyclaire: nitpick, could you make this multi-line instead of this super long line?

gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb
cjyclaire: I'd +1 for this, any follow up or discussion on the advice?

gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb
```diff
@@ -166,6 +182,20 @@ def stub_responses(operation_name, *stubs)
       end
     end
 
+    # Allows you to access all of the requests that the stubbed client has made
+    #
+    # @return [Array] Returns an array of the api requests made
```
cjyclaire: If not create a delicate class for it, could you document that each element in the array is actually a hash containing request info? And supported keys for those hashes?

gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb
YashoSharma: There was some brief follow up in the outdated comments after Alex cross checked against aws-record

gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb
YashoSharma: I added documentation for everything inside of the log hash, but since the params hash can be variable based on whatever operation the client made I don't think enumerating the supported keys for it is feasible

PR #1830, opened 2018-07-11
Encode Custom service name if encoded

cjyclaire:

This PR will remain pending until original encode strategy is confirmed.

### DIFF
--- a/build_tools/spec/custom_service_spec.rb
+++ b/build_tools/spec/custom_service_spec.rb
@@ -22,6 +22,23 @@ module BuildTools
       expect(cs.svc_name).to eq("SdkTestApi")
     end
 
+    it 'format svc_name with spaces' do
+      cs = CustomService.new(
+        service_name: "Sdk teSt api",
+        default_endpoint: "https://foo.com",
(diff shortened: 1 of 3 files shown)


cjyclaire:
This is now ready for review

build_tools/spec/custom_service_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    it 'decode svc_name properly' do
+      cs = CustomService.new(
+        service_name: "Sdk%20%20test%20Api", 
+        default_endpoint: "https://foo.com",
+        model_path: "foo/path"
+      )
+      expect(cs.svc_name).to eq("SdkTestApi")
+    end
+
```
awood45: I'd actually like us to add more tests. For example, service names with spaces, and possibly some service names which should fail by intention. This covers the basics, I'd like to get a bit creative on cases.

cjyclaire: Added more tests in the latest commit, unfortunately, it would be tricker for failed scenarios since  that will happened at Kernel require time later at build times

PR #1839, opened 2018-07-21
moving towards a more 'frozen strings literal friendly' ruby

plentz:
trying to eliminate all places that modify string literals.

we still can't run the build using `RUBYOPT="--enable-frozen-string-literal" rake`, because of [kramdown](https://github.com/gettalong/kramdown) (cc @gettalong), but that's something I'm going to address tomorrow.

### DIFF
--- a/gems/aws-eventstream/lib/aws-eventstream/bytes_buffer.rb
+++ b/gems/aws-eventstream/lib/aws-eventstream/bytes_buffer.rb
@@ -1,5 +1,7 @@
+# frozen_string_literal: true
+
 module Aws
-  module EventStream 
+  module EventStream
 
     # @api private
     class BytesBuffer
@@ -35,7 +37,7 @@ module Aws
       end
 
       def write(bytes)
-        @data <<= bytes
+        @data += bytes
         bytes.bytesize
       end
       alias_method :<<, :write
--- a/gems/aws-sdk-core/lib/aws-sdk-core/shared_config.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/shared_config.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module Aws
 
   # @api private
@@ -283,7 +285,7 @@ module Aws
       unless (@parsed_credentials && @parsed_credentials[profile]) ||
           (@parsed_config && @parsed_config[profile])
         msg = "Profile `#{profile}' not found in #{@credentials_path}"
-        msg << " or #{@config_path}" if @config_path
+        msg += "or #{@config_path}" if @config_path
         raise Errors::NoSuchProfileError.new(msg)
       end
     end
--- a/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/helper.rb
+++ b/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/helper.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require 'set'
 # kramdown
 require 'kramdown'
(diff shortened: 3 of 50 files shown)


gettalong:
Why would you use the `--enable-frozen-string-literal`Ruby option instead of the the `# frozen_string_literal: true` comment? What's the benefit

The latter is completely under your control whereas the former is not. So the option can only meaningfully be used if *everything* is under your control. If you use gems, you are out of luck.

As for kramdown: I will change it to use the frozen string literal comment some time in the future (also see https://github.com/gettalong/kramdown/pull/512).

plentz:
@gettalong thanks for the quick response! the option was more of a _desired_ thing than a _required_ thing. And it's better to have everything moving in the "right" direction (maybe put that as an allowed-to-fail build step).

plentz:
I think that the branch is ready now. I've put the `# frozen_string_literal: true` just on the files that it caused a problem to appear, to make it easier to review - but we can add it to all files - and I think it's the best option.

One spec is failing and I don't know what's cause, but it doesn't appear to be related to my changes.

awood45:
Tests appear to be passing now, just reviewing for possible sharp edges. Thanks for submitting!

gems/aws-eventstream/lib/aws-eventstream/bytes_buffer.rb
```diff
@@ -35,7 +37,7 @@ def readbyte
       end
 
       def write(bytes)
-        @data <<= bytes
+        @data += bytes
```
awood45: Is this change related to freezing string literals, or just an extra change you wanted to make added in?

gems/aws-sdk-core/lib/aws-sdk-core/shared_config.rb
```diff
@@ -283,7 +285,7 @@ def validate_profile_exists(profile)
       unless (@parsed_credentials && @parsed_credentials[profile]) ||
           (@parsed_config && @parsed_config[profile])
         msg = "Profile `#{profile}' not found in #{@credentials_path}"
-        msg << " or #{@config_path}" if @config_path
+        msg += "or #{@config_path}" if @config_path
```
awood45: Same comment on relation to the PR intention.

gems/aws-eventstream/lib/aws-eventstream/bytes_buffer.rb
plentz: yep, is related to freezing strings. but tests only break here if we add `frozen_string_literal` to:

```
rspec ./build_tools/aws-sdk-code-generator/spec/interfaces/client/event_stream_spec.rb
rspec ./build_tools/aws-sdk-code-generator/spec/protocols_spec.rb
rspec ./gems/aws-eventstream/spec/bytes_buffer_spec.rb
rspec ./gems/aws-eventstream/spec/decoder_spec.rb
rspec ./gems/aws-eventstream/spec/encoder_spec.rb
```

but I added in this PR just what needs to be changed to fix the problems I've found. I think that we can run

```
pragmater --add --comments "# frozen_string_literal: true" --includes "Gemfile" "Guardfile" "Rakefile" ".gemspec" "config.ru" "bin/**/*" "**/*.rake" "**/*.rb"
````

after merging this to ensure that everything stays compatible ;) [and that's how I found this kind of change]

gems/aws-sdk-core/lib/aws-sdk-core/shared_config.rb
plentz: yes, this one is directly related to the change we're doing to the string in the line above

mullermp:
Is this change still desirable? Matz has said that [he's abandoning making frozen string literals the default for Ruby 3](https://bugs.ruby-lang.org/issues/11473#note-53).

mullermp:
Duplicate of #1736 and closing for now. We would like to add rubocop and these would be added for us.

PR #1482, opened 2017-04-11
Fix typo (http_reed_timeout)

kriskhaira:

PR #1844, opened 2018-07-27
Revert "Fix typo (http_reed_timeout) (#1482)" which is intended typo

wktk:
I think this section is an example for typo in a parameter and we should not fix this typo since it is intended.


awood45:
Good catch, thanks!

PR #1848, opened 2018-08-06
Add method to enable generating eks token

smcavallo:
Add method for getting a presigned_url for the get_caller_identity sts service

For ruby users who have tools/scripts/applications which interact with EKS, we'd like an equivalent to
https://github.com/kubernetes-sigs/aws-iam-authenticator
This should provide that equivalent.

This exposes the presigned_url for get_caller_identity which will allow generation of the EKS token.
Ex.
```
    resp = client.get_caller_identity_presigned_url({
        headers: {"X-K8s-Aws-Id" => 'my-eks-cluster'},
      })
    eks_token = "k8s-aws-v1." + Base64.urlsafe_encode64(resp).chomp("==")
```

Signed-off-by: S.Cavallo <smcavallo@hotmail.com>


awood45:
It's important to note that the client itself is generated code, and this PR if accepted would be written over by the next STS release. If you'd like to expose presigner functionality for STS, I'm open to it, but I'd follow the pattern we used for S3 of having a specific presigner class.

smcavallo:
@awood45 - thank you!  This has now been refactored as a Presigner class.
I used the presigner classes from S3 and Polly as examples.

I definitely would like to expose the presigner functionality for STS.  The new EKS services all use auth tokens which are based on the get_caller_identity presigned_url.  It would be nicer to have a helper method in the sdk to retrieve a token, but creating a presigner for get_caller_identity will at least get part of the way there.

gems/aws-sdk-core/spec/aws/sts/client_spec.rb
```diff
@@ -15,15 +15,15 @@ module STS
           Seahorse::Client::Response.new(context: context)
         end
         resp = client.assume_role_with_web_identity(
-          role_arn: 'role-arn',
-          role_session_name: 'role-session-name',
-          web_identity_token: 'token'
+            role_arn: 'role-arn',
```
awood45: Can we fix the formatting on this file? It was correct before.

mullermp:
Sorry that this PR was dropped. I am going through and picking up old reviews and issues. I fixed up some docs/formatting of the code and pushed to your fork. I will go and merge this. Thanks for the contribution.

mullermp:
I had to revert this change. It broke code generation.

lwoggardner:
Hi AWS team,  I have a use-case to validate the caller of a lambda function by passing a presigned caller identity.  I would be ideal if the documentation for this included an example of how to validate the URL via a raw Net:HTTP call  (including passing matching headers).

mullermp:
@lwoggardner Thanks for your feedback on the PR. I'd be happy to include any additional documentation if you've already figured out how to do that. Personally I've never had such a use case and I don't have the available bandwidth at the moment.

PR #1863, opened 2018-08-22
use waiter and avoid double delete at s3 integration test

cjyclaire:

Tested locally


cjyclaire:
After approve will port be to v2 as well

Issue #1866, opened 2018-08-27
The API request record also saves built-but-not-sent requests from aws-sdk-s3 presigning


janko-m:
### Issue description

I have been using the new API requests record, and I noticed that when I make a presigned URL, the corresponding API call is saved in the record, even though the request itself was not sent.

I assume the place which triggers the "fake" API request recording is here: https://github.com/aws/aws-sdk-ruby/blob/8dbe883f810271a2132b7a754ac79faf61a66a2b/gems/aws-sdk-s3/lib/aws-sdk-s3/presigner.rb#L51-L54

Since in my project I want to be diligent with how many API requests I'm actually sending, it would be ideal if `Aws::S3::Presigner#presigned_url` didn't add anything to the API request record.

### Gem name ('aws-sdk', 'aws-sdk-resources' or service gems like 'aws-sdk-s3') and its version

* aws-sdk-core 3.24.1
* aws-sdk-s3 1.17.0

### Version of Ruby, OS environment

* Ruby 2.5.1

### Code snippets / steps to reproduce

```rb
require "aws-sdk-s3"

client = Aws::S3::Client.new(stub_responses: true)
presigner = Aws::S3::Presigner.new(client: client)

presigner.presigned_url(:get_object, bucket: "my-bucket", key: "foo")

p client.api_requests[0][:operation_name]
```
```
:get_object
```

awood45:
This is a good callout, I hadn't thought of this. I'm happy to take a pull request for this as well.

cjyclaire:
PR #1867 is opened addressing this issue

awood45:
So thinking about this, I'm a little concerned that changing this could break customers who have started to write test expectations against this. Would it meet your needs if we instead lifted whether or not the request was presigner based to the top level of the API requests object? Or, possibly, if we allowed you to filter them out with an optional param?

cjyclaire:
PR has been merged, now you can try:
```ruby
client.api_requests(exclude_presign: true).size.zero?
# => true
```

This will be out in the next release, closing : )

janko-m:
Thanks for the quick fix! Ok, the additional option makes sense to me for preserving backwards compatibility.

PR #1867, opened 2018-08-27
Added an option for exclude presigned url not_sent request usage in api_request counts

cjyclaire:

Addressing #1866

### DIFF
--- a/gems/aws-sdk-s3/spec/presigner_spec.rb
+++ b/gems/aws-sdk-s3/spec/presigner_spec.rb
@@ -55,6 +55,13 @@ module Aws
           presigner = Presigner.new(client: stub_client)
           presigner.presigned_url(:get_object, bucket: 'bkt', key: 'k')
           expect(stub_client.api_requests(exclude_presign: true).size).to be_zero
+
+          stub_client.list_buckets
+          stub_client.list_buckets
+          presigner.presigned_url(:get_object, bucket: 'bkt', key: 'k')
+          stub_client.list_buckets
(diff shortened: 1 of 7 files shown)


awood45:
We should also carefully call out in release notes how tests could break when updating to this - if customers are relying on this to be collected for presigned requests, they'll need to adapt.

awood45:
On second thought, hold on merging this. I don't want to cause surprise test breakage if there is another way to avoid this.

gems/aws-sdk-s3/spec/presigner_spec.rb
```diff
@@ -43,6 +43,20 @@ module S3
...
+          presigner = Presigner.new(client: stub_client)
+          presigner.presigned_url(:get_object, bucket: 'bkt', key: 'k')
+          expect(stub_client.api_requests.size).to eq(1)
+        end
+
+        it 'can be excluded from being tracked as an api request' do
+          stub_client = Aws::S3::Client.new(stub_responses: true, exclude_presign: true)
+          presigner = Presigner.new(client: stub_client)
+          presigner.presigned_url(:get_object, bucket: 'bkt', key: 'k')
+          expect(stub_client.api_requests.size).to be_zero
```
awood45: I actually imagined this as an option you could pass in to the `api_requests` method itself. Should be easy to replace the attr_reader with a method that takes an optional param to do this. That way, you don't need to decide at client construction time.

cjyclaire:
Also addressed doc issue mentioned in #1868

PR #1877, opened 2018-09-18
[Bug Fix] fix retry for error happened after initial success signal

cjyclaire:

(seahorse)`:response_target` [plugin](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/lib/seahorse/client/plugins/response_target.rb#L41) (90) has a higher priority over (s3 encryption) `:decrypt_handler` [plugin](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-s3/lib/aws-sdk-s3/encryption/decrypt_handler.rb#L55)(default: 50). When registering error callbacks using `on_error`, response_target callback will be triggered before decrypt_handler one.

Thus after an initial 200 response, once data chunk write [starts](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-s3/lib/aws-sdk-s3/encryption/io_decrypter.rb), if an error occurs, `context.http_response.body` will be failed to be updated with https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-s3/lib/aws-sdk-s3/encryption/decrypt_handler.rb#L57. Thus could cause problem at [`#reset`](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/lib/seahorse/client/http/response.rb#L152) (before each retry attempt), where body is not truncate or rewind properly.

PR propose a fix that avoids body update at response target if the body is an io_decryptor
Open to spec enhancement.


gems/aws-sdk-core/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Update `:response_target` plugin error handling
```
awood45: I would expand on this. What GitHub issue does it solve, or, what kind of error does this address? Helps users identify if they need to pick up this upgrade in particular.

gems/aws-sdk-core/lib/seahorse/client/plugins/response_target.rb
```diff
@@ -41,7 +41,9 @@ def add_event_listeners(context, target)
             context.http_response.on_error do
               body = context.http_response.body
               File.unlink(body) if ManagedFile === body
-              context.http_response.body = StringIO.new
+              unless context.http_response.body.respond_to?(:io)
```
awood45: It may be worth adding a comment explaining why this conditional is important for later reference if looking at this code months later.

gems/aws-sdk-core/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Fix the bug at `:response_target` plugin error callback that removes special body object before its error callback can be triggered. (Bug: s3 decrypt response body is not truncated/rewind properly causing malformed body write when error retry happened after initial success request)
```
awood45: Revised:
```
* Issue - Fixes a bug in the `:response_target` plugin error callback. Under certain circumstances a special body object can be removed before its error callback is triggered, breaking retry logic.
```

PR #1894, opened 2018-10-20
Add waiters for rds cluster snapshots

bpilkerton:
Adds two new waiters for RDS cluster snapshots -- snapshot available and snapshot deleted.

### DIFF
--- a/apis/rds/2014-10-31/waiters-2.json
+++ b/apis/rds/2014-10-31/waiters-2.json
@@ -170,6 +170,91 @@
           "argument": "DBSnapshots[].Status"
         }
       ]
+    },
+    "DBClusterSnapshotAvailable": {
+      "delay": 30,
+      "operation": "DescribeDBClusterSnapshots",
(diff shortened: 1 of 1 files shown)


pchaganti:
👍

mullermp:
Apologies for a late response. This code is auto generated, so I cannot merge this as is. I can apply these changes upstream so that they are built into all of the SDKs; have you tested these changes locally? In a quick sanity test, I was able to use this waiter for db_cluster_snapshot_available but was unable to get it working for db_cluster_snapshot_deleted.

mullermp:
I've merged these changes upstream, although I am concerned that the deletion waiters for RDS do not work (for any SDK) because RDS#Describe* operations will never list a resource with status="deleted" and thus the waiter will never return.

mullermp:
Good news: I've fixed the deleted waiters for RDS as well. I will attempt to get those changes merged in and deployed.

PR #1931, opened 2018-12-05
Added support for encryption context to S3 encryption client

ankane:
Hi, this adds support for KMS [Encryption Context](https://docs.aws.amazon.com/kms/latest/developerguide/encryption-context.html) for S3 uploads with the `Aws::S3::Encryption::Client`. This allows you to know which files are being decrypted.

```ruby
client = Aws::S3::Encryption::Client.new(kms_key_id: "alias/my-key")

client.put_object(
  body: File.read("test.txt"),
  bucket: "my-bucket",
  key: "test.txt",
  encryption_context: {key: "test.txt"}
)

resp = client.get_object(
  bucket: "my-bucket",
  key: "test.txt",
  encryption_context: {key: "test.txt"}
)
```

Currently, the materials description is used as the context and stored in the envelope as `x-amz-matdesc`. By design, the additional encryption context is not stored in the envelope.

An open question is whether the user-supplied encryption context should:

1. Replace materials description context entirely
2. Merge with materials description context, with user-supplied context taking precedence (this is what the PR does)

This feature was asked about in the Java SDK as well (https://github.com/aws/aws-sdk-java/issues/1631), so it'd probably be good to standardize the approach across languages.

Lastly, I wasn't sure the best way to test, so let me know if it can be improved. Thanks!


alextwoods:
We have just released a V2 of the encryption client - and I agree that the KMS Encryption context should support user provided values.  We want to leave the V1 client unchanged, but I'll migrate these changes into the V2 client and update this PR.

ankane:
Great, thanks @alextwoods!

mullermp:
@ankane should also review :)

ankane:
+1 to the new name

alextwoods:
@ankane  - I've gone ahead and merged this.  Thanks for your contribution!

ankane:
Awesome, thanks for adding it!

PR #1938, opened 2018-12-09
Added support for authenticated encryption

ankane:
Hi, this adds support for authenticated encryption (AES-GCM). Currently, the gem can decrypt AES-GCM. This allows it to encrypt as well.

```ruby
Aws::S3::Encryption::Client.new(
  kms_key_id: "alias/my-key",
  authenticated_encryption: true
)
```

The Java SDK uses [CryptoMode](https://aws.amazon.com/blogs/developer/amazon-s3-client-side-authenticated-encryption/) for this. In the future, a `:strict` option can be added to only allow for authenticated decryption.

screen shot 2018-12-09 at 1 38 46 pm

### DIFF
--- a/gems/aws-sdk-s3/CHANGELOG.md
+++ b/gems/aws-sdk-s3/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Feature - Adds support for authenticated encryption.
+
 1.30.0 (2018-12-04)
 ------------------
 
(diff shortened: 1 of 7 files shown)


alextwoods:
Thanks for the contribution and apologies for the delay in responding.   We have just released a V2 of the encryption client (see #2336 ) which adds support (and in fact requires) authenticated encryption using AES-GCM for both the cek algorithm and the key wrapping algorithm.  The V2 client provides support for decrypting objects encrypted with the V1 client as well as compatibility/interop with other language SDKs.  

Since this is supported through the V2 client, I'm going to close out this PR.

ankane:
Awesome, thanks for adding it, and glad to hear it's required :tada:

PR #1944, opened 2018-12-25
FEATURE: Add support to yield the response if block given

vinothkannans:
It will be helpful to get `etag` value like below

```ruby
is_success = obj.upload_file(file, options) do |response|
    etag = response.etag
end
```


cjyclaire:
Thanks for the contribution! Could you also add a test case for your changes?

vinothkannans:
@cjyclaire test cases are added.

mullermp:
Apologies for not getting to this sooner. I will provide more detailed code comments.

gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
```diff
@@ -305,7 +305,8 @@ def upload_file(source, options = {})
         uploader = FileUploader.new(
           multipart_threshold: uploading_options.delete(:multipart_threshold),
           client: client)
-        uploader.upload(source, uploading_options.merge(bucket: bucket_name, key: key))
+        response = uploader.upload(source, uploading_options.merge(bucket: bucket_name, key: key))
```
mullermp: Please add some documentation to this method about using the block feature. Also please limit this line to 80 characters like in [master](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb#L317). You should be able to rebase against the latest changes.

(we plan to add back Rubocop which is why I'm starting to tackle violations slowly..)

gems/aws-sdk-s3/spec/object/upload_file_spec.rb
```diff
@@ -94,6 +94,12 @@ module S3
             object.upload_file(ten_meg_file.path)
           end
 
+          it 'yields the response to the given block' do
```
mullermp: I don't think two tests make sense - this feature should work irregardless of large objects or small objects. You can move this test "up one level" out of the 'describe' block and in its own 'it' block, then delete the other test.

vinothkannans:
@mullermp the changes are done now.

mullermp:
These look good! Thanks for fixing that.

One more thing - I'm sorry I forgot to mention this, because I forgot myself.. - you'll need to add an entry to the aws-sdk-s3/CHANGELOG.md file so that your change will be included in the next version. Please add this line under the "unreleased changes" section:

`* Feature - Add support to yield the response in #upload_file if a block is given`

After you add that, I can merge this. :)

mullermp:
Since I botched up the review for that changelog, I went ahead and did it for you. Thanks for the contribution!

PR #1942, opened 2018-12-20
[H2] Support H2 service/operations

cjyclaire:
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

This PR is aimed for async usage for operations over HTTP2 protocol (bi-direction/uni-direction)

~~CI currently failed as expected~~
~~[WIP]~~

~~TODOs~~
1. ~~event signal param validation~~
2. ~~retry story~~
3. ~~code gem docs for async operations/shapes/usage~~
4. ~~dependency story~~
5. ~~CI~~

PR #1946, opened 2018-12-29
[H2] Support event sigv4 signing

cjyclaire:

this PR separate `aws-sigv4` gem and `aws-eventstream` gem diff for easier review of #1942 
after review complete, this PR will be merged before #1942


cjyclaire:
This PR is ready for review

gems/aws-sigv4/lib/aws-sigv4/signer.rb
```diff
@@ -243,6 +244,24 @@ def sign_request(request)
         )
       end
 
+      # Signs a event and returns signature headers and prior signature
```
awood45: Let's flesh out this documentation with documentation of parameters and the return value. This method is doing a lot of stuff (for one example, explaining how we're changing the encoding at the end with the unpacking).

gems/aws-sigv4/lib/aws-sigv4/signer.rb
```diff
@@ -375,6 +394,19 @@ def string_to_sign(datetime, canonical_request)
         ].join("\n")
       end
 
+      def event_string_to_sign(datetime, headers, payload, prior_signature, encoder)
+        encoded_headers = encoder.encode_headers(
+          Aws::EventStream::Message.new(headers: headers, payload: payload)).read
```
awood45: indentation here is a bit hard to follow, I'd at least most the last `).read` down to its own line.

gems/aws-sigv4/spec/signer_spec.rb
```diff
@@ -270,6 +271,29 @@ module Sigv4
 
       end
 
+      context '#sign_event' do
```
awood45: Are there any partial event receive cases we need to handle here, or is that abstracted out before we get to this phase? Defensive coding here is a good idea.

cjyclaire: Yeah, currently no. This event signature v4 signing is currently used for event sending only and received events only requires eventstream decoding.

cjyclaire:
Ready for another review  🎉

PR #1953, opened 2019-01-06
Replace Fixnum with Integer in HashFormatter code generator

janko:
The `Fixnum` class has been deprecated in newer Ruby versions, the `Integer` superclass should now be used. This gets rid of the related deprecation warnings when running tests.


cjyclaire:
Thanks for the contribution! merging

PR #1962, opened 2019-01-15
Retain all provided headers when signing a request

### LINKED ISSUE #1961: Custom headers provided to the sign_request method are not present in the signed output headers
### Issue description

The request headers provided to `Aws::Sigv4#sign_request` are not retained in the signed output headers. When a user supplies the request object and headers to this method, they get back a `Signature` object that has the correct `authorization` header, but lacks any of the custom headers that were part of the provided request.

You can see that the canonical request includes those custom headers in the "signed headers" section, but again those headers are not retained in the output headers that include authorization.

This means if the user issues a request with only the headers found in the output hash, they'll get a signature mismatch because the custom headers they provided were not present. A workaround would be to manually re-add those headers back into the returned `Signature` before executing a request.

aws-sigv4 1.0.3

This is not a Ruby/OS specific issue.

```ruby
signature = Signer.new(options).sign_request(
  http_method: 'PUT',
  url: 'https://domain.com',
  headers: {
    'Foo' => 'foo',
    'Bar' => 'bar  bar',
    'Bar2' => '"bar  bar"',
    'Content-Length' => 9,
    'X-Amz-Date' => '20120101T112233Z',
  }
)
# this fails
expect(signature.headers['foo']).to eq('foo')
```

JAndritsch:
Fixes #1961

### DIFF
--- a/gems/aws-sigv4/spec/signer_spec.rb
+++ b/gems/aws-sigv4/spec/signer_spec.rb
@@ -268,6 +268,24 @@ module Aws
           expect(signature.headers['authorization']).to eq('AWS4-HMAC-SHA256 Credential=akid/20120101/REGION/SERVICE/aws4_request, SignedHeaders=bar;bar2;foo;host;x-amz-content-sha256;x-amz-date, Signature=4a7d3e06d1950eb64a3daa1becaa8ba030d9099858516cb2fa4533fab4e8937d')
         end
 
+        it 'retains all custom headers from the provided request' do
+          signature = Signer.new(options).sign_request(
+            http_method: 'PUT',
+            url: 'https://domain.com',
+            headers: {
+              'Foo' => 'foo',
+              'Bar' => 'bar  bar',
+              'Bar2' => '"bar  bar"',
+              'Content-Length' => 9,
+              'X-Amz-Date' => '20120101T112233Z',
+            }
+          )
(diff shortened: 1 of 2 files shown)


gems/aws-sigv4/spec/signer_spec.rb
```diff
@@ -268,6 +268,24 @@ module Sigv4
           expect(signature.headers['authorization']).to eq('AWS4-HMAC-SHA256 Credential=akid/20120101/REGION/SERVICE/aws4_request, SignedHeaders=bar;bar2;foo;host;x-amz-content-sha256;x-amz-date, Signature=4a7d3e06d1950eb64a3daa1becaa8ba030d9099858516cb2fa4533fab4e8937d')
         end
 
+        it 'retains all custom headers from the provided request' do
```
awood45: I want to make sure we include a test that provides the `:unsigned_headers` value as well, ensuring that we do not sign those headers. Say, `'Bar2'` or similar.

The situation I want to avoid is where we suddenly start signing banned headers, which then get modified by proxies, and then cause breaking changes down the line.

cjyclaire:
Thanks for the contribution!  This can be a breaking change for current behavior, could you make this behavior (signing all headers by default) an option instead for the #sign_request?

mullermp:
Closing as we have not heard back. Reopen if needed!

PR #1963, opened 2019-01-17
Allow nil for url argument of Aws::CloudFront::CookieSigner#singed_cookie

knjko:
Thanks for the project!

Currently `Aws::CloudFront::CookieSigner#signed_cookie` requires url argument, but it seems unnecessary to pass url when passing policy. `Aws::CloudFront::Signer#signature` also expects either `policy` or `resource` (and `expires`).

https://github.com/aws/aws-sdk-ruby/blob/a6626a1e2f86c5f7c9cb19650b8df46a7305a6fb/gems/aws-sdk-cloudfront/lib/aws-sdk-cloudfront/signer.rb#L79-L90

https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-setting-signed-cookie-custom-policy.html

### DIFF
--- a/gems/aws-sdk-cloudfront/spec/cookie_signer_spec.rb
+++ b/gems/aws-sdk-cloudfront/spec/cookie_signer_spec.rb
@@ -29,10 +29,7 @@ module Aws
                 }
               ]
           }
-          cookie = signer.signed_cookie(
-            "http://abc.cloudfront.net/images/image.jpg",
-            policy: policy.to_json
-          )
+          cookie = signer.signed_cookie(nil, policy: policy.to_json)
           expect(cookie['CloudFront-Policy']).to eq('eyJTdGF0ZW1lbnQiOlt7IlJlc291cmNlIjoiaW1hZ2VzL2ltYWdlLmpwZyIsIkNvbmRpdGlvbiI6eyJJcEFkZHJlc3MiOnsiQVdTOlNvdXJjZUlwIjoiMTAuNTIuMTc2LjAvMjQifSwiRGF0ZUxlc3NUaGFuIjp7IkFXUzpFcG9jaFRpbWUiOjEzNTcwMzQ0MDB9fX1dfQ__')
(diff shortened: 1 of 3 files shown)


gems/aws-sdk-cloudfront/spec/cookie_signer_spec.rb
```diff
@@ -29,10 +29,7 @@ module CloudFront
                 }
               ]
           }
-          cookie = signer.signed_cookie(
-            "http://abc.cloudfront.net/images/image.jpg",
-            policy: policy.to_json
-          )
+          cookie = signer.signed_cookie(nil, policy: policy.to_json)
```
cjyclaire: could you add a separate test case instead of editing existing one?

knjko: @cjyclaire Thanks for reviewing! Could you check it again?

gems/aws-sdk-cloudfront/spec/cookie_signer_spec.rb
```diff
@@ -37,6 +37,21 @@ module CloudFront
           expect(cookie['CloudFront-Signature']).to eq('n4V7xum3wA-w1PaCMyEMpWVXdfw-Yt8I26RpZJKc~Nk8yQh8LYOxewItGJXFq0BxnKuSEKoEVYVTFEteFAGKXwhkbC7K~JfL83aroPbRagjyG-V9Y5wMLccBAzMj5nHXxjvjlOu541VUR-RlR0KK106HT4-Hp1c~nyOmXs4R5mU_')
           expect(cookie['CloudFront-Key-Pair-Id']).to eq('CF_KEYPAIR_ID')
         end
```
cjyclaire: nit-pick, one empty line between 2 suites?

knjko: @cjyclaire  Thanks! I added it ➕

cjyclaire:
Thanks so much! merged, will be included in the next release ;)

PR #1971, opened 2019-01-31
[BuildTools] Support structure with no members key

cjyclaire:

rebuild all service, no diff


cjyclaire:
Yep, I rebuild all services with `rake build` and no diff is generated

PR #1990, opened 2019-03-08
Use target client in Object#copy_to when possible

Manfred:
When calling `source.copy_to(target)` when source and target are both `Aws::S3::Object`s we need to use the `Client` instance of the target instead of the source to make the request land on the correct endpoint (and thus region).

### DIFF
--- a/gems/aws-sdk-s3/CHANGELOG.md
+++ b/gems/aws-sdk-s3/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Change `Aws::S3::Object#copy_to` to use the target's client so the request is sent to the correct region endpoint.
+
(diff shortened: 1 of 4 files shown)


mullermp:
Hello. Thank you for the pull request (and sorry for the late response, I am new around here). Are you trying to do a cross region copy? You should use copy_from and not copy_to, as it is not designed for cross region copy.

https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb#L30

https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb#L73

I will close this assuming that copy_from will work from you. Please reopen if there is an issue!

PR #1999, opened 2019-03-19
Correct gemspec metadata

ekohl:
This changes Github URLs to use https to avoid a redirect. It also fixes two typos in descriptions.


cjyclaire:
Thanks for the contribution!

PR #2017, opened 2019-04-10
Deduplicate empty strings in seahorse shapes

casperisfine:
While doing some memory profiling, I noticed that `ShapeRef` instances were holding onto a lot of duplicated empty strings:

```
Retained String Report
      1636  ""
       583  [REDACTED]
       401  /tmp/bundle/ruby/2.5.0/gems/aws-sdk-core-3.48.2/lib/seahorse/model/shapes.rb:18
       401  /tmp/bundle/ruby/2.5.0/gems/aws-sdk-core-3.48.2/lib/seahorse/model/shapes.rb:20
       ....
```

By calling `.freeze` on them, we allow MRI to intern them.

### DIFF
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,7 +1,7 @@
 Unreleased Changes
 ------------------
 
-* Small memory retention reduction.
+* Issue - Small memory retention reduction.
 
 3.48.3 (2019-03-26)
 ------------------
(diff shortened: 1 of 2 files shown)


cjyclaire:
Thanks for the contribution! Mind also add a changelog entry [here](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/CHANGELOG.md) for tracking this change at release? : D

casperisfine:
@cjyclaire done!

PR #2028, opened 2019-04-21
Add PriorRequestNotComplete to throttling errors

itkq:
As other sdks like aws-sdk-js and aws-sdk-go handle
PriorRequestNotComplete as the throttling error,
handle it as the throttling error in aws-sdk-ruby as well.


cjyclaire:
Thanks for the contribution! Merging

Issue #2040, opened 2019-05-06
Cannot stub CloudFront responses due to is_truncated error


jeisen:
Please fill out the sections below to help us address your issue

### Issue description
When trying to construct responses for CloudFront's list_* methods that include an `is_truncated` field, the stubbing API rejects any input because of the following error:

`no member 'cloud_front_origin_access_identity_list.is_truncated' in struct`

It appears that the field is being removed in the `remove_paging_tokens` method. Just before it enters that method, the generated struct looks like this:

`#<struct Aws::CloudFront::Types::ListCloudFrontOriginAccessIdentitiesResult cloud_front_origin_access_identity_list=#<struct Aws::CloudFront::Types::CloudFrontOriginAccessIdentityList marker="string", next_marker="string", max_items=0, is_truncated=false, quantity=0, items=[]>>`

The code then removes the entire stubbed result list, resulting in this:

`#<struct Aws::CloudFront::Types::ListCloudFrontOriginAccessIdentitiesResult cloud_front_origin_access_identity_list=nil>`

At that point, the attempt to set the stub's `is_truncated` value to false in the next step fails.

This error occurs on at least the following:
* `list_cloud_front_origin_access_identities`
* `list_distributions`
* `list_streaming_distributions`

### Gem name ('aws-sdk', 'aws-sdk-resources' or service gems like 'aws-sdk-s3') and its version
aws-sdk (3.0.1)
aws-sdk-core (3.50.0)
aws-sdk-cloudfront (1.15.0)

### Version of Ruby, OS environment
Ruby 2.3.8
macOS Mojave (10.14.3 (18D109))

### Code snippets / steps to reproduce
```
client = Aws::CloudFront::Client.new(:stub_responses => true)
access_identities = [{"id" => "foo", "s3_canonical_user_id" => "bar"}]
data = {"cloud_front_origin_access_identity_list" => {"items" => access_identities }}
client.stub_data(:list_cloud_front_origin_access_identities, data)
```

cjyclaire:
Thanks for the report! I can reproduce this locally, will take a look shortly

cjyclaire:
a short summary is current SDK default stubs for nested path `more_results` is not supported, PR opened with the fix is under review,

cjyclaire:
[...]

PR #2043, opened 2019-05-07
Handles paginator stubs with expression

cjyclaire:

for paginators with `token`, it could be a `jmespath` expression to the member, default empty stub for those nested structured paginators shouldn't be breaking.

addressing #2040


cjyclaire:
pending, working on populating default token and more result value for the nesting struct as well

cjyclaire:
pending, s3 has a different patten in "tokens"

cjyclaire:
current change would be introducing breaking changes:
```
Aws> Aws::RDS::Client.new(stub_responses: true).describe_engine_default_parameters(db_parameter_group_family: 'foo')
```
before:
```
=> #<struct Aws::RDS::Types::DescribeEngineDefaultParametersResult engine_defaults=nil>
```
after
```
=> #<struct Aws::RDS::Types::DescribeEngineDefaultParametersResult
 engine_defaults=
  #<struct Aws::RDS::Types::EngineDefaults db_parameter_group_family="String", marker="String", parameters=[]>>
```

though we are fixing the right thing there

cjyclaire:
The s3 case has been fixed, ready for review

Issue #2046, opened 2019-05-15
cannot load such file -- aws-sdk-core/plugins/transfer_encoding.rb


raykov:
### Issue description
Wrong dependency in `aws-sdk-sqs.gemspec`. 

```bash
cannot load such file -- aws-sdk-core/plugins/transfer_encoding.rb
```

Guys, you recently updates `aws-sdk-core` and `aws-sdk-sqs`

* You have added a new plugin to `sdk-core` [transfer_encoding](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/lib/aws-sdk-core/plugins/transfer_encoding.rb)
* And used this plugin in new version of `sdk-sqs` [require 'aws-sdk-core/plugins/transfer_encoding.rb'](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-sqs/lib/aws-sdk-sqs/client.rb#L26)
* **BUT** you left old dependency in [aws-sdk-sqs.gemspec](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-sqs/aws-sdk-sqs.gemspec#L26) to `'aws-sdk-core', '~> 3', '>= 3.48.2'` (**This plugin was just added and doesn't exist in old versions**)
* **NOW** it should be `'aws-sdk-core', '~> 3', '>= 3.52.0'`
* And if someone updates to new version of sdk-sqs (with sdk-core locked to v3.48.2) -> he has troubles. 

### Gem name 

`aws-sdk-sqs` v1.14.0
probably other gems which are using `transfer_encoding` plugin from `sdk-core`

cjyclaire:
Thanks for the contribution! PR has been merged will be out in the next release

scottsuch:
Did this go out with the `v2.11.275` release?

cjyclaire:
@scottsuch This should be affecting v3 only, not for v2.

Today's v3 release is out, all latest versions of service gems right now are with the change.
closing, feel free to reopen if you see further issues, thanks!

gauravonekingslane:
hi 
I am getting the error - `cannot load such file ruby/2.3.0/gems/aws-sdk-core-3.54.2/lib/aws-sdk-core/log/formatter (LoadError)` when updating `aws-sdk` to version '~> 3'.
[...]

PR #2047, opened 2019-05-15
Update MINIMUM_CORE_VERSION to 3.52.0

raykov:
**Description**
In recent update a new plugin transfer_encoding was added to sdk-core.
This plugin is in use in many gems which are using sdk-core.
But minimum version of sdk core was stayed on old version 3.48.2.
Since we are introduced a new pluging and used it everywhere we have to
bump minimal version to 3.52.0. Because new plugin was introduced in
this version.

**Changes**
 * Update MINIMUM_CORE_VERSION to version "3.52.0" 
 * Regenerate gems

**Related issue**
 * https://github.com/aws/aws-sdk-ruby/issues/2046

### DIFF
--- a/build_tools/services.rb
+++ b/build_tools/services.rb
@@ -8,7 +8,7 @@ module BuildTools
     MANIFEST_PATH = File.expand_path('../../services.json', __FILE__)
 
     # Minimum `aws-sdk-core` version for new gem builds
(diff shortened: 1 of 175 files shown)


cjyclaire:
Thanks for the fix! looks good to me, merging

Issue #1831, opened 2018-07-11
Parsing exception raised when ServiceUnavailableException returned


adammw:
### Issue description
Parsing exception raised when ServiceUnavailableException returned

### Gem versions
```
    aws-eventstream (1.0.1)
    aws-partitions (1.94.0)
    aws-sdk-core (3.22.0)
      aws-eventstream (~> 1.0)
      aws-partitions (~> 1.0)
      aws-sigv4 (~> 1.0)
      jmespath (~> 1.0)
    aws-sdk-firehose (1.4.0)
      aws-sdk-core (~> 3)
      aws-sigv4 (~> 1.0)
    aws-sdk-kinesis (1.3.0)
      aws-sdk-core (~> 3)
      aws-sigv4 (~> 1.0)
    aws-sigv4 (1.0.2)
```

### Version of Ruby, OS environment
Ruby MRI 2.3.3
Debian Linux

### Code snippets / steps to reproduce
Calling PutRecords on a Kinesis client occasionally fails due to AWS API returning `<ServiceUnavailableException/>` which the gem fails to parse as it attempts to decode the exception as JSON.

Backtrace:
```
EncodingError: Empty input at line 1, column 1 [parse.c:928] in '<ServiceUnavailableException/>
/fluentd/vendor/bundle/ruby/2.3.0/gems/aws-sdk-core-3.22.0/lib/aws-sdk-core/json.rb:25:in `load'
/fluentd/vendor/bundle/ruby/2.3.0/gems/aws-sdk-core-3.22.0/lib/aws-sdk-core/json.rb:25:in `load'
/fluentd/vendor/bundle/ruby/2.3.0/gems/aws-sdk-core-3.22.0/lib/aws-sdk-core/json/error_handler.rb:17:in `extract_error'
/fluentd/vendor/bundle/ruby/2.3.0/gems/aws-sdk-core-3.22.0/lib/aws-sdk-core/xml/error_handler.rb:22:in `error'
/fluentd/vendor/bundle/ruby/2.3.0/gems/aws-sdk-core-3.22.0/lib/aws-sdk-core/json/error_handler.rb:9:in `block in call'
/fluentd/vendor/bundle/ruby/2.3.0/gems/aws-sdk-core-3.22.0/lib/seahorse/client/response.rb:45:in `block in on'
[...]

PR #2085, opened 2019-07-16
Handle EncodingError coming from oj gem

stiak:
Also adds some test coverage for the `Aws::Json` module.

Fixes: https://github.com/aws/aws-sdk-ruby/issues/1831

For context, I'm using the `aws-sdk-kinesis` library which sometimes receives `<ServiceUnavailableException/>` when writing records.  The SDK seems to translate these into failure responses when I *don't* have the `oj` gem installed, but with it present the `EncodingError` bubbles all the way out of the SDK.

### DIFF
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Handle `EncodingError` when using Oj gem [Github Issue](https://github.com/aws/aws-sdk-ruby/issues/1831)
+
 3.59.0 (2019-07-03)
 ------------------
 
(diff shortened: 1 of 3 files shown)


cjyclaire:
Thanks for the contribution! Changes looks good to me, could you also update test to cover the "EncodingError" case if possible?

stiak:
> Thanks for the contribution! Changes looks good to me, could you also update test to cover the "EncodingError" case if possible?

The `context 'invalid JSON'` covers the "EncodingError" case.  Without the changes EncodingError is raised.  Should I update the context description?

stiak:
I've added another test case so there is one for each exception type that Oj raises.

stiak:
@cjyclaire Any other changes I should make on this one?

cjyclaire:
Merged! Thanks for the contribution, will be included in the next release : )

PR #2088, opened 2019-07-18
Allow running state while waiting for getting replication task stopped

mozamimy:
The current implementation of DMS ReplicationTaskStopped waiter handles 'running' state
as a failure while waiting for getting target replication task stopped.

However, I guess that following state transition and use case is natural.

```
[stopped]
   ↓ Use StartReplicationTask API
[starting]
   ↓ Wait with ReplicationTaskRunning waiter.
[running]
   ↓ Wait with ReplicationTaskStopped waiter.
[stopping]
```

This change allows the above use case of ReplicationTaskStopped waiter.

### DIFF
--- a/apis/dms/2016-01-01/waiters-2.json
+++ b/apis/dms/2016-01-01/waiters-2.json
@@ -193,12 +193,6 @@
                 },
                 {
                     "argument":"ReplicationTasks[].Status",
-                    "expected":"running",
-                    "matcher":"pathAny",
-                    "state":"failure"
-                },
-                {
-                    "argument":"ReplicationTasks[].Status",
                     "expected":"failed",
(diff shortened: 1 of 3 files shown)


mullermp:
Hi mozamimy, unfortunately we can't just merge in this change because these files are generated and will need to be changed upstream. Because this is an API change, I would have to review this API internally with the database migration service team to see if this is an appropriate change to their service. We need to do this because it would affect all of the AWS SDKs and not just Ruby. I will start that process and try to keep you updated.

mullermp:
@mozamimy could you also provide any relative code or context on how you are trying to use the waiters?

mullermp:
I am closing this as the DMS team did not want this change.

PR #2115, opened 2019-09-04
Fix default retries documentation

mullermp:
Fixes #1773

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
@@ -51,8 +51,9 @@ module Aws
       super
     end
 
-    # @return [Integer] The number of times to retry failed attempts to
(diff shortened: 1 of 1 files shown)


gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
```diff
@@ -51,8 +51,9 @@ def initialize options = {}
       super
     end
 
-    # @return [Integer] The number of times to retry failed attempts to
-    #   fetch credentials from the instance metadata service. Defaults to 0.
+    # @return [Integer] Number of times to retry when retrieving credentials
+    #   from the instance metadata service. Defaults to 0 when initialized from
+    #   the {Aws::CredentialProviderChain} class.
```
cjyclaire: Probably we can directly point out this is from: when resolving from the default credential chain.
There won't be much scenarios when developers are doing the initialization on their own or they are aware of there is a module for that.

PR #2130, opened 2019-10-01
Deprecate credential methods

mullermp:
This PR does two big things.

1) Start the deprecation process of credential accessor methods on the credential provider.

2) Adds stronger printing for deprecations and includes a version printed to the console that we intend to make minor backwards incompatible changes.

As per @cjyclaire's recommendation, we can do this in two stages. The first being this PR to add the warnings and the second to remove the methods.

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core/deprecations.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/deprecations.rb
@@ -35,33 +35,39 @@ module Aws
   # @api private
   module Deprecations
 
-    # @param [Symbol] method_name The name of the deprecated method.
+    # @param [Symbol] method The name of the deprecated method.
     #
     # @option options [String] :message The warning message to issue
     #   when the deprecated method is called.
     #
-    # @option options [Symbol] :use The name of an use
-    #   method that should be used.
+    # @option options [String] :use The name of a method that should be used.
     #
-    def deprecated(method_name, options = {})
+    # @option options [String] :version The version that will remove the
+    #   deprecated method.
+    #
+    def deprecated(method, options = {})
 
       deprecation_msg = options[:message] || begin
-        msg = "DEPRECATION WARNING: called deprecated method `#{method_name}' "
-        msg << "of an #{self}"
-        msg << ", use #{options[:use]} instead" if options[:use]
+        msg = "#################### DEPRECATION WARNING ####################\n"
+        msg << "Called deprecated method `#{method}` of #{self}."
+        msg << " Use `#{options[:use]}` instead.\n" if options[:use]
+        if options[:version]
+          msg << "Method `#{method}` will be removed in #{options[:version]}."
+        end
+        msg << "\n#############################################################"
         msg
       end
 
-      alias_method(:"deprecated_#{method_name}", method_name)
+      alias_method(:"deprecated_#{method}", method)
 
       warned = false # we only want to issue this warning once
 
-      define_method(method_name) do |*args,&block|
+      define_method(method) do |*args, &block|
         unless warned
           warned = true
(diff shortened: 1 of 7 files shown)


gems/aws-sdk-dynamodb/lib/aws-sdk-dynamodb/client.rb
```diff
@@ -190,7 +190,7 @@ class Client < Seahorse::Client::Base
     #
     #     @see https://www.awsarchitectureblog.com/2015/03/backoff.html
     #
-    #   @option options [Integer] :retry_limit (3)
+    #   @option options [Integer] :retry_limit (10)
```
cjyclaire: this file should not be included in the diff?

gems/aws-sigv4/lib/aws-sigv4/signer.rb
```diff
@@ -202,7 +202,7 @@ def initialize(options = {})
       #
```
cjyclaire: would need to add a changelog for sigv4 gem as well

gems/aws-sdk-core/lib/aws-sdk-core/deprecations.rb
```diff
@@ -35,33 +35,39 @@ module Aws
...
-    #   method that should be used.
+    # @option options [String] :use The name of a method that should be used.
     #
-    def deprecated(method_name, options = {})
+    # @option options [String] :version The version that will remove the
+    #   deprecated method.
+    #
+    def deprecated(method, options = {})
 
       deprecation_msg = options[:message] || begin
```
cjyclaire: Can we have this PR in 2 steps?
1. add stronger deprecation waring for those methods with a version number/deadline when they will be removed
2. perform removal after reaching the deadline

The `:message` parameter would be able to help provide additional messagings

gems/aws-sigv4/lib/aws-sigv4/signer.rb
mullermp: The signer changes do not have any new functionality or fixes. The slight refactor is just included. Next version bump of that gem will include that refactor.

gems/aws-sdk-dynamodb/lib/aws-sdk-dynamodb/client.rb
mullermp: Discussed offline.

gems/aws-sdk-core/lib/aws-sdk-core/deprecations.rb
mullermp: I agree. I arbitrarily chose 3.75 as the version. I will update the PR.

gems/aws-sigv4/lib/aws-sigv4/signer.rb
```diff
@@ -656,20 +656,16 @@ def uri_escape_path(string)
         self.class.uri_escape_path(string)
       end
 
-      def get_credentials
+      def fetch_credentials
         credentials = @credentials_provider.credentials
-        if credentials_set?(credentials)
+        if credentials.access_key_id && credentials.secret_access_key
```
mgwidmann: I'm finding it impossible with the previous version as well as this version to not produce the deprecation warning. Is this intentional? It seems as best I can tell this code is simultaneously deprecating `#access_key_id` and `#secret_access_key` as well as still using the methods internally.

mullermp: In section of code, I removed the credentials_set? method as it was only called once in this location. There are no deprecations going on in the sigv4 signer - this was a small refactor that I included into the change.

We are only deprecating those methods for credential provider objects, not the provider itself. On line 660, we're fetching credentials via the provider and then calling those methods on credentials.

PR #2132, opened 2019-10-04
Remove deprecated credential methods from credential providers

mullermp:
This change removes deprecated methods. We will merge this in as core version 3.75.


edma2:
Given that this change is backwards incompatible, wasn't it supposed to be released with a major version bump? According to the [documentation](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/index.html#Versioning): 

> You can safely express a dependency on a major version and expect all minor and patch versions to be backwards compatible.

mullermp:
@edma2 For the [change prior to this](https://github.com/aws/aws-sdk-ruby/pull/2130/files), the documentation indicated that they were to be removed in version 2 but never were. Even with that miss, it should have been removed in version 3 (prior to my time).

```
    # @deprecated Deprecated in 2.1.0. This method is subject to errors
    #   from a race condition when called against refreshable credential
    #   objects. Will be removed in 2.2.0.
```

We did so because they were buggy to begin with, and that they had been marked as deprecated for such a long time. This was communicated not just in the comments but as well as log statements. I'm sorry for this inconvenience, but you can fix it just by calling the `credentials` method on the provider.

Issue #2120, opened 2019-09-14
S3::Object::upload_file does not allow thread_count configuration 


gregpardo:
Please fill out the sections below to help us address your issue

### Issue description
Our system was configured to use 5 threads and was hanging on the default configuration of multipart uploads. The docs do not state that you can pass a thread_count option to the Object::upload_file method. Even after you do pass you will get a params validation error unless your Client has 'params_validation: false' passed to it.  Object::download_file does seem to support thread_count option without any issues. Seems like this would be a common thing you may need to configure for upload_file as well. 

### Gem name ('aws-sdk', 'aws-sdk-resources' or service gems like 'aws-sdk-s3') and its version
aws-sdk-s3    1.48.0

### Version of Ruby, OS environment
2.6.4  on Alpine Linux 3.10
Puma running Rails on 5 threads

### Code snippets / steps to reproduce

Causes error when running 5 threads
```ruby
client = Aws::S3::Client.new
bucket = Aws::S3::Resource.new(client: client).bucket(some_bucket)
obj = bucket.object(some_object)
obj.upload_file(file_path)
```

Causes error when passing thread_count
```ruby
client = Aws::S3::Client.new
bucket = Aws::S3::Resource.new(client: client).bucket(some_bucket)
obj = bucket.object(some_object)
obj.upload_file(file_path, thread_count: 5)
```

Does not cause error
```ruby
client = Aws::S3::Client.new(validate_params: false)
bucket = Aws::S3::Resource.new(client: client).bucket(some_bucket)
obj = bucket.object(some_object)
obj.upload_file(file_path, thread_count: 5)
```

cjyclaire:
We do support `:thread_count` for `upload_file`,  we should make our parameter doc clearer on that, if your file is less than 15M (default), it won't trigger multipart uploader, this could be why you don't see the `:thread_count` working. This threshold value is also configurable by `: multipart_threshold `.

PR #2133, opened 2019-10-04
Fix up documentation around customized S3 object

### LINKED ISSUE #1967: Passing in a closed `Tempfile` to `Aws::S3::FileUploader#upload` raises
### Issue description

`Aws::S3::FileUploader#upload` does not properly handle closed a closed `Tempfile`. I triggered this issue by passing a closed `Tempfile` to `Aws::S3::Object#upload_file`. Read the section below for more details.

aws-sdk-s3 1.30.1

Ubuntu Linux, ruby 2.4.5p335

The issue is a bit tricky and may actually be a bug in the Ruby VM itself, but should be relatively easy to work around in the S3 gem.

Before diving into the S3 gem itself, please note this behavior from Ruby:

```
[25] pry(main)> n=Tempfile.new('hi'); n.write('hi'); n.close; n.size
=> 2
[26] pry(main)> File.size(n)
IOError: closed stream
from (pry):26:in `size'
```

In other words: `Tempfile#size` works fine, but `File.size` raises. Why this happens is a bit tricky and may be a bug in the Ruby VM.

The issue is that internally in the Ruby VM a call to `File.size` will check if the file is open and raise if it is not. 

See the following pieces of code:
1.  First, the call to `File.size` (which calls `rb_stat`): https://github.com/ruby/ruby/blob/ruby_2_4/file.c#L1953-L1973
2.  Next, the call to `rb_stat` (which calls `GetOpenFile`): https://github.com/ruby/ruby/blob/ruby_2_4/file.c#L1007-L1022
3.  Next, the call to `GetOpenFile` (which calls `rb_io_check_closed`): https://github.com/ruby/ruby/blob/ruby_2_4/include/ruby/io.h#L120
4.  Finally the code in `rb_io_check_closed` which triggers the raise: https://github.com/ruby/ruby/blob/ruby_2_4/io.c#L638-L645

In contrast, the `Tempfile#size` method itself will first check if the file is open or closed and avoid the raise in `File.size` by passing in the `Tempfile#path` to `File.size` instead. See this code from `Tempfile#size`: https://github.com/ruby/ruby/blob/ruby_2_4/lib/tempfile.rb#L224-L232

Compare that code to the code in `Aws::S3::FileUploader`: https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-s3/lib/aws-sdk-s3/file_uploader.rb#L26-L36

`Aws::S3::FileUploader#upload` triggers the raise because it does not do the same check that `Tempfile#size` is doing.

My suggestion on how to resolve this would be one of the following:
1.  If this behavior is desired, update the comment to indicate that only an open `Tempfile` can be passed to `upload_file`, or
2.  Update the code so that `Tempfile#size` is called in `Aws::S3::FileUploader#upload` which avoids this issue entirely

mullermp:
This change updates documentation around Object#upload_file, as well as some general code clean-up via Rubocop and pending tests.

I bounced back and forth on adding a check for Tempfiles to determine its size, but decided that the doc updates made more sense. A change would allow closed Tempfiles to be uploaded but not closed Files, and that was inconsistent.

Closes #1967, #2120

### DIFF
--- a/gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
+++ b/gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
@@ -1,14 +1,14 @@
 module Aws
   module S3
     class Object
-
       alias size content_length
 
       # Copies another object to this object. Use `multipart_copy: true`
       # for large objects. This is required for objects that exceed 5GB.
       #
-      # @param [S3::Object, S3::ObjectVersion, S3::ObjectSummary, String, Hash] source
-      #   Where to copy object data from. `source` must be one of the following:
+      # @param [S3::Object, S3::ObjectVersion, S3::ObjectSummary, String, Hash]
+      #   source Where to copy object data from. `source` must be one of the
+      #   following:
       #
       #   * {Aws::S3::Object}
       #   * {Aws::S3::ObjectSummary}
@@ -103,8 +103,9 @@ module Aws
         ObjectCopier.new(self, options).copy_to(target, options)
       end
 
-      # Copies and deletes the current object. The object will only be
-      # deleted if the copy operation succeeds.
+      # Copies and deletes the current object. The object will only be deleted
+      # if the copy operation succeeds.
+      #
       # @param (see Object#copy_to)
       # @option (see Object#copy_to)
       # @return [void]
@@ -129,10 +130,7 @@ module Aws
           client.config.credentials,
           client.config.region,
           bucket_name,
-          {
-            key: key,
-            url: bucket.url,
-          }.merge(options)
+          { key: key, url: bucket.url }.merge(options)
         )
       end
 
@@ -186,10 +184,10 @@ module Aws
       #
       def presigned_url(http_method, params = {})
         presigner = Presigner.new(client: client)
-        presigner.presigned_url("#{http_method.downcase}_object", params.merge(
-          bucket: bucket_name,
-          key: key,
-        ))
+        presigner.presigned_url(
+          "#{http_method.downcase}_object",
+          params.merge(bucket: bucket_name, key: key)
+        )
       end
 
       # Returns the public (un-signed) URL for this object.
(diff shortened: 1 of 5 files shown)


gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
```diff
@@ -320,29 +328,31 @@ def upload_file(source, options = {})
...
       #  `get_range` mode allows `chunk_size` parameter to configured in
       #  customizing each range size in multipart_download,
       #  By default, `auto` mode is enabled, which performs multipart_download
       #
-      # @option options [String] chunk_size required in get_range mode
+      # @option options [String] chunk_size required in get_range mode.
       #
-      # @option options [String] thread_count Customize threads used in multipart
-      #   download, if not provided, 10 is default value
+      # @option options [String] thread_count (10) Customize threads used in the
```
cjyclaire: still have string here?

gems/aws-sdk-s3/spec/object_spec.rb
```diff
@@ -3,6 +3,13 @@
 module Aws
   module S3
     class Object
+
+      pending '#copy_from'
```
cjyclaire: pending?

gems/aws-sdk-s3/spec/object_spec.rb
```diff
@@ -14,6 +21,17 @@ class Object
         end
 
       end
+
```
cjyclaire: also here?

gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
mullermp: Yeah - I changed it (locally) as well as some other stuff. I will update soon.

cjyclaire:
Make sure fix CI before merging

Issue #2114, opened 2019-09-03
Custom endpoint for SQS (VPC Endpoint) does not work.


hhk1989:
### Issue description

In my account, I have a VPC Interface Endpoint for SQS and a lambda function in the same VPC. The lambda job (`Ruby 2.5` runtime) tries to send an SQS message to an SQS queue that the lambda execution role has permissions to, by using the VPC Endpoint. For this, I do something like the following

```
sqs_client = Aws::SQS::Client.new(credentials: ####, endpoint: 'https://vpce-<id>.sqs.us-west-2.amazonaws.com')
sqs_client.send_message(queue_url: 'https://sqs.us-west-2.amazonaws.com/<account_id>/test-queue', message_body: 'test')
```

I notice that the lambda always times out at the `send_message` call.

I notice that I am able to make other calls like STS (for `assume_role`) and SNS (for `publish`), from the lambda using their respective VPC Endpoints for STS and SNS. The VPC Endpoints are all configured the same (security groups, etc.) using terraform code.

I also noticed that I am able to make the `send_message` call from an AWS SDK CLI, in an EC2 instance in the same VPC. So, the following does work.

`aws sqs send-message --queue-url 'https://sqs.us-west-2.amazonaws.com/<account_id>/test-queue ' --endpoint-url 'https://vpce-<id>.sqs.us-west-2.vpce.amazonaws.com ' --message-body "this works"`

This makes me believe that setting the custom endpoint for SQS is what is not working here. 

### Gem name ('aws-sdk', 'aws-sdk-resources' or service gems like 'aws-sdk-s3') and its version

`aws-sdk-sqs 1.11.0`

### Version of Ruby, OS environment

`Ruby 2.5` lambda runtime.
`Amazon Linux` (https://docs.aws.amazon.com/lambda/latest/dg/lambda-ruby.html)

cjyclaire:
Thanks for the information!

Some Quick follow-up questions that might help debugging

1. could you try set `:http_wire_trace` to true at the client so http wire log information can be available in lambda logs?
2. using cli is outside from lambda environment, are you also seeing the timeout if making the call with Ruby SDK outside from lambda(such as from a local script?), this can help decoupling if it has to do with lambda environment configuration.

hhk1989:
Hello, thanks for your response.

1. I set `:http_wire_trace` as indicated and got the following response.

```
sqs_client = Aws::SQS::Client.new(
        credentials: Aws::Credentials.new(
[...]

PR #2148, opened 2019-10-29
Support QueueUrl Plugin override

cjyclaire:

Addressing #2114 

```
Aws::SQS::Client.new(disable_queue_url_override: true)
```
will disable endpoint and region override from the `:queue _url` provided

Note:

Pending on sync with Java team

### DIFF
--- a/gems/aws-sdk-sqs/lib/aws-sdk-sqs/plugins/queue_urls.rb
+++ b/gems/aws-sdk-sqs/lib/aws-sdk-sqs/plugins/queue_urls.rb
@@ -4,10 +4,20 @@ module Aws
       # @api private
       class QueueUrls < Seahorse::Client::Plugin
 
+        option(:disable_queue_url_override,
+              default: false,
+              doc_type: 'Boolean',
+              docstring: <<-DOCS
+Set to `true` to disable region and endpoint override from
+`:queue_url` value, default to `false`
+              DOCS
+        )
+
         class Handler < Seahorse::Client::Handler
 
           def call(context)
-            if queue_url = context.params[:queue_url]
+            disable = !!context.config.disable_queue_url_override
+            if !disable && (queue_url = context.params[:queue_url])
               update_endpoint(context, queue_url)
               update_region(context, queue_url)
             end
--- a/gems/aws-sdk-sqs/spec/client/queue_urls_spec.rb
+++ b/gems/aws-sdk-sqs/spec/client/queue_urls_spec.rb
@@ -59,6 +59,16 @@ module Aws
               client.send(method, params.merge(queue_url: url))
             }.to_not raise_error
           end
+
+          it 'does not override endpoint and region when disabled' do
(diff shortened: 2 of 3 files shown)


cjyclaire:
CI passing the 2.1 failing is transient issue as it's succeeding in the other run

hhk1989:
I wonder if we could do something like this, rather than having to disable the plugin. That would make it more default to use custom endpoints, rather than having to remember to pass in `disable_queue_url_override` as well.

In the queue_url plugin, we check if an endpoint is provided by the end user. If not, then go ahead and update the endpoint and region based on the queue_url. If yes, then do not make those changes.

cjyclaire:
@hhk1989 Thanks for the feedback! Appreciate that, personally thought about that too, was hesitate to do so because it could potentially breaking people providing a customized endpoint but depending on current overwrite, it's hard to tell who might be replying on this behavior. Adding an option would make it safer for people who are certain that they want to enable this behavior.

In the next major version (where breaking changes are allowed), we probably should consider making this override turned off by default when customized endpoint is provided. Thoughts?

hhk1989:
That's a fair point, it's prudent to not break existing workflows. However, could we also keep this consistent for the Java SDK ? As in for now, we introduce the `disable_queue_url_override` option, but in the next major version we  turn it off by default when custom endpoint is provided.

cjyclaire:
@hhk1989 Sounds good! I will sync with Java team to make sure we are on the same page

gems/aws-sdk-sqs/lib/aws-sdk-sqs/plugins/queue_urls.rb
```diff
@@ -4,10 +4,20 @@ module Plugins
...
+Set to `true` to disable region and endpoint override from
+`:queue_url` value, default to `false`
+              DOCS
+        )
+
         class Handler < Seahorse::Client::Handler
 
           def call(context)
-            if queue_url = context.params[:queue_url]
+            disable = !!context.config.disable_queue_url_override
```
mullermp: Avoid use of double negation (as per Rubocop). Use .nil?

gems/aws-sdk-sqs/spec/client/queue_urls_spec.rb
```diff
@@ -59,6 +59,16 @@ module SQS
               client.send(method, params.merge(queue_url: url))
             }.to_not raise_error
           end
+
+          it 'does not override endpoint and region when disabled' do
+            endpoint = 'https://vpc-123234.sqs.us-west-2.vpc.amazonaws.com'
+            url = 'https://sqs.us-east-1.amazonaws.com/1234567890/demo'
+            client = Client.new(stub_responses: true, endpoint: endpoint, disable_queue_url_override: true)
```
mullermp: Check for 80 chars!

cjyclaire:
closing, will create a separate PR for tracking as the issue is queue url plugin is parsing out the wrong region from vpc endpoint queue

PR #2149, opened 2019-10-30
Ensure locking around `@entries`. Fixes #2142.

### LINKED ISSUE #2142: thread safety issues in endpoint_cache.rb
```ruby
    # @param [String] key
    # @param [Hash] value
    def []=(key, value)
      @mutex.synchronize do
        # delete the least recent used endpoint when cache is full
        unless @entries.size < @max_entries
          old_key, _ = @entries.shift
          self.delete_polling_thread(old_key)
        end
        # delete old value if exists
        @entries.delete(key)
        @entries[key] = Endpoint.new(value.to_hash)
      end
    end

    # checking whether an unexpired endpoint key exists in cache
    # @param [String] key
    # @return [Boolean]
    def key?(key)
      if @entries.key?(key) && (@entries[key].nil? || @entries[key].expired?)
        self.delete(key)
      end
      @entries.key?(key)
    end
```

This does not look thread safe to me. If multiple threads are accessing @entries, `#key?` can be called without locking around `@entries`. I don't believe those are thread-safe operations.

ioquatix:

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core/endpoint_cache.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/endpoint_cache.rb
@@ -60,10 +60,12 @@ module Aws
     # @param [String] key
     # @return [Boolean]
     def key?(key)
-      if @entries.key?(key) && (@entries[key].nil? || @entries[key].expired?)
-        self.delete(key)
+      @mutex.synchronize do
+        if @entries.key?(key) && (@entries[key].nil? || @entries[key].expired?)
+          @entries.delete(key)
+        end
+        @entries.key?(key)
       end
-      @entries.key?(key)
     end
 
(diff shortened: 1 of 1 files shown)


mullermp:
Thanks for the contribution!

ioquatix:
I am happy to provide feedback on other thread related issues but perhaps AWS can consider sponsoring my open source work so I can dedicate more time to it. Let me know if this is something worth exploring.

mullermp:
I may take you up on that if we ever run into any thread safe issues, and thanks for bringing this to our attention.

Regarding sponsorships, I've never looked into it, I'll ask around internally.

ioquatix:
During my review I found other potentially unsafe code. Sponsorship or consulting are two options available if you want more of my time. Thanks for investigating it internally.

PR #2158, opened 2019-11-08
Support S3 IAD regional option

cjyclaire:

Pending:
* changelog (partition, s3, core)
* ~~s3 re-build diff~~
* rebase from master
* core/s3 min requirement on partition gem version
* ~~dualstack regional behavior discussion~~

Ready for review


gems/aws-partitions/lib/aws-partitions/endpoint_provider.rb
```diff
@@ -38,9 +38,22 @@ def initialize(rules)
...
+      #   `legacy` (global endpoint) or `regional` mode for using regional
+      #   endpoint for S3 'us-east-1' region
       # @api private Use the static class methods instead.
-      def resolve(region, service, sts_regional_endpoints = nil)
-        "https://" + endpoint_for(region, service, sts_regional_endpoints)
+      def resolve(
+        region,
+        service,
+        sts_regional_endpoints = 'legacy',
+        s3_us_east_1_regional_endpoint = 'legacy'
```
mullermp: I see that you set defaults for both this, a different resolve method, and endpoint_for. Could we centralize them somewhere? If the default are ever changed, it must be done in multiple places.

gems/aws-partitions/lib/aws-partitions/endpoint_provider.rb
```diff
@@ -80,9 +98,16 @@ def endpoint_for(region, service, sts_regional_endpoints)
           region = service_cfg.fetch("partitionEndpoint", region)
         end
 
+        # Check for s3 IAD regional behavior
+        s3_iad_regional = service == 's3' &&
+          region == 'us-east-1' &&
+          s3_us_east_1_regional_endpoint == 'regional'
+
```
mullermp: Is this variable necessary? In either case, wrap the boolean expression in parenthesis before assigning or checking `unless`.

gems/aws-sdk-core/spec/aws/shared_config_spec.rb
```diff
@@ -146,5 +146,25 @@ module Aws
...
 
+    context 's3_us_east_1_regional_endpoint selection' do
+
+      it 'can resolve s3_us_east_1_regional_endpoint from config file' do
+        config = SharedConfig.new(
+          config_path: mock_config_file,
+          config_enabled: true,
+          profile_name: "s3_iad_regional"
+        )
+        expect(config.s3_us_east_1_regional_endpoint).to eq('regional')
```
mullermp: What happens with 'regional' or 'legacy' is not provided?

gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/iad_regional_endpoint.rb
```diff
@@ -0,0 +1,31 @@
...
+region, defaults to `legacy` mode, using global endpoint.
+          DOCS
+          resolve_iad_regional_endpoint(cfg)
+        end
+
+        private
+
+        def self.resolve_iad_regional_endpoint(cfg)
+          env_mode = ENV['AWS_S3_US_EAST_1_REGIONAL_ENDPOINT']
+          env_mode = nil if env_mode == ''
```
mullermp: Why is setting to nil necessary in this case? What about the following?

```
mode = ENV['AWS_S3_US_EAST_1_REGIONAL_ENDPOINT'] ||
  Aws.shared_config.s3_us_east_1_regional_endpoint(profile: cfg.profile) ||
  'legacy'

# should we raise an error if it's neither legacy nor regional?

mode == 'regional'
```

gems/aws-sdk-core/spec/aws/shared_config_spec.rb
cjyclaire: Confirmed with Jason offline, will error out, will update

gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/iad_regional_endpoint.rb
cjyclaire: Confirmed with Jason offline, will error out, will update

cjyclaire:
This PR is currently pending (not mergeable yet):
 * final logic ordering check with `partitions.json`
* changelog entry in `aws-partition`, `aws-sdk-core` and `aws-sdk-s3` gem
* [Minimum version requirement](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/aws-sdk-core.gemspec#L15) of partition gem in `aws-sdk-core`
* [Minimum version requirement](https://github.com/aws/aws-sdk-ruby/blob/master/build_tools/services.rb#L10) of core gem in `aws-sdk-s3`

cjyclaire:
ready for review again :) @mullermp

gems/aws-sdk-s3/spec/client/iad_regional_endpoint_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+            stub_responses: true,
+            region: 'us-east-1'
+          )
+          expect(client.config.s3_us_east_1_regional_endpoint).to eq('regional')
+          resp = client.list_buckets
+          expect(resp.context.http_request.endpoint.to_s).to eq(
+            'https://s3.us-east-1.amazonaws.com/')
+        end
+
+        it 'uses client option before ENV' do
```
mullermp: Probably redundant given that is expected behavior of the plugin stack. Fine to keep!

PR #2175, opened 2019-11-22
Fix CI (jruby/rake test)

mullermp:


Gemfile
```diff
@@ -57,12 +52,17 @@ group :test do
...
       gem 'oj', '1.3.0'
     else
-      gem 'oj' unless RUBY_VERSION == '1.9.3'
+      if RUBY_VERSION == '1.9.3'
+        # oj drop support for Ruby under 2.0 since 3.3.5
+        gem 'oj', '<= 3.3.4'
+      else
+        gem 'oj'
+      end
     end
```
mullermp: These are grouped under :test dependencies.. but these should be labeled as core shouldn't they?

Issue #1778, opened 2018-05-07
Lots of warnings generated by this gem


ioquatix:
When running Ruby with warnings enabled, the `aws-sdk-core` gem generates a lot of warnings. These warnings should be fixed and may represent bugs or unexpected collisions. You should add `--warnings` to your `.rspec` and ensure that no warnings are generated during testing.

```
/home/samuel/.rvm/gems/ruby-2.5.0/gems/aws-sdk-core-3.20.2/lib/seahorse/client/plugin.rb:125: warning: method redefined; discarding old doc_default
/home/samuel/.rvm/gems/ruby-2.5.0/gems/aws-sdk-core-3.20.2/lib/aws-sdk-sts/client_api.rb:55: warning: assigned but unused variable - expiredIdentityTokenMessage
/home/samuel/.rvm/gems/ruby-2.5.0/gems/aws-sdk-core-3.20.2/lib/aws-sdk-sts/client_api.rb:58: warning: assigned but unused variable - idpCommunicationErrorMessage
/home/samuel/.rvm/gems/ruby-2.5.0/gems/aws-sdk-core-3.20.2/lib/aws-sdk-sts/client_api.rb:59: warning: assigned but unused variable - idpRejectedClaimMessage
/home/samuel/.rvm/gems/ruby-2.5.0/gems/aws-sdk-core-3.20.2/lib/aws-sdk-sts/client_api.rb:60: warning: assigned but unused variable - invalidAuthorizationMessage
/home/samuel/.rvm/gems/ruby-2.5.0/gems/aws-sdk-core-3.20.2/lib/aws-sdk-sts/client_api.rb:61: warning: assigned but unused variable - invalidIdentityTokenMessage
/home/samuel/.rvm/gems/ruby-2.5.0/gems/aws-sdk-core-3.20.2/lib/aws-sdk-sts/client_api.rb:62: warning: assigned but unused variable - malformedPolicyDocumentMessage
/home/samuel/.rvm/gems/ruby-2.5.0/gems/aws-sdk-core-3.20.2/lib/aws-sdk-sts/client_api.rb:64: warning: assigned but unused variable - packedPolicyTooLargeMessage
/home/samuel/.rvm/gems/ruby-2.5.0/gems/aws-sdk-core-3.20.2/lib/aws-sdk-sts/client_api.rb:65: warning: assigned but unused variable - regionDisabledMessage
[...]

PR #2179, opened 2019-11-28
Add `.rspec` with `--warnings`.

ioquatix:

### DIFF
--- a/gems/aws-sdk-autoscaling/lib/aws-sdk-autoscaling/activity.rb
+++ b/gems/aws-sdk-autoscaling/lib/aws-sdk-autoscaling/activity.rb
@@ -21,6 +21,7 @@ module Aws::AutoScaling
       @id = extract_id(args, options)
       @data = options.delete(:data)
       @client = options.delete(:client) || Client.new(options)
+      @waiter_block_warned = false
     end
 
     # @!group Read-Only Attributes
--- a/.rspec
+++ b/.rspec
@@ -1,0 +1,1 @@
+--warnings
--- a/build_tools/aws-sdk-code-generator/spec/spec_helper.rb
+++ b/build_tools/aws-sdk-code-generator/spec/spec_helper.rb
@@ -138,7 +138,7 @@ module SpecHelper
 
     def model_path(model, models_dir)
       path = "#{models_dir}/#{model}.json"
-      File.exists?(path) ? path : nil
+      File.exist?(path) ? path : nil
     end
   end
 end
--- a/build_tools/aws-sdk-code-generator/templates/resource_class.mustache
+++ b/build_tools/aws-sdk-code-generator/templates/resource_class.mustache
@@ -30,6 +30,7 @@ module {{module_name}}
       @data = Aws::EmptyStructure.new
       {{/shape}}
       @client = options.delete(:client) || Client.new(options)
+      @waiter_block_warned = false
     end
 
     # @!group Read-Only Attributes
--- a/build_tools/changelog.rb
+++ b/build_tools/changelog.rb
@@ -74,7 +74,7 @@ module BuildTools
     private
 
     def read
-      if File.exists?(path)
+      if File.exist?(path)
         File.open(path, 'r', encoding: 'UTF-8') { |f| f.read }
       else
         ''
--- a/build_tools/custom_service.rb
+++ b/build_tools/custom_service.rb
@@ -54,7 +54,7 @@ module BuildTools
 
     def model_path(model_dir)
       path = File.expand_path("#{model_dir}/service-2.json", __FILE__)
-      File.exists?(path) ? path : nil
+      File.exist?(path) ? path : nil
     end
 
     def gem_dependencies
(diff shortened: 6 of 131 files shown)


ioquatix:
https://github.com/gettalong/kramdown/issues/634

ioquatix:
```
/Users/samuel/Documents/ioquatix/aws-sdk-ruby/gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb:187: warning: Net::HTTPResponse#header is obsolete
/Users/samuel/Documents/ioquatix/aws-sdk-ruby/gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb:187: warning: Net::HTTPResponse#header is obsolete
/Users/samuel/Documents/ioquatix/aws-sdk-ruby/gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb:187: warning: Net::HTTPResponse#header is obsolete
/Users/samuel/Documents/ioquatix/aws-sdk-ruby/gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb:187: warning: Net::HTTPResponse#header is obsolete
/Users/samuel/Documents/ioquatix/aws-sdk-ruby/gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb:187: warning: Net::HTTPResponse#header is obsolete
/Users/samuel/Documents/ioquatix/aws-sdk-ruby/gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb:187: warning: Net::HTTPResponse#header is obsolete
```

Might require some more attention - beyond my time availability.

ioquatix:
Unfortunately https://github.com/aws/aws-sdk-ruby/pull/2179 has been held up for almost 2 years. But it's responsible for ~20k warnings when running specs.

ioquatix:
Please ensure you merge the changes to `.rspec` this time. https://github.com/aws/aws-sdk-ruby/issues/1778#issuecomment-434106447 said there would be a follow up but there never was. Because as you can see, without constant attention, it's easy for new warnings to show up. There are many issues where people are submitting fixes for warnings, but they should never have occurred in the first place. Thanks.

gems/aws-sdk-autoscaling/lib/aws-sdk-autoscaling/activity.rb
```diff
@@ -21,6 +21,7 @@ def initialize(*args)
       @id = extract_id(args, options)
       @data = options.delete(:data)
       @client = options.delete(:client) || Client.new(options)
+      @waiter_block_warned = false
```
mullermp: These service files are generated. I see that you added this to the mustache template as well. Were these added manually or the result of `rake build`?

mullermp:
@ioquatix Sorry that it was dropped. I can merge this, but I'd like to do so after re:Invent to avoid any hidden problems. I hope that's okay. I have changes queued up for next week and I'd like to minimize any impact.

gems/aws-sdk-autoscaling/lib/aws-sdk-autoscaling/activity.rb
ioquatix: `rake build`

mullermp:
Everything caught on fire today anyway, so I'll just go and merge this now. Thanks again for the contribution.

mullermp:
@ioquatix I am reverting the --warnings change. While squashing warnings is helpful (I will run it from time to time), we are unfortunately pinned to an old version of mustache and it makes it difficult to diagnose spec failures and slows down CI. We're trying to get off mustache and instead use Smithy (https://awslabs.github.io/smithy/) but that work would be a version 4 that only follows a version 2 deprecation.

ioquatix:
Mustache is releasing an update hopefully fixing all warnings.

mullermp:
@ioquatix is the update to the older version that supports Ruby 1.9.3?

Gemfile:
```
  gem 'mustache', '0.99.8' # pinned to support Ruby 1.9.3
```

Someone would need to create a branch (such as v0.99.9) starting at v0.99.8, fix the warnings, and then publish. We could change the pinned dependency to ~> 0 because version 1.0.0 is ruby > 2. Going forward, if there were more things to fix, it would need to have a version such as 0.99.9.1, 0.99.9.2, etc.

mullermp:
Adding to this, I would LOVE to drop Ruby 1.9.3 support. As well as deprecating V2. I'll start those discussions internally after re:Invent.

ioquatix:
@mullermp Do you have an update?

mullermp:
I was able to get leadership on board with deprecating version 2 (blog: https://aws.amazon.com/blogs/developer/deprecation-schedule-for-aws-sdk-for-ruby-v2/)

Runtimes are still an open question - I think that there will be some global SDK maintenance policy in the future that will allow us to make those decisions. Based on the data, the usage of V3 on Ruby 1.9->2.2 is very low and I want to make that case for dropping support.

ioquatix:
Awesome! Thanks for the update.

PR #2185, opened 2019-12-04
Fix access point ARN retries

mullermp:

### DIFF
--- a/gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/bucket_arn.rb
+++ b/gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/bucket_arn.rb
@@ -25,7 +25,7 @@ the S3 ARN.
           def call(context)
             bucket_member = _bucket_member(context.operation.input.shape)
             if bucket_member && (bucket = context.params[bucket_member])
-              _resolved_bucket, resolved_region, arn = BucketARN.resolve_arn!(
+              _resolved_bucket, _resolved_region, arn = BucketARN.resolve_arn!(
                 bucket,
                 context.config.region,
                 context.config.s3_use_arn_region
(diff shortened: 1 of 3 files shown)


gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/s3_signer.rb
```diff
@@ -199,15 +199,23 @@ def build_v4_signer(options = {})
           end
 
           def new_hostname(context, region)
-            bucket = context.params[:bucket]
-            if region == 'us-east-1'
-              "#{bucket}.s3.amazonaws.com"
```
mullermp: I think this qualifies as "old code". I've removed it because we already force regionalization in EndpointProvider here: https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-partitions/lib/aws-partitions/endpoint_provider.rb#L88

gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/bucket_arn.rb
```diff
@@ -25,15 +25,14 @@ class Handler < Seahorse::Client::Handler
...
-              _resolved_bucket, resolved_region, arn = BucketARN.resolve_arn!(
+              _resolved_bucket, _resolved_region, arn = BucketARN.resolve_arn!(
                 bucket,
                 context.config.region,
                 context.config.s3_use_arn_region
               )
               if arn
                 if arn.resource.start_with?('accesspoint')
                   validate_config!(context.config)
-                  context.config.region = resolved_region
```
mullermp: I think this was incorrectly being set. The warning would show:

> S3 client configured for "us-west-2" but the bucket "arn:aws:s3:us-west-2:995319131876:accesspoint/mamuller-us-west-2" is in "us-west-2"; Please configure the proper region to avoid multiple unnecessary redirects and signing attempts

For the first request (that would fail and be retried) it should be set to 'us-east-1' and then set to 'us-west-2' on the second attempt rather than "patching" the current context in place.

PR #2188, opened 2019-12-08
Fix sts presigner region endpoints

lwoggardner:

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-sts/presigner.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-sts/presigner.rb
@@ -51,7 +51,7 @@ module Aws
         )
 
         url = Aws::Partitions::EndpointProvider.resolve(
-          req.context.config.region, 'sts'
+            req.context.config.region, 'sts', false
         )
         url += "/?#{param_list}"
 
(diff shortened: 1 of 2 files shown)


gems/aws-sdk-core/lib/aws-sdk-sts/presigner.rb
```diff
@@ -51,7 +51,7 @@ def get_caller_identity_presigned_url(options = {})
         )
 
         url = Aws::Partitions::EndpointProvider.resolve(
-          req.context.config.region, 'sts'
+          req.context.config.region, 'sts', false
```
mullermp: Thanks for the contribution.

I think this should be set to `'regional'` instead of `false` as it should be a string type: https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-partitions/lib/aws-partitions/endpoint_provider.rb#L42

gems/aws-sdk-core/spec/aws/sts/presigner_spec.rb
```diff
@@ -21,15 +21,15 @@ module STS
           )
           pre = Presigner.new(client: client)
 
-          expected_url = 'https://sts.amazonaws.com/?Action=GetCallerIdentity'\
+          expected_url = 'https://sts.us-west-2.amazonaws.com/?Action=GetCallerIdentity'\
```
mullermp: Please also wrap this to 80 characters now that this line is longer!

mullermp:
We have a release going out soon, so I actually went and made these changes for you so that I wouldn't have to release another version with the fix (since the presigner was not shipped yet anyway).

https://github.com/aws/aws-sdk-ruby/commit/41beac4cc244ef687f895648337f7720ec6773ad

PR #2200, opened 2020-01-02
S3/client: fix small typo in put_bucket_policy method comment

skipcloud:
The method comment had a backtick out of place, i.e. `<code>PutBucketPolic</code>y`

This small patch just fixes it.


mullermp:
Hey -- thanks for the contribution. Unfortunately this JSON file comes from an upstream source, so if we merge this change as is, it will be erased on the next S3 release. I've went ahead and informed the S3 doc writers about this and they will add a commit upstream to fix it.

PR #2202, opened 2020-01-08
Reuse connections even if http_wire_trace is true

abicky:
ConnectionPool.pool_options always includes a new logger if http_wire_trace is true, so ConnectionPool.for always returns a new instance. Connections will be reused by this commit even if http_wire_trace is true.

### DIFF
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Reuse connections even if `http_wire_trace` is true.
+
 3.86.0 (2019-12-13)
 ------------------
 
--- a/gems/aws-sdk-core/lib/seahorse/client/net_http/connection_pool.rb
+++ b/gems/aws-sdk-core/lib/seahorse/client/net_http/connection_pool.rb
@@ -18,6 +18,7 @@ module Seahorse
 
         @pools_mutex = Mutex.new
         @pools = {}
+        @default_logger = Logger.new($stdout)
 
         OPTIONS = {
           http_proxy: nil,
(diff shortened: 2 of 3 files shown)


mullermp:
Hi! Thanks for the contribution! Just a quick question - if I'm understanding this change, this is intended to reuse **loggers** and not about reusing **connections** correct?

Also, could you please comment with:

"By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license."

abicky:
> this is intended to reuse loggers and not about reusing connections correct?

No. In [ConnectionPool.for](https://github.com/aws/aws-sdk-ruby/blob/d9b535440172be3e50423bb9a750cb88b22c7f02/gems/aws-sdk-core/lib/seahorse/client/net_http/connection_pool.rb#L213-L218), an existing connection pool will be reused only if `@pools` has the key returned from `ConnectionPool.pool_options`. In other words, if `ConnectionPool.pool_options` returns different values even though the same arguments are passed, `ConnectioonPool.for` always create a new instance. Consequently, any connections are never reused. I've changed the behavior by reusing `logger` to reuse connections.

For example, the following script consumes file descriptor without limitation and `SocketError` will be raised because every method call creates a new connection:

```ruby
require 'aws-sdk-ecs'

client = Aws::ECS::Client.new(http_wire_trace: true)
loop do
  client.list_clusters
  sleep 0.1
end
```

I've changed the test to clarify the purpose https://github.com/aws/aws-sdk-ruby/compare/c2a6c99cc80e86e055aab3ee3989e115e0adc368..cde79b8277919e788837b0ca831b95289b2a7204.

mullermp:
That makes sense to me! I am happy to merge this, but first, please comment: "By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license."

mullermp:
Thanks for the contribution!

PR #2208, opened 2020-01-15
fix ruby 2.7 warning

timoschilling:
fixes
```
lib/seahorse/client/net_http/connection_pool.rb:309: warning: `if' at the end of line without an expression
```

### DIFF
--- a/gems/aws-sdk-core/lib/seahorse/client/net_http/connection_pool.rb
+++ b/gems/aws-sdk-core/lib/seahorse/client/net_http/connection_pool.rb
@@ -306,10 +306,7 @@ module Seahorse
           now = Aws::Util.monotonic_milliseconds
           @pool.each_pair do |endpoint,sessions|
             sessions.delete_if do |session|
-              if
-                session.last_used.nil? or
-                now - session.last_used > http_idle_timeout * 1000
-              then
+              if session.last_used.nil? or now - session.last_used > http_idle_timeout * 1000
                 session.finish
                 true
               end


gems/aws-sdk-core/lib/seahorse/client/net_http/connection_pool.rb
```diff
@@ -306,10 +306,7 @@ def _clean
           now = Aws::Util.monotonic_milliseconds
           @pool.each_pair do |endpoint,sessions|
             sessions.delete_if do |session|
-              if
-                session.last_used.nil? or
-                now - session.last_used > http_idle_timeout * 1000
-              then
+              if session.last_used.nil? or now - session.last_used > http_idle_timeout * 1000
```
nbekirov: Also, what would you say about converting the `or` to `||`?

I'm not sure if any particular style guide is followed but maybe [No and or or](https://github.com/rubocop-hq/ruby-style-guide#no-and-or-or) is a good rule to have in mind to avoid operator precedence related bugs.

mullermp: Eventually, my goal is to add Rubocop to this package, so +1 on using `||` but I won't block this PR on that. :)

mullermp: Would also prefer to stay within 80 characters, but also won't block this PR on that!

timoschilling: I think `||`, an line limit and rubocop are good ideas. I use them all in my private and work projects. But I think as I meet the current code requirements and the goal of this PR is fixing Ruby 2.7, this PR should keep as it is.

There are already line loger as this in this file.

My experience is that a 80 char limit is to hard, you should think about 120 chars, which is for example the limit of the github code view without scrolling.

nbekirov: @mullermp Sorry, didn't mean to disturb the PR flow! I was just confused while recently looking at this and trying to figure if there is an accepted style guide around. 

Rubocop is cool to have, but for large and legacy projects like this one, you may want to consider combining it with [pronto](https://github.com/prontolabs/pronto) to analyze only the PR diff.

mullermp:
Thanks for the contribution!

PR #2217, opened 2020-01-30
Remove JSON dependency

casperisfine:
I know that gem is deprecated and all, and we're trying to get rid of it, but in the meantime this dependency on `json ~> 1.4` is causing us lots of trouble.

@awood45 since you relaxed the dependency on Nokogiri, would it be possible to do the same with JSON?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - JSON - Removed the dependency on the json gem because it is a default gem since ruby 1.9.
+
 1.67.0 (2017-04-03)
 ------------------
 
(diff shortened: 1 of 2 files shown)


mullermp:
Version 1 has been deprecated since 2015. Please use Version 3. I will not be changing this.

alextwoods:
@casperisfine  - Just wanted to understand a little bit more about your use case.  What issue is the json depednency causing you?  Is there something blocking you from upgrading to Version 3 of the SDK?  

Unfortunately dropping the json dependency would break users of Ruby < 1.9.  Do you have anyway of overriding the usage locally?

casperisfine:
> What issue is the json depednency causing you?

It locks us to a fairly old version of `json`.

> Is there something blocking you from upgrading to Version 3 of the SDK?

Lots of legacy code, we're migrating but it's a lot of works so it's not going to happen short term.

> Unfortunately dropping the json dependency would break users of Ruby < 1.9. 

The alternative is to add a laxer dependency, e.g. `>= 1.4`.

> Do you have anyway of overriding the usage locally?

I'll just fork the gem and use it as a git source.

Issue #2122, opened 2019-09-20
Aws::CognitoIdentityProvider::Client initiate_auth exposes user password in logs


miqs1992:
Please fill out the sections below to help us address your issue

### Issue description
In a rails 6 application, I'm using Cognito to manage user authentication. I recently noticed that calls' logs are automatically added to the logs but there is an issue - Cognito initiate_auth method with `auth_flow: "USER_PASSWORD_AUTH"` is exposing the password provided by the user:
```
[Aws::CognitoIdentityProvider::Client 200 1.012806 0 retries] initiate_auth(client_id:"[FILTERED]",auth_flow:"USER_PASSWORD_AUTH",auth_parameters:{"USERNAME"=>"test@gmail.com","PASSWORD"=>"Test1234!"})
```
 
### Gem name ('aws-sdk', 'aws-sdk-resources' or service gems like 'aws-sdk-s3') and its version
aws-sdk-core 3.67.0
aws-sdk-cognitoidentityprovider 1.25.0

### Version of Ruby, OS environment
ruby 2.6.3
Mac OS 10.13.6

### Code snippets / steps to reproduce
my code: 
```
COGNITO_CLIENT = Aws::CognitoIdentityProvider::Client.new
COGNITO_CLIENT.initiate_auth(
  client_id: ENV['AWS_COGNITO_CLIENT_ID'],
  auth_flow: 'USER_PASSWORD_AUTH',
  auth_parameters: { 'USERNAME' => username, 'PASSWORD' => password }
)
```

The issue is related to the filter methods in https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/lib/aws-sdk-core/log/param_filter.rb

```
def filter_hash(values)
  filtered = {}
  values.each_pair do |key, value|
    filtered[key] = @filters.include?(key) ? '[FILTERED]' : filter(value)
  end
  filtered
end
```

During parsing `:auth_parameters` hash keys of this hash are actually string written in the capital letters, so they do not match symbols form `SENSITIVE` constant.

My suggestion is to include `'PASSWORD'` or `:auth_parameters` in `SENSITIVE` constant or to add some parsing to the key in this method to make it more universal like:
```
def filter_hash(values)
  filtered = {}
  values.each_pair do |key, value|
    key = key.downcase.to_sym
    filtered[key] = @filters.include?(key) ? '[FILTERED]' : filter(value)
  end
  filtered
end
```
This change produces log like:
```
[...]

Issue #2082, opened 2019-07-01
Error filtering in aws-sdk-transcribestreamingservice overly restrictive


TJNII:
### Issue description

When using the aws-sdk-transcribestreamingservice gem I noticed exceptions and log messages have important information redacted out of them.

The following was the exception text when too large of a block size was used:

```
{:message=>"[FILTERED]", :event_type=>:bad_request_exception}
```

By enabling http debugging I was able to see that :message should have shown the following, non-sensitive information:

```
{\"Message\":\"Your stream is too big. Reduce the frame size and try your request again.\"}
```

Furthermore when printing events to get a better sense of what events are available I noticed the event types were also being redacted:

```
[...]

PR #2219, opened 2020-01-31
feat(aws-sdk-core): case-insensitive log filtering

jessedoyle:
resolves: https://github.com/aws/aws-sdk-ruby/issues/2122

* There are AWS API calls that may require alternate casing for
  sensitive parameters (i.e. Cognito `initiate_auth` with an
  uppercased `PASSWORD` parameter).
* To be safe, perform a case-insensitive comparison when looping
  through parameter names to filter.
* Add test cases for `Aws::Log::ParamFilter#filter` that explicitly
  check for a case-insensitive comparison when filtering log
  parameters.

### DIFF
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Perform a case-insensitive comparison when filtering sensitive parameters from logs
+
 3.90.0 (2020-02-12)
 ------------------
 
--- a/gems/aws-sdk-core/lib/aws-sdk-core/log/param_filter.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/log/param_filter.rb
@@ -7,7 +7,7 @@ module Aws
 
       # A managed list of sensitive parameters that should be filtered from
       # logs. This is updated automatically as part of each release. See the
-      # `tasks/sensitive.rake` for more information.
+      # `tasks/update-sensitive-params.rake` for more information.
       #
       # @api private
       # begin
@@ -31,7 +31,7 @@ module Aws
       def filter_hash(values)
         filtered = {}
         values.each_pair do |key, value|
-          filtered[key] = @filters.include?(key) ? '[FILTERED]' : filter(value)
+          filtered[key] = @filters.any? { |f| f.to_s.casecmp(key.to_s).zero? } ? '[FILTERED]' : filter(value)
         end
         filtered
       end
--- a/gems/aws-sdk-core/spec/aws/log/param_filter_spec.rb
+++ b/gems/aws-sdk-core/spec/aws/log/param_filter_spec.rb
@@ -1,0 +1,60 @@
+require_relative '../../spec_helper'
+
+module Aws
+  module Log
+    describe ParamFilter do
+      describe '#filter' do
+        context 'with an array' do
+          context 'with a filtered parameter name' do
+            it 'filters lowercase parameter names' do
+              expect(subject.filter([{ password: 'p@assw0rd'}])).to eq([{ password: '[FILTERED]'} ])
+            end
+
+            it 'filters uppercase parameter names' do
+              expect(subject.filter([{ PASSWORD: 'p@assw0rd'}])).to eq([{ PASSWORD: '[FILTERED]'} ])
+            end
+
+            it 'filters mixed-case parameter names' do
+              expect(subject.filter([{ Password: 'p@assw0rd'}])).to eq([{ Password: '[FILTERED]'} ])
+            end
+          end
+        end
+
+        context 'with a Struct' do
+          context 'with a filtered parameter name' do
+            it 'filters lowercase parameter names' do
+              instance = Struct.new(:password).new('p@assw0rd')
+              expect(subject.filter(instance)).to eq(password: '[FILTERED]')
+            end
+
+            it 'filters uppercase parameter names' do
+              instance = Struct.new(:PASSWORD).new('p@assw0rd')
+              expect(subject.filter(instance)).to eq(PASSWORD: '[FILTERED]')
+            end
+
+            it 'filters mixed-case parameter names' do
+              instance = Struct.new(:Password).new('p@assw0rd')
+              expect(subject.filter(instance)).to eq(Password: '[FILTERED]')
+            end
+          end
+        end
+
+        context 'with a hash' do
+          context 'with a filtered parameter name' do
+            it 'filters lowercase parameter names' do
+              expect(subject.filter(password: 'p@ssw0rd')).to eq(password: '[FILTERED]')
+            end
+
+            it 'filters uppercase parameter names' do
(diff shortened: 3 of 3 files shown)


mullermp:
Thanks for opening up this pull request. See my comment on the main issue here: https://github.com/aws/aws-sdk-ruby/issues/2122#issuecomment-580916837 but this change should not be necessary. The service team has already fixed the problem by marking the field as sensitive. In general we prefer service teams to mark fields that way rather than handle this with client code in all of the SDKs.

jessedoyle:
Hi @mullermp - thanks for the response! I understand why it makes sense to have the service teams define their own sensitive parameters.

Do you have any idea of an ETA for the updates? 

I believe this is a fairly critical security issue as it's very easy to accidentally expose plaintext user passwords in log files when using`Aws::CognitoIdentityProvider::Client#initiate_auth`.

AWS logging can be enabled by gems such as [`aws-sdk-rails`](https://github.com/aws/aws-sdk-rails/blob/master/lib/aws-sdk-rails.rb#L34) unintentionally - which could lead to password exposure.

mullermp:
It's planned for mid February. I can't give details on their feature release, but what I can tell you is that I've checked that the change addressing this issue has been made and it is staged in that release.

mullermp:
I agree with you it's a critical security issue. I wish this was closed faster but the cognito team wanted to lump it in with their other feature. Without it, all of the SDKs have this problem. Thanks for understanding.

mullermp:
I was partially wrong on this, I'm sorry. The service team did update the params to be sensitive, but the upcase does bite us here. I think this is a side effect of Ruby styling (snake_case, lowercase) and how we implement the param filtering. I'm asking Cognito about supporting lowercase params (like other services..)

For implementation, I also looked at changing the rake task (https://github.com/aws/aws-sdk-ruby/blob/master/tasks/update-sensitive-params.rake) and adding 'USERNAME' or 'PASSWORD' to the SENSITIVE list. The initialize method of `ParamFilter` takes additional arguments (https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/lib/aws-sdk-core/log/param_filter.rb#L18) so perhaps another option is creating a plugin for Cognito to initialize a new ParamFilter and add 'USERNAME' or 'PASSWORD' that way.

Before I merge this, I wanted to bring these options up with you. One side effect of this change is that if two services define the same key (but different cased variations) then in both cases, even if a field is not marked as sensitive, will be treated as sensitive. This is sort of related to https://github.com/aws/aws-sdk-ruby/issues/2082 in that regard.

jessedoyle:
Hey @mullermp - thanks for the response.

My initial expectation with sensitive parameters is _any_ permutation of case should also filter out the parameter value.

If an AWS service used any of the following parameters:

```ruby
{
  Password: '...',
  password: '...',
  PASSWORD: '...'
}
```

, then I believe the user would reasonably expect that the parameters should be filtered from logs.

That's why I initially implemented the [`casecmp`](https://ruby-doc.org/core-2.4.0/String.html#method-i-casecmp) method to perform a case-insensitive search - because sensitive parameters _should_ be case-insensitive. 

It's better to be less risky and filter the parameters even if it's a file-positive than to expose extremely sensitive values.

But that's just my 2 cents!

jessedoyle:
I'd also be happy to get this PR back up-to-date with master if you'd consider merging!

mullermp:
Yes please. I can do a mini-review if you don't mind a little house cleaning.

mullermp:
Also, mostly agree on the false positives. Right now we have no way to undo those, so it's a pain point to consider.

gems/aws-sdk-core/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Feature - Perform a case-insensitive comparison when filtering sensitive parameters from logs
```
mullermp: Could you mark this as an `* Issue` instead of a `* Feature` ? I'd classify this as a bug.

gems/aws-sdk-core/spec/aws/log/param_filter_spec.rb
```diff
@@ -0,0 +1,25 @@
+require_relative '../../spec_helper'
+
+module Aws
+  module Log
+    describe ParamFilter do
+      describe '#filter' do
```
mullermp: Does it make sense to add other cases while you're here? One for struct/array? I won't block you on that, but it'd be helpful. :)

gems/aws-sdk-core/lib/aws-sdk-core/log/param_filter.rb
```diff
@@ -31,7 +31,7 @@ def filter(value)
       def filter_hash(values)
         filtered = {}
         values.each_pair do |key, value|
-          filtered[key] = @filters.include?(key) ? '[FILTERED]' : filter(value)
+          filtered[key] = @filters.any? { |f| f.to_s.casecmp(key.to_s).zero? } ? '[FILTERED]' : filter(value)
```
mullermp: What happens in the Array case? Would we potentially want to filter those?

mullermp:
Line 10 mentions `tasks/sensitive.rake`. Could you kindly also update that to `tasks/update-sensitive-params.rake`?

gems/aws-sdk-core/CHANGELOG.md
jessedoyle: For sure!

gems/aws-sdk-core/spec/aws/log/param_filter_spec.rb
jessedoyle: I'd be happy to - the more test cases the merrier :)

gems/aws-sdk-core/lib/aws-sdk-core/log/param_filter.rb
jessedoyle: I'll write a test for the array case to see what happens and adjust according.

jessedoyle:
@mullermp - Do you mind if I amend my last commit and force-push?

mullermp:
@jessedoyle It's your fork, you don't have to ask! ;)

gems/aws-sdk-core/lib/aws-sdk-core/log/param_filter.rb
jessedoyle: In the array case, hash/Struct/nested arrays elements are filtered due to the recursive `#filter` call.

If the object is a primitive value (say a `String`), then the value is returned as-is.

I have a basic test case incoming.

gems/aws-sdk-core/spec/aws/log/param_filter_spec.rb
jessedoyle: Done!

jessedoyle:
@mullermp - Thanks - just amended my last commit.

mullermp:
Cool, waiting for CI. Thanks!

jessedoyle:
Hold on - I forgot to update the mention to `tasks/sensitive.rake`. Will do one more amendment.

jessedoyle:
Good to go now - once CI is done!

mullermp:
I should be able to merge this tomorrow before our release.

PR #2223, opened 2020-02-10
Honor explicit profile in client when credentials from AWS_ environment variables are present

### LINKED ISSUE #2165: Ruby SDK does not honor explicit profile in client when credentials from AWS_ environment variables are present
Please fill out the sections below to help us address your issue

### Issue description

Ruby SDK does not honor explicit profile in client when credentials from AWS_ environment variables are present whereas Golang SDK supports this.

I have AWS_ environment variables setup.

*test.rb*
```ruby
require 'aws-sdk-core'
p Aws::STS::Client.new(profile: ENV['PROFILE']).get_caller_identity
```

```bash
aws_credentials_env $ bundle exec ruby test.rb
#<struct Aws::STS::Types::GetCallerIdentityResponse user_id="AIDAT6A**************", account="2706********", arn="arn:aws:iam::2706********:user/noel.georgi">
aws_credentials_env $ PROFILE=static bundle exec ruby test.rb
#<struct Aws::STS::Types::GetCallerIdentityResponse user_id="AIDAT6A**************", account="2706********", arn="arn:aws:iam::2706********:user/noel.georgi">
```

*shared.go*
```golang
package main

import (
	"fmt"
	"os"

	"github.com/aws/aws-sdk-go/aws/session"
	"github.com/aws/aws-sdk-go/service/sts"
)

func main() {
	// create a session
	opts := session.Options{}
	opts.Profile = os.Getenv("PROFILE")
	opts.SharedConfigState = session.SharedConfigEnable
	session, err := session.NewSessionWithOptions(opts)
	if err != nil {
		fmt.Println(err)
	}
	// create a service client
	stsClient := sts.New(session)
	// create a request
	request := sts.GetCallerIdentityInput{}
	// call the service operation
	id, err := stsClient.GetCallerIdentity(&request)
	if err != nil {
		fmt.Println(err)
	} else {
		// handle the response
		fmt.Printf("AccountId: %s, Arn: %s, UserId: %s", *id.Account, *id.Arn, *id.UserId)
		fmt.Println()
	}
}
```

```bash
aws_caller_identity $ go run shared.go
AccountId: 2706********, Arn: arn:aws:iam::2706********:user/noel.georgi, UserId: AIDAT6A**************
aws_caller_identity $ PROFILE=static go run shared.go
AccountId: 4118********, Arn: arn:aws:iam::4118********:user/noel.georgi, UserId: AIDAV7Y**************
```

Golang SDK honor's passing in explicit `profile` even with ENV vars set, whereas ruby SDK uses the ENV vars only. 

Side note: Ruby SDK also does not support `credential_source` as **Enviroment**

alextwoods:

Fixes #2165


ferricoxide:
Sorry to resurrect this, but…

Trying to get a GitLab installation to not be single-threaded when working with S3 (opened an [issue with GitLab](https://gitlab.com/gitlab-org/gitlab/-/issues/482770), but have gotten no response on it, yet). Haven't found anything in _their_ documentation to suggest that they're explicitly using this feature. Looking at _this_ code (which I'm _hoping_ their S3 storage-driver uses), I'm only seeing references to `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment variables and that it will also scan `~/.aws/credentials`) in the README. Not seeing indications that it would make use of permissions from the EC2's attached instance role, nor am I seeing a method for telling it to use a custom config-file (i.e.., `AWS_CONFIG_FILE`)

Overall, my goal here is to see if I can force GitLab to reference my performance-optimizations from `AWS_CONFIG_FILE` so that, instead of being a single-threaded process, when GitLab reads from or writes to S3 it does so in both a mult-threaded manner and that it takes advantage of the the common runtime client (absent using those, transfers are happening at about 130MiB/s vs the 1000MiB/s that one achieves when using the AWS CLI).

alextwoods:
What performance optimizations do you have in your config file?  Unfortunately the Ruby SDK does not support the common runtime S3 client.  We do support multi-threaded uploads and downloads, but only through the Resource interface such as [Object#download_file](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/Object.html#download_file-instance_method) and calling `client.get_object` or `client.put_object` won't give you multithreaded upload/downloads.

mullermp:
AWS_CONFIG_FILE is supported, and so are credentials from your ec2 assumed role. Those should be mentioned in the readme and if not we can touch that up.

ferricoxide:
@alextwoods sed:
> What performance optimizations do you have in your config file? Unfortunately the Ruby SDK does not support the common runtime S3 client. We do support multi-threaded uploads and downloads, but only through the Resource interface such as [Object#download_file](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/Object.html#download_file-instance_method) and calling `client.get_object` or `client.put_object` won't give you multithreaded upload/downloads.

Bummer. What I've found, when attempting to transfer (with the AWS CLI) a multi-hundreds gigabyte TARfile created as an export (or as a full backup) is that, unless one uses the CRT, the transfer is very slow, even if one uses the non-CRT tuning-options. For a reference, when testing with a 500GiB TAR-file, _with_ CRT, the transfer runs at 1GiB/s; without CRT – but with suitable values for number of threads and chunk-sizes – I can get it to run at 200-300MiB/s. Ultimately, it was AWS Support's recommendation to switch to CRT. Dunno if that's a limitation of the AWS CLI that I _assumed_ carried over to the Ruby-based implementation (used by GitLab's S3 interfaces). Granted, my assumption that GitLab's [FOG](https://github.com/fog/fog)-based implementation is using this project's capabilities could be in error.

Suppose I should go hit up the FOG project to validate my assumptions. I think I ended up in _this_ thread due to a Google search that was thin on (non "sponsored" hits).

ferricoxide:
@mullermp sed:
> AWS_CONFIG_FILE is supported, and so are credentials from your ec2 assumed role. Those should be mentioned in the readme and if not we can touch that up.

I assumed that EC2 instance-roles were supported by the Ruby SDKs (since the GitLab documentation had provided requisite guidance for setting up it's Ruby-based S3 driver to use it). However, when I'd looked for `AWS_CONFIG_FILE` in either the SDK's documentation or a quick search (using GitHub's project-level search capability against this project), I didn't see it in the code. Thus, asked here.

Thank you for confirming. 😄

PR #2232, opened 2020-02-21
Presign a WebSocket URL - wss://domain.com/mqtt

scott-stewart:
Before this commit, it was not possible to presign a WebSocket url. The
method to determine if the url was "standard" only took into account
http:80 and https:443 as standard. If a wss url was provided for
presign, it resulted in the signed header host including a trailing
colon and prevented the signautre from working.

For example, before this commit, a call to presign_url with url:
'wss://domain.com', would result in a signed header host of:
'domain.com:' instead of 'domain.com', which would prevent the signature from working.

Here is a simple sample of code that works with this pull request, but results in an invalid presigned_url without it

```
def generate_presig_url
    signer = Aws::Sigv4::Signer.new(service: 'iotdevicegateway',
      region: ENV['AWS_REGION'],
      access_key_id: ENV['AWS_ACCESS_KEY_ID'],
      secret_access_key: ENV['AWS_SECRET_ACCESS_KEY'])
    signer.presign_url(http_method: 'GET',
      url: 'wss://' + ENV['AWS_IOT_ENDPOINT_HOST'].to_s + '/mqtt',
      expires_in: 3600)
  end
```
I fully appreciate the question over if WebSockets are an actual URI scheme or subset of HTTP, and that the Ruby URI library doesn't treat WSS or WS as an official scheme, and therefore when WSS or WS urls are requested for presign in the Ruby library, the default port is not evaluated to 80 or 443, which is what causes the bug of a signed host record of "domain.com:" instead of "domain.com" when presigning default port WSS urls. But, I think the choice of using URI to parse the URL into parts is an implantation detail, and would expect that a wss request could get a valid presign_url. Some services use WebSocket and accept SigV4 presigns (i.e. AWS IOT supports MQTT over WebSockets and the url to connect is WSS). Other AWS SDK Sigv4 libraries (Node for example) result in valid, usable presigned wss urls.

### DIFF
--- a/gems/aws-sigv4/lib/aws-sigv4/signer.rb
+++ b/gems/aws-sigv4/lib/aws-sigv4/signer.rb
@@ -558,7 +558,9 @@ module Aws
 
       def standard_port?(uri)
         (uri.scheme == 'http' && uri.port == 80) ||
-        (uri.scheme == 'https' && uri.port == 443)
+        (uri.scheme == 'https' && uri.port == 443) ||
+        (uri.scheme == 'ws' && (uri.port == 80 || uri.port.nil?)) ||
+        (uri.scheme == 'wss' && (uri.port == 443 || uri.port.nil?))
       end
 
       # @param [File, Tempfile, IO#read, String] value
--- a/gems/aws-sigv4/lib/aws-sigv4/signer.rb
+++ b/gems/aws-sigv4/lib/aws-sigv4/signer.rb
@@ -549,18 +549,12 @@ module Aws
       end
 
       def host(uri)
-        if standard_port?(uri)
+        # Handles known and unknown URI schemes; default_port nil when unknown.
+        if uri.default_port == uri.port
           uri.host
         else
           "#{uri.host}:#{uri.port}"
         end
-      end
-
-      def standard_port?(uri)
-        (uri.scheme == 'http' && uri.port == 80) ||
-        (uri.scheme == 'https' && uri.port == 443) ||
-        (uri.scheme == 'ws' && (uri.port == 80 || uri.port.nil?)) ||
-        (uri.scheme == 'wss' && (uri.port == 443 || uri.port.nil?))
       end
 
       # @param [File, Tempfile, IO#read, String] value
--- a/gems/aws-sigv4/spec/signer_spec.rb
+++ b/gems/aws-sigv4/spec/signer_spec.rb
@@ -143,6 +143,38 @@ module Aws
           expect(signature.headers['host']).to eq('domain.com:123')
         end
 
+        it 'includes WS port in Host when not 80' do
+          signature = Signer.new(options).sign_request(
+            http_method: 'GET',
+            url: 'ws://domain.com:123'
+          )
+          expect(signature.headers['host']).to eq('domain.com:123')
+        end
+
+        it 'includes WSS port in Host when not 443' do
+          signature = Signer.new(options).sign_request(
+            http_method: 'GET',
+            url: 'wss://domain.com:123'
+          )
+          expect(signature.headers['host']).to eq('domain.com:123')
+        end
+
+        it 'does not include trailing colon in HOST if WS default port' do
+          signature = Signer.new(options).sign_request(
+            http_method: 'GET',
+            url: 'ws://domain.com'
+          )
+          expect(signature.headers['host']).to eq('domain.com')
+        end
+
+        it 'does not include trailing colon in HOST if WSS default port' do
+          signature = Signer.new(options).sign_request(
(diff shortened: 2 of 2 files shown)


alextwoods:
Thanks for the contribution!  We'll review and discuss with some of the other SDK teams to ensure this is consistent behavior.

mullermp:
> I fully appreciate the question over if WebSockets are an actual URI scheme or subset of HTTP, and that the Ruby URI library doesn't treat WSS or WS as an official scheme,

This is a good question. I have drafted a change for the `uri` library in Ruby to support WebSockets.

gems/aws-sigv4/lib/aws-sigv4/signer.rb
```diff
@@ -558,7 +558,9 @@ def host(uri)
 
       def standard_port?(uri)
         (uri.scheme == 'http' && uri.port == 80) ||
-        (uri.scheme == 'https' && uri.port == 443)
+        (uri.scheme == 'https' && uri.port == 443) ||
+        (uri.scheme == 'ws' && (uri.port == 80 || uri.port.nil?)) ||
+        (uri.scheme == 'wss' && (uri.port == 443 || uri.port.nil?))
       end
```
mullermp: After some discussion with @alextwoods, we think that it would be better to remove the `standard_port?` method.

The `host` method could be rewritten:

```
      def host(uri)
        if uri.default_port == uri.port # no longer calls standard_port?
          uri.host
        else
          "#{uri.host}:#{uri.port}"
        end
      end
```
This checks the URI for the default port and the port from the URI. If it's default, send the host, otherwise, include the port. Since `nil == nil => true` it would work for new schemes.

Pros:
 * In the future when other schemes are added, no updates are necessary.
 * Uses the `Generic#default_port` [method](https://github.com/ruby/uri/blob/master/lib/uri/generic.rb#L40)

Cons:
 * Assumes other schemes will inherit from `Generic` though this is probably a safe assumption.

What do you think?

gems/aws-sigv4/spec/signer_spec.rb
```diff
@@ -143,6 +143,38 @@ module Sigv4
...
+
+        it 'includes WSS port in Host when not 443' do
+          signature = Signer.new(options).sign_request(
+            http_method: 'GET',
+            url: 'wss://domain.com:123'
+          )
+          expect(signature.headers['host']).to eq('domain.com:123')
+        end
+
+        it 'does not include trailing colon in HOST if WS default port' do
```
mullermp: I like that you made these tests about trailing colon, but I would see if you can refactor/delete these tests with the new changes. The previous tests were for HTTP, and now WebSocket, but we can have cases for: "scheme default port and uri port are the same", "scheme default port and uri port are different", "scheme not recognized and uri port not provided", and "scheme not recognized and uri port provided". You can rename those tests (or even use an RSpec 'context' block for 'scheme not recognized by URI module') if you'd like.

gems/aws-sigv4/lib/aws-sigv4/signer.rb
scott-stewart: I like that approach way better. My initial assumption was to do something similar in the host method and skip the standard_port method all together, but then figured the pull request might be more likely to get accepted by appending to the existing approach. I'm all for this take instead.

mullermp: Better (backwards compatible) code is more likely to get accepted than hacky code :)

gems/aws-sigv4/spec/signer_spec.rb
```diff
@@ -127,20 +127,53 @@ module Sigv4
...
+              url: 'abcd://domain.com'
+            )
+            expect(signature.headers['host']).to eq('domain.com')
+
+          end
+
+          it 'includes port in Host when uri port provided' do
+            signature = Signer.new(options).sign_request(
+              http_method: 'GET',
+              url: 'abcd://domain.com:123'
```
mullermp: Let's hope `abcd://` never becomes a real protocol! XD

scott-stewart: "Let's hope abcd:// never becomes a real protocol!" I knew we couldn't use ws/wss for the test because hopefully the URI pull request will get merged shortly. I originally chose "xyz", and then decided to check if it was a protocol, and found that it was! Thanks for holding my had through the better implementation!

mullermp:
I forgot to mention a changelog entry. I just pushed one: https://github.com/aws/aws-sdk-ruby/commit/5518cdbffa8add2c4fed25d4b368bd5e09aab455

This should go out in tomorrow's release.

PR #2235, opened 2020-02-25
Fix stuff

mullermp:
Rename client metrics
Fix DDB documentation bug

### DIFF
--- a/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/client_constructor.rb
+++ b/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/client_constructor.rb
@@ -24,8 +24,12 @@ module AwsSdkCodeGenerator
     private
 
     def documented_plugin_options(plugins)
+      i = 0
       plugins.map(&:options).flatten.select(&:documented?).sort_by do |opt|
-        [opt.required ? 'a' : 'b', opt.name]
+        # Stable sort, first required options, then sort by name, then if
+        # two plugins of the same name, use an incrementer.
+        # options.fetch(:plugins) will be ordered.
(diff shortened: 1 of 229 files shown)


build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/client_constructor.rb
```diff
@@ -25,7 +25,8 @@ def initialize(options)
 
     def documented_plugin_options(plugins)
       plugins.map(&:options).flatten.select(&:documented?).sort_by do |opt|
-        [opt.required ? 'a' : 'b', opt.name]
+        # Determine sorting - required first, and add overrides at the end
+        [opt.required ? 'a' : 'b', opt.name, opt.override ? 'b' : 'a']
```
alextwoods: The plugins are passed to this class in the order that they are applied to the generated client, which is the order that defaults will be resolved in as well.  We should use that ordering rather than relying on whether one of them is required or not.

gems/aws-sdk-accessanalyzer/lib/aws-sdk-accessanalyzer/client.rb
```diff
@@ -21,8 +21,8 @@
 require 'aws-sdk-core/plugins/stub_responses.rb'
 require 'aws-sdk-core/plugins/idempotency_token.rb'
 require 'aws-sdk-core/plugins/jsonvalue_converter.rb'
-require 'aws-sdk-core/plugins/client_metrics_plugin.rb'
-require 'aws-sdk-core/plugins/client_metrics_send_plugin.rb'
+require 'aws-sdk-core/plugins/client_metrics.rb'
```
mullermp: Test Comment

PR #2049, opened 2019-05-16
Support Modeled Exceptions

cjyclaire:
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

TODO, need to add changelog entry once review complete

This PR adds support for Modeled exceptions from service model (which contains no-empty members in their error shape definition), introducing

1. code generation for modeled error class
2. xml and json error handler supporting error parsing
3. unit tests for different protocol mocking errors

sample generated diff will be attached in comments

code generation change => `/build_tools/`
error parsing changes => `aws-sdk-core/lib/aws-sdk-core`
  * gems/aws-sdk-core/lib/aws-sdk-core/json/error_handler.rb
  * gems/aws-sdk-core/lib/aws-sdk-core/xml/error_handler.rb

PR #2238, opened 2020-02-27
Consider empty member exception structs as errors

mullermp:
Not too sure on this change, but it appears errors are not generated in line with v2. Maybe broken behavior from https://github.com/aws/aws-sdk-ruby/pull/2049 ?

Errors: https://docs.aws.amazon.com/sdk-for-ruby/v2/api/Aws/ElastiCache/Errors.html
No errors: https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/ElastiCache/Errors.html

Errors: https://docs.aws.amazon.com/sdk-for-ruby/v2/api/Aws/S3/Errors.html
No errors: https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/Errors.html

### DIFF
--- a/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/error_list.rb
+++ b/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/error_list.rb
@@ -9,7 +9,7 @@ module AwsSdkCodeGenerator
       @errors = @api['shapes'].inject([]) do |es, (name, shape)|
         # only generate error shape with non empty members
         # excluding event shapes marked as error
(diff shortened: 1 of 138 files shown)


build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/error_list.rb
```diff
@@ -31,8 +31,7 @@ def initialize(options)
 
     def non_empty_error_struct?(shape)
```
mullermp: I will clean up this name.

PR #2246, opened 2020-03-03
Correct and retry clock skew errors

### LINKED ISSUE #2005: Feature request: Automatic clock skew correction?
Is there any reason not to implement something akin to [https://github.com/aws/aws-sdk-js/pull/706](https://github.com/aws/aws-sdk-js/pull/706)? Recently experienced errors due to clock skew and it would be convenient to have a permanent solution at the SDK level. Thanks.

alextwoods:

Fixes #2005


gems/aws-sdk-core/lib/aws-sdk-core/plugins/retries/clock_skew.rb
```diff
@@ -0,0 +1,64 @@
...
+
+      # @api private
+      class ClockSkew
+
+        CLOCK_SKEW_THRESHOLD = 5 * 60 # five minutes
+
+        def initialize
+          @mutex = Mutex.new
+          @system_clock_offset = 0
+          @endpoint_clock_corrections = Hash.new(0)
```
mullermp: Minor - align params

gems/aws-sdk-core/lib/aws-sdk-core/plugins/retries/error_inspector.rb
```diff
@@ -106,14 +118,20 @@ def modeled_throttling?
...
         def retryable?(context)
           (expired_credentials? && refreshable_credentials?(context)) ||
             throttling_error? ||
             checksum? ||
             networking? ||
             server? ||
             endpoint_discovery?(context) ||
-            modeled_retryable?
+            modeled_retryable? ||
+            clock_skew?(context)
```
mullermp: Does any ordering matter here? I would think that this should be after `server?` -- also maybe `modeled_retryable` should be moved up.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/retry_errors.rb
```diff
@@ -183,6 +198,20 @@ def self.resolve_adaptive_retry_wait_to_fill(cfg)
         value == 'true'
       end
 
+      def self.resolve_correct_clock_skew(cfg)
+        value = ENV['AWS_CORRECT_CLOCK_SKEW']
+        value = Aws.shared_config.correct_clock_skew(profile: cfg.profile) if value.nil?
+        value = false if value.nil?
```
mullermp: Talked offline about this

gems/aws-sdk-core/lib/aws-sdk-core/plugins/retry_errors.rb
```diff
@@ -196,6 +225,11 @@ def call(context)
           error_inspector = Retries::ErrorInspector.new(response.error, response.context.http_response.status_code)
 
           request_bookkeeping(context, response, error_inspector)
+
+          # Clock skew needs to be updated from the response even when
+          # the request is not retryable
+          config.clock_skew.update_clock_skew(context) if error_inspector.clock_skew?(context)
```
mullermp: You may not like this, but when `ErrorInspector` checks clock skew, we should potentially do the updating there, similar to endpoint discovery. Or even that can be refactored out to something like this.

gems/aws-sdk-core/spec/aws/plugins/retries/error_inspector_spec.rb
```diff
@@ -227,6 +227,43 @@ def inspector(error, http_status_code = 404)
           end
         end
       end
+
+      describe '#clock_skew?' do
+        let(:context) { double("context", config: double("config", clock_skew: double('clock_skew', clock_skewed?: clock_skewed))) }
```
mullermp: Minor - multiline

gems/aws-sdk-core/spec/aws/plugins/signature_v4_spec.rb
```diff
@@ -221,6 +221,65 @@ module Plugins
         end
 
       end
+
+      describe 'clock skew correction' do
+        let(:clock_skew) { Retries::ClockSkew.new }
+        let(:now) { Time.now }
+        let(:utc) { now.utc }
+
+        before(:each) {
```
mullermp: Minor - do .. end

gems/aws-sdk-resources/bin/aws-v3.rb
```diff
@@ -185,6 +185,7 @@ def run_with_pry
 
 def run_with_irb
   require 'irb'
+  puts Time.now
```
mullermp: Remove

gems/aws-sdk-core/lib/aws-sdk-core/plugins/retries/error_inspector.rb
alextwoods: In theory order should not matter here - I guess in theory we want to put the most likely errors first (to take advantage of short circuiting).  However, because the check for endpoint discovery actually will do something, the order *could* matter in some cases (see other discussion).  I've updated the order a bit.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/retry_errors.rb
alextwoods: I strongly disagree that `clock_skew?` should do any updating.  I'd prefer to extract the logic for updating the end point cache from the Error Inspector.

mullermp: I agree that the predicate method shouldn't update anything. If we can refactor that out (at least for the retries changes) that would be nice.

PR #2254, opened 2020-03-09
Oj compatibility with mimic json mode

### LINKED ISSUE #2247: aws-sdk-core implicitly depends on Oj gem
According to https://github.com/aws/aws-sdk-ruby/blob/a99d0abf9e61a0f4b848a60043cc4e8db00c8971/gems/aws-sdk-core/lib/aws-sdk-core/json.rb#L41

`aws-core-sdk` will try to use OJ if its installed, and the version of this dependency is not controlled at all. When error raised from Oj, it also tries to convert `Oj::ParseError` to `Aws::Json::ParseError `
https://github.com/aws/aws-sdk-ruby/blob/a99d0abf9e61a0f4b848a60043cc4e8db00c8971/gems/aws-sdk-core/lib/aws-sdk-core/json.rb#L26

https://github.com/ohler55/oj/issues/584
However, according to the Author of Oj, Oj is supposed to raise `Json::ParserError`, and raising `Oj::ParserError` is the behavior caused by a bug in a previous version. 

Context:
Same issue from https://github.com/aws/aws-sdk-ruby/pull/2085 but still unfixed for us since we use a different Oj version.

Suggested changes could be:
- also rescue JSON::ParserError even if Oj is installed
- declare the Oj dependency explicitly
...

mullermp:
Fixes #2247

### DIFF
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,11 +1,12 @@
 Unreleased Changes
 ------------------
 
+* Issue - Rescue from `JSON::ParserError` when using `Oj.mimic_JSON`. (#2247)
+
 3.91.0 (2020-03-09)
 ------------------
 
 * Feature - Updated Aws::STS::Client with the latest API changes.
-
 * Feature - Add `standard` and `adaptive` retry modes.
 
 3.90.1 (2020-02-14)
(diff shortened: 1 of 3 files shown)


mullermp:
Haven't been able to really make a test for this... `Oj.mimic_JSON` will modify the JSON module and muck up the environment.

PR #2257, opened 2020-03-11
Stack conn pool

Bobochka:


alextwoods:
Can you provide a little bit more context and reasoning behind this change?

From the changelog it sounds like you're trying to have the connection pool return the least recently used connections.  In the current implementation it already does that.  When we get a new connection we do `@pool[endpoint].shift` and when we return the connection we return it to the end of the pool with  `@pool[endpoint] << session`.

PR #2269, opened 2020-03-19
remove mutated string declarations

midnight-wonderer:
These gems are the reason why I cannot set  
RUBYOPT=--enable=frozen-string-literal

Can I introduce just some styling changes, leaving the functionality the same?
This pull request does just that.

P.S. I will let Travis do his jobs first; too lazy to test locally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: ruby
 
 before_install:
   - gem install bundler -v '< 2'
-  - "find . -type f -iname '*.rb' -exec sed -i '1s/^/# frozen_string_literal: true\\n/' \"{}\" +;"
 
 dist: trusty
 
@@ -25,7 +24,10 @@ env:
   - PURE_RUBY=1
   - KITCHEN_SINK=1
 
-script: rake build && rake test:spec
+script: >
+  bundle exec rake build &&
+  find . -not -path './vendor/*' -type f -iname '*.rb' -exec sed -i -z '1s/^\(\s*#\s*encoding[^\n]*\n\)\?/\1# frozen_string_literal: true\n/' '{}' +; &&
+  bundle exec rake test:spec
 
 bundler_args: --without docs repl
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: ruby
 
 before_install:
   - gem install bundler -v '< 2'
-  - "find . -type f -iname '*.rb' -exec sed -i '1s/^/# frozen_string_literal: true\\n\\n/' \"{}\" +;"
+  - "find . -type f -iname '*.rb' -exec sed -i '1s/^/# frozen_string_literal: true\\n/' \"{}\" +;"
 
 dist: trusty
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: ruby
 
 before_install:
   - gem install bundler -v '< 2'
+  - "find . -type f -iname '*.rb' -exec sed -i '1s/^/# frozen_string_literal: true\\n\\n/' \"{}\" +;"
 
 dist: trusty
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,6 @@ env:
 
 script: >
   (bundle exec rake build) &&
-  (find . -not -path './vendor/*' -type f -iname '*.rb' -exec sed -i -z '1s/^\(\s*#\s*encoding[^\n]*\n\)\?/\1# frozen_string_literal: true\n/' '{}' +;) &&
   (bundle exec rake test:spec)
 
 bundler_args: --without docs repl
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,8 @@
 language: ruby
+cache: bundler
+
+# to clear bundler cache:
+#   rm -r ./vendor
 
 before_install:
   - gem install bundler -v '< 2'
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,4 @@
 language: ruby
-cache: bundler
-
-# to clear bundler cache:
-#   rm -r ./vendor
 
 before_install:
   - gem install bundler -v '< 2'
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,9 +29,9 @@ env:
   - KITCHEN_SINK=1
 
 script: >
-  bundle exec rake build &&
-  find . -not -path './vendor/*' -type f -iname '*.rb' -exec sed -i -z '1s/^\(\s*#\s*encoding[^\n]*\n\)\?/\1# frozen_string_literal: true\n/' '{}' +; &&
-  bundle exec rake test:spec
+  (bundle exec rake build) &&
+  (find . -not -path './vendor/*' -type f -iname '*.rb' -exec sed -i -z '1s/^\(\s*#\s*encoding[^\n]*\n\)\?/\1# frozen_string_literal: true\n/' '{}' +;) &&
+  (bundle exec rake test:spec)
 
 bundler_args: --without docs repl
 
--- a/build_tools/spec/region_spec.rb
+++ b/build_tools/spec/region_spec.rb
@@ -33,7 +33,6 @@ describe 'ensure no hard-coded region' do
 
       file = File.open(path, 'r', encoding: 'UTF-8', &:read)
       lines = file.lines.to_a
-      lines_inserted = 2 # for string immutability tests
 
       it "#{path} has no hard-coded region" do
         lines.each_with_index do |val, idx|
@@ -42,9 +41,7 @@ describe 'ensure no hard-coded region' do
 
           # skip known whitelists
           next if whitelist[key] &&
-                  (((idx - lines_inserted)..idx).include?(
-                    whitelist[key][File.basename(path)]
-                  ) ||
+                  (whitelist[key][File.basename(path)] == idx ||
                   whitelist[key][File.basename(path)] == 'SKIP_FILE')
 
           # If we use \w+ we will get false positives for uid fields
--- a/build_tools/spec/region_spec.rb
+++ b/build_tools/spec/region_spec.rb
@@ -33,6 +33,7 @@ describe 'ensure no hard-coded region' do
 
       file = File.open(path, 'r', encoding: 'UTF-8', &:read)
       lines = file.lines.to_a
+      lines_inserted = 2 # for string immutability tests
 
       it "#{path} has no hard-coded region" do
         lines.each_with_index do |val, idx|
@@ -41,7 +42,9 @@ describe 'ensure no hard-coded region' do
 
           # skip known whitelists
           next if whitelist[key] &&
-                  (whitelist[key][File.basename(path)] == idx ||
+                  (((idx - lines_inserted)..idx).include?(
+                    whitelist[key][File.basename(path)]
+                  ) ||
                   whitelist[key][File.basename(path)] == 'SKIP_FILE')
 
           # If we use \w+ we will get false positives for uid fields
--- a/gems/aws-partitions/lib/aws-partitions/partition.rb
+++ b/gems/aws-partitions/lib/aws-partitions/partition.rb
@@ -21,8 +21,8 @@ module Aws
         if @regions.key?(region_name)
           @regions[region_name]
         else
-          msg = "invalid region name #{region_name.inspect}; valid region "
-          msg << "names include #{@regions.keys.join(', ')}"
+          msg = "invalid region name #{region_name.inspect}; valid region "\
+            "names include #{@regions.keys.join(', ')}"
           raise ArgumentError, msg
         end
       end
@@ -45,8 +45,8 @@ module Aws
         if @services.key?(service_name)
           @services[service_name]
         else
-          msg = "invalid service name #{service_name.inspect}; valid service "
-          msg << "names include #{@services.keys.join(', ')}"
+          msg = "invalid service name #{service_name.inspect}; valid service "\
+            "names include #{@services.keys.join(', ')}"
           raise ArgumentError, msg
         end
       end
--- a/gems/aws-partitions/lib/aws-partitions/partition.rb
+++ b/gems/aws-partitions/lib/aws-partitions/partition.rb
@@ -22,7 +22,7 @@ module Aws
           @regions[region_name]
         else
           msg = "invalid region name #{region_name.inspect}; valid region "\
-            "names include #{@regions.keys.join(', ')}"
+                "names include #{@regions.keys.join(', ')}"
           raise ArgumentError, msg
         end
       end
@@ -46,7 +46,7 @@ module Aws
           @services[service_name]
         else
           msg = "invalid service name #{service_name.inspect}; valid service "\
-            "names include #{@services.keys.join(', ')}"
+                "names include #{@services.keys.join(', ')}"
           raise ArgumentError, msg
         end
       end
--- a/gems/aws-sdk-core/lib/aws-sdk-core/plugins/retry_errors.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/plugins/retry_errors.rb
@@ -167,8 +167,8 @@ a clock skew correction and retry requests with skewed client clocks.
         # Raise if provided value is not one of the retry modes
         if value != 'legacy' && value != 'standard' && value != 'adaptive'
           raise ArgumentError,
-                'Must provide either `legacy`, `standard`, or `adaptive` for '\
-                'retry_mode profile option or for ENV[\'AWS_RETRY_MODE\']'
+            'Must provide either `legacy`, `standard`, or `adaptive` for '\
+            'retry_mode profile option or for ENV[\'AWS_RETRY_MODE\']'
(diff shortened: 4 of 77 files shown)


.travis.yml
```diff
@@ -2,6 +2,7 @@ language: ruby
 
 before_install:
   - gem install bundler -v '< 2'
+  - "find . -type f -iname '*.rb' -exec sed -i '1s/^/# frozen_string_literal: true\\n\\n/' \"{}\" +;"
```
mullermp: This is clever - though we actually want to add Rubocop to our repo and it would be beneficial to put this on every file. Is that something you'd be willing to try?

gems/aws-partitions/lib/aws-partitions/partition.rb
```diff
@@ -21,8 +21,8 @@ def region(region_name)
         if @regions.key?(region_name)
           @regions[region_name]
         else
-          msg = "invalid region name #{region_name.inspect}; valid region "
-          msg << "names include #{@regions.keys.join(', ')}"
+          msg = "invalid region name #{region_name.inspect}; valid region "\
+            "names include #{@regions.keys.join(', ')}"
```
mullermp: As per Rubocop guidance, i believe that these strings should be aligned such as:

```
msg = "stuff"\
      "other stuff"
```

midnight-wonderer: Working on it!

.travis.yml
midnight-wonderer: The original theme of this pull request is "least intrusive".
But if you are OK, after I finished playing with Travis, I can:
- commit what `sed` changes
- remove all the hacks
- Find a way to enable `Style/FrozenStringLiteralComment` and let rubocop do the jobs.

How does it sound?

mullermp: Sounds fine with me! I anticipate a lot of effort, so be aware... this is a big repo.

midnight-wonderer:
@mullermp
I am done with Travis; ruby 1.9 is the tough one.  
Do you want to finish this pull request before the Rubocop, or everything at once?  

At this point, the pull request is still easy to review. The test is working.  
If you decide to merge this chunk first, let me know after the maintainers finished reviewing, I will remove all the hacks that I did before merging.

Currently, no hacks = no CI testing.

mullermp:
We understand the frustration about Ruby 1.9. As a customer, what is your opinion on dropping Ruby 1.9?

We will review this. I need some time to look through it carefully to make sure the logic around the string mutation still works.

midnight-wonderer:
I don't use Ruby 1.9 obviously, nor I know about the statistics, so I can't say anything definite.

However, in my opinion, support to the most recent EOL version is the furthest gem maintainers should aim for; this means, from [this page](https://www.ruby-lang.org/en/downloads/branches/), it is now at version 2.3.  
People continue running an older version, have far more severe issues than unable to update some AWS libraries.

mullermp:
@midnight-wonderer For V3, Ruby 1.9 use is very low. Much higher for V2. I suspect that they are legacy applications. In general I agree with you. We try to support as much as we reasonably can. That said, if I could wave a wand, I'd love to drop anything below 2.3 (can at least use safe navigation in our development too...). I'll continue to push internally for a deprecation policy.

.travis.yml
```diff
@@ -24,7 +28,10 @@ env:
   - PURE_RUBY=1
   - KITCHEN_SINK=1
 
-script: rake build && rake test:spec
+script: >
+  (bundle exec rake build) &&
+  (find . -not -path './vendor/*' -type f -iname '*.rb' -exec sed -i -z '1s/^\(\s*#\s*encoding[^\n]*\n\)\?/\1# frozen_string_literal: true\n/' '{}' +;) &&
```
mullermp: I'm thinking we can leave this out. It was certainly useful for testing! I'd rather see this magic comment committed to the files.

build_tools/spec/region_spec.rb
```diff
@@ -41,7 +42,9 @@ def whitelist
 
           # skip known whitelists
           next if whitelist[key] &&
-                  (whitelist[key][File.basename(path)] == idx ||
+                  (((idx - lines_inserted)..idx).include?(
+                    whitelist[key][File.basename(path)]
+                  ) ||
```
mullermp: Would this change still be necessary? When the files are committed with the magic comment, all of the numbers above will need a +2 (see that you're doing it dynamically here)

gems/aws-eventstream/lib/aws-eventstream/bytes_buffer.rb
```diff
@@ -57,7 +57,7 @@ def tell
       end
 
       def clear!
-        @data = ''
+        @data = ::String.new
```
mullermp: Should these be `String.new`? Unless I'm missing something, I'm not sure why these are prefixed with the empty namespace `::`. There are other places where this is changed too.

gems/aws-eventstream/lib/aws-eventstream/bytes_buffer.rb
```diff
@@ -7,7 +7,7 @@ class BytesBuffer
       # This Util class is for Decoder/Encoder usage only
       # Not for public common bytes buffer usage
       def initialize(data)
-        @data = data
+        @data = data.dup
```
mullermp: I'm assuming this was necessary because we are manipulating `@data` in this file? What would need to be changed so that `dup` isn't necessary? Ditto on the other `dup` usages.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/retry_errors.rb
```diff
@@ -167,8 +167,8 @@ def self.resolve_retry_mode(cfg)
         # Raise if provided value is not one of the retry modes
         if value != 'legacy' && value != 'standard' && value != 'adaptive'
           raise ArgumentError,
-                'Must provide either `legacy`, `standard`, or `adaptive` for '\
-                'retry_mode profile option or for ENV[\'AWS_RETRY_MODE\']'
+            'Must provide either `legacy`, `standard`, or `adaptive` for '\
```
mullermp: `Rubocop` would yell at this change doesn't it?

gems/aws-sdk-core/lib/aws-sdk-core/stubbing/protocols/query.rb
```diff
@@ -20,18 +20,18 @@ def stub_error(error_code)
         private
 
         def build_body(api, operation, data)
-          xml = []
-          builder = Aws::Xml::DocBuilder.new(target: xml, indent: '  ')
+          builder = Aws::Xml::DocBuilder.new(target: [], indent: '  ')
           builder.node(operation.name + 'Response', xmlns: xmlns(api)) do
-            if rules = operation.output
+            rules = operation.output
```
mullermp: BTW - Rubocop is okay with assignments in conditionals as long as they are wrapped in parenthesis - `if (rules = operation.output)`

gems/aws-sdk-core/spec/aws/xml/doc_builder_spec.rb
```diff
@@ -5,22 +5,22 @@ module Aws
...
+      let(:result) { ::String.new }
 
       let(:options) { { target: result, indent: '' } }
 
       let(:xml) { DocBuilder.new(options) }
 
       it 'creates empty xml documents' do
         xml.node('Xml')
-        expect(result).to eq('<Xml/>')
+        expect(xml.target).to eq('<Xml/>')
```
mullermp: Why were these changed?

gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/md5s.rb
```diff
@@ -47,7 +47,9 @@ def md5(value)
           end
 
           def update_in_chunks(digest, io)
-            while chunk = io.read(CHUNK_SIZE, buffer ||= "")
```
mullermp: Minor: why not just change `""` to `String.new`!

tasks/test.rake
```diff
@@ -4,17 +4,21 @@ require 'rspec/core/rake_task'
...
 
+vendor_selector = lambda do |spec_file|
+  spec_file.start_with?('vendor/')
+end
+
 desc 'Executes every spec file, one at a time.'
 task 'test:spec:each' do
   failures = []
-  Dir.glob("**/spec/**/*_spec.rb").each do |spec_file|
+  ::Dir.glob('**/spec/**/*_spec.rb').reject(&vendor_selector).each do |spec_file|
```
mullermp: Ditto on `::`

.travis.yml
```diff
@@ -1,4 +1,8 @@
 language: ruby
+cache: bundler
+
+# to clear bundler cache:
+#   rm -r ./vendor
```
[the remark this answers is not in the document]
midnight-wonderer: I just enabled the cache for bundler. The changes would reduce Travis's worker runtime for about 1 minute per build. (Like 20 minutes per job)
The downside is that the cache itself can be invalid which will cause other unrelated builds including builds for other pull requests to fail.

For example, I once accidentally apply the magic comment to 3rd party gem, which causes all of my builds to fail. The solution is to wipe the cache, and it took some time to figure out where the cache is.
So I leave the comment where I think it is handy. However, if you have a better place or prefer not documenting anything at all, I can make a change with the comment.

.travis.yml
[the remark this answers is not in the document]
midnight-wonderer: I am with you on removing this hack before merging. The hack cause stack traces to shift which would cause confusion for others. But for now, it serves as a makeshift solution, so we can review the changes with better confidence.

build_tools/spec/region_spec.rb
midnight-wonderer: No, it does not.  
This should be removed before merging. It is required for now to make the hack work.

.travis.yml
midnight-wonderer: The hack should be removed before merging IMO. The shift in stack traces would cause confusion.

gems/aws-eventstream/lib/aws-eventstream/bytes_buffer.rb
midnight-wonderer: This is my coding style.  
`::` prefix in ruby is an absolute way to address modules.

`String` can be relative.
~~~ruby
module ModuleA
  class String
  end
  class ClassB
    def initialize
      String.new # would refer to the class we just defined.
      ::String.new # is absolute, in this context it would be the same as ''.dup
    end
  end
end
~~~

In this pull request, I just want to be on the safe side, so I used absolute addressing.  
If this is against your style guides you can request a change.

gems/aws-eventstream/lib/aws-eventstream/bytes_buffer.rb
midnight-wonderer: I will try to figure that out.  
To be honest, the code around buffers is quite messy. I don't understand everything yet.

gems/aws-sdk-core/spec/aws/xml/doc_builder_spec.rb
midnight-wonderer: I tried many things around buffers.  
This is a failed experiment where I attempt to make the string usage of the class immutable.  

Nonetheless, if we use `DocBuilder.new(target: result)` interface, and test the `result` input, it is implied that we wholeheartedly accept mutable strings. That is why I leave the change.

The actions are on me I will investigate these whole buffer and doc builder mess and come up with a better solution/explanation.

gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/md5s.rb
midnight-wonderer: Because the buffer is come up from nowhere and unused anywhere.  
So "The Boy Scout Rule" applied.

tasks/test.rake
midnight-wonderer: Ditto on the explanation.

However, in this place, the `::` is unnecessary; it is just my style.  
I default to absolute addressing and only use relative ones when intentionally wanted to use relative modules.

If you prefer consistency on the module addressing, and relative addressing is your default, I can always change it to be the same. Just ping me with the confirmation.

gems/aws-eventstream/lib/aws-eventstream/bytes_buffer.rb
mullermp: Ah yeah, I see. We generally prefer the relative styling if that's okay.

gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/md5s.rb
mullermp: Makes sense to me! If it's for sure unused, then it's okay to be removed.

mullermp:
Just a friendly ping on this. I will have bandwidth this week to take a look at the fixes.

midnight-wonderer:
Something come up on my end. The PR is not abandoned.

Here the current update:
With further investigation on ByteBuffer, I believe that the original author was too obsessed with Ruby IO interface, and want to implement the interface on internal objects, even if it is just a wrapper around string.

ByteBuffer violates IO interface in the end. Because it expose `bytesize` method, which is not parts of IO interface. And a normal streamable IO object would not be able to implement such method without streaming everything into buffers.
The code around there would be much cleaner with String.

Maybe the IO mocking was put there, despite any advantage gained, in hope that someone in the future can come up with an actual IO stream implementations.

midnight-wonderer:
Here the summarizes what was changed:

After I rebased the branch for my `BytesBuffer` elimination, I found that most of the code can work without mutability modifications. (no `dup` needed)  
If mutated strings are evil, `BytesBuffer` must be the root of all evils. 😀
So I remove most of my changes.

Regarding `Aws::Xml::DocBuilder`, I decided to revert the changes done because there is no way to implements immutability to this class without breaking its public interface.  
`target` is required to implements `<<` method, which implies mutability.  
Some FYI here: if the `target` supplied is a frozen string, the class will breaks.  
This should not be a problem because
* In `Aws` code base the `target` option is always `Array`
* On the consumer side `<<` method implementation implies mutability, if the value supplied breaks the class, it would break a long time ago before my changes.

The changes were simpler than the last you see; I hope this will be easier to review too.  

If this is OK, the only pending operation for me for this pull request is to remove the hacks used for testing.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/retry_errors.rb
midnight-wonderer: I haven't enabled `Rubocop` for myself yet.  
I align strings based on the review comment alone.  
Do you want to change something?

mullermp:
Thanks, we will take a look at this! This is a bigger change to review.

tasks/test.rake
```diff
@@ -4,17 +4,21 @@ require 'rspec/core/rake_task'
 ## spec / unit tests
 ##
 
+vendor_selector = lambda do |spec_file|
+  spec_file.start_with?('vendor/')
```
mullermp, line 8: Is this a temporary change that goes along with the travis change?

.travis.yml
```diff
@@ -1,4 +1,8 @@
 language: ruby
+cache: bundler
```
mullermp, line 2: I think it'd be better to leave this out - caching can be evil sometimes.

gems/aws-sdk-core/lib/aws-sdk-core/xml/doc_builder.rb
```diff
@@ -6,7 +6,7 @@ class DocBuilder
       # @option options [String] :pad ('')
       # @option options [String] :indent ('')
       def initialize(options = {})
-        @target = options[:target] || ''
+        @target = options[:target] || String.new
```
alextwoods, line 9: Why String.new here?  I assume because it needs to be mutable - if so, add a comment.

tasks/test.rake
midnight-wonderer: This is not temporary; it is for Travis cache.

To elaborate on it, when you use
~~~yml
cache: bundler
~~~
on Travis it would install gems inside `vendor/` instead of global system location, and the caching module would pick up files from there.

Since this repository use
~~~Ruby
Dir.glob('**/spec')
~~~
to search for testing, it would pick up 3rd party spec files as well, since the files are now located under `./vendor`.

This filter has been written to allow spec files to exist under `./vendor` without interfering with the actual tests intended.

.travis.yml
midnight-wonderer: Roger that

tasks/test.rake
mullermp: Ah, I see. We should definitely not be running those tests. I would assume they only exist in gems/* and build_tools.

mullermp:
Waiting on @alextwoods to approve and we can merge this. We don't have any releases planned until late next week. This PR might also need CHANGELOG entries on some gems that have behavior changes, so holding off on merging.

mullermp:
We will merge this on Friday - We have some planned updates coming this week and we want to limit any interference.

mullermp:
Thanks for the contribution @midnight-wonderer !

PR #2270, opened 2020-03-19
Default to regional STS endpoints

alextwoods:

### DIFF
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,7 +1,7 @@
 Unreleased Changes
 ------------------
 
-* Feature - Default to regional STS endpoints
+* Issue - Change the default of sts_regional_endpoints from 'legacy' to 'regional'.
 
 3.91.1 (2020-03-10)
 ------------------
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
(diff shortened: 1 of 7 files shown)


gems/aws-sdk-core/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Feature - Default to regional STS endpoints
```
mullermp: I think this can be an issue and not a feature. Also mention what option is changing (`sts_regional_endpoints`)

gems/aws-sdk-core/spec/aws/credential_resolution_chain_spec.rb
```diff
@@ -84,7 +84,7 @@ module Aws
           'AR_TOKEN'
         )
         client = ApiHelper.sample_rest_xml::Client.new(
-          profile: 'ar_web_identity', region: 'us-west-2'
+          profile: 'ar_web_identity', region: 'us-east-1'
```
mullermp: Why `us-east-1` here? Even using us-west-2 and setting regional we should expect us-west-2 in the url.

alextwoods: The point was to have a consistent region across all of the tests to avoid having to mock responses from multiple endpoints - about half of them were us-east-1 and half were us-west-2 so I arbitrarily picked us-east-1, but I can easily change that if you have a preference.

mullermp: No preference! I see what you're doing now.

PR #2274, opened 2020-03-23
Cognito identity credential

### LINKED ISSUE #2267: Aws::CognitoIdentity::Types::Credentials doesn't work with Aws::Sigv4::Signer due to secret_access_key/secret_key attribute name mismatch
  
**Describe the bug**
When using credentials obtained through the [enhanced (simplified) Cognito AuthFlow](https://docs.aws.amazon.com/cognito/latest/developerguide/authentication-flow.html), the Elasticsearch client throws the following error:
```
undefined method `secret_access_key' for #<Aws::CognitoIdentity::Types::Credentials:0x00007fc2b28802f8>
Did you mean?  secret_key
```

aws-sdk-cognitoidentity (1.18.0)
[elasticsearch](https://github.com/elastic/elasticsearch-ruby) (6.8.1)
[faraday_middleware-aws-sigv4](https://github.com/winebarrel/faraday_middleware-aws-sigv4) (0.3.0)

ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-darwin19]

Here's a fragment from my script showing the major calls:

```ruby
# Get/Create & Get Identity ID for this app in our Identity Pool

cognito_client = Aws::CognitoIdentity::Client.new(
  region: @region
)

app_identity = cognito_client.get_id({
  account_id: sandbox_account_id,
  identity_pool_id: identity_pool_id,
  logins: {
    identity_provider_name => token
  }
})

# Use app Identity ID to assume appropriate role based on scopes

assumed_role_session = cognito_client.get_credentials_for_identity(
  identity_id: app_identity.identity_id,
  logins: {
    identity_provider_name => token
  }
)

# Call ES with assumed role credentials

assumed_role_client = ElasticsearchClient.new(
  credentials_provider: assumed_role_session,
  es_endpoint: @es_endpoint
)

assumed_role_results = assumed_role_client.cluster_health
```
And here's the instantiation of the Elasticsearch::Client:

```ruby
def initialize(**options)
    es_endpoint = options.fetch(:es_endpoint) { "https://vpc-workbench-es-sandbox-binwrzkohx6nuhcyetss4jdofe.us-east-1.es.amazonaws.com" }
    service = options.fetch(:service) { "es" }
    region = options.fetch(:region) { "us-east-1" }
    
    credentials_provider = options[:credentials_provider]

    @client = Elasticsearch::Client.new(url: es_endpoint) { |f|
      f.request :aws_sigv4,
        service: service,
        region: region,
        credentials_provider: credentials_provider

      f.adapter Faraday.default_adapter
    }
  end
```

I can extract a relevant example script and put it in a separate repo if needed.

**Expected behavior**
I expected the [Aws::Sigv4::Signer](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/Sigv4/Signer.html#initialize-instance_method) to be able to consume either the `Aws::CognitoIdentity::Types::GetCredentialsForIdentityResponse` object directly as a `:credentials_provider`, or the `Aws::CognitoIdentity::Types::GetCredentialsForIdentityResponse.credentials` object as a `:credentials` object. Both, however, result in the above error.

**Additional context**
This error occurs because the response [Aws::CognitoIdentity::Types::Credentials](https://github.com/aws/aws-sdk-ruby/blob/949907d33b40022d1e1f36d30b424a85ed26b717/gems/aws-sdk-cognitoidentity/lib/aws-sdk-cognitoidentity/types.rb#L183) cognito credentials object exposes the secret access key through an attribute called `secret_key`, rather than `secret_access_key`. The standard for this attribute elsewhere in the SDK (e.g. the top-level [Aws::Credentials](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/Credentials.html) object) seems to be `secret_access_key`, which the [Aws::Sigv4::Signer](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/Sigv4/Signer.html#initialize-instance_method) consumes without issue.

I'm not sure if this is an interface inconsistency or if I'm misusing these objects.

alextwoods:
Closes #2267

## Overview
Adds a new CognitoIdentityCredentials object to support Cognito's [enhanced flow](https://aws.amazon.com/blogs/mobile/understanding-amazon-cognito-authentication-part-4-enhanced-flow/), similar to the existing [AssumeRoleWebIdentityCredentials](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/AssumeRoleWebIdentityCredentials.html).  This is also loosely based on the JS version: [cognito_identity_credentials.js](https://github.com/aws/aws-sdk-js/blob/master/lib/credentials/cognito_identity_credentials.js)

Usage:
```ruby
Aws.config.update(credentials: CognitoIdentityCredentials.new(identity_pool_id: pool_arn, 
    logins: {'graph.facebook.com' => auth_token})

client = Aws::S3::Client.new  # now uses the CognitoIdentityCredentials defined above

creds = CognitoIdentityCredentials.new(identity_pool_id: pool_arn, 
    logins: {'graph.facebook.com' => auth_token},
    tokens_expired: ->(provider) { true if tokens_expired? },
    before_refresh: ->(provider) { provider.logins = new_logins }
)

```

### Contribution


gems/aws-sdk-cognitoidentity/lib/aws-sdk-cognitoidentity/customizations/cognito_identity_credentials.rb
```diff
@@ -0,0 +1,102 @@
...
+module Aws
+  module CognitoIdentity
+    # An auto-refreshing credential provider
+    # TODO: Add good documentation....
+    class CognitoIdentityCredentials
+
+      include CredentialProvider
+      include RefreshingCredentials
+
+      # @option options [String] :identity_id the Cognito identity_id.  Required
```
mullermp: For Yard, I think you need `#param [Hash] options` or something similar above this.

gems/aws-sdk-cognitoidentity/lib/aws-sdk-cognitoidentity/customizations/cognito_identity_credentials.rb
```diff
@@ -0,0 +1,102 @@
...
+      # unless identity_pool_id is given. A unique
+      # identifier in the format REGION:GUID
+      #
+      # @option options [String] :identity_pool_id Required unless identity_id is
+      # provided. An identity pool ID in the format REGION:GUID.
+      #
+      # @option options [Hash<String,String>] :logins A set of optional
+      # name-value pairs that map provider names to provider tokens.
+      # The name-value pair will follow the syntax
+      # "provider_name": "provider_user_identifier".
```
mullermp: Do some general formatting here (and in other places). Indent two spaces for each additional line on the same documentation line. (i.e. two spaces at name-values, etc)

gems/aws-sdk-cognitoidentity/lib/aws-sdk-cognitoidentity/customizations/cognito_identity_credentials.rb
```diff
@@ -0,0 +1,102 @@
...
+      # that do not support role customization.
+      #
+      # @option options [STS::Client] :client
+      def initialize(options = {})
+        @identity_pool_id = options.delete(:identity_pool_id)
+        @identity_id = options.delete(:identity_id)
+        @custom_role_arn = options.delete(:custom_role_arn)
+        @logins = options.delete(:logins) || {}
+
+        unless @identity_pool_id || @identity_id
```
mullermp: Don't add extra validations to an 'unless' because of De Morgan's Law (can have bad side effect).

gems/aws-sdk-cognitoidentity/lib/aws-sdk-cognitoidentity/customizations/cognito_identity_credentials.rb
```diff
@@ -0,0 +1,102 @@
...
+        end
+
+        @client = options[:client] || CognitoIdentity::Client.new(
+          options.merge(credentials: false)
+        )
+        super
+      end
+
+      # TODO: This does not work - new calls refresh
+      def self.from_get_credentials_for_identity_response(resp)
```
mullermp: `from_get..`

gems/aws-sdk-cognitoidentity/lib/aws-sdk-cognitoidentity/customizations/cognito_identity_credentials.rb
```diff
@@ -0,0 +1,102 @@
...
+        refresh!
+      end
+
+      # @return [String]
+      def identity_id
+        return @identity_id if @identity_id
+
+        @identity_id = @client
+                       .get_id(identity_pool_id: @identity_pool_id)
+                       .identity_id
```
mullermp: `@identity_id || @client. .. .. `

gems/aws-sdk-cognitoidentity/lib/aws-sdk-cognitoidentity/customizations/cognito_identity_credentials.rb
```diff
@@ -0,0 +1,102 @@
...
+      end
+
+      private
+
+      def refresh
```
mullermp: Does this have to be public to work? I'm unsure.

gems/aws-sdk-cognitoidentity/lib/aws-sdk-cognitoidentity/plugins/wrap_cognito_credentials.rb
```diff
@@ -0,0 +1,52 @@
...
+appropriate for us in other clients.
+        DOCS
+
+        # @api private
+        class Handler < Seahorse::Client::Handler
+
+          def call(context)
+            response = @handler.call(context)
+            params = context.params
+            # TODO: This creates an infite loop
```
mullermp: Let's leave the plugin out of the PR for now..

gems/aws-sdk-cognitoidentity/lib/aws-sdk-cognitoidentity/customizations/cognito_identity_credentials.rb
alextwoods: It should be private - users of the class should only ever call the refresh! method define in the RefreshingCredentials which wraps refresh in a mutex.synchronize.

gems/aws-sdk-cognitoidentity/lib/aws-sdk-cognitoidentity/plugins/wrap_cognito_credentials.rb
alextwoods: Yeah - agreed for now.  Removing it.

gems/aws-sdk-cognitoidentity/lib/aws-sdk-cognitoidentity/customizations/cognito_identity_credentials.rb
```diff
@@ -0,0 +1,100 @@
...
+        @credentials = Credentials.new(
+          resp.credentials.access_key_id,
+          resp.credentials.secret_key,
+          resp.credentials.session_token
+        )
+        @expiration = resp.credentials.expiration
+      end
+
+      def near_expiration?
+        (@tokens_expired && @tokens_expired.call(self)) ||
```
gradybarrett: The way I see it there's a one-way dependency here. Expired tokens affect the ability to get a new set of credentials, but don't affect the usability of credentials.

The concern I have with @tokens_expired is that it couples the token refresh with the creds refresh, forcing consumers to consider both when supplying this proc. I can see that causing confusion and unintended side-effects. That coupling would make sense if the token expiration was tied to the credential's expiration, but that's not the case.

It might be simpler and more flexible to just have @before_refresh as you have it above. It's executed every time the refresh is called, which is fine because the app can then choose how efficient/intelligent they want to be about refreshing their token within the supplied proc's logic.

alextwoods: The case I was trying to address with this was a change in credentials - eg from the unauthenticated to authenticated state.  That would necessitate updating the token in the logins and refreshing the credentials.  Otherwise, you are correct that its a one way dependency and token refresh shouldn't be coupled with credentials refresh.  

I had originally wanted to handle the chase of changed credentials through something that would trigger a refresh of credentials when logins was updated.  Something like:
```ruby
       def logins=(logins)
        @logins = logins
        refresh!
      end
```

However, this doesn't handle the case of `credentials.logins['graph.facebook.com'] = new_token` and I couldn't easily find a way to support that aside from either 1) freezing the logins hash and requiring updates through `credentials.logins = {'graph.facebook.com'=>new_token}` or 2) having logins extend hash and include custom set logic w. a callback. 

Anyway - I do agree with you and will remove the `tokens_expired` and think about a solution for updates to logins.

gradybarrett: Oh, I see. My assumption was that updating a particular login's token would be done by the consumer inside the @before_refresh proc. Assuming of course that, like you have it, the CognitoIdentityCredentials object is available within that context so that people can modify the `logins` attr directly.

If you can find an elegant way to help facilitate that I'm game, but I think leaving it all to the user is totally acceptable.

Alternatively, if instead of, or in addition to, the generic @before_filter, there was a token-specific proc parameter with expected return value(s) - probably the entire logins hash - could you just set logins based on that proc's return value directly in the `refresh` function? As a bonus, if you could find a way to allow users to modify logins, without leaking other CognitoIdentityCredentials details that might be good. Maybe that's not worth the trouble though...

alextwoods: After having thought through this a little bit more I've settled for keeping the fairly generic before_refresh proc as is that has the CognitoIdentityCredentials object available so that logins can be modified directly and in whatever way the user needs.  I've updated the documentation to reflect this and cover the case of how to update logins directly.

gradybarrett: Fantastic. I'm always a fan of simplicity.

gems/aws-sdk-cognitoidentity/lib/aws-sdk-cognitoidentity/customizations.rb
```diff
@@ -1,7 +1 @@
-# WARNING ABOUT GENERATED CODE
-#
-# This file is generated. See the contributing for info on making contributions:
-# https://github.com/aws/aws-sdk-ruby/blob/master/CONTRIBUTING.md
-#
-# WARNING ABOUT GENERATED CODE
-
+require 'aws-sdk-cognitoidentity/customizations/cognito_identity_credentials'
```
mullermp: Our travis tests run `rake build` but be sure that this file isn't overwritten (generate code warning has me nervous).

gems/aws-sdk-cognitoidentity/lib/aws-sdk-cognitoidentity/customizations/cognito_identity_credentials.rb
```diff
@@ -0,0 +1,132 @@
...
+    #
+    # This provider gets credentials using the
+    # {Client.get_credentials_for_identity} service operation, which
+    # requires either an `identity_id` or an `identity_pool_id` (Amazon Cognito
+    # Identity Pool ID), which is used to call {Client.get_id} to
+    # obtain an `identity_id` automatically.
+    #
+    # In addition, if this credential provider is used to provide authenticated
+    # login, the `logins` map may be set to the tokens provided by the
+    # respective identity providers. See {constructor} for an example on
```
mullermp: Is this `{constructor}` tag linking correctly? I think in our Yard code we expect Constructor but do double check.

gems/aws-sdk-cognitoidentity/lib/aws-sdk-cognitoidentity/customizations/cognito_identity_credentials.rb
```diff
@@ -0,0 +1,132 @@
...
+      # @option options [String] :identity_pool_id Required unless identity_id
+      #   is provided. A Amazon Cognito
+      #   Identity Pool ID)in the format REGION:GUID.
+      #
+      # @option options [Hash<String,String>] :logins A set of optional
+      #   name-value pairs that map provider names to provider tokens.
+      #   The name-value pair will follow the syntax
+      #   "provider_name": "provider_user_identifier".
+      #
+      # @option options [String] :custom_role_arn (optional) The Amazon Resource
```
mullermp: in Yard, `(optional)` will actually lay out a default of "optional". If you want to say something is required, I believe it's `[required, String]` Otherwise, if not required, it's assumed optional.

gems/aws-sdk-cognitoidentity/lib/aws-sdk-cognitoidentity/customizations/cognito_identity_credentials.rb
```diff
@@ -0,0 +1,132 @@
...
+    # obtain an `identity_id` automatically.
+    #
+    # In addition, if this credential provider is used to provide authenticated
+    # login, the `logins` map may be set to the tokens provided by the
+    # respective identity providers. See {constructor} for an example on
+    # creating a credentials object with proper property values.
+    #
+    # ## Refreshing Credentials from Identity Service
+    #
+    # the CognitoIdentityCredentials will auto-refresh the AWS credentials from
```
mullermp: Minor: capital T here - also line 30 remove the space next to "supported".

gems/aws-sdk-cognitoidentity/lib/aws-sdk-cognitoidentity/customizations/cognito_identity_credentials.rb
```diff
@@ -0,0 +1,132 @@
...
+
+      # @return [CognitoIdentity::Client]
+      attr_reader :client
+
+      # @return [Hash<String,String>]
+      attr_accessor :logins
+
+      # @return [String]
+      def identity_id
+        @identity_id ||= @client
```
mullermp: Nit: formatting

gems/aws-sdk-cognitoidentity/spec/cognito_identity_credentials_spec.rb
```diff
@@ -0,0 +1,126 @@
+
+require_relative 'spec_helper'
+
+module Aws
+  module CognitoIdentity
+    describe CognitoIdentityCredentials do
+
+      let(:client) {
```
mullermp: Nit: do .. end

gems/aws-sdk-cognitoidentity/spec/cognito_identity_credentials_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+          region: 'us-west-2',
+          stub_responses: true
+        )
+      }
+
+      let(:in_one_hour) { Time.now + 60 * 60 }
+
+      let(:expiration) { in_one_hour }
+
+      let(:cognito_creds) {
```
mullermp: Nit: do .. end

gems/aws-sdk-cognitoidentity/spec/cognito_identity_credentials_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+        double('credentials',
```
mullermp: Nit:

```
double(
   'thing',
)
```

gems/aws-sdk-cognitoidentity/spec/cognito_identity_credentials_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+
+      let(:identity_id) { 'identity_id' }
+      let(:identity_pool_id) { 'pool_id' }
+
+      let(:resp) { double('client-resp', credentials: cognito_creds) }
+
+      describe '#initialize' do
+
+        it 'constructs a client with passed arguments when not given' do
+          expect(CognitoIdentity::Client).to receive(:new)
```
mullermp: By the way, no need to mock clients. I think in many of our tests we actually just make them.

gems/aws-sdk-cognitoidentity/spec/cognito_identity_credentials_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+          expect(creds.client).to be(client)
+        end
+
+        it 'raises an argument error when identity_pool_id and identity_id are missing' do
+          expect { CognitoIdentityCredentials.new }.to raise_error(ArgumentError)
+        end
+
+        it 'sets the client when passed in and does not create a new one' do
+          test_client = client # force construction
+          expect(CognitoIdentity::Client).not_to receive(:new)
```
mullermp: Ditto -- you can always do `expect(creds.client).to be(client)` (where `be` will actually test that the object is the same)

gems/aws-sdk-cognitoidentity/lib/aws-sdk-cognitoidentity/customizations.rb
alextwoods: Yeah - I was nervous about that.  It doesn't over write (note: other gem customization files do not have the generated src warning).

gems/aws-sdk-cognitoidentity/lib/aws-sdk-cognitoidentity/customizations/cognito_identity_credentials.rb
alextwoods: This should actually be {#initialize}.  Updated.

gems/aws-sdk-cognitoidentity/spec/cognito_identity_credentials_spec.rb
alextwoods: Here I was testing explicitly for new being called on the client.

PR #2277, opened 2020-04-01
Remove string mutation in EventStream::BytesBuffer

midnight-wonderer:
This is a sub pull request related to https://github.com/aws/aws-sdk-ruby/pull/2269.  
In the pull request, @mullermp asked, can we remove weird `dup` from the code and implement the functionality without mutating a string?

After taking a close inspection. I noticed that this whole `BytesBuffer` thing does not make sense.  
`BytesBuffer` is essentially a class implementing [Ruby's IO interface](https://ruby-doc.org/core-2.7.0/IO.html) but with `bytesize` method added.

The IO implemented in `BytesBuffer` is not streamable, because it needs to know `bytesize`, which implied the whole data have to be buffered in memory to operate correctly.  
Standard `String` class would do the exact same thing, with way better readability on the consumer side.

The old implementation use `IO.read` everywhere instead of just using `String`, which adds unnecessary complexity. And AFAIK, not apparent benefits gained.


mullermp:
Thanks for this.

Somehow Travis isn't showing? I think your changes would break for Ruby 1.9.

mullermp:
I think I fixed travis, next time you make a commit it should show up.

Current build: https://travis-ci.org/github/aws/aws-sdk-ruby/builds/669543679

midnight-wonderer:
There is a decision to make here.  
The reason why I touch and change the syntax there is because I ran it in Ruby 2.7, and the syntax that consumers of the class use are now deprecated in preparation for Ruby 3.0.  
In Ruby 2.7 there are warnings showing up when you ran the code.

More info here:  
https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

Either syntax has an issue of its own.  
In this pull request, do we still want to support Ruby 1.9?  
(The answer is pretty much guessable, but I want a confirmation anyway, just to serve as a feedback for dropping 1.9)

mullermp:
I (personally) would like to drop Ruby 1.9 but we can't do so just yet.

midnight-wonderer:
As expected, but at least the dilemma has been recorded here.  
  
Here you go.  
The breaking change has been reverted.

mullermp:
Thanks. Taking a look. Just trying to be sure this isn't breaking at all.

gems/aws-eventstream/lib/aws-eventstream/decoder.rb
```diff
@@ -63,26 +63,31 @@ class Decoder
...
 
       # bytes of prelude part, including 4 bytes of
       # total message length, headers length and crc checksum of prelude
       PRELUDE_LENGTH = 12
+      private_constant :PRELUDE_LENGTH
 
       # bytes of total overhead in a message, including prelude
       # and 4 bytes total message crc checksum
-      OVERHEAD_LENGTH = 16
+      CRC32_LENGTH = 4
```
alextwoods, line 75: Is this change needed?  I'd be a little worried about a user having depended on the OVERHEAD_LENGTH constant.

gems/aws-eventstream/lib/aws-eventstream/decoder.rb
```diff
@@ -63,26 +63,31 @@ class Decoder
       include Enumerable
 
       ONE_MEGABYTE = 1024 * 1024
+      private_constant :ONE_MEGABYTE
 
       # bytes of prelude part, including 4 bytes of
       # total message length, headers length and crc checksum of prelude
       PRELUDE_LENGTH = 12
+      private_constant :PRELUDE_LENGTH
```
alextwoods, line 71: I like the private_constant....we should have done these before.  However, since this was previously exposed to customers and this class isn't listed as api private we should leave these constants exposed unfortunately.

gems/aws-eventstream/lib/aws-eventstream/decoder.rb
```diff
@@ -63,26 +63,31 @@ class Decoder
...
       def initialize(options = {})
         @format = options.fetch(:format, true)
-        @message_buffer = BytesBuffer.new('')
+        @message_buffer = ''
       end
 
-      # @return [BytesBuffer]
-      attr_reader :message_buffer
+      # exposed for testing
+      def _message_buffer
```
alextwoods, line 88: Similar comment here - leaves this attr_reader :message_buffer and change the documentation of return type.

Since this was exposed previously and isn't api private, we should keep it exposed.  It is changing return types so could still be breaking, but string should expose (mostly) the same methods as the ByteBuffer.

midnight-wonderer:
@alextwoods I will work on the change request.  
However, let me explain my rationale on the matters.

The changes related to constant were made on the `Decoder` class.  
Which roughly works like this:
Input --> Decoder --> Output

* The information the constants provided contain inside the `Input`.  
* The result provided by the execution of this class does not contain the information.  
* The conclusion, only possible use cases for the constants are for use internally in the class itself.

The same things cannot be said about the `Encoder` class.

As for the `message_buffer`, while the method itself does not mark with `API private`, the return type does.  
If we want to be truly backward compatible, we probably don't change the return type.  
Also, the method does not name properly for consumers.  
The only reason the method is there is to serve as a shortcut for testing.

My plan for the substituted type would be `StringIO`, by the way.

If you find my rationale compelling, even after I made the changes requested, I would happily revert the changes back.

alextwoods:
Apologies for all the back and forth on this and thanks so much for making those changes!

I agree with your logic on removing those from the interface of the Decoder so I'm happy for you to revert your most recent change. However, rather than exposing the `_message_buffer` method for testing, you can make it private and in the specs use send(:message_buffer).

mullermp:
Thanks for updating. Will take a look at this today.

alextwoods:
Thanks for updating those - looks good!  I've added changelog entries for these (which are required by our build system) and approved.

mullermp:
Merged. :) Thanks for the contribution. Looking forward to progress on the other PR!

PR #2282, opened 2020-04-08
Fix for S3 error responses with 200 OK and empty body

alextwoods:


gems/aws-sdk-s3/spec/client_spec.rb
```diff
@@ -714,30 +714,52 @@ module S3
...
+              context.http_response.signal_headers(200, {})
+              context.http_response.signal_data("\r\n")
+              context.http_response.signal_done
+              Seahorse::Client::Response.new(context: context)
+            end
+            resp = client.send(operation_name, {
+              bucket: 'bucket',
+              key: 'key'
+            }.merge(params))
+            expect(resp.error).not_to be_nil
```
trevorrowe, line 759: I feel like we should assert here more than not null. I would probably assert which error is being raised in case we have an error parsing which is unexpected.

gems/aws-sdk-s3/spec/client_spec.rb
```diff
@@ -714,30 +714,52 @@ module S3
...
+            expect(resp.data).to be(nil)
+          end
+
+          it 'handles 200 http response with empty body as error' do
+            client.handlers.remove(
+              Seahorse::Client::Plugins::RaiseResponseErrors::Handler
+            )
+            client.handle(step: :send) do |context|
+              context.http_response.signal_headers(200, {})
+              context.http_response.signal_data("\r\n")
```
trevorrowe, line 751: Did you confirm how S3 sends whitespace while computing the results? I would like to ensure the singaled data matches what the service would send.

mullermp: Talked offline - yes it sends the xml preamble, followed by \r\n as well as other junk, then finally the result structure.

gems/aws-sdk-core/lib/aws-sdk-core/xml/error_handler.rb
```diff
@@ -6,7 +6,7 @@ class ErrorHandler < Seahorse::Client::Handler
 
       def call(context)
         @handler.call(context).on(300..599) do |response|
-          response.error = error(context)
+          response.error = error(context) unless response.error
```
mullermp, line 9: Because this is on 300...599 would this case ever happen? I thought because errors are for 200..299?

gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/http_200_errors.rb
```diff
@@ -27,12 +27,25 @@ def check_for_error(context)
...
               S3::Errors.error_class(error_code).new(context, error_message)
-            elsif xml.strip == ''
-              Seahorse::Client::NetworkingError.new(
-                S3::Errors
+            elsif incomplete_response?(xml, context)
+              S3::Errors
                   .error_class('InternalError')
-                  .new(context, 'Empty response body')
-              )
+                  .new(context, 'Incomplete or empty response body')
```
mullermp: Will this be retryable? It looks to be.

gems/aws-sdk-s3/spec/client_spec.rb
alextwoods: Updated to be the specific error.

gems/aws-sdk-core/lib/aws-sdk-core/xml/error_handler.rb
alextwoods: This handler happens after the Http200Errors handler, which in the case of an empty body sets the status to 500.  The <Error> case in the body was actually working because this plugin creates the error from the code, but in the case of no body, it was overwriting the error with a generic error (which causes the spec to fail).

gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/http_200_errors.rb
alextwoods: Yeah - its retryable.  The spec checks that gets retried.

mullermp:
LGTM

gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/http_200_errors.rb
```diff
@@ -27,6 +27,27 @@ def check_for_error(context)
...
+                  .new(context, 'Empty response body')
+              )
+            end
+          end
+
+          private
+
+          # An S3 response is considered complete only if all output shapes
+          # expected in the body are present
+          def incomplete_response?(xml, context)
```
trevorrowe, line 43: I don't believe it is necessary to search the body for each of the expected top-level members of the document. If the XML document is opened but not closed correctly it would fail to parse. A minor simplification might be to check only for the document closing element.

alextwoods: I've updated the PR with a similar approach - checking for the start of a valid xml tag.  If an xml tag is started but the body is incomplete, all of the XML engines will raise an xml ParsingError (except Oga, in the case of an incomplete closing tag.  see: https://gitlab.com/yorickpeterse/oga/-/issues/203).

Issue #2281, opened 2020-04-07
Provide better diagnostics when empty credentials are provided


p-mongo:
**Is your feature request related to a problem? Please describe.**

If a Client is created with empty credentials and subsequently used for an operation, the error produced is: `The security token included in the request is invalid. (Aws::STS::Errors::InvalidClientTokenId)`.

For example:

<pre>
credentials = Aws::Credentials.new('', '')
client = Aws::STS::Client.new(credentials: credentials, region: 'us-east-2')
p client.get_caller_identity
</pre>

The error message produced does not indicate the problem (empty credentials). It is actually misleading because it implies that some piece of information was supplied ("security token", which is not a terribly well defined term either) and was incorrect, but the user in this case did *not* provide any information.

To provide helpful diagnostics to users of my applications, I need to check that credentials are not empty.

**Describe the solution you'd like**

I would like the various AWS SDK clients to check whether the credentials are empty, and produce an error with a message stating so. If credentials are not changeable on an existing client, I would like this diagnostics to happen upon client instantiation, otherwise whenever a signature is calculated.

**Describe alternatives you've considered**

A client application can perform this validation. But this means every client application needs to perform it whenever user credentials are obtained, or whenever any Client is constructed.

[ ] :wave: I may be able to implement this feature request

**Additional context**

Reproduce code: https://github.com/p-mongo/tests/tree/master/aws-empty-credentials

mullermp:
Thanks for the feature request. In general, the SDK prefers validation as late as possible (such as during signature signing). There are also many edge cases here for customers using other credential providers such as `AssumeRoleCredentials` and `InstanceProfileCredentials` where some service is being called before a `Credentials` object is created. I think also this might be a breaking change for users to raise errors earlier than they have expected? We will definitely consider this one, and I'm happy to look at any pull requests.

p-mongo:
Between the contents of the message and when it is produced, to me the contents is much more important, i.e. having the signer fail with an error like "Credentials were not provided". I understand that later diagnostics can be desirable in various situations.

PR #2288, opened 2020-04-15
Throw a MissingCredentialsError when credentials are empty/blank

alextwoods:
Fixes: #2281

### DIFF
--- a/gems/aws-sigv4/lib/aws-sigv4/signer.rb
+++ b/gems/aws-sigv4/lib/aws-sigv4/signer.rb
@@ -654,9 +654,7 @@ module Aws
 
       def fetch_credentials
         credentials = @credentials_provider.credentials
-        # ensure credentials are not blank
-        if credentials.access_key_id =~/[^[:space:]]/ &&
-           credentials.secret_access_key =~ /[^[:space:]]/
+        if credentials.set?
           credentials
         else
           raise Errors::MissingCredentialsError,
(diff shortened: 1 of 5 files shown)


gems/aws-sigv4/lib/aws-sigv4/signer.rb
```diff
@@ -654,7 +654,9 @@ def uri_escape_path(string)
 
       def fetch_credentials
         credentials = @credentials_provider.credentials
-        if credentials.access_key_id && credentials.secret_access_key
+        # ensure credentials are not blank
```
mullermp, line 657: Why not `credentials.set?` If you can't use `.set?` I think regex is a bad idea.. would prefer `nil?` and `empty?`

alextwoods: yeah - I was trying to cover the case of blank (eg: '  '), but I agree that's not necessary.  Using `credentials#set?` method now which should work for all credentials/credential providers.

alextwoods:
Ive moved the test out of the S3 client and into core (and added a test both for nil and empty).

mullermp:
Verify that instance profile credentials (and others) still work in these cases.

alextwoods:
The `#set?` method works on all of the credential providers (including InstanceProfileCredentials, ProcessCredentials, ECSCredentials, and the new CognitoIdentityCredentials, ect) - all of these are CredentialProviders that return return `Aws::Credentials`.

alextwoods:
I don't believe we need a changelog for core - I have added specs that test cilent, but I think only the sigv4 gem needs an update.

PR #2289, opened 2020-04-15
Add presigned request; force virtual_host to https

mullermp:
Adds a presigned request method that returns the headers needed for the presigned URL rather than hoisting them onto the query string.

~~Opens up the presigned request and presigned url methods to be used with non-object APIs such as `list_buckets`.~~

Forces virtual_host to use HTTPS unless configured not to.


gems/aws-sdk-s3/spec/presigner_spec.rb
```diff
@@ -162,14 +144,161 @@ module S3
...
+        it 'hoists x-amz-* headers to the query string' do
+          signer = Presigner.new(client: client)
+          url = signer.presigned_url(
+            :put_object, bucket: 'aws-sdk', key: 'foo', acl: 'public-read'
+          )
+          expect(url).to match(/x-amz-acl=public-read/)
+        end
+      end
+
+      describe '#presigned_request' do
```
mullermp, line 173: These tests are exactly the same as presigned_url except for the last. Shared examples is difficult here because of the return types.

alextwoods: Yeah, it is a lot of duplication.  I think we get could shared examples working w/ the different return types using some sort of wrapper... but I dont think its really worth it here - as is, the tests are fairly readable.

gems/aws-sdk-s3/lib/aws-sdk-s3/presigner.rb
```diff
@@ -138,23 +198,26 @@ def sign_but_dont_send(req, expires_in, scheme, time, unsigned_headers)
...
-          # hoist x-amz-* headers to the querystring
           query = http_req.endpoint.query ? http_req.endpoint.query.split('&') : []
-          http_req.headers.keys.each do |key|
-            if key.match(/^x-amz/i)
-              value = Aws::Sigv4::Signer.uri_escape(http_req.headers.delete(key))
-              key = Aws::Sigv4::Signer.uri_escape(key)
+          http_req.headers.each do |key, value|
+            next unless key =~ /^x-amz/i
+
+            value = Aws::Sigv4::Signer.uri_escape(value)
```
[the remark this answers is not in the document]
mullermp: That makes sense.

mullermp:
I wonder if it makes sense to deprecate presigned_url in favor of presigned_request in a new major version.

PR #2295, opened 2020-05-05
Taking AWS_REGION env as default region

### LINKED ISSUE #2112: Feature request: pre-populate Aws.config from environment
### Issue description

Would it be feasible to pre-populate `Aws.config` from the runtime environment?  For example, if `ENV["AWS_REGION"]` is set (and non-empty), then by default `Aws.config[:region] == ENV["AWS_REGION"]`.  Likewise for any settings picked up from "~/.aws" or elsewhere.

This would enable in-app configuration code like the following:
```ruby
Aws.config[:region] ||= "us-east-1"
```

Where a reasonable default is provided, but can overridden from the outside, without code modification.

Additionally, it may be useful when one wants to test what config values are being picked up from the environment, for example, when debugging.

`aws-sdk`

any

N/A

sivagollapalli:
FIX #2112 issue

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core.rb
@@ -110,6 +110,7 @@ module Aws
     def config=(config)
       if Hash === config
         @config = config
+        Aws.config.update({region: ENV['AWS_REGION']}) unless Aws.config.has_key?(:region)
       else
         raise ArgumentError, 'configuration object must be a hash'
       end


alextwoods:
The SDK will already look for region (and other config values) in the ENV before other sources (See this [dev guide](https://docs.aws.amazon.com/sdk-for-ruby/v3/developer-guide/setup-config.html#aws-ruby-sdk-setting-credentials) for an overview of the order of preference or to see the code itself: [CredentialProviderChain](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/lib/aws-sdk-core/credential_provider_chain.rb)).

mullermp:
This doesn't fix the issue unfortunately. :( Thanks for opening the PR.

sivagollapalli:
@mullermp No problem. I will try to fix.

alextwoods:
@sivagollapalli - I don't think we should be changing the default behavior of Aws.config here (without a major version bump) as this could break current users.  Aws.config is easy to update in application code, so users are not blocked from doing this themselves if its behavior they want.

sivagollapalli:
@alextwoods Would that mean are we discarding this issue #2112 or we are postponing till next major release?

PR #2300, opened 2020-05-08
Fix CloudFront signer URL spliting

palkan:
Fix URL signing when a url contains `://` in its path (which is valid for URLs).

### DIFF
--- a/gems/aws-sdk-cloudfront/CHANGELOG.md
+++ b/gems/aws-sdk-cloudfront/CHANGELOG.md
@@ -1,7 +1,7 @@
 Unreleased Changes
 ------------------
 
-* Issue - Split on first instance of `://` in `Signer` to handle 3rd party s3:// urls.
+* Issue - Split on first instance of `://` in `Aws::CloudFront::Signer` to handle 3rd party s3:// urls.
 
 1.27.0 (2020-05-07)
 ------------------
--- a/gems/aws-sdk-cloudfront/CHANGELOG.md
+++ b/gems/aws-sdk-cloudfront/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Split on first instance of `://` in `Signer` to handle 3rd party s3:// urls.
+
 1.27.0 (2020-05-07)
 ------------------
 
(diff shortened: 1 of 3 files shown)


mullermp:
Thanks for the code contribution. I'm not against this change, but is this a default in cloud front to put s3:// links as part of the URL? This does not seem right..

palkan:
The example use-case could be found in [Imgproxy](https://imgproxy.net) server: it could serve files [directly from S3](https://docs.imgproxy.net/#/serving_files_from_s3), and uses `s3://` for that as a part of the query.

mullermp:
The example on https://imgproxy.net/ looks like it's HTML escaping the nasa.gov link:

```
https:%2F%2Fwww.nasa.gov%2Fsites%2Fdefault%2Ffiles%2Fthumbnails%2Fimage%2Fpia22228.jpg
```

What does the URL look like for an S3 url? I assume it should be escaping `s3://` the same way and if not then it sounds like unintended behavior on their end?

palkan:
> I assume it should be escaping s3:// the same way and if not then it sounds like unintended behavior on their end?

That makes sense to me. Though the `imgproxy` gem doesn't do that. Maybe, just a bug) Though it works fine without escaping, I only encountered this issue when started using signed urls.

Anyway, although common sense tells that urls do not have `://` in their path, they can, and sometimes that works 🙂

mullermp:
Could you open an issue on their repo and reference this issue first?

mullermp:
I need to check if cloudfront actually allows these kinds of urls..

palkan:
> Could you open an issue on their repo and reference this issue first?

Done: https://github.com/imgproxy/imgproxy.rb/issues/48

mullermp:
Added some correspondence on that issue. Still trying to check with cloudfront on our side to see if this change is fine.

ajredniwja:
Checked with the service team, these kind of URLs are not supported by CloudFront.

mullermp:
Closing because even merging this change will not work with cloud front. Image proxy service will need to fix on their end if you wish to continue using it.

palkan:
> even merging this change will not work with cloud front

That's not true. Everything works as expected with this patch. If I haven't checked it in the real application I wouldn't propose a fix.

**That's a bug**.

mullermp:
@ajredniwja can we get reconfirmation from the service team over these types of urls?

mullermp:
It doesn't make sense that someone on cloudfront would say they don't work, but I'm inclined to believe you. Regarding the change itself, it's pretty safe, so I'm just going to merge it. I'm sorry for the back and forth. On imgproxy's end, I do think the S3 url should be encoded.

mullermp:
I added a changelog entry so that this will get released today. Thanks for the contribution.

PR #2305, opened 2020-05-13
[#2304] Raise argument error if parameter is empty

### LINKED ISSUE #2304: Aws::S3::Client#delete_object deletes the bucket when key is set to empty string

```
aws-sdk-s3 (1.61.1)
```  

**Describe the bug**
The documentation for `Aws::S3::Client#delete_object` specifies that [`:key` is a required parameter](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/Client.html#delete_object-instance_method). However, if the `key` is an empty string (`""`) but the `bucket` name is provided, parameter validation passes and the client attempts to delete the bucket.

In my case, this occurred because the key was derived from a response from an external service. The field from which the key was derived happened to be an empty key in the response. Fortunately, the bucket was _not_ empty and an `Aws::S3::Errors::BucketNotEmpty` error was raised. However, a bucket that happened to be empty would have been inadvertently deleted.

```
ruby 2.4.3p205 (2017-12-14 revision 61247) [x86_64-linux]
```

```ruby
client = AWS::S3::Client.new
client.delete_object(bucket: 'my-bucket', key: '')
#=> Aws::S3::Errors::BucketNotEmpty: The bucket you tried to delete is not empty
```

**Expected behavior**
`Aws::S3::Client#delete_object` should not attempt to the delete the bucket when the `key` is set to an empty string (`""`).

The following is what happens if you specify a `bucket` but pass a `key` that is `nil`:
```ruby
client = AWS::S3::Client.new
client.delete_object(bucket: 'my-bucket', key: nil)
#=> ArgumentError: missing required parameter params[:key]
```
Perhaps a similar parameter validation error could be raised when the key is an empty string.

**Additional context**
I realize there's some ambiguity as to what the behavior really _should_ be here based on how one interprets the word "required" in the docs. If nothing else, maybe the documentation could be updated to be clear that this could occur. That said, considering the alternative is accidentally deleting a bucket (even an empty one), I think raising an `ArgumentError` might be more appropriate.

sivagollapalli:
Fix #2304 

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core/param_validator.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/param_validator.rb
@@ -61,7 +61,7 @@ module Aws
         if @validate_required
           shape.required.each do |member_name|
             input_eventstream = ref.shape.member(member_name).eventstream && @input
-            empty_param = values[member_name].blank? if values[member_name].nil?
+            empty_param = values[member_name].to_s.empty? if values[member_name].nil?
             if empty_param && !input_eventstream
               param = "#{context}[#{member_name.inspect}]"
               errors << "missing required parameter #{param}"
--- a/gems/aws-sdk-core/lib/aws-sdk-core/param_validator.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/param_validator.rb
@@ -61,7 +61,8 @@ module Aws
         if @validate_required
           shape.required.each do |member_name|
             input_eventstream = ref.shape.member(member_name).eventstream && @input
-            if (values[member_name].nil? || values[member_name].empty?) && !input_eventstream
+            empty_param = values[member_name].blank? if values[member_name].nil?
+            if empty_param && !input_eventstream
               param = "#{context}[#{member_name.inspect}]"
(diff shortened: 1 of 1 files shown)


mullermp:
I don't know if this is the right fix. See comment [here](https://github.com/aws/aws-sdk-ruby/issues/2304#issuecomment-628700423). There may be clients and operations outside of S3 that allow for required parameters to be blank strings. I'm not sure how to check for that. This problem is limited to REST XML (`DELETE /` versus `DELETE /object`). I think some validation is needed somewhere in the XML classes and not on the param validator.

mullermp:
Thanks for giving this a try. Our team discussed this and we're going to move forward with fixing this at the request level and not the param validation level. This is similar to how the javascript SDK fixes the issue (https://github.com/aws/aws-sdk-js/commit/940c56b1cf03aaaa13bd85e0c8e07f885e1fbc9d) specifically the bit about `if (shape['location'] === 'uri')` which checks only require parameters part of the URI.

Closing in favor of https://github.com/aws/aws-sdk-ruby/pull/2306/.

PR #2306, opened 2020-05-14
Require rest endpoint params

### LINKED ISSUE #2304: Aws::S3::Client#delete_object deletes the bucket when key is set to empty string

mullermp:
Fixes #2304

Also allows for s3 presigner to presign list_objects.. can maybe lump that change in.

Specs will need fixing.


mullermp:
```
Aws> s3.delete_object(bucket: 'mamuller-us-west-2', key: '')
ArgumentError: :key must not be blank
from /Users/mamuller/workplace/aws-sdk-ruby/gems/aws-sdk-core/lib/aws-sdk-core/rest/request/endpoint.rb:40:in `param_value_for_placeholder'
```

gems/aws-sdk-core/spec/aws/xml/error_handler_spec.rb
```diff
@@ -135,10 +135,10 @@ module Xml
       end
 
       it 'handles route53 error format with message' do
-        stub_request(:post, 'https://route53.amazonaws.com/2013-04-01/hostedzone//rrset/').
+        stub_request(:post, 'https://route53.amazonaws.com/2013-04-01/hostedzone/ResourceId/rrset/').
           to_return(:status => 400, :body => route53_error)
         expect {
-          route53.change_resource_record_sets
+          route53.change_resource_record_sets(hosted_zone_id: "ResourceId")
```
alextwoods, line 141: My fault on this - but use single quotes.

PR #2307, opened 2020-05-14
Fix printing of exception in a thread in passing spec

alextwoods:
Ruby 2.5 added a default value for Thread.report_on_exception to true (https://blog.bigbinary.com/2018/04/18/ruby-2-5-enables-thread-report_on_exception-by-default.html). 
its not supported before ruby 2.4.  And ruby 2.4 had a default value of false, which was changed to true in 2.5+

Alternative:
Use Rspec around hooks (eg: https://relishapp.com/rspec/rspec-core/v/3-4/docs/hooks/around-hooks#define-a-global-%60around%60-hook)

### DIFF
--- a/gems/aws-sdk-core/spec/shared_spec_helper.rb
+++ b/gems/aws-sdk-core/spec/shared_spec_helper.rb
@@ -28,6 +28,17 @@ RSpec.configure do |config|
     stub_request(:put, "http://169.254.169.254#{token_path}").to_raise(SocketError)
 
     Aws.shared_config.fresh
-
   end
+end
+
+def without_thread_report_on_exception
+  if Thread.respond_to?(:report_on_exception)
+    current_value = Thread.report_on_exception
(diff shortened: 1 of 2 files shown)


gems/aws-sdk-core/spec/shared_spec_helper.rb
```diff
@@ -28,6 +28,17 @@
...
+  end
+end
 
+def without_thread_report_on_exception
+  if Thread.respond_to?(:report_on_exception)
+    current_value = Thread.report_on_exception
+    Thread.report_on_exception = false
   end
+
+  yield
```
mullermp, line 40: I think a before/after hook is simpler here rather than yielding. I would also think that saving the previous value is a bit overkill.. we don't rely on the exception reporting anyway regardless of Ruby versions. I believe that setting this value only applies for the spec and not for all specs.. but I'm not 100% sure on that one.

PR #2308, opened 2020-05-17
Fix ruby 2.7 warnings

hugopeixoto:
There were three classes of warnings:

- two seemingly simple fixes:
  - an `if` statement with no clause on the same line;
  - a redundant regexp (`[\w_]`);
- three unused variable `resp` in generated code;
- method redefinition warnings in the customizations code.

To handle the unused variables, I had to change `resp: true` in
`ResourceActionCode` to `false` in some scenarios. This now looks a bit
complicated.

I silenced the method redefinition warnings by using a technique used by
rails:

https://edgeapi.rubyonrails.org/classes/Module.html#method-i-silence_redefinition_of_method

I ran `rake build` and looked at the changes, they looked fine. I was not sure if these should be included, so I did include them.

`aws-sdk-s3.gemspec` was bumped to `3.96.1`. I am also unsure if this should have been included in the commit.

I added a changelog entry to the s3 gem, since that's the one I was focusing on. The other ones may still contain warnings, even though some of them were fixed by the code generator changes.

### DIFF
--- a/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/resource_action_code.rb
+++ b/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/resource_action_code.rb
@@ -16,7 +16,9 @@ module AwsSdkCodeGenerator
         parts << ResourceBatchBuilder.new(resource: @resource).build
         parts << "#{resource_type}::Collection.new([batch], size: batch.size)"
       elsif resource_action?
-        parts << client_request
+        parts << client_request(
+          @resource['identifiers'].any? { |i| i['source'] == 'response' } || @resource['path']
+        )
         parts << ResourceBuilder.new(resource: @resource, request_made: true).build
       else
         parts << client_request
@@ -27,8 +29,8 @@ module AwsSdkCodeGenerator
 
     private
 
-    def client_request
-      ResourceClientRequest.build(request: @request, resp: true, streaming: @streaming)
+    def client_request(resp = true)
+      ResourceClientRequest.build(request: @request, resp: resp, streaming: @streaming)
     end
 
     def resource_type
--- a/build_tools/spec/region_spec.rb
+++ b/build_tools/spec/region_spec.rb
@@ -14,7 +14,7 @@ def whitelist
     },
     's3' => {
       'location_constraint.rb' => 12,
-      'bucket.rb' => 143,
+      'bucket.rb' => 145,
       'presigned_post.rb' => 587,
       'iad_regional_endpoint.rb' => 'SKIP_FILE'
     },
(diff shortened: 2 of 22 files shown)


mullermp:
Thanks for opening this. It would be helpful if you could include the exact warnings that come up. The unused variable changes make sense. I'm not sure about the method redefinition part.. is there a better solution? For the changelog entries, it may make sense to either patch version bump service gems or just have the changes picked up on the next release automatically. I'm leaning towards the latter case for that.

hugopeixoto:
> It would be helpful if you could include the exact warnings that come up

These are the warnings I'm fixing:

```
[...]/aws-sdk-s3-1.64.0/lib/aws-sdk-s3/plugins/expect_100_continue.rb:16:
  warning: `if' at the end of line without an expression
[...]/aws-sdk-s3-1.64.0/lib/aws-sdk-s3/plugins/http_200_errors.rb:30:
  warning: character class has duplicated range: /<[\w_]/
[...]/aws-sdk-s3-1.64.0/lib/aws-sdk-s3/resource.rb:75:
  warning: assigned but unused variable - resp
[...]/aws-sdk-s3-1.64.0/lib/aws-sdk-s3/bucket.rb:523: 
  warning: assigned but unused variable - resp
[...]/aws-sdk-s3-1.64.0/lib/aws-sdk-s3/multipart_upload.rb:277:
  warning: assigned but unused variable - resp
[...]/aws-sdk-s3-1.64.0/lib/aws-sdk-s3/customizations/bucket.rb:9:
  warning: method redefined; discarding old initialize
[...]/aws-sdk-s3-1.64.0/lib/aws-sdk-s3/bucket.rb:20:
  warning: previous definition of initialize was here
[...]/aws-sdk-s3-1.64.0/lib/aws-sdk-s3/customizations/object.rb:62:
  warning: method redefined; discarding old copy_from
[...]/aws-sdk-s3-1.64.0/lib/aws-sdk-s3/object.rb:662:
  warning: previous definition of copy_from was here
[...]/aws-sdk-s3-1.64.0/lib/aws-sdk-s3/customizations/object_summary.rb:11:
  warning: method redefined; discarding old copy_from
[...]/aws-sdk-s3-1.64.0/lib/aws-sdk-s3/object_summary.rb:413:
  warning: previous definition of copy_from was here
```

> I'm not sure about the method redefinition part.. is there a better solution?

With the current ruby targets, I don't think so. If ruby 1.9.3 support were dropped, the whole costumizations mechanism could be replaced with creating a new module and prepending it instead of reopening the class, but this is only available in ruby 2.0.0 (we'd need prepend to be able to use `super`).

> For the changelog entries, it may make sense to either patch version bump service gems or just have the changes picked up on the next release automatically. I'm leaning towards the latter case for that.

Yeah, I don't think this is important enough to warrant a version bump by itself. Does this mean that no changelog entries are needed? Should I remove the s3 one?

mullermp:
Yes, you can remove all CHANGELOG entries. I can get back to this soon.

hugopeixoto:
I have updated the PR to remove the CHANGELOG entry.

mullermp:
Thanks for the contribution. Merging.

PR #2310, opened 2020-05-18
Add region validation using regex

mullermp:


gems/aws-sdk-core/lib/aws-sdk-core/errors.rb
```diff
@@ -208,12 +208,40 @@ class InvalidProcessCredentialsPayload < RuntimeError; end
...
+        msg = 'No region was provided. Configure the `:region` option or '\
+          "export the region name to ENV['AWS_REGION']"
         super(msg)
       end
     end
 
+    # Raised when a client is contsructed and the region is not valid.
+    class InvalidRegionError < ArgumentError
+      def initialize(*args)
+        super(<<-MSG)
```
alextwoods, line 220: This is a fairly long exception message.  Do we have similar length messages anywhere else?  I do think it may be useful to customers though.  Maybe add it to the MissingRegionError as well?

gems/aws-sdk-core/lib/aws-sdk-core/errors.rb
```diff
@@ -208,12 +208,40 @@ class InvalidProcessCredentialsPayload < RuntimeError; end
...
+
+Known AWS regions include (not specific to this service):
+
+#{possible_regions}
+        MSG
+      end
+
+      private
+
+      def possible_regions
```
alextwoods, line 236: This is duplicated - and seems like it should probably live somewhere other than in individual error classes

gems/aws-sdk-core/lib/aws-sdk-core/plugins/regional_endpoint.rb
```diff
@@ -47,21 +43,27 @@ class RegionalEndpoint < Seahorse::Client::Plugin
       end
 
       def after_initialize(client)
-        if client.config.region.nil? or client.config.region == ''
+        if client.config.region.nil? || client.config.region == ''
           raise Errors::MissingRegionError
         end
+
+        # check region format follows pattern <continent>-<direction>-<number>
```
alextwoods, line 50: The comment doesn't quite match the check that we're doing.  We're only checking for a valid label/sub-domain format and normally aws regions follow the <continent>-<direction>-<number> format.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/regional_endpoint.rb
```diff
@@ -47,21 +43,27 @@ class RegionalEndpoint < Seahorse::Client::Plugin
...
+        unless client.config.region =~ /^[a-zA-Z]([a-zA-Z0-9\-]*[a-zA-Z0-9])?$/
```
alextwoods, line 51: Could shorten this by either: making it case insensitive or using bracketed character classes.  eg: [:alpha:] and [:alnum:]

gems/aws-sdk-core/lib/aws-sdk-core/errors.rb
mullermp: Mostly taken from here: https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/lib/aws-sdk-core/errors.rb#L226

gems/aws-sdk-core/lib/aws-sdk-core/errors.rb
mullermp: It is duplicated. I can't really re-use it here (don't want to change how errors are inherited). This logic could maybe live in aws-partitions somewhere if you have any ideas.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/regional_endpoint.rb
mullermp: Pending whatever the teams decide...

gems/aws-sdk-core/lib/aws-sdk-core/plugins/regional_endpoint.rb
mullermp: what's a good comment I can put here for my future self?

mullermp:
Waiting on input on whether to validate the region or the host.

PR #2316, opened 2020-05-26
Ignore expect header for V4 signature

murlock:
When ruby SDK is used with nginx as reverse proxy,
the header Expect is not forwarded to the endpoint
and request is rejected with Invalid Signature.

This patch will avoid to use the expect header for V4.

### DIFF
--- a/gems/aws-sigv4/CHANGELOG.md
+++ b/gems/aws-sigv4/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Don't use `expect` header to compute Signature.
+
 1.1.3 (2020-04-27)
 ------------------
 
--- a/gems/aws-sigv4/lib/aws-sigv4/signer.rb
+++ b/gems/aws-sigv4/lib/aws-sigv4/signer.rb
@@ -123,6 +123,7 @@ module Aws
         @unsigned_headers = Set.new((options.fetch(:unsigned_headers, [])).map(&:downcase))
         @unsigned_headers << 'authorization'
         @unsigned_headers << 'x-amzn-trace-id'
+        @unsigned_headers << 'expect'
         [:uri_escape_path, :apply_checksum_header].each do |opt|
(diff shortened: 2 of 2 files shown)


mullermp:
Hey - trying to understand your use case a little better. Will the `unsigned_headers` feature work for you?

https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sigv4/lib/aws-sigv4/signer.rb#L105-L107

murlock:
Hi,

Yes, it make sense to use `unsigned_headers` (I've missed it when I've checked).
I've updated the PR.

mullermp:
I guess what I mean to say is.. couldn't you pass in 'expect' as an unsigned header when signing?

Can I understand your use case a little more? How are you using the sigv4 gem? Is it implicitly (from a service gem) or are you constructing an object yourself?

Solvik:
the Ruby SDK, used by `fluentd` uses the `Expect` header for signature.
It's a behavior that is different from other SDK, like the python one.

When used with nginx in front of AWS S3 or any other S3-compatible solution, the authentication is refused by the sigv4 hmac because it's missing

mullermp:
Thanks for the context. Could you link me the GitHub repo of the exact package you are using?

I'm not opposed to merging this change. My only concern is that this could have some down stream impact on an AWS service and it would be difficult to verify that. If at all possible, I'd prefer to make the change in the 3rd party package you are using. In that package, we would add the Expect header to the unsigned headers param there. The unsigned headers hardcoded in the sigv4 gem are necessary for some AWS services.

murlock:
Thanks @Solvik for your help
@mullermp botocore already skip this header (https://github.com/boto/botocore/blob/develop/botocore/auth.py#L49)

mullermp:
Ah ok! If boto already does it, it gives us higher confidence that it's okay, and probably safe to merge. Let me do a quick double check and this should go out tomorrow.

mullermp:
Looks good. Thanks for the contribution!

PR #2317, opened 2020-05-26
endpoint_discovery defaults

alextwoods:
Set endpoint_discovery default value to true for services with an operation that requires it. Change behavior of endpoint_discovery plugin: raise an error on operations that require endpoint discovery when disabled.

### DIFF
--- a/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/plugin_list.rb
+++ b/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/plugin_list.rb
@@ -15,17 +15,6 @@ module AwsSdkCodeGenerator
 
     private
 
-    def compute_plugin_options(class_name, options)
-      plugin_options = const_get(class_name).options
-      if class_name == 'Aws::Plugins::EndpointDiscovery' &&
-         options[:endpoint_discovery_required]
-        plugin_options.each do |opt|
-          opt.doc_default = true if opt.name == :endpoint_discovery
-        end
-      end
-      plugin_options
-    end
-
     def compute_plugins(options)
       plugins = {}
       plugins.update(options[:async_client] ? default_async_plugins : default_plugins)
@@ -38,11 +27,10 @@ module AwsSdkCodeGenerator
       plugins.map do |class_name, path|
         path = "./#{path}" unless path[0] == '/'
         Kernel.require(path)
-        plugin_options = compute_plugin_options(class_name, options)
 
         Plugin.new(
           class_name: class_name,
-          options: plugin_options,
+          options: const_get(class_name).options,
           path: path)
       end
     end
--- a/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/plugin_list.rb
+++ b/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/plugin_list.rb
@@ -15,6 +15,17 @@ module AwsSdkCodeGenerator
 
     private
 
+    def compute_plugin_options(class_name, options)
+      plugin_options = const_get(class_name).options
+      if class_name == 'Aws::Plugins::EndpointDiscovery' &&
+         options[:endpoint_discovery_required]
+        plugin_options.each do |opt|
+          opt.doc_default = true if opt.name == :endpoint_discovery
+        end
+      end
+      plugin_options
+    end
+
     def compute_plugins(options)
       plugins = {}
       plugins.update(options[:async_client] ? default_async_plugins : default_plugins)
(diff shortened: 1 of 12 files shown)


build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/plugin_list.rb
```diff
@@ -15,6 +15,17 @@ def each(&block)
 
     private
 
+    def compute_plugin_options(class_name, options)
+      plugin_options = const_get(class_name).options
+      if class_name == 'Aws::Plugins::EndpointDiscovery' &&
```
alextwoods, line 20: I don't love having the logic for this here because its specific to the business logic of one plugin.  Ideally - this should be pushed into the plugin itself.  And one option would be to allow doc_default to be a block that takes the `options`.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/endpoint_discovery.rb
```diff
@@ -151,8 +155,12 @@ def _discover_endpoint(ctx, required)
 
       def self.resolve_endpoint_discovery(cfg)
         env = ENV['AWS_ENABLE_ENDPOINT_DISCOVERY']
+        default = cfg.api.operations.any? do |_, o|
+          o.endpoint_discovery && o.endpoint_discovery['required']
+        end
```
mullermp, line 160: Wasn't this already determined above in `endpoint_discovery_required`?

gems/aws-sdk-core/lib/seahorse/client/plugin.rb
```diff
@@ -123,11 +123,12 @@ def initialize(name, options = {})
         attr_writer :doc_default
         attr_accessor :docstring
 
-        def doc_default
+        def doc_default(options)
           if @doc_default.nil? && !default.is_a?(Proc)
             default
           else
-            @doc_default
+            puts "doc_default for: #{name}, @doc_default=#{@doc_default}"
```
mullermp, line 130: remove

gems/aws-sdk-core/lib/aws-sdk-core/plugins/endpoint_discovery.rb
```diff
@@ -4,10 +4,10 @@ module Plugins
     class EndpointDiscovery < Seahorse::Client::Plugin
 
       option(:endpoint_discovery,
-        default: false,
+        doc_default: Proc.new { |options| options[:endpoint_discovery_required] },
```
mullermp, line 7: Why is this a Proc?

gems/aws-sdk-core/spec/aws/plugins/endpoint_discovery_spec.rb
```diff
@@ -228,7 +242,7 @@ module Plugins
           region: 'us-east-1',
           endpoint_discovery: true,
           active_endpoint_cache: true,
-          endpoint_cache_poll_interval: 1,
+          endpoint_cache_poll_interval: 0.01,
```
mullermp, line 245: Why change these?

alextwoods: This test was sleeping for 5 seconds to give the 1 second cache poll time to run in a seperate thread.  Lowering this value and then decreasing sleep saves 4.9 seconds.  Its not required, just was annoying me that these specs were taking so long.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/endpoint_discovery.rb
alextwoods: To allow it to be called dynamically during code generation

gems/aws-sdk-core/lib/aws-sdk-core/plugins/endpoint_discovery.rb
alextwoods: This is the confusing part of using a dynamic default (ie, the `do |cfg| resolve_endpoint_discovery(cfg) end`.  [Plugin#add_option](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/lib/seahorse/client/plugin.rb#L9) and [Configuration#add_option](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/lib/seahorse/client/configuration.rb#L106) let us either have a static default or a dynamic (block) default.  The dynamic default is called by the Configuration::DefaultResolver during build! (which is called during [Client::Base#initiailize](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/lib/seahorse/client/base.rb#L19)) and doesn't have access to the doc_default that is resolved during code gen...

I don't like the disconnect between the doc_default and the behavior of the dynamic default - I'm open to ideas on how to connect them, but I don't have any great ones.

PR #2319, opened 2020-05-28
Convert ENV['AWS_MAX_ATTEMPTS'] string value to int

jesseadams:

With ruby 2.7.1 I am unable to set `AWS_MAX_ATTEMPTS` to a custom value with an environment variable. 

```
# Wont work because it is a String and not an Integer
irb(main):001:0> ENV['AWS_MAX_ATTEMPTS'] = '300'

# Also fails with TypeError
irb(main):002:0> ENV['AWS_MAX_ATTEMPTS'] = 300
        5: from /usr/bin/irb:23:in `<main>'
        4: from /usr/bin/irb:23:in `load'
        3: from /usr/lib/ruby/gems/2.7.0/gems/irb-1.2.1/exe/irb:11:in `<top (required)>'
        2: from (irb):2
        1: from (irb):2:in `[]='
TypeError (no implicit conversion of Integer into String)
```

This was likely missed with tests because of the stubbing of `ENV` here: https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/spec/shared_spec_helper.rb#L19

This simple change keeps tests passing and will enable you to set `AWS_MAX_ATTEMPTS` to a String in ruby, which then gets converted into an integer when resolved.

Thanks for your time!

### DIFF
--- a/gems/aws-sdk-core/spec/aws/plugins/retry_errors_spec.rb
+++ b/gems/aws-sdk-core/spec/aws/plugins/retry_errors_spec.rb
@@ -6,109 +6,99 @@ module Aws
     describe RetryErrors do
       let(:client) { RetryErrorsSvc::Client.new(stub_responses: true) }
 
-      context 'ENV is stubbed' do
-        before(:each) do
-          stub_const('ENV', {})
-        end
-
-        it 'can configure retry_mode with shared config' do
-          allow_any_instance_of(Aws::SharedConfig)
-            .to receive(:retry_mode).and_return('standard')
-          expect(client.config.retry_mode).to eq('standard')
-        end
-
-        it 'can configure retry_mode using ENV with precedence over config' do
-          allow_any_instance_of(Aws::SharedConfig)
-            .to receive(:retry_mode).and_return('standard')
-          ENV['AWS_RETRY_MODE'] = 'adaptive'
-          expect(client.config.retry_mode).to eq('adaptive')
-        end
-
-        it 'raises when retry_mode is not legacy, standard, or adaptive' do
-          ENV['AWS_RETRY_MODE'] = 'peccy'
-          expect { client }.to raise_error(ArgumentError)
-        end
-
-        it 'uses the handler when retry_mode is standard' do
-          client = RetryErrorsSvc::Client.new(retry_mode: 'standard',
-                                              region: 'us-west-2')
-          expect(client.handlers.entries.map(&:handler_class)).
-            to include(RetryErrors::Handler)
-        end
-
-        it 'uses the handler when retry_mode is adaptive' do
-          client = RetryErrorsSvc::Client.new(retry_mode: 'adaptive',
-                                              region: 'us-west-2')
-          expect(client.handlers.entries.map(&:handler_class))
-            .to include(RetryErrors::Handler)
-        end
-
-        it 'defaults config.max_attempts to 3' do
-          expect(client.config.max_attempts).to eq(3)
-        end
-
-        it 'can configure max_attempts with shared config' do
-          allow_any_instance_of(Aws::SharedConfig)
-            .to receive(:max_attempts).and_return(5)
-          expect(client.config.max_attempts).to eq(5)
-        end
-
-        it 'defaults config.adaptive_retry_wait_to_fill to true' do
-          expect(client.config.adaptive_retry_wait_to_fill).to eq(true)
-        end
-
-        it 'can configure adaptive_retry_wait_to_fill using ENV with precedence over config' do
-          ENV['AWS_ADAPTIVE_RETRY_WAIT_TO_FILL'] = 'false'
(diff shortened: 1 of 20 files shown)


mullermp:
Thanks for opening this pull request. I think you can call `.to_i` directly instead of using `&&` with another expression. 

After we merged this, I found a method that should have probably been used for booleans (https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/lib/aws-sdk-core/util.rb#L59) I think that this plugin should maybe use `str_2_bool` for AWS_ADAPTIVE_RETRY_WAIT_TO_FILL and other similar boolean-like options. In this case, what do you think about a `str_2_int` method that we can use in this instance too?

mullermp:
I would also want to fix the test to catch this error before merging, too.

jesseadams:
@mullermp - We only want to convert it to an integer if it is set. If it is unset, `nil` will convert to `0` and then fallback methods of resolving the value won't be tried. Unless I am incorrect?

mullermp:
> @mullermp - We only want to convert it to an integer if it is set. If it is unset, `nil` will convert to `0` and then fallback methods of resolving the value won't be tried. Unless I am incorrect?

Ah yeah! Sorry about that.

jesseadams:
@mullermp - I updated the tests to not stub `ENV` for the three related tests in question. Before my change:

```
Finished in 31.91 seconds (files took 1.67 seconds to load)
1224 examples, 1 failure

Failed examples:

rspec ./spec/aws/plugins/retry_errors_spec.rb:88 # Aws::Plugins::RetryErrors ENV is not stubbed can configure max_attempts using ENV with precedence over config
```

After my change:

```
Finished in 32.16 seconds (files took 1.69 seconds to load)
1224 examples, 0 failures
```

gems/aws-sdk-core/spec/shared_spec_helper.rb
```diff
@@ -15,9 +15,6 @@
 #
 RSpec.configure do |config|
   config.before(:each) do
-
-    stub_const('ENV', {})
```
mullermp, line 19: I'd prefer to keep this here rather than anyone having to remember to add stubbing to individual spec files when necessary.. this seems like it has an important job to potentially be overlooked (to prevent real credentials from being used).

I also think that here lies the actual problem.. rather than stubbing ENV as a hash (because it's not a hash - it's [hash-LIKE](https://docs.ruby-lang.org/en/master/ENV.html)) we should actually be clearing the ENV with `ENV.clear` or better if there's a way to `stub_const('ENV', somehow_make_a_new_ENV_here)`.

gems/aws-sdk-core/spec/aws/plugins/retry_errors_spec.rb
```diff
@@ -6,97 +6,109 @@ module Plugins
...
 
-      it 'can configure correct_clock_skew using ENV with precedence over config' do
-        ENV['AWS_CORRECT_CLOCK_SKEW'] = 'true'
-        expect(client.config.correct_clock_skew).to eq(true)
-      end
+      context 'ENV is not stubbed' do
+        it 'can configure max_attempts using ENV with precedence over config' do
+          allow_any_instance_of(Aws::SharedConfig)
+            .to receive(:max_attempts).and_return(3)
+          ENV['AWS_MAX_ATTEMPTS'] = '1'
```
mullermp, line 91: I think it makes sense to have three tests here - tests for '1', '-1', and 1. We currently have 1 and -1, so we would need to make the change accordingly.

mullermp:
Thanks for taking the initiative on this. Please see my feedback. :) Happy to also contribute here.

jesseadams:
@mullermp - I am totally open to you collaborating on this. I am out of time today but I can take another look tomorrow. 

Regarding the test cases: '1' and '-1' make sense. 1 does not as you will immediately get a `TypeError` in ruby when you try to set the value. I don't think this is a useful test for the scope of this utility. We also have test case for 'string' which also makes sense.

I was just trying to make the least amount of impact/changes while still being able to prove out the failure and my fix. If you have a preferred refactor or approach to accomplish this I am more than willing to give it a shot or let you hijack this PR and take it over the finish. My ruby is honestly fairly rusty these days.

Thanks!

mullermp:
No worries! I went ahead and updated your branch. I think these changes are less intrusive. It was unfortunate on our part that ENV was stubbed as a hash, allowing integers to be set when in reality they never could be. It'd be better instead of using `ENV.clear` to stub ENV somehow but I wasn't sure how to construct a new ENV object. I thought about using `double()` but I would have to clear ENV anyway.

jesseadams:
@mullermp - Awesome, thanks for your prompt work on this. I appreciate it.

mullermp:
Thanks for the contribution! (and for opening this issue)

PR #2322, opened 2020-06-02
Http checksum required trait support

mullermp:
Consume the http checksum required trait instead of managing a hardcoded list


gems/aws-sdk-core/lib/aws-sdk-core/plugins/http_checksum.rb
```diff
@@ -0,0 +1,56 @@
...
+          elsif value.respond_to?(:read)
+            md5 = OpenSSL::Digest::MD5.new
+            update_in_chunks(md5, value)
+            md5.base64digest
+          else
+            OpenSSL::Digest::MD5.digest(value).base64digest
+          end
+        end
+
+        def update_in_chunks(digest, io)
```
alextwoods, line 37: I believe there is an opportunity for us to avoid reading the body twice here.  siv4 requires the `x-amz-content-sha256` header to be set and will read the body to compute it.  However, since we are already doing so here, we can compute and set it here as well.

mullermp: Can you link to where that's done? It maybe makes sense to do that here, but if it's part of the sigv4 gem then maybe not.

mullermp: Talked offline - will be doing this later when other modeled checksums are supported.

Issue #2105, opened 2019-08-22
FILTERED showing up in yml export


brianbedard:
Please fill out the sections below to help us address your issue

### Issue description
When exporting job settings utilizing get_job https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/MediaConvert/Client.html#get_job-instance_method, the word FILTERED is showing up in the output. It was originally HLS_GROUP_SETTINGS when the job ran successfully, but the export is replacing that when it's output through the api call. 

### Gem name ('aws-sdk', 'aws-sdk-resources' or service gems like 'aws-sdk-s3') and its version
require 'aws-sdk-s3'
require 'aws-sdk-mediaconvert'
### Code snippets / steps to reproduce
output_groups:
- custom_name: HLSOutput
  name: **"[FILTERED]"**
  output_group_settings:
    cmaf_group_settings:
    dash_iso_group_settings:
    file_group_settings:
    hls_group_settings:
      ad_markers:
      base_url:
      caption_language_mappings:
      caption_language_setting: OMIT
      client_cache: ENABLED
      codec_specification: RFC_4281
      destination: s3://sample-video-output/HLS-Test-Video
      destination_settings:
      directory_structure: SINGLE_DIRECTORY
      encryption:
      manifest_compression: NONE
      manifest_duration_format: INTEGER
      min_final_segment_length: 0.0
      min_segment_length: 0
      output_selection: MANIFESTS_AND_SEGMENTS
      program_date_time: EXCLUDE
      program_date_time_period: 600
      segment_control: SEGMENTED_FILES
      segment_length: 3
      segments_per_subdirectory:
      stream_inf_resolution: INCLUDE
      timed_metadata_id_3_frame: PRIV
      timed_metadata_id_3_period: 10
      timestamp_delta_milliseconds:
    ms_smooth_group_settings:
    type: **"[FILTERED]"**
  outputs:

cjyclaire:
Thanks for the report! this happened because SDK helps [filter out](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/lib/aws-sdk-core/log/param_filter.rb#L14) sensitive parameters from logging, `to_s`, `to_hash` and wire traces to avoid those information exposed in public accidentally.

However, currently SDK is filtering those parameters across all services even if it's only consider to be sensitive for one specific service, similar issue: #2082

[...]

PR #2325, opened 2020-06-09
Sensitive param filtering per request/response type

### LINKED ISSUE #2312: list_tags_for_resource for Managed Instance does not return the list of tags
  
**Describe the bug**
Invoking 
```
    resource_id: "mi-1234567890",
  )
```

Returns:
```
```

```
{
    "TagList": [
        {
            "Key": "ImageUrl",
            "Value": "XYZ"
        }
    ]
}
```

**Gem name**

```
aws-sdk-ssm (1.77.0)
aws-sdk-core (3.96.0)
```

Invoking 
```
    resource_id: "mi-1234567890",
  )
```

**Expected behavior**

**Additional context**

mullermp:
Filters sensitive params only if they are marked in the type (via codegen)
Provide an option to disable sensitive param filtering if needed.

Fixes #2312, #2105, #2082


mullermp:
This doesn't work for all services.. will need to touch code gen.

gems/aws-sdk-core/lib/aws-sdk-core/log/param_filter.rb
```diff
@@ -11,33 +11,39 @@ class ParamFilter
...
       # end
 
       def initialize(options = {})
-        @filters = Set.new(SENSITIVE + Array(options[:filter]))
+        filters = options[:filter] || {}
```
mullermp, line 18: To not break current customers, we should "splat" these to all services in the hash if it's an array.

gems/aws-sdk-core/lib/aws-sdk-core/log/formatter.rb
```diff
@@ -121,7 +121,8 @@ def _operation(response)
 
       def _request_params(response)
         params = response.context.params
-        @param_formatter.summarize(@param_filter.filter(params))
+        service = response.context.config.api.metadata['serviceId']
```
mullermp, line 124: Will need to touch code gen... this doesn't match all services all the time.

tasks/update-sensitive-params.rake
```diff
@@ -1,27 +1,23 @@
...
       if shape['type'] == 'structure' && shape['members']
         shape['members'].each_pair do |member_name, member_ref|
-          if 
-            member_ref['sensitive'] || 
-            svc.api['shapes'][member_ref['shape']]['sensitive']
-          then
+          if member_ref['sensitive'] || svc.api['shapes'][member_ref['shape']]['sensitive']
             name = AwsSdkCodeGenerator::Underscore.underscore(member_name).to_sym
-            sensitive << name
+            sensitive[svc.name] << name unless sensitive[svc.name].include?(name)
```
alextwoods, line 10: Rather than using `svc.name` here consider using the serviceId, eg: `svc.api['metadata']['serviceId']`

gems/aws-sdk-core/lib/aws-sdk-core/log/param_filter.rb
```diff
@@ -11,33 +11,47 @@ class ParamFilter
...
+        # Support backwards compatibility for Array
+        if filters.is_a?(Array)
+          @filters = SENSITIVE.dup
+          @filters.update(@filters) { |_k, v| v + Array(filters) }
+        elsif filters.is_a?(Hash)
+          @filters = SENSITIVE.merge(filters)
```
alextwoods, line 24: I think this is an issue that it will override existing params for that service.  Possibly this is useful.... and since the hash is new its not necessarily breaking behavior.  But I think its inconsistent with the way it works for filter arrays (ie. additive).  In either case I think the documentation could use an update to clarify the behavior.

mullermp: We also want a way to disable .. either by overriding or another option. Previously the array was monkey patched in one of the issues.

mullermp: I don't think I can get an option to work because of when the object is created so I think nothing can be additive.

mullermp:
@jessedoyle Adding you to this PR for visibility. A while back you made these changes: https://github.com/aws/aws-sdk-ruby/pull/2219

The TL;DR here -

* Sensitive params will be a hash per service instead of a giant array across all services
* Sensitive params can be disabled via an option to the log formatter (default to enabled)
* No longer needing to do case insensitive checks - Cognito will be updating their API to mark AuthParameters as sensitive. The original fix (as you remember) was not correct because it relied on keys in a hash to be marked sensitive instead of the entire member - it does not guard against additional auth parameters added by cognito.

Let me know what you think.

mullermp:
Putting this down for now -- It'd be nice to have a fully code-gen solution here.. Perhaps a class called `Aws::<Service>::TypesFormatters::<Operation>` that could be applied at the operation level.

mullermp:
Current draft needs more work, it's WIP

mullermp:
I've messed up the Git history :/

PR #2328, opened 2020-06-11
Enable frozen_string_literals

casperisfine:
### Context

We profile the memory usage of our application on a regular basis using `memory_profiler`, and while it probably doesn't amount to much in total, the `aws-sdk` gems is resposible for a lot of noise in the reports. Just a couple examples but there are lots of these:

```
 314  "header"
   1  /tmp/bundle/ruby/2.7.0/gems/aws-sdk-s3-1.65.0/lib/aws-sdk-s3/client_api.rb:1080
   .... snip 305 lines of aws-sdk-s3/client_api.rb
   1  /tmp/bundle/ruby/2.7.0/gems/aws-sdk-s3-1.65.0/lib/aws-sdk-s3/client_api.rb:886
   1  /tmp/bundle/ruby/2.7.0/gems/bootsnap-1.4.6/lib/bootsnap/compile_cache/iseq.rb:19
   1  /tmp/bundle/ruby/2.7.0/gems/fastly-2.5.2/lib/fastly/util.rb:6
   1  /tmp/bundle/ruby/2.7.0/gems/kramdown-1.13.2/lib/kramdown/parser/html.rb:200
   1  /tmp/bundle/ruby/2.7.0/gems/kramdown-1.13.2/lib/kramdown/parser/html.rb:202
   1  /tmp/bundle/ruby/2.7.0/gems/kramdown-1.13.2/lib/kramdown/parser/html.rb:205
   1  /tmp/bundle/ruby/2.7.0/gems/kramdown-1.13.2/lib/kramdown/parser/html.rb:38
   1  /tmp/bundle/ruby/2.7.0/gems/kramdown-1.13.2/lib/kramdown/parser/html.rb:59
 
 112  "uri"
   1  /tmp/bundle/ruby/2.7.0/gems/aws-sdk-s3-1.65.0/lib/aws-sdk-s3/client_api.rb:1005
   .... snip 110 lines of aws-sdk-s3/client_api.rb
   1  /tmp/bundle/ruby/2.7.0/gems/aws-sdk-s3-1.65.0/lib/aws-sdk-s3/client_api.rb:1011
```

In the end 300 instances of `"header"` is about 12KB of memory usage, not a lot, but it's also super easy to fix, and it adds on.

### This PR

I simply edited the code generator templates to prepend `# frozen_string_literal: true` and then ran `rubocop -a --only Style/FrozenStringLiteralComment`. I fully expect CI to fail in various ways, but it will probably be easy to fix in followup commits.

And regardless I wanted to open the discussion on this change.

### DIFF
--- a/build_tools/aws-sdk-code-generator/templates/apig_endpoint_class.mustache
+++ b/build_tools/aws-sdk-code-generator/templates/apig_endpoint_class.mustache
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module {{module_name}}
   module Plugins
 
--- a/build_tools/aws-sdk-code-generator/templates/async_client_class.mustache
+++ b/build_tools/aws-sdk-code-generator/templates/async_client_class.mustache
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 {{#generated_src_warning}}
 {{generated_src_warning}}
 {{/generated_src_warning}}
--- a/build_tools/aws-sdk-code-generator/templates/authorizer_class.mustache
+++ b/build_tools/aws-sdk-code-generator/templates/authorizer_class.mustache
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module {{module_name}}
   module Plugins
 
--- a/build_tools/aws-sdk-code-generator/templates/client_api_module.mustache
+++ b/build_tools/aws-sdk-code-generator/templates/client_api_module.mustache
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 {{#generated_src_warning}}
 {{generated_src_warning}}
 {{/generated_src_warning}}
--- a/build_tools/aws-sdk-code-generator/templates/client_class.mustache
+++ b/build_tools/aws-sdk-code-generator/templates/client_class.mustache
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 {{#generated_src_warning}}
 {{generated_src_warning}}
 {{/generated_src_warning}}
--- a/build_tools/aws-sdk-code-generator/templates/errors_module.mustache
+++ b/build_tools/aws-sdk-code-generator/templates/errors_module.mustache
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 {{#generated_src_warning}}
 {{generated_src_warning}}
 {{/generated_src_warning}}
--- a/build_tools/aws-sdk-code-generator/templates/event_streams_module.mustache
+++ b/build_tools/aws-sdk-code-generator/templates/event_streams_module.mustache
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 {{#generated_src_warning}}
 {{generated_src_warning}}
 {{/generated_src_warning}}
--- a/build_tools/aws-sdk-code-generator/templates/features/env.mustache
+++ b/build_tools/aws-sdk-code-generator/templates/features/env.mustache
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 {{#generated_src_warning}}
 {{generated_src_warning}}
 {{/generated_src_warning}}
--- a/build_tools/aws-sdk-code-generator/templates/features/smoke_step_definitions.mustache
+++ b/build_tools/aws-sdk-code-generator/templates/features/smoke_step_definitions.mustache
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 {{#generated_src_warning}}
 {{generated_src_warning}}
 {{/generated_src_warning}}
--- a/build_tools/aws-sdk-code-generator/templates/gemspec.mustache
+++ b/build_tools/aws-sdk-code-generator/templates/gemspec.mustache
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 {{#generated_src_warning}}
 {{generated_src_warning}}
 {{/generated_src_warning}}
--- a/build_tools/aws-sdk-code-generator/templates/resource_class.mustache
+++ b/build_tools/aws-sdk-code-generator/templates/resource_class.mustache
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 {{#generated_src_warning}}
 {{generated_src_warning}}
 {{/generated_src_warning}}
--- a/build_tools/aws-sdk-code-generator/templates/root_resource_class.mustache
+++ b/build_tools/aws-sdk-code-generator/templates/root_resource_class.mustache
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 {{#generated_src_warning}}
 {{generated_src_warning}}
 {{/generated_src_warning}}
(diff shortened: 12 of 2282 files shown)


casperisfine:
Ok, so surprisingly it doesn't break CI at all. However one thing that would need to be fixed is that the comment made it into documentation snippets: https://github.com/aws/aws-sdk-ruby/blob/38413080881246798b56c0a25b74b0cf162fbc88/gems/aws-sdk-s3/lib/aws-sdk-s3/client.rb#L4568-L4570

I'm unsure how to avoid that.

mullermp:
This change will have a huge blast radius just FYI..

casperisfine:
I understand the sentiment, but for having migrated large amount of code to `frozen_string_literals` in the last few years, I don't really think so TBH.

Usually what breaks when you turn it on, is the code you call, not the code that calls you. So it's more of a problem when enabled in apps, rather than in lower level dependencies like AWS.

mullermp:
We had a series of PRs recently where CI was hacked with `sed` to add frozen string literal and some changes were required because we were mutating strings. It's possible that all infractions were handled in that PR allowing you to make these changes.

https://github.com/aws/aws-sdk-ruby/pull/2269
https://github.com/aws/aws-sdk-ruby/pull/2277

mullermp:
Perhaps @midnight-wonderer would be interested in this PR

casperisfine:
Indeed. I was actually surprised to see 0 errors on CI, but this PRs explains it. @midnight-wonderer already made the code compatible with it.

mullermp:
I agree -- and again, this is very risky from our perspective. I'm not against the change. There's a lot to review here and a lot that could break.

It would be easier from our side if we can review this in chunks. To start, I think just the generator changes (before 3176 files have been modified..) would be ideal. Then we can get plugins/build_tools/etc..

casperisfine:
> this is very risky from our perspective.

I totally understand.

>  To start, I think just the generator changes (before 3176 files have been modified..) would be ideal 

Sure. I'll do that.

midnight-wonderer:
Thanks for pinging.  
I totally forgot about this, busy with something else recently.

Matz himself abandon the idea of making every string immutable by default about a year ago.
https://bugs.ruby-lang.org/issues/11473
I think the risks and incompatibilities outweigh the benefits.
(See some discussion there for contexts.)

So the magic comment is what we live on forever, presumably.  
Another possibility I can think of is that instead of migrating the whole set of libraries, reviewing magic comments per case basis might be less risky.

And since we can memory-profile the code, just like what OP does in this pull request, we can be sure that the magic comment truly has benefits. The pull request like those should take a shorter time to review, and significantly less risk.

What do you say?

casperisfine:
> Matz himself abandon the idea of making every string immutable by default about a year ago.

Sure, but that's not really a factor in my PR. I'd just like this noise gone.

> I think the risks and incompatibilities outweigh the benefits.

IMO the risks are very small. As for the benefits, I suppose I can memory profile the gem in isolation to see how much impact it has on memory footprint.

> reviewing magic comments per case basis might be less risky.

Yeah, if this PR is rejected, I already plan to submit an alternative one that target the biggest sources of string duplications. But I'd rather get rid of the issue entirely at once.

casperisfine:
Here's the profile of `aws-sdk-s3` only: https://gist.github.com/casperisfine/48af65a9c5055ba95035713ff311b399

As a quick summary, before:

```
Total retained:  2.33 MB (22054 objects)
```

after:
```
Total retained:  2.22 MB (19297 objects)
```

So 110KB out of 2 330KB, about a 5% reduction. Not huge for sure, but not too shabby either.

What I'd really like to point out though is how the duplicated string are obviously used as constants:

```ruby
Retained String Report
       307  "header"
       109  "uri"
        97  "Bucket"
        67  "querystring"
        59  "POST"
        55  "/"
        50  "KeyId"
        41  "message"
        37  "GET"
        37  "Key"
        29  "PUT"
        29  "http://s3.amazonaws.com/doc/2006-03-01/"
        25  "Boolean"
        24  "x-amz-request-payer"
        23  "Prefix"
        21  "Content-MD5"
        20  "x-amz-request-charged"
        19  "Content-MD5 header will now be automatically computed and injected in associated operation's Http request."
        17  "x-amz-server-side-encryption-customer-algorithm"
        16  ""
        16  "x-amz-server-side-encryption-customer-key-MD5"
        15  "DELETE"
        14  "Status"
        14  "versionId"
        13  "ETag"
        13  "GrantTokens"
        13  "Owner"
        13  "Tag"
        13  "x-amz-server-side-encryption"
        13  "x-amz-server-side-encryption-aws-kms-key-id"
        12  "Event"
        11  "CustomKeyStoreId"
        11  "IsTruncated"
        11  "Marker"
        11  "Policy"
        11  "StorageClass"
        10  "/{Bucket}/{Key+}"
        10  "Expires"
        10  "Id"
        10  "Name"
        10  "NextMarker"
        10  "x-amz-version-id"
         9  "Filter"
         9  "ID"
         9  "LastModified"
         9  "Rule"
         9  "id"
         9  "x-amz-server-side-encryption-customer-key"
         8  "Credentials"
         8  "Limit"
```

mullermp:
https://github.com/aws/aws-sdk-ruby/pull/2328#issuecomment-642621355

Is this still an issue? When I get some time later I can check it out.

casperisfine:
@mullermp I think so yes.

mullermp:
I think these first changes are good. Running `rake build:aws-sdk-s3` for example, the doc strings aren't updated. Did you by chance accidentally update `build_tools/aws-sdk-code-generator/templates/method.mustache` on your first iteration?

build_tools/aws-sdk-code-generator/templates/apig_endpoint_class.mustache
```diff
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
```
mullermp, line 1: If you could, please insert a newline after all of these.

casperisfine: Done!

casperisfine:
> I think we're good to run `rake build` here.

You mean in this PR?

mullermp:
Yep! Try running `rake build:aws-sdk-s3` and verify it looks correct to you too. Then go ahead and run `rake build` and commit it to this PR.

casperisfine:
@mullermp done! I chose to do it in a distinct commit, so that the actual template changes are easier to review.

mullermp:
Will wait for CI to finish and merge. Could we open another PR for any specific services/customizations?

casperisfine:
> Could we open another PR for any specific services/customizations?

Sorry, I'm not sure what you are referring to? Was this message addressed to me?

mullermp:
Yes, it was addressed to you. I was assuming that you'd want to add frozen string literal to hand-managed code such as plugins, customizations, etc? If not then this is fine - codegen covers most of the infractions.

casperisfine:
I see. I'm really only familiar with a couple components like S3. I'll take a look the the hand-managed code tomorrow.

mullermp:
Sure - I would recommend doing separate PRs for S3 and DynamoDB (given the larger amount of customizations/plugins).

mullermp:
Thanks for the contribution!

PR #2329, opened 2020-06-11
Bump the aws-sdk-core dependency version

sternik:
Changed because new plugins have been added to `aws-sdk-core`.


alextwoods:
Thanks for doing this!  We can add this change to the services.rb in the build-tools to ensure that the minimum core version is set when gems are generated (I'll push this update in a minute).

I'm looking into whether we need to add changelogs to all of the gems in order to publish new versions of them.  At minimum we may want to yank the gems that have been published since the http_checksum plugin went out and publish new versions of those.

mullermp:
We are very sorry about this. We will start yanking gems.

sufyanadam:
Thanks. Confirming that my issues have been resolved now.

PR #2333, opened 2020-06-16
Add frozen_string_literals comment in aws-sdk-s3

casperisfine:

I simply ran:

```bash
rubocop -a --only Style/FrozenStringLiteralComment,Layout/EmptyLineAfterMagicComment gems/aws-sdk-s3/lib/
```

### DIFF
--- a/build_tools/spec/region_spec.rb
+++ b/build_tools/spec/region_spec.rb
@@ -13,9 +13,9 @@ def whitelist
       'stub_responses.rb' => 19
     },
     's3' => {
-      'location_constraint.rb' => 12,
-      'bucket.rb' => 145,
-      'presigned_post.rb' => 587,
+      'location_constraint.rb' => 14,
+      'bucket.rb' => 147,
+      'presigned_post.rb' => 589,
       'iad_regional_endpoint.rb' => 'SKIP_FILE'
     },
     's3control' => {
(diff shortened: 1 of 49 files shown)


mullermp:
You'll need to update the line configuration in `build_tools/spec/region_spec.rb`. The test exists to check for hardcoded region values. Since you added the magic comment, the line number needs to be increased by 2.

build_tools/spec/region_spec.rb
```diff
@@ -9,17 +9,17 @@ def whitelist
...
+      'signature_v4.rb' => 37,
+      'stub_responses.rb' => 21
     },
     's3' => {
-      'location_constraint.rb' => 12,
-      'bucket.rb' => 145,
-      'presigned_post.rb' => 587,
+      'location_constraint.rb' => 14,
+      'bucket.rb' => 147,
+      'presigned_post.rb' => 589,
```
mullermp, line 18: Ahh.. you only need to touch these ones! The core/s3control tests may fail now.

mullermp:
Thanks for the contribution!

casperisfine:
@mullermp there's 714 files left https://github.com/Shopify/aws-sdk-ruby/commit/remaining-frozen-literals would you rather merge them all in one bulk or should I do several batches?

mullermp:
Whatever is easiest to review would be preferred. If there are any other gems with large amounts of customization, those can be split out..

PR #2334, opened 2020-06-16
Add frozen_string_literals comment aws-sdk-dynamodb*

casperisfine:

I simply ran:

```bash
rubocop -a --only Style/FrozenStringLiteralComment,Layout/EmptyLineAfterMagicComment gems/aws-sdk-dynamo*/lib
```

### DIFF
--- a/gems/aws-sdk-dynamodbstreams/lib/aws-sdk-dynamodbstreams/customizations.rb
+++ b/gems/aws-sdk-dynamodbstreams/lib/aws-sdk-dynamodbstreams/customizations.rb
@@ -1,7 +1,0 @@
-# WARNING ABOUT GENERATED CODE
-#
-# This file is generated. See the contributing for info on making contributions:
-# https://github.com/aws/aws-sdk-ruby/blob/master/CONTRIBUTING.md
-#
-# WARNING ABOUT GENERATED CODE
-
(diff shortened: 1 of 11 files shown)


gems/aws-sdk-dynamodbstreams/lib/aws-sdk-dynamodbstreams/customizations.rb
```diff
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
```
mullermp, line 1: This file is code generated. Did you add it manually or update any code generation?

casperisfine: Manually, which suggest I might have missed a template somewhere?

mullermp: Checking out the build logic - it looks like there is no view for customizations (https://github.com/aws/aws-sdk-ruby/blob/master/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/code_builder.rb#L61)

I think that this generated code warning has been committed but then later removed for customizations files. I erased the contents of the file and ran `rake build:aws-sdk-dynamodbstreams` and it remained as a blank file. I think you can erase the contents here.

casperisfine: Done 🔥

mullermp:
Thanks for the contribution!

casperisfine:
Nah thanks to you!

PR #2335, opened 2020-06-16
Add customization for authtype until cognito updates their model

mullermp:
Fixes #1751

I'm not sure why this was done to begin with? Cognito still needs to update their model but this is a stop gap for now.

### DIFF
--- a/build_tools/customizations.rb
+++ b/build_tools/customizations.rb
@@ -193,5 +193,14 @@ module BuildTools
       end
     end
 
+    api('CognitoIdentityProvider') do |api|
+      operations = %w(
+        InitiateAuth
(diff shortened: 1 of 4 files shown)


mullermp:
Cognito team was finally able to track down where `authtype: 'none'` is set on their APIs.. This will be scrapped in favor of an SDK-wide fix, as well as fixes for the CognitoIdentity model for these operations:

```
GetCredentialsForIdentity
GetId
GetOpenIdToken
UnlinkIdentity
```

mullermp:
Changed my mind on waiting for cognito team. No harm merging this now with a temporary fix and removing it later when it's fixed properly.

PR #2337, opened 2020-06-16
Add frozen_string_literals comment in aws-sdk-core

casperisfine:

I simply ran:

```bash
rubocop -a --only Style/FrozenStringLiteralComment,Layout/EmptyLineAfterMagicComment gems/aws-sdk-core/
```

I also adjusted the line numbers in `region_spec.rb`.

That's pretty much the last big gem remaining, after that it's a couple files per gem.

@mullermp

### DIFF
--- a/gems/aws-sdk-core/features/env.rb
+++ b/gems/aws-sdk-core/features/env.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 # WARNING ABOUT GENERATED CODE
 #
 # This file is generated. See the contributing for info on making contributions:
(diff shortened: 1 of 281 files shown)


gems/aws-sdk-core/features/env.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 # WARNING ABOUT GENERATED CODE
```
casperisfine, line 3: I suppose this might be dead code as well?

mullermp: I don't think so.. I'll investigate soon.

mullermp: It looks used. Let's keep it for now - if it's blown away we can investigate.

mullermp:
Thanks for the contribution!

PR #2338, opened 2020-06-18
Add the last missing frozen_string_literals comments

casperisfine:
Followup: #2328

Since there isn't any big gem left, but many groups of a few files in many different gems, I opted for submitted all the last changes as a single PR.

I simply ran:

```bash
rubocop -a --only Style/FrozenStringLiteralComment,Layout/EmptyLineAfterMagicComment
```

I also adjusted the line numbers in `region_spec.rb`.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 source 'https://rubygems.org'
 
 if RUBY_VERSION == '1.9.3'
--- a/Rakefile
+++ b/Rakefile
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 $REPO_ROOT = File.dirname(__FILE__)
 $GEMS_DIR = "#{$REPO_ROOT}/gems"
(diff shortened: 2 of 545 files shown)


mullermp:
Uh oh, checking the build output, in both ec2 and s3 clients, this is shown in the docs:

```
     # @example Streaming a file from disk
+    #   # frozen_string_literal: true
+    #
```

Hopefully this wasn't merged in the previous CRs?

casperisfine:
Hopefully not, but that's the issue I witnessed in the original PR. If you have a few pointers as to where these snippets are extracted I could take a look at updating the code so that it ignores the `# frozen_string_literal: true`.

mullermp:
Ah! I think because the examples in `doc-src/examples/` are loaded as comments. You may want to exclude changes to those files.

mullermp:
Only s3 and ec2 have examples there, which makes sense.

casperisfine:
Makes total sense.I just removed it.

mullermp:
Thanks for the contribution!

PR #2341, opened 2020-06-19
Put SENSITIVE_PARAMS on request/response types instead of a giant list

### LINKED ISSUE #2312: list_tags_for_resource for Managed Instance does not return the list of tags

mullermp:
This code generates SENSITIVE_PARAMS on each request/response type. The log formatter will look for the type and apply sensitive fields if needed. This prevents over filtering when using a giant list when members can be shared across services.

This also provides an option to disable sensitive param filtering.

I messed up the history on #2325

Fixes #2312, #2105, #2082 


mullermp:
Huh... why did github just carry over everything?

PR #2342, opened 2020-06-19
Filter sensitive params per service/type rather than a global list

### LINKED ISSUE #2312: list_tags_for_resource for Managed Instance does not return the list of tags

mullermp:
Filters sensitive params only if they are marked in the type (via codegen)
Provide an option to disable sensitive param filtering if needed.

Fixes #2312, #2105, #2082

### DIFF
--- a/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/types_module.rb
+++ b/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/types_module.rb
@@ -269,11 +269,6 @@ module AwsSdkCodeGenerator
         def empty?
           @empty
         end
-
-        # @return [Boolean]
-        def sensitive_params?
-          @sensitive_params.any?
-        end
       end
 
       class StructMember
--- a/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/types_module.rb
+++ b/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/types_module.rb
@@ -48,9 +48,14 @@ module AwsSdkCodeGenerator
           if shape['eventstream']
             list
           elsif struct_type?(shape)
+            struct_members = struct_members(shape)
+            sensitive_params = struct_members.select(&:sensitive).map do |m|
+              m.member_name.to_sym
+            end
             list << StructClass.new(
               class_name: shape_name,
-              members: struct_members(shape),
+              members: struct_members,
+              sensitive_params: sensitive_params,
               documentation: struct_class_docs(shape_name)
             )
           else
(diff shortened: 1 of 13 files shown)


gems/aws-sdk-core/spec/aws/log/formatter_spec.rb
```diff
@@ -13,6 +13,17 @@ def format(pattern, options = {})
         Formatter.new(pattern, options).format(response)
       end
 
+      before do
```
mullermp, line 16: This is really gross - and also puts out many redefinitions.

gems/aws-sdk-core/spec/aws/log/formatter_spec.rb
```diff
@@ -13,6 +13,17 @@ def format(pattern, options = {})
...
+        class OperationRequestType < Struct
+          include Aws::Structure
+        end
+        OperationRequest = Seahorse::Model::Shapes::StructureShape.new(name: 'OperationRequest')
+        response.context.operation = Seahorse::Model::Operation.new.tap do |o|
+          o.input = Seahorse::Model::Shapes::ShapeRef.new(shape: OperationRequest)
+        end
+        OperationRequest.struct_class = OperationRequestType
+      end
```
mullermp, line 25: This is really gross and puts out many redefinitions

gems/aws-sdk-core/lib/aws-sdk-core/log/param_filter.rb
```diff
@@ -6,40 +6,40 @@
...
+        else values
         end
       end
 
       private
 
-      def filter_hash(values)
+      def filter_hash(values, type)
+        filters = @additional_filters
+        if type.const_defined?('SENSITIVE')
```
alextwoods, line 26: Check type nil since its the default on filter. eg: `if type && type.const_defined?`

gems/aws-sdk-core/spec/aws/log/formatter_spec.rb
```diff
@@ -7,12 +7,36 @@ module Aws
...
+          )
+        end
+        resp
+      end
 
       def format(pattern, options = {})
         Formatter.new(pattern, options).format(response)
       end
 
+      before do
```
alextwoods, line 36: Unused?

mullermp:
eventstream needs investigation, this could break

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/types_module.rb
```diff
@@ -253,23 +262,34 @@ def initialize(options)
...
+        attr_accessor :sensitive_params
+
         # @return [Boolean]
         def empty?
           @empty
         end
 
+        # @return [Boolean]
+        def sensitive_params?
+          @sensitive_params.any?
```
alextwoods, line 275: I don't think this is a problem - but if sensitive_params isn't set at all on line 247, this will give an error.  To be safe I'd suggest having fetch have a default of []?

mullermp:
Needs a plan for bumping versions..

Issue #2326, opened 2020-06-10
Retry streaming S3 object downloads using range


alextwoods:
**Is your feature request related to a problem? Please describe.**
Copying from #1617 : 

When downloading S3 objects with a block, aws-sdk-s3 internally creates a BlockIO to be the writable response target. Normally aws-sdk-s3 retries any failed downloads in case of network errors, but not for streaming downloads. This is because BlockIO is not truncatable, unlike a response target like StringIO or Tempfile, and aws-sdk-s3 needs it to be because it retries the whole download from the start.

**Describe the solution you'd like**
s3 get_object supports the range parameter. On retry of streaming bodies for the get_object request, we can use the number of received so far to add a range parameter, thus avoiding re-downloading data that has already been sent.

PR #2343, opened 2020-06-22
Streaming s3 retry

alextwoods:
This is an attempt to implement #2326 - and support for retry S3 get_objects that are streaming w/ a Proc using range to not ge-download redundant bytes.  

Issues:
* need to set range directly on the header (This does get correctly factored into the signiture).  Doesn't work to set it on the context.params[:range] because the request is already built when retried.
* Does not work if range is passed to get_object initially.  Range is a complex spec and would need to be parsed.


mullermp:
Needs testing of some sort.. looks good otherwise.

gems/aws-sdk-s3/spec/plugins/streaming_retry_spec.rb
```diff
@@ -0,0 +1,153 @@
+# frozen_string_literal: true
+
+require_relative '../spec_helper'
+
+module Aws
+  module S3
+    module Plugins
+      describe StreamingRetry do
+        let(:creds) { Aws::Credentials.new('akid', 'secert') }
```
mullermp, line 9: Super important spelling error

alextwoods: The funny thing is that I copied this from another spec.  I've gone and updated this across the package - it seems to have made it into a few places...

uthark:
Breaks with `uninitialized constant Aws::S3::Plugins::RetryableBlockIO::Forwardable`:
Opened PR to fix: https://github.com/aws/aws-sdk-ruby/pull/2348

PR #2345, opened 2020-06-23
Don't overwrite `host` header in sigv4 signer if given.

### LINKED ISSUE #2339: AWS::Sigv4::Signer overwrites 'host' header
  
**Describe the bug**
`AWS::Sigv4::Signer#sign_request`, when provided a `:headers` hash, overwrites the 'host' header. In my use case, I need to manually set the host in order to invoke a private API Gateway. Sending a request with a different host after creating the signature will invalidate the signature. I don't believe the `:unsigned_headers` option will help me here, because the 'host' header is [one of the ones that must be signed](https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html)

`aws-sigv4 (~> 1.0)`

`ruby 2.6.5p114 (2019-10-01 revision 67812) [x86_64-darwin18]`

Steps to reproduce the behavior (please share code or minimal repo)
```ruby
require 'aws-sigv4'

signed_request = Aws::Sigv4::Signer.new(
  service: 'execute-api',
  region: 'us-east-1',
  access_key_id: '',
  secret_access_key: ''
).sign_request(
  http_method: 'POST',
  url: 'http://wwww.domain.com/path',
  headers: { 'host' => 'host-header-value' },
  body: ''
)

signed_request.headers['host']
# => "wwww.domain.com"
```
**Expected behavior**
Expected the signer to create a signature using use the given 'host' value, if it exists 

**Screenshots**
N/A

**Additional context**
I've solved this locally by monkey-patching the method at the offending line here: https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sigv4/lib/aws-sigv4/signer.rb#L222

```ruby
sigv4_headers['host'] = headers['host'] || host(url)
```

Will post a PR with a patch shortly too, but I wanted to post this issue first to confirm that it is actually a bug and not just a misunderstanding of this method.

aphillips19:
Closes #2339 

Makes sigv4's `Signer#sign_request` keep the original Host header if that header was given, instead of overwriting it with a host header from the URL. 

This also updates `Signer#presign_url`, which looks like it would have the same issue. 

### DIFF
--- a/gems/aws-sigv4/spec/signer_spec.rb
+++ b/gems/aws-sigv4/spec/signer_spec.rb
@@ -146,7 +146,7 @@ module Aws
 
         context 'when a Host header is provided' do
           
-          let(:headers) { 'host' => 'custom-host-value' }
+          let(:headers) { {'host' => 'custom-host-value'} }
 
           it 'uses the original Host header' do
             signature = Signer.new(options.merge(headers: headers)).sign_request(request)
--- a/gems/aws-sigv4/spec/signer_spec.rb
+++ b/gems/aws-sigv4/spec/signer_spec.rb
@@ -146,12 +146,12 @@ module Aws
 
         context 'when a Host header is provided' do
           
-          let(:headers) { {'host' => 'custom-host-value'} }
+          let(:headers) { { headers: { 'host' => 'custom-host-value' } } }
 
           it 'uses the original Host header' do
-            signature = Signer.new(options.merge(headers: headers)).sign_request(request)
+            signature = Signer.new(options).sign_request(request.merge(headers))
 
-            expect(signature.headers['host']).to eql(custom-host-value)
+            expect(signature.headers['host']).to eql('custom-host-value')
           end
         end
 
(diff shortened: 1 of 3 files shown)


gems/aws-sigv4/spec/signer_spec.rb
```diff
@@ -144,6 +144,17 @@ module Sigv4
           expect(signature.headers['host']).to eq('domain.com')
         end
 
+        context 'when a Host header is provided' do
+          
+          let(:headers) { 'host' => 'custom-host-value' }
```
alextwoods, line 149: It looks like you need to wrap this in an additional {}, ie: `let(:headers) {  { 'host' => 'custom-host-value' } } `

mullermp:
Actually fix tests :D

aphillips19:
> Actually fix tests :D

Thank you for actually actually fixing this! I had trouble running tests locally and was relying on the CI pipeline to catch my errors

mullermp:
To run tests locally, you can use `rake test:spec`

mullermp:
Thanks for the contribution!

Issue #2103, opened 2019-08-19
Transfer Acceleration support for presigned post


adrenalin23:
# Issue description
Hello,
I'm using Transfer Acceleration. It works well with `Aws::S3::Presigner` if we specify `use_accelerate_endpoint` option:
```ruby
require 'aws-sdk-s3'

signer = Aws::S3::Presigner.new()
url = signer.presigned_url(:put_object, bucket: 'my_bucket', key: 'my_key', use_accelerate_endpoint: true)
```
But I can not find the way how to use Transfer Acceleration with `Aws::S3::PresignedPost`:
```ruby
Aws::S3::PresignedPost.new(creds, bucket_region, bucket_name, key: object_key, signature_expiration: expires, expires: expires)
```
I didn't find any docs how to implement it. Is it possible to use Transfer Acceleration with presigned post?
### Gem name ('aws-sdk', 'aws-sdk-resources' or service gems like 'aws-sdk-s3') and its version
aws-sdk-s3-1.36.0

### Version of Ruby
ruby 2.5.1

cjyclaire:
Thanks for the feedback!
This is a known [feature request](https://github.com/aws/aws-sdk-ruby/blob/master/FEATURE_REQUESTS.md#add-accelerate-endpoint-support-to-presigned-post-requests), tracking a a "+1" for the feature

adrenalin23:
@cjyclaire Great, thanks a lot for your feedback!

QMuncaster:
Also interested in this but `FEATURE_REQUESTS.md` 404s for me. Would be great to get another "+1" @cjyclaire .

cjyclaire:
@QMuncaster Thanks for chime in! We removed the feature request file and keep issue open for better visibility now :D your comments here totally works!

QMuncaster:
Great, thanks!

Could I also confirm that setting `use_accelerate_endpoint: true` on the S3 client isn't a workaround? The pre-signed POST url doesn't change, so I assumed not.

mullermp:
Sorry for the radio silence on this issue. A workaround is to pass `url: mybucket.s3-accelerate.amazonaws.com` into the presigned post class.

simonrussell:
I don't know if there's a documentation issue, but for me the behaviour hasn't been fixed -- creating a client with `use_accelerate_endpoint: true` still presigns POSTs with the non-accelerated endpoint.

mullermp:
[...]

PR #2352, opened 2020-07-07
Presigned post accelerate

mullermp:
Add #2103 

Also did some refactors

### DIFF
--- a/gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/accelerate.rb
+++ b/gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/accelerate.rb
@@ -24,6 +24,7 @@ each bucket. [Go here for more information](http://docs.aws.amazon.com/AmazonS3/
             :create_bucket, :list_buckets, :delete_bucket
           ]
           # Need 2 handlers so that the context can be set for other plugins
+          # and to remove :use_accelerate_endpoint from the params.
           handlers.add(
             OptionHandler, step: :initialize, operations: operations
           )
--- a/gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/accelerate.rb
+++ b/gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/accelerate.rb
@@ -3,36 +3,43 @@
 module Aws
   module S3
     module Plugins
-
       # Provides support for using `Aws::S3::Client` with Amazon S3 Transfer
       # Acceleration.
       #
       # Go here for more information about transfer acceleration:
       # [http://docs.aws.amazon.com/AmazonS3/latest/dev/transfer-acceleration.html](http://docs.aws.amazon.com/AmazonS3/latest/dev/transfer-acceleration.html)
       class Accelerate < Seahorse::Client::Plugin
-
-        option(:use_accelerate_endpoint,
+        option(
+          :use_accelerate_endpoint,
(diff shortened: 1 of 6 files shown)


gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/accelerate.rb
```diff
@@ -3,94 +3,80 @@
...
           DOCS
 
         def add_handlers(handlers, config)
           operations = config.api.operation_names - [
-            :create_bucket, :list_buckets, :delete_bucket,
+            :create_bucket, :list_buckets, :delete_bucket
           ]
-          handlers.add(OptionHandler, step: :initialize, operations: operations)
-          handlers.add(AccelerateHandler, step: :build, priority: 0, operations: operations)
+          # Need 2 handlers so that the context can be set for other plugins
```
alextwoods, line 26: Thats not quite true - we need the first handler to remove the option from the params during initialization so that it doesn't cause serialization to input to raise an ArgumentError.  I don't believe this is used by any other plugins

gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/accelerate.rb
```diff
@@ -3,94 +3,80 @@
...
-            endpoint.scheme = 'https'
-            endpoint.port = 443
-            endpoint.host = "#{bucket_name}.s3-accelerate.amazonaws.com"
-            context.http_request.endpoint = endpoint.to_s
-            # s3 accelerate endpoint doesn't work with 'expect' header
-            context.http_request.headers.delete('expect')
-          end
-
-          def use_combined_accelerate_dualstack_endpoint(context)
+          def use_accelerate_endpoint(context, dualstack)
```
alextwoods, line 60: Nice cleanup

gems/aws-sdk-s3/lib/aws-sdk-s3/presigned_post.rb
```diff
@@ -582,15 +611,16 @@ def check_required_values!
...
+          url.host = "#{@bucket_name}.#{url.host}"
         else
-          url.path = '/' + @bucket_name
+          url.path = "/#{@bucket_name}"
         end
         if @bucket_region == 'us-east-1'
           # keep legacy behavior by default
           url.host = Plugins::IADRegionalEndpoint.legacy_host(url.host)
         end
+        url.host = "#{@bucket_name}.s3-accelerate.amazonaws.com" if @accelerate
```
alextwoods, line 623: Does this ever conflict with the legacy behavior above for us-east-1?

mullermp: It shouldn't. s3 accelerate isn't regional. The legacy behavior will override the host so I have to do this after that. `legacy_host` basically does a gsub of `us-east-1` out of the host.

gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/accelerate.rb
mullermp: It's used by dualstack and others (quite possibly ARN support in the future too). I will update for both cases.

mullermp:
Those tests are only for `Aws::S3::Presigner`

PR #2353, opened 2020-07-07
Request Callbacks: on_chunk_sent

### LINKED ISSUE #648: Progress callback for S3Object

alextwoods:
Fixes #648 

Adds a new plugin, `RequestCallback` which will call a provided block with size as each chunk is read by the request.  This applies to all operations on all Clients EXCEPT async clients.  

## Testing/Usage
```ruby
res = s3_client.put_object(
  bucket: test_bucket,
  key: test_key, 
  body: huge_body, 
  progress_callback: Proc.new { |read_bytes, total| puts "Progress: #{100.0 * read_bytes / total}%"} 
)
```

Outputs:
```
Progress: 20.507685375256596%
Progress: 41.01537075051319%
Progress: 61.52305612576979%
Progress: 82.03074150102638%
Progress: 100.0%

```


gems/aws-sdk-core/lib/seahorse/client/plugins/progress_callback.rb
```diff
@@ -0,0 +1,63 @@
...
+          def call(context)
+            if (callback = context[:progress_callback])
+              context.http_request.body = ProgressTrackingBody.new(context.http_request.body, callback)
+            end
+            @handler.call(context)
+          end
+        end
+
+        handler(OptionHandler, step: :initialize)
+        handler(ProgressCallbackHandler, step: :sign, priority: 0)
```
mullermp, line 59: Why the :sign step and not the default :build step?

alextwoods: Added comments for this as well - it needs to the be sign step, and a low priority because it needs to go after all of the plugins which read the request body (eg content_md5 and sigv4).

gems/aws-sdk-core/lib/seahorse/client/plugins/request_callback.rb
```diff
@@ -0,0 +1,87 @@
...
+          def call(context)
+            if (callback = context[:on_chunk_sent])
+              context.http_request.body = ReadCallbackIO.new(context.http_request.body, callback)
+              add_event_listeners(context)
+            end
+            @handler.call(context)
+          end
+
+          def add_event_listeners(context)
+
```
mullermp, line 68: space

gems/aws-sdk-core/lib/seahorse/client/plugins/request_callback.rb
```diff
@@ -0,0 +1,87 @@
...
+            context.http_response.on_headers do
+              body = context.http_request.body
+              if body.is_a? ReadCallbackIO
+                context.http_request.body = body.io
+              end
+            end
+          end
+        end
+
+        handler(OptionHandler, step: :initialize)
```
mullermp, line 79: Most of our options are in the build step. Why initialize here?

gems/aws-sdk-core/lib/seahorse/client/plugins/request_callback.rb
```diff
@@ -0,0 +1,87 @@
...
+        # This handler needs to go late in the call stack
+        # other plugins including Sigv4 and content_md5 read the request body
+        # and rewind it
+        handler(ReadCallbackHandler, step: :sign, priority: 0)
```
mullermp, line 83: Why the sign step here?

gems/aws-sdk-core/spec/aws/request_callback_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+require_relative '../spec_helper'
+
+# Although this plugin lives in Seahorse it needs to be tested against
+# the Retry Plugin in the AWS namespace
+module Seahorse
+  module Client
+    module Plugins
+      describe RequestCallback do
+
```
mullermp, line 11: space

gems/aws-sdk-core/spec/aws/request_callback_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+        let(:client_class) do
+          ApiHelper.sample_service.const_get(:Client)
+        end
+
+        let(:callback) { Proc.new {|chunk, read, total| @call_count += 1} }
+        let(:client) { client_class.new(on_chunk_sent: callback, region: 'us-west-1', endpoint: 'http://foo.com') }
```
mullermp, line 17: multiline

gems/aws-sdk-core/spec/aws/request_callback_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+              {status: 500, body: 'error'},
+              {status: 200, body: ''}
+            )
+
+          client.example_operation
+          expect(@call_count).to eq(2)
+        end
+
+        it 'it can be used as a parameter on the operation' do
+
```
mullermp, line 50: space

gems/aws-sdk-core/lib/seahorse/client/plugins/request_callback.rb
mullermp: NVM I see your other response.

gems/aws-sdk-core/lib/seahorse/client/plugins/request_callback.rb
alextwoods: Added a comment - same reason the OptionHandler in the accelerate plugin is on intialize - it needs to remove `:on_chunk_sent` from the parameters before build.

gems/aws-sdk-core/lib/seahorse/client/plugins/request_callback.rb
```diff
@@ -0,0 +1,87 @@
...
+When a Proc object is provided, it will be used as callback when each chunk 
+of the request body is sent. It will be called with three arguments: the chunk,
+the number, of bytes read from the body, and the total number of 
+bytes in the body.
+          DOCS
+
+        # @api private
+        class OptionHandler < Client::Handler
+          def call(context)
+            if context.params.is_a?(Hash)
```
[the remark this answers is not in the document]
alextwoods: Never in the case of a service client.  However, since this plugin is in Seahorse rather than AWS, its possible (see the ResponseTarget plugin).

gems/aws-sdk-core/lib/seahorse/client/plugins/request_callback.rb
```diff
@@ -0,0 +1,87 @@
...
+              on_chunk_sent = context.params.delete(:on_chunk_sent)
+            end
+            on_chunk_sent = context.config.on_chunk_sent if on_chunk_sent.nil?
```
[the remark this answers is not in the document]
alextwoods: Yeah, thats a good question.  I don't think it makes sense from a progress reporting use case to have it as client config.  However, I could see someone wanting to record information about every byte sent?  Additionally - I couldn't think of another way to expose this as an option in the documentation easily without having it as a config param.  Thoughts?

gems/aws-sdk-core/lib/seahorse/client/plugins/request_callback.rb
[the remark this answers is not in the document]
alextwoods: Just updating this for posterity - this is actually not because its a Seahorse plugin.  This check is useful because the `OptionHandler` is in the `:initialize` step, which happens before `:validate`.  If someone calls `client.operation('not_a_hash')` then `params` will be a string rather than a hash at this point.  This is not a valid call, and the `ParamValidator` will raise a user friendly `ArgumentError`.

PR #2357, opened 2020-07-13
Pass headers to streaming block

### LINKED ISSUE #1913: Feature request: Retrieving data from response headers when streaming S3 objects
In my open source project I'm using `Aws::S3::Object#get` with a block to stream the S3 object on-demand:

```rb
chunks = object.enum_for(:get, **options)
chunks #=> #<Enumerator>
chunks.next # first chunk of data
chunks.next # next chunk of data
```

I'm wrapping this functionality into a custom IO-like object, so that people can stream using the `IO#read` semantics, which under the hood retrieves subsequent chunks of data when needed.

I'd like the content length to be known at the start of streaming, which would be returned when the user called `#size`. Currently I have to make an additional `#head_object` request to retrieve `Aws::S3::Object#content_length`. However, the `#get_object` request that `Object#get` makes already retrieves response headers before it starts streaming, which I think hold all the same information that response headers for a `#head_object` call would hold.

So, it would be convenient if `Object#content_length`, `#content_type` and other data was filled in on `Object#get` as soon as the response headers are received (before the response body is read). That way I could avoid the additional HEAD request to S3 for this use case, to retrieve data that we technically already have, but which can only be retrieved once the whole S3 object has been downloaded.

In code, I would like this behaviour:

```rb
object.get do |chunk|
  # this currently makes an additional HEAD request, but doesn't have to,
  # because it could fill this info in from the response headers of the
  # `#get_object` call that we already retrieved at this point
  object.content_length
end
```

I think it should be possible to somehow use the notification about response headers that Seahorse transmits: https://github.com/aws/aws-sdk-ruby/blob/e5ef6389dcf36ff23586f5891b0914458a50c5c7/gems/aws-sdk-core/lib/seahorse/client/net_http/handler.rb#L80

alextwoods:
Fixes #1913 

### DIFF
--- a/doc-src/examples/s3/client/get_object/03_streaming_data_to_a_block.rb
+++ b/doc-src/examples/s3/client/get_object/03_streaming_data_to_a_block.rb
@@ -1,6 +1,8 @@
 # WARNING: yielding data to a block disables retries of networking errors
+# However truncation of the body will be retried automatically using a range request
 File.open('/path/to/file', 'wb') do |file|
-  s3.get_object(bucket: 'bucket-name', key: 'object-key') do |chunk|
+  s3.get_object(bucket: 'bucket-name', key: 'object-key') do |chunk, headers|
+    # headers['content-length']
     file.write(chunk)
   end
 end
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,7 @@
 Unreleased Changes
 ------------------
 
+* Feature - Add headers to the `ResponseTarget` callback.  A block passed as the response target on a streaming method will be called with the `chunk` and `headers`.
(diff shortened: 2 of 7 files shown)


doc-src/examples/s3/client/get_object/03_streaming_data_to_a_block.rb
```diff
@@ -1,6 +1,8 @@
 # WARNING: yielding data to a block disables retries of networking errors
+# However truncation of the body will be retried automatically using a range request
 File.open('/path/to/file', 'wb') do |file|
-  s3.get_object(bucket: 'bucket-name', key: 'object-key') do |chunk|
+  s3.get_object(bucket: 'bucket-name', key: 'object-key') do |chunk, headers|
+    # headers['content-length']
```
mullermp, line 5: Was this intended to be commented in the example?

alextwoods: Yeah - intended to document the parameter, but its not actually used here.

Issue #1907, opened 2018-11-05
doesn't yet use .aws/config ca_bundle


mmattice:
Please fill out the sections below to help us address your issue

### Issue description
aws-cli uses ca_bundle in `~/.aws/config` [aws-cli usage](https://github.com/aws/aws-cli/blob/69208fe970e2b9302e4ce157c68e98deeb6a9dd1/awscli/customizations/globalargs.py#L69)
aws-sdk-ruby does not

### Gem name ('aws-sdk', 'aws-sdk-resources' or service gems like 'aws-sdk-s3') and its version
aws-sdk-core 3.36.0

### Version of Ruby, OS environment
ruby 2.5.3p105 (2018-10-18 revision 65156) [x86_64-darwin17]

### Code snippets / steps to reproduce
Be behind corporate SSL MITM firewall.  Use `ca_bundle` in `~/.aws/config` to allow aws-cli to work.  Try to run something ruby based, get failures due to lack of use of configured bundle.

```
	 7: from /usr/local/lib/ruby/gems/2.5.0/gems/aws-sdk-core-3.36.0/lib/seahorse/client/net_http/connection_pool.rb:96:in `session_for'
	 6: from /usr/local/lib/ruby/gems/2.5.0/gems/aws-sdk-core-3.36.0/lib/seahorse/client/net_http/connection_pool.rb:297:in `start_session'
	 5: from /usr/local/Cellar/ruby/2.5.3/lib/ruby/2.5.0/delegate.rb:83:in `method_missing'
	 4: from /usr/local/Cellar/ruby/2.5.3/lib/ruby/2.5.0/net/http.rb:915:in `start'
	 3: from /usr/local/Cellar/ruby/2.5.3/lib/ruby/2.5.0/net/http.rb:920:in `do_start'
	 2: from /usr/local/Cellar/ruby/2.5.3/lib/ruby/2.5.0/net/http.rb:981:in `connect'
	 1: from /usr/local/Cellar/ruby/2.5.3/lib/ruby/2.5.0/net/protocol.rb:44:in `ssl_socket_connect'
/usr/local/Cellar/ruby/2.5.3/lib/ruby/2.5.0/net/protocol.rb:44:in `connect_nonblock': SSL_connect returned=1 errno=0 state=error: certificate verify failed (self signed certificate in certificate chain) (Seahorse::Client::NetworkingError)
```

mullermp:
[...]

PR #2361, opened 2020-07-16
Support AWS_CA_BUNDLE and ca_bundle config option

mullermp:
Adds #1907 


gems/aws-sdk-core/lib/seahorse/client/plugins/net_http.rb
```diff
@@ -21,7 +21,10 @@ class NetHttp < Plugin
 
         option(:ssl_verify_peer, default: true, doc_type: 'Boolean', docstring: '')
 
-        option(:ssl_ca_bundle, default: nil, doc_type: String, docstring: '')
+        option(:ssl_ca_bundle, doc_type: String, docstring: '') do |cfg|
```
alextwoods, line 24: Does this need to be updated in the h2 plugin as well?

mullermp: Probably, yeah, i'll do that.

grim-fendango:
Bump

mullermp:
Thanks for the bump. The status of this PR is to check if this is something we should be changing and if it could possibly break customers. Currently only CLI supports this ENV variable. It's common for CLI and Ruby to be used together.

gems/aws-sdk-core/lib/seahorse/client/plugins/h2.rb
```diff
@@ -28,11 +28,14 @@ class H2 < Plugin
         DOCS
 
         # SSL Context
-        option(:ssl_ca_bundle, default: nil, doc_type: String, docstring: <<-DOCS)
+        option(:ssl_ca_bundle, default: nil, doc_type: String, docstring: <<-DOCS) do |cfg|
 Full path to the SSL certificate authority bundle file that should be used when
 verifying peer certificates. If you do not pass `:ssl_ca_directory` or `:ssl_ca_bundle`
 the system default will be used if available.
         DOCS
+          ENV['AWS_CA_BUNDLE'] ||
```
tareksamni, line 36: My understanding is that it's better (good practice) to use `fetch` method for such logic. Something like:

```
ENV.fetch('AWS_CA_BUNDLE', Aws.shared_config.ca_bundle(profile: cfg.profile)) if cfg.respond_to?(:profile)
```

mullermp: I don't think that can be done here because we can only fall back to the shared config if the config responds to profile.

tareksamni: I'm still using `if cfg.respond_to?(:profile)` which guarantees that `we can only fall back to the shared config if the config responds to profile`. It's just the way we fall back, instead of using `||` I was encouraging the the PR author to use `Hash#fetch`.

alextwoods: I'd agree in general its preferred to use `fetch` for fallback ( since cases where the value are set to false will fallback with `||`).   However I think you're code is incorrect, and needs to be:
```ruby
 ENV.fetch('AWS_CA_BUNDLE', (Aws.shared_config.ca_bundle(profile: cfg.profile) if cfg.respond_to?(:profile))) # parens around the default value w/ if
```

However, for ENV the values are restricted to strings so the issues with `||` don't apply.  The rest of the code uses `||` with ENV so I'd prefer to keep that style consistent here.

alextwoods (approved):

gems/aws-sdk-core/lib/seahorse/client/plugins/h2.rb
mullermp: Yes agreed. When I meant "I don't think that can be done here" it was because the if statement was outside of ENV.fetch. It adds extra complexity here in this case.

grim-fendango (approved):

grim-fendango:
bump

PR #2367, opened 2020-07-20
Dup params to prevent modification of user provided params

### LINKED ISSUE #2366: Occur `FrozenError` when passed frozen hash to `DynamodbClient.resource.create_table`
  
**Describe the bug**

When passed frozen hash to `DynamodbClient.resource.create_table`, be raised `FrozenError`.
This happened since aws-sdk-core v3.104.0 and it doesn't happen before v3.103.0.

aws-sdk-core (3.104.0) and aws-sdk-dynamodb (1.51.0).

ruby 2.6.6p146 (2020-03-31 revision 67876) [x86_64-linux]

```ruby 
TABLE_DEFINITION = {
  table_name: "test_table",
  key_schema: [
    {
      attribute_name: "activity_id",
      key_type: "HASH"
    }
  ],
  attribute_definitions: [
    {
      attribute_name: "activity_id",
      attribute_type: "S"
    }
  ]
}.freeze

DynamodbClient.resource.create_table(TABLE_DEFINITION)
# => /home/y-yagi/.rbenv/versions/2.6.6/lib/ruby/gems/2.6.0/gems/aws-sdk-core-3.104.0/lib/seahorse/client/plugins/request_callback.rb:67:in `delete': can't modify frozen Hash (FrozenError)

```

**Expected behavior**

The exception doesn't be raised.

**Additional context**

Maybe releated with #2353.

alextwoods:
Fixes #2366 

### DIFF
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Duplicate params to ensure user provided params are not modified.  Fixes #2366
+
 3.104.0 (2020-07-15)
 ------------------
 
--- a/gems/aws-sdk-core/lib/seahorse/client/request_context.rb
+++ b/gems/aws-sdk-core/lib/seahorse/client/request_context.rb
@@ -19,7 +19,7 @@ module Seahorse
         @operation = options[:operation]
         @authorizer = options[:authorizer]
         @client = options[:client]
-        @params = options[:params].dup || {}
+        @params = options[:params] ? options[:params].dup : {}
         @config = options[:config]
         @http_request = options[:http_request] || Http::Request.new
         @http_response = options[:http_response] || Http::Response.new
--- a/gems/aws-sdk-core/lib/seahorse/client/request_context.rb
+++ b/gems/aws-sdk-core/lib/seahorse/client/request_context.rb
@@ -19,7 +19,7 @@ module Seahorse
         @operation = options[:operation]
         @authorizer = options[:authorizer]
         @client = options[:client]
-        @params = options[:params] || {}
+        @params = options[:params] ? options[:params].dup : {}
         @config = options[:config]
         @http_request = options[:http_request] || Http::Request.new
         @http_response = options[:http_response] || Http::Response.new
--- a/gems/aws-sdk-core/lib/seahorse/client/request_context.rb
+++ b/gems/aws-sdk-core/lib/seahorse/client/request_context.rb
@@ -19,7 +19,7 @@ module Seahorse
         @operation = options[:operation]
         @authorizer = options[:authorizer]
         @client = options[:client]
-        @params = options[:params] ? options[:params].dup : {}
+        @params = options[:params].dup || {}
         @config = options[:config]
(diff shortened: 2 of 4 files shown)


numberten:
Unsure how much you care about compatibility between different gems of varying versions, but this change broke my existing S3 code (using `gem 'aws-sdk-s3', '1.60.2'`, but should apply to any version prior to #2289).

Specifically, when signing a presigned url with the `expires_in` parameter, if the [`ParamValidator`](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/lib/aws-sdk-core/plugins/param_validator.rb) finds an `expires_in` key while validating a `PutObjectRequest` it will raise an error:
> ArgumentError: unexpected value at params[:expires_in]

Prior to the changes in #2289, `build_request()` is called before `expires_in()`: https://github.com/aws/aws-sdk-ruby/blob/8ed508960aad9fc1923a892b252f55313314da11/gems/aws-sdk-s3/lib/aws-sdk-s3/presigner.rb#L86-L90

Since `expires_in()` calls `params.delete(:expires_in)` it removes the key in both `params` as well as `req.context.params`: https://github.com/aws/aws-sdk-ruby/blob/8ed508960aad9fc1923a892b252f55313314da11/gems/aws-sdk-s3/lib/aws-sdk-s3/presigner.rb#L103-L117

With this PR applied `build_request()` makes a copy of `params` such that `params.object_id != req.context.params.object_id`. The call to `expires_in()` no longer changes `req.context.params` which causes it to fail validation via `ParamValidator`.

None of this is a problem if all gems are kept up-to-date because the newest `aws-sdk-s3` version calls `expires_in()` prior to `build_request()`: https://github.com/aws/aws-sdk-ruby/blob/009292ceefa5c6152cdd8296326f7f6cc0e99f4f/gems/aws-sdk-s3/lib/aws-sdk-s3/presigner.rb#L137-L139

I'm leaving this comment more as a solution for others that find themselves with the edge case described above.

alextwoods:
Thanks for the detailed report!  The change was meant to be backwards and forwards compatible - I just missed this case.  I'll investigate a possible fix.

lizthegrey:
Thanks @alextwoods and @numberten, we just had our chef bootstrap unexpectedly broken by this and were stumped until we bisected and found you'd stumbled into this before us and written up the bug super clearly!

I can confirm reverting this pull in our unpacked gem cache restores things to working order, but a longer term solution is desirable. FYI @daiyi @toshok @martin308 who worked the incident with me.

mullermp:
@numberten When using an earlier core version, the change in #2289 doesn't break you, correct?

numberten:
> @numberten When using an earlier core version, the change in #2289 doesn't break you, correct?

I never tried a new `aws-sdk-s3` (incorporating changes from #2289) with an earlier core. But from the code it doesn't appear that it would have broken me (at least not in this way).

alextwoods:
Unfortunately I believe the only way to fix this is to revert #2367 - I don't like that plugins modify user provided parameters but I don't see an alternative fully forward/backward compatible solution (and this prevents us from accepting frozen parameters in some cases.  See: #2366). 
PR to revert:  #2368

mullermp:
We're really sorry about this. This is not something we intended.

PR #2390, opened 2020-08-14
Delete reference to deleted file

ianfixes:
`FEATURE_REQUESTS.md` was deleted in a0d4497c44359d74942a46114ccc28a29205ed6c so it shouldn't be referenced here

Omitting CHANGELOG.md entry as this is a typo fix in a non-code file.


mullermp:
Thanks

PR #2392, opened 2020-08-14
Determine the resource type

ianfixes:
Attempt to render a resource type according to the documentation:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html

There's got to be a better way to get the resource type of a given ARN, but I haven't been able to find it.

>By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core/arn.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/arn.rb
@@ -59,6 +59,29 @@ module Aws
     # @return [String]
     attr_reader :resource
 
+    # Determine the "official" name of a resource
+    #
+    # This function attempts to render a resource name in this form
+    # https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html
+    #
+    # Barring an API call that accomplishes this, we guess the capitalization by introspection
+    # on the AWS SDK class names
+    def resource_type
+      # look for a match in the defined constants, to get proper capitalization. default capitalize first letter
+      guess_capitalization = Proc.new do |a_module, member_name|
+        a_module.constants.map(&:to_s).find(member_name.method(:capitalize)) { |t| member_name.casecmp(t) == 0 }
+      end
+
+      # extract the portion of resource type that's contained in the ARN. split on either : or /
+      local_type = resource.split(/[:\/]/)[0]
+
+      # look for a match in the defined constants, to get proper capitalization.  Same for resource
+      official_service = guess_capitalization.call(Aws, service)
+      official_resource = guess_capitalization.call(Object.const_get(official_service), local_type
+
+      ["AWS", official_service, official_resource].join("::")
+    end
+    
     # Validates ARN contains non-empty required components.
     # Region and account_id can be optional.
     #


mullermp:
Hey, thanks for opening this pull request. Unfortunately I don't think it's a good idea to merge this. When this class was created, the concept of parsing resources for ARNs was left out intentionally because it currently differs per service what the resource format can or would be (your example refers to just cloud formation). For example, in general, the ARN can have a colon or slash, and some may contain paths. When and if resource portions of an ARN were formally defined, we didn't want to have to make a potentially breaking change.

I can double check with the SDK teams about this. If this were to be implemented, the implementation would be more light weight (i.e. just splitting colon or slash and potentially populating a Resource class).

ianfixes:
No problem, I understand -- if it were straightforward, this is the class where it would be already.  And you proved correct my hunch, that there isn't an existing way to get this data.  This doesn't block me, I can always just put this function (or any repeatable way to get a resource_type identifier) in my own code and change it later.  

If helps a conversation with the SDK team, this is my

### Use case

Several documents exist about the purpose, importance, and best practices for tagging AWS resources:

* [awsstatic.com whitepapers "Tagging Best Practices" / "Implement an Effective AWS Resource Tagging Strategy" [pdf]](https://d1.awsstatic.com/whitepapers/aws-tagging-best-practices.pdf)
  >  if a significant portion of your AWS resources are missing tags used for cost allocation, your cost analysis and reporting
process will be more complicated and time-consuming, and probably less accurate.
* [metricly.com "AWS Tagging Best Practices – The Ultimate Guide"](https://www.metricly.com/aws-tagging-best-practices/)
  > In an ideal world, an organization would determine a tagging strategy, establish a method of applying tags which mitigates user errors, and only then begin deploying resources into their AWS account. The real world is seldom ideal
* [cloudforecast.io "AWS Tagging Best Practices Guide: Part 1 of 3"](https://www.cloudforecast.io/blog/aws-tagging-best-practices/)
  >  you have to consider how tags will change over time
* [amazon.com "Tagging AWS resources": "Best Practices" ](https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html#tag-best-practices)
  > Automation tags are used to opt in or opt out of automated tasks or to identify specific versions of resources to archive, update, or delete.

**TL;DR there are financial and resouce-deleting consequences to having tags that are missing and incorrect, and on top of that the tagging strategy itself may need to change over time**

The recommended long-term solution to this problem seems to be: 
1. to rely on CloudFormation templates "to apply tags to certain resource types upon creation"
2. to configure the AWS Service Catalog to "specify required tags as well as their range of allowable values"
3. "For resources not created via AWS CloudFormation, you can still implement automation to automatically propagate tags from related resources. [... e.g.] use CloudWatch Events to trigger a Lambda function to copy tags any time objects are created"
4. to "**employ reactive tag governance approaches** to identify resources that are not properly tagged and correct them. Reactive tag governance approaches include programmatically using tools such as the Resource Tagging API, AWS Config rules, and **custom scripts**" (emphasis mine)

As a DevOps engineer, I want to develop those "custom scripts" in Ruby.  In addition to providing some "reactive governance",  auditing the current state of tagging against current or future tagging strategy specs, detecting errors, and remediating them immediately is a valuable ability.  (And when needed, faster than rolling out remediation through CI/CD or rule adjustments).  **Writing these scripts becomes simpler with an ability to determine a resource's type** based solely on ARN, because tag-related operations can be centralized in the codebase.

----

I appreciate how quickly you responded to this, and thanks for all you're doing here.

mullermp:
So if I'm understanding correctly, your use case is to inspect and parse any service ARN (such as one from S3), and determine a CloudFormation specific resource type (in the form `service-provider::service-name::data-type-name`) so that you can tag your resources?

ianfixes:
@mullermp Yes, that's exactly right.  I want write code in terms of the official `service-provider::service-name::data-type-name` values.  It will give me the ability to look up some internal rules on "what should the tags for this type be?" and report or adjust accordingly.

mullermp:
I see. Just to clarify, this ARN class is part of `aws-sdk-core`. I wouldn't call those resource types as "official values" but they seem to only be standard in cloud formation and not across other services. An example would be [S3's access point ARN](https://docs.aws.amazon.com/AmazonS3/latest/dev/using-access-points.html) such as `arn:aws:s3:us-west-2:123456789012:accesspoint/test` where the resource_type part of the ARN is "accesspoint" and resource_name is "test".

ianfixes:
Sounds like I might be opening a can of worms.  I am indeed making the assumption that "the official `service-provider::service-name::data-type-name` values" align with the resource type as specified in a CloudFormation template.  

Am I hearing that "it's more complex than that"?  That would be a bit daunting.

mullermp:
Not quite. The current documentation on ARNs is [here](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html). In the case of any ARN, `resource` method in the [ARN class](https://github.com/ianfixes/aws-sdk-ruby/blob/008f99790f95f85526a36050ca06d5f64fe93296/gems/aws-sdk-core/lib/aws-sdk-core/arn.rb#L44) would return a string that may be just some resource-id or a resource-type/resource-id, but the small list here is not complete. That's all this class was intended to do, was to just parse out the six segments of the ARN. How we parse the `resource` is the open question here, and what makes an ARN "valid" (including the resource portion)? In this example: `arn:aws:iam::123456789012:user/Development/product_1234/*` the resource type would be "user" and the resource id would be "/Development/product_1234/*".. but is it? A new S3 feature that's in the works will have two "types" in its resource portion of the ARN, so any generic functionality that would have parsed the resource into a singular resource-type and resource-id would have been incorrect. Leaving that to customers is the safest option for backwards compatibility.

Your case is slightly different. You want to take an ARN of some sort and derive some cloud formation specific identifier (AWS::Service::DataType). This is not just limited to ARN input, it could also be derived from the module namespaces from the  Clients (i.e. Aws::S3::Types::Bucket (Ruby class) => "AWS::S3::Bucket" (string used by cloud formation). Perhaps some functionality could be added to the cloud formation gem to facilitate with getting those.. but it would maybe involve keeping a list that could be out of date, because I don't see an API to retrieve a current set. The documentation in `[update_stack](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/CloudFormation/Client.html#update_stack-instance_method)` suggests that you can use wildcards (i.e. `AWS::S3::*`) unless you know the exact resource type.

mullermp:
What happens if a cloud formation resource type is set to some unknown resource? (let's say S3 has a new feature called super bucket, so what happens with AWS::S3::SuperBucket, as opposed to AWS::S3::Bucket)

ianfixes:
> What happens if a cloud formation resource type is set to some unknown resource? (let's say S3 has a new feature called super bucket, so what happens with AWS::S3::SuperBucket, as opposed to AWS::S3::Bucket)

It sounds like you're suggesting that CloudFormation is able to create resources even if it doesn't know what the types are, which sounds ... frightening.  I had assumed there was some built-in knowledge of it because `cfn-lint` checks that resource types are known.


> Your case is slightly different. You want to take an ARN of some sort and derive some cloud formation specific identifier (AWS::Service::DataType).

Well, I am looking for any way to ask AWS for the type of resource that an ARN describes, such that the answer is both definitive and consistent.  I had been under the flawed assumption that CloudFormation's names for these resource types were "standard across services" -- thank you for setting me straight on that, even though it's back to the drawing board.


> Perhaps some functionality could be added to the cloud formation gem to facilitate with getting those.. but it would maybe involve keeping a list that could be out of date, because I don't see an API to retrieve a current set. 

It sounds like this might be an upstream problem, do you know if there's an issue/ticket I could subscribe to for creating such an API feature?  I saw a reference to a `aws cloudformation list-type-versions --type "RESOURCE" ` [on this other issue](https://github.com/aws-cloudformation/cloudformation-cli/issues/366#issuecomment-564690074), but this command apparently isn't available in my CLI.

mullermp:
> It sounds like you're suggesting that CloudFormation is able to create resources even if it doesn't know what the types are, which sounds ... frightening. I had assumed there was some built-in knowledge of it because `cfn-lint` checks that resource types are known.

I wasn't suggesting that.. I wasn't actually sure, I was genuinely asking. I'm glad it checks the types.

>I saw a reference to a aws cloudformation list-type-versions --type "RESOURCE" on this other issue, but this command apparently isn't available in my CLI.

It's listed in the Ruby SDK too, so if you don't have it in CLI, you may need to update your CLI version to the latest and I'm sure it will be there.

> Well, I am looking for any way to ask AWS for the type of resource that an ARN describes, such that the answer is both definitive and consistent.
> It sounds like this might be an upstream problem, do you know if there's an issue/ticket I could subscribe to for creating such an API feature? 

So playing around with their API, there is a way to list out all of the AWS types using the [`list_types`](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/CloudFormation/Client.html#list_types-instance_method) API

`cloudformation.list_types(visibility: 'PUBLIC')`

It will list all of the types listed on that documentation page as well as their ARN and type name string.

You can use [`describe_type`](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/CloudFormation/Client.html#describe_type-instance_method) to single out one of them:

`cloudformation.describe_type(type: 'RESOURCE', arn: 'arn:aws:cloudformation:us-west-2::type/resource/AWS-CloudFront-CloudFrontOriginAccessIdentity')`

This API seems to *only* accept other cloud formation ARNs, so if you're able to use cloud formation ARNs, you can get the type this way, and it seems to be their preferred method.

What this API *doesn't do* is allow you to pass an S3 access point ARN and retrieve AWS::S3::AccessPoint type. If that is the type of ARN that you must work with, we'd have to submit a feature request and hope they prioritize it.

ianfixes:
I meant to reply to this sooner, but the end of the summer months was a busy time 😅 

I submitted this PR thinking that there was a simple way to unify the ARNs, AWS types, and Ruby types... it sounds like I was pretty far off there.  And even if we were to put resources toward resolving it in any "proper" sense (with direct support from AWS), I would have to come up with an alternate solution in the meantime.

So in the meantime I am working on my own alternate solution.  I'm still interested in pursuing a feature request for this, but I'm not sure where to submit that nor the subtleties of what I'm actually asking for.  (Or to say it another way, you have a better vocabulary for what would go into such a request than I do).  If you make that feature request, can you link it here?

mullermp:
We can make a feature request for this, feel free to open it as an issue and you can tag this PR as correspondence. It sounds like what you want is a utility in the cloudformation gem that converts any service ARN into a cloudformation type identifier.

input (iam ARN): `arn:aws:iam::123456789012:user/Development/product_1234/*`
output (cloudformation type): `AWS::IAM::User`

input (s3 ARN): `arn:aws:s3:::my_corporate_bucket/*`
output (cloudformation type): `AWS::S3::Bucket`

PR #2394, opened 2020-08-17
SSOCredentials Provider

alextwoods:


gems/aws-sdk-core/lib/aws-sdk-core/sso_credentials.rb
```diff
@@ -0,0 +1,120 @@
...
+    #   login to the SSO directory. This is also sometimes referred to as
+    #   the "User Portal URL"
+
+    # @option options [SSO::Client] :client Optional `SSO::Client`.  If not
+    #   provided, a client will be constructed.
+    def initialize(options = {})
+      client_opts = {}
+
+      missing_keys = SSO_REQUIRED_OPTS.select { |k| options[k].nil? }
+      unless missing_keys.empty?
```
mullermp, line 55: you could use `.any?` here

gems/aws-sdk-core/lib/aws-sdk-core/sso_credentials.rb
```diff
@@ -0,0 +1,120 @@
...
+      @sso_region = options.delete(:sso_region)
+      @sso_role_name = options.delete(:sso_role_name)
+      @sso_account_id = options.delete(:sso_account_id)
+
+      # validate we can read the token file
+      read_cached_token
+
+      options.each_pair do |key, value|
+        client_opts[key] = value
+      end
```
mullermp, line 69: At this point, why make `client_opts`? Once you've pulled out any real options (using `delete`) you can just use `options[:client]` or `SSO::Client.new(options)`

gems/aws-sdk-core/spec/aws/sso_credentials_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+        expect(SSO::Client).not_to receive(:new)
+
+        mock_token_file(sso_start_url, cached_token)
+
+        creds = SSOCredentials.new(sso_opts.merge(client: test_client))
+        expect(creds.client).to be(test_client)
+      end
+    end
+
+    # describe '#identity_id' do
```
mullermp, line 100: is this intended to be uncommented?

gems/aws-sdk-core/spec/aws/credential_resolution_chain_spec.rb
```diff
@@ -93,6 +93,36 @@ module Aws
...
+      it 'prefers sso credentials over assume role' do
+        expect(SSOCredentials).to receive(:new).with(
+          sso_start_url: 'START_URL',
+          sso_region: 'us-east-1',
+          sso_account_id: 'SSO_ACCOUNT_ID',
+          sso_role_name: 'SSO_ROLE_NAME'
+        ).and_return(
+          double(
+            'creds',
+            set?: true,
```
mullermp, line 105: wouldn't `set?` return true already because you are mocking credentials?

gems/aws-sdk-sso/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 Unreleased Changes
 ------------------
 
+1.7.0 (2020-09-01)
+------------------
+
+* Feature - SSO has been moved into `aws-sdk-core` to support usage of SSOCredentials.
```
mullermp, line 7: I'd mention here what version of core now has this change. Be careful though I think our changelog verifier test will look for "x.y.z" regex and fail (it's expecting to find 1.7.0 (2020-09-01) for instance but would find 3.105.0). I can't remember if I fixed that test or not, but worst case you could say 3.105.

gems/aws-sdk-sso/lib/aws-sdk-sso.rb
```diff
@@ -1,52 +1,5 @@
 # frozen_string_literal: true
 
-# WARNING ABOUT GENERATED CODE
-#
-# This file is generated. See the contributing guide for more information:
-# https://github.com/aws/aws-sdk-ruby/blob/master/CONTRIBUTING.md
-#
-# WARNING ABOUT GENERATED CODE
-
+# This gem is a STUB - SSO has been moved into aws-sdk-core
```
mullermp, line 3: ditto on version

gems/aws-sdk-core/lib/aws-sdk-core/sso_credentials.rb
alextwoods: I wanted to provide which keys were missing in the error message so need to select all of them.

gems/aws-sdk-core/spec/aws/sso_credentials_spec.rb
alextwoods: These were intended to be deleted.

gems/aws-sdk-core/spec/aws/credential_resolution_chain_spec.rb
alextwoods: No - the set? method gets called on this double and if its not mocked it will fail.

PR #2402, opened 2020-08-27
Cross Region Copy for neptune/docdb + tests

mullermp:


build_tools/customizations.rb
```diff
@@ -120,27 +120,32 @@ def dynamodb_example_deep_transform(subsegment, keys)
...
+    %w[RDS Neptune DocDB].each do |service|
+      operations = %w[
         CopyDBClusterSnapshotMessage
         CreateDBClusterMessage
-      ).each do |shape_name|
-        api['shapes'][shape_name]['members']['DestinationRegion'] = {"shape" => "String", "documented" => false }
-        api['shapes'][shape_name]['members']['SourceRegion'] = {"shape" => "String"}
+      ]
+      if service == 'RDS'
+        operations += %w[
```
alextwoods, line 130: Rather than having RDS add to this list, It might be better to have a map from service => operations that is static.

gems/aws-sdk-docdb/lib/aws-sdk-docdb/plugins/cross_region_copying.rb
```diff
@@ -0,0 +1,66 @@
...
+      #
+      # This parameter is required by RDS when copying an encrypted snapshot
+      # across regions. This plugin will be skipped if the `:pre_signed_url`
+      # parameter is provided by the user.
+      class CrossRegionCopying < Seahorse::Client::Plugin
+        # @api private
+        class Handler < Seahorse::Client::Handler
+          def call(context)
+            params = context.params
+            if params.is_a?(Hash) &&
```
alextwoods, line 19: If both `source_region` and `pre_signed_url` are present, this will break right?

gems/aws-sdk-docdb/lib/aws-sdk-docdb/plugins/cross_region_copying.rb
```diff
@@ -0,0 +1,66 @@
...
+              expires_in: 3600
+            ).to_s
+          end
+        end
+
+        handler(
+          Handler,
+          step: :initialize,
+          operations: [
+            :copy_db_cluster_snapshot
```
alextwoods, line 60: Any easy way to have this be set by the customization we do in the generator instead of maintained seperatly here?  I know we don't intend to support additional operations... but I am concerned about the ability for these to get out of sync.

gems/aws-sdk-docdb/spec/client_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+            )
+
+            expect(resp.context.params[:pre_signed_url]).to match(
+              /01eda84cb84ff1558373f4759aaf76aa4b7be8664241a58f6906ae842a0a9d74/
+            )
+          end
+        end
+      end
+
+      # context 'create_db_cluster' do
```
alextwoods, line 69: delete

build_tools/customizations.rb
mullermp: these three services share an API (same endpoint), so if we were to do this, operations would be duplicated. i think it makes sense to just make a special case for RDS and refactor away later if needed

gems/aws-sdk-docdb/spec/client_spec.rb
mullermp: keeping it here for now, it should be modeled soon.

gems/aws-sdk-docdb/lib/aws-sdk-docdb/plugins/cross_region_copying.rb
mullermp: the customization in the generator is just middleware to insert API reference changes prior to generation. this list is for the handler list and is separate.

gems/aws-sdk-docdb/lib/aws-sdk-docdb/plugins/cross_region_copying.rb
mullermp: no, it shouldn't. the customization "models" source region and is treated as valid by the param validator. there are tests for this case for each operation on each service.

Zenix27 (approved):

Issue #1683, opened 2017-12-20
Message verifier for SNS does not use configured http_proxy


brettburley:
### Issue description
The aws-sdk-sns gem's `MessageVerifier` does not use the `http_proxy` configured in `Aws.config`. Since it's using `Net::HTTP` directly, it will use any `ENV` variables, but there doesn't appear to be any other way to configure a proxy for it to use otherwise.

See: https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-sns/lib/aws-sdk-sns/message_verifier.rb

I think I would expect the  the `Aws.config.http_proxy` option to apply to all network requests made.

### Gem name
aws-sdk-sns 1.1.0

cjyclaire:
Thanks for the feedback. The reason is that this verifier module doesn't use an SNS client, it doesn't make any client call but making http call directly to get pem, that's why `Aws.config` is never used for this module. 

To support this request, SDK might somehow need to extend the message verifier to use the client http interface.

awood45:
We could also expose a way to pass options into Net::HTTP, for an alternative approach.

mullermp:
Reopening - deprecating usage of Feature Requests backlog markdown file.

PR #2417, opened 2020-10-02
Use configured http_proxy for SNS MessageVerifier

mullermp:
Adds #1683


alextwoods (commented):
Change looks good, but would be good to add specs

alextwoods (approved):

alextwoods (approved):

PR #2424, opened 2020-10-20
No need to send a public method "Object#extend"

amatsuda:
`Object#extend` has been defined as a public method since the last century (I don't exactly know when, but it already was a public method even on version 1.0.0).
So we don't need to call it via `send`.

----

1. Can I just submit this without editing the changelog? I don't want to claim for a credit for a trivial change like this...


alextwoods (approved):
Good catch - and thanks for the contribution!

PR #2427, opened 2020-10-27
Tagged Unions

alextwoods:
**Summary**: Add support for the union trait - adds additional behavior to existing Structure member types.

There are two types of behavior modified:
1. Unions as inputs - when used as an input, unions validate that exactly one member is set.
2. Unions as outputs - there are two additional methods added: `#member` which returns the name of the union's member that is set and `#value` which returns the set value.  Additionally, `:unknown` is added as a member on unions.

These methods make it easy to use union output types in case statements:
```ruby
item_value = `ddb.get_item(params).item['value']
case item_value.member
when :s then handle_string_member(item_value.value)
when :ss then handle_ss_member(item_value.value)
when :unknown then handle_unknown_value(item_value)
end
```

**NOTE**:  (do not merge as is) This PR contains a modification to dynamodb api to add `union` to the dynamodb `AttributeValue` type to demonstrate the changes from code gen.

**TODO**: 
* Documentation cleanups

### DIFF
--- a/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/client_operation_documentation.rb
+++ b/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/client_operation_documentation.rb
@@ -100,6 +100,9 @@ module AwsSdkCodeGenerator
           if member_ref['jsonvalue']
             docstring = docstring.to_s + "<p><b>SDK automatically handles json encoding and base64 encoding for you when the required value (Hash, Array, etc.) is provided according to the description.</b></p>"
           end
+          if member_ref['union']
+            docstring = docstring.to_s + "<p>UNION DOCUMENTATION</p>"
+          end
           YardOptionTag.new(
             name: Underscore.underscore(member_name),
             ruby_type: Api.ruby_input_type(member_ref, api, operation),
--- a/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/client_operation_documentation.rb
+++ b/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/client_operation_documentation.rb
@@ -101,7 +101,7 @@ module AwsSdkCodeGenerator
             docstring = docstring.to_s + "<p><b>SDK automatically handles json encoding and base64 encoding for you when the required value (Hash, Array, etc.) is provided according to the description.</b></p>"
           end
           if member_ref['union']
-            docstring = docstring.to_s + "<p>UNION DOCUMENTATION</p>"
+            docstring = docstring.to_s + "<p>This is a union type and you must set exactly one of the members.</p>"
           end
           YardOptionTag.new(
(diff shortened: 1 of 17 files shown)


build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/client_operation_documentation.rb
```diff
@@ -100,6 +100,9 @@ def option_tags(operation, api)
           if member_ref['jsonvalue']
             docstring = docstring.to_s + "<p><b>SDK automatically handles json encoding and base64 encoding for you when the required value (Hash, Array, etc.) is provided according to the description.</b></p>"
           end
+          if member_ref['union']
+            docstring = docstring.to_s + "<p>This is a union type and you must set exactly one of the members.</p>"
```
mullermp, line 104: I think use a passive voice here. Something like "Unions have many members, but only one member can contain a value at any given time"

build_tools/aws-sdk-code-generator/templates/types_module.mustache
```diff
@@ -18,6 +18,9 @@ module {{module_name}}
       {{/members}}
       SENSITIVE = {{sensitive_params}}
       include Aws::Structure
+      {{#union?}}
+      include Aws::Structure::Union
```
mullermp, line 22: Is there a solution that exists without a union module?

mullermp (approved):

mullermp:
Nice! looks good

build_tools/aws-sdk-code-generator/templates/types_module.mustache
alextwoods: The Union structure is being used for two things: 1) to provide some methods and 2) for "type" checking.  It could be replaced by code-genning that information directly.  eg:

```
{{#union?}}
      def member
        self.members.select { |k| self[k] }.first
      end

      def value
        self[member] if member
      end

     def union?
       true
     end
{{^union?}}
    def union?
      false
    end
{{#/union?}}
```

mullermp:
"Additionally, :unknown is added as a member on unions"

alextwoods:
> "Additionally, :unknown is added as a member on unions"
> 
> What happens when a member is named "unknown"?

Yeah, this is a good question.  The parsing code (for both json and xml) first checks if a member ref is defined and if so uses it.  However.... this would end up being confusing in the case where an actual new type is sent since the `unknown` member would still end up getting set from it.  I see a few options:
1. Check for conflicting member name during code gen.  If it conflicts.... we have a few options: a) consider it an error and fail or b) use a different name for this particular type.  To some extent because we have delegation on our response objects we have a similar problem with name conflicts.
2. Take a completely different approach to handling unknowns - instead of adding it as a member, it would need to be set differently (logic added to the Union module) and then extra logic added to the member and value methods to handle it.  But it would work around this issue.

mullermp:
Is method_missing a good idea?

PR #2430, opened 2020-11-04
No need to include Seahorse::Model::Shapes into Aws::Stubbing::Protocols::RestXml

amatsuda:
Because `Aws::Stubbing::Protocols::RestXml`'s parent class `Aws::Stubbing::Protocols::Rest` already includes this very module `Seahorse::Model::Shapes`. https://github.com/aws/aws-sdk-ruby/blob/6ca1851/gems/aws-sdk-core/lib/aws-sdk-core/stubbing/protocols/rest.rb#L10

1. I don't think I want to add a changelog for this trivial one line deletion that wouldn't cause any behavior change. I mean, I'm happy if the patch just gets merged.


alextwoods (approved):
Thanks for your contribution and good cleanup!

PR #2452, opened 2020-12-11
Remove dependency on webrick for ruby 3.0.0 compatibility

casperisfine:
Webrick has been removed from the stdlib in Ruby 3.0.0: https://bugs.ruby-lang.org/issues/17303

This breaks the legacy signer. Instead we can use `URI::DEFAULT_PARSER.escape` which is the original `URI.escape` but without the deprecation warning. It is undocumented, but is unlikely to be removed soon, so for legacy code it makes sense I think.

Alternatively if you still want to use webrick, we can add it to the gemspec: https://rubygems.org/gems/webrick

cc @mullermp

### DIFF
--- a/gems/aws-sdk-s3/lib/aws-sdk-s3/legacy_signer.rb
+++ b/gems/aws-sdk-s3/lib/aws-sdk-s3/legacy_signer.rb
@@ -4,7 +4,6 @@ require 'set'
 require 'time'
 require 'openssl'
 require 'cgi'
-require 'webrick/httputils'
 require 'aws-sdk-core/query'
 
 module Aws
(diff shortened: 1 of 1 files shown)


gems/aws-sdk-s3/lib/aws-sdk-s3/legacy_signer.rb
```diff
@@ -157,33 +156,12 @@ def signed_querystring_params(endpoint)
...
-        #         ERB::Util.url_encode(s),
-        #         URI.encode_www_form_component(s),
-        #         WEBrick::HTTPUtils.escape_form(s),
-        #         WEBrick::HTTPUtils.escape(s),
-        #         URI.escape(s),
-        #       ]
-        #       next if e.uniq.length == 1
-        #       puts("%5s %5s %5s %5s %5s %5s %5s" % ([s.inspect] + e))
-        #     }
-        #
```
mullermp, line 183: I'm not against making this change. However, can we keep this script documentation? I would touch up the documentation to mention that webrick was used but dropped for ruby 3 support. This kind of historical documentation (especially something critical such as signing) is helpful.

Here is the updated script.

```
# (0..255).each {|c|
#   s = [c].pack("C")
#   e = [
#     CGI.escape(s),
#     ERB::Util.url_encode(s),
#     URI.encode_www_form_component(s),
#     WEBrick::HTTPUtils.escape_form(s),
#     WEBrick::HTTPUtils.escape(s),
#     URI.escape(s),
#     URI::DEFAULT_PARSER.escape(s)
#   ]
#   next if e.uniq.length == 1
#   puts("%5s %5s %5s %5s %5s %5s %5s %5s" % ([s.inspect] + e))
# }
```

mullermp (changes requested):

gems/aws-sdk-s3/lib/aws-sdk-s3/legacy_signer.rb
casperisfine: Of course. I assumed that since `URI::DEFAULT_PARSER.escape` is pretty much an alias it wasn't necessary anymore. But it's true that this comment was helpful for me so I added it back.

mullermp:
Thanks for the contribution!

mullermp:
Agh, I forgot to mention we need a CHANGELOG entry. I'll take care of it.

PR #2457, opened 2020-12-30
add rexml dependency to support ruby v3

tonytonyjan:
> The following default gems are now bundled gems.
>
>  * rexml
>  * rss

ref: https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/


alextwoods:
Thanks for the contribution!  It looks like tests are failing for Ruby 1.9.3, which we unfortunately still need to support:
```
/home/travis/.rvm/gems/ruby-1.9.3-p551/gems/rexml-3.2.4/lib/rexml/xpath.rb:73: syntax error, unexpected tPOW, expecting ')'

      parser = XPathParser.new(**options)
```

However, it looks like this may be happening without this change as well.  In previous runs, rexml was not being installed at all (eg: https://travis-ci.com/github/aws/aws-sdk-ruby/jobs/459953938).  But that same change when re-run now does install rexml 3.2.4 (see: https://travis-ci.com/github/aws/aws-sdk-ruby/jobs/459953937).  It looks like it has been added as an indirect dependency.  `rexml >= 0` was added as a dependency to [crack](https://rubygems.org/gems/crack/versions/0.4.5) on Dec 27 2020, after which all of our CI started failing (we depend on crack indirectly from cucumber).    Its a quick [fix to our Gemfile](https://github.com/aws/aws-sdk-ruby/pull/2459/commits/726956ba642f4950f6df44db61fc1ed2930fa9fd) to address that issue ([build output](https://travis-ci.com/github/aws/aws-sdk-ruby/jobs/468340805) w/ fix).

That being said - we still have an issue I believe that adding a dependency on rexml breaks Ruby 1.9.3 and I'm not quite sure what we should do about that.

mullermp:
Related: https://github.com/ruby/rexml/issues/38

alextwoods:
We discussed this a bit more in: [ruby/rexml#38](https://github.com/ruby/rexml/issues/38) and #2459 - and given the compatibility issues with Ruby 1.9 have decided against including rexml as an explicit dependency.   rexml is currently used as the last resort [xml engine](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/lib/aws-sdk-core/xml/parser.rb#L70) so I've added an error message when none can be loaded.

In almost all cases, other gems are already depending on at least one of the xml libraries we support so most users should not be impacted by rexml not being an explicit dependency (and in those cases, they can add it, or select any of the supported xml libraries).  

If you have a specific use case where this is required as a dependency in the sdk, lets chat!  Otherwise, I'm going to close this PR - but thank you for opening it and kicking off such an important discussion!

patbl:
Now that the required Ruby version is 2.7+, would you reconsider adding `rexml` as a requirement (especially since it's been a default or bundled gem since Ruby 2.6)?

https://stdgems.org/rexml/

richardwang1124:
Hey @patbl, we recently merged a [PR](https://github.com/aws/aws-sdk-ruby/pull/3412) to add `rexml` as a dependency which should go out with today's release.

patbl:
> Hey @patbl, we recently merged a [PR](https://github.com/aws/aws-sdk-ruby/pull/3412) to add `rexml` as a dependency which should go out with today's release.

This PR was reverted and there was no release that day. I asked why [here](https://github.com/aws/aws-sdk-ruby/pull/3415#issuecomment-5542792033).

Issue #2455, opened 2020-12-27
aws-sdk-core-3.110.0/lib/seahorse/client/h2/connection.rb throws /http/2/compressor.rb:333:in `initialize': wrong number of arguments (given 1, expected 0)


rgaufman:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [X] I've gone though [Developer Guide for v3](https://docs.aws.amazon.com/sdk-for-ruby/v3/developer-guide/welcome.html) and [API reference](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/index.html)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-ruby) for answers
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-ruby/issues) and didn't find any solution
  
**Describe the bug**

This is my code:
```
require 'aws-sdk'
require 'net-http2'

Aws.config.update(
  credentials: Aws::Credentials.new('REMOVED', 'REMOVED')
)

async_client = Aws::TranscribeStreamingService::AsyncClient.new(
  region: 'eu-west-2',
)
```

It throws the following exception:
```
$ ruby transcribe.rb
/opt/homebrew/lib/ruby/gems/3.0.0/gems/http-2-0.10.2/lib/http/2/compressor.rb:333:in `initialize': wrong number of arguments (given 1, expected 0) (ArgumentError)
	from /opt/homebrew/lib/ruby/gems/3.0.0/gems/http-2-0.10.2/lib/http/2/connection.rb:79:in `new'
	from /opt/homebrew/lib/ruby/gems/3.0.0/gems/http-2-0.10.2/lib/http/2/connection.rb:79:in `initialize'
	from /opt/homebrew/lib/ruby/gems/3.0.0/gems/http-2-0.10.2/lib/http/2/client.rb:28:in `initialize'
	from /opt/homebrew/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.110.0/lib/seahorse/client/h2/connection.rb:45:in `new'
	from /opt/homebrew/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.110.0/lib/seahorse/client/h2/connection.rb:45:in `initialize'
[...]

PR #2459, opened 2021-01-04
Add Ruby 3.0 to CI matrix.

alextwoods:


alextwoods:
Ruby 3 is currently failing.   Issue is related to: #2455.

Fix has been pushed to master in http2:  https://github.com/igrigorik/http-2/pull/156

MSP-Greg:
For the AR amongst us...

JFYI, using:
```
ruby: [2.3, 2.4, 2.5, 2.6, 2.7, 3.0, jruby-9.2]
```
The yaml is parsed as numbers, which if okay, but there are issues with trailing zeros in what appears to be a float, eg, 2.2.10 would be okay, but 3.0 or 2.10 would not.  Hence, using the following shows '3.0' in the jobs list:
```
ruby: [2.3, 2.4, 2.5, 2.6, 2.7, '3.0', jruby-9.2]
```

igrigorik (approved):
👍

gems/aws-sdk-core/aws-sdk-core.gemspec
```diff
@@ -17,6 +17,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency('aws-partitions', '~> 1', '>= 1.239.0') # necessary for STS & S3 regional
   spec.add_dependency('aws-sigv4', '~> 1.1') # necessary for making Aws::STS API calls
   spec.add_dependency('aws-eventstream', '~> 1', '>= 1.0.2') # necessary for binary eventstream
+  spec.add_dependency('rexml', '>= 0')
```
mullermp, line 20: Talked offline about this - I think instead, customers should depend on any XML parser that we support (bring your own parser), and we should not force dependencies that may not be necessary and to avoid potential breakage. In general we try to own and control our own dependencies.

alextwoods: I agree with this.  The consequence however is that customers will need to specify (directly or indirectly) an xml library in order to use the SDK. In many cases rexml or another library will be brought in by other libraries anyway so this may not impact many users.  However, I will add descriptive error messages - currently we assume that there will always be at least one xml engine available and dont handle the missing case (since rexml was previously always available).

gems/aws-sdk-core/lib/aws-sdk-core/xml/parser.rb
```diff
@@ -70,6 +70,11 @@ def set_default_engine
           [:ox, :oga, :libxml, :nokogiri, :rexml].each do |name|
             @engine ||= try_load_engine(name)
           end
+          unless @engine
+            raise 'Unable to find a compatible xml library. ' \
+            ' Make sure you have installed or added to your Gemfile one of ' \
```
mullermp, line 75: nit: "Ensure that you have installed ox, oga, ... etc"

mullermp (approved):

gems/aws-sdk-core/aws-sdk-core.gemspec
arothian: Is there potential for reversing this decision? It seems odd to have an undeclared dependency, that wouldn't be discovered until runtime, and only when using Ruby 3+. 

Given that Ruby 3 still includes `rexml` as a bundled gem, it should generally be available everywhere anyhow.

IMO this is a breaking change that is a bit odd for minor version bump.

alextwoods: Adding it as an explicit dependency breaks older rubies, which we must still support.  It's also not a required dependency: we support a number of xml engines, and will use whichever a client has loaded.    I don't believe that not including a new dependency is a breaking change for existing users - it might, in a few cases, require changes in a users code when they are upgrading from an older ruby version to 3.

PR #2460, opened 2021-01-08
DynamoDBStreams simple attributes

### LINKED ISSUE #1212: Parse DynamoDB Stream events?
Hey

Is there a way to parse DynamoDB Stream events using the SDK? I receive a JSON from a DynamoDB stream in a process, and I would like to parse the `NewImage` into a "normal" hash, without the types `S`, `M` etc as root keys.

The JSON I receive:

``` json
{
  "eventID":"",
  "eventName":"MODIFY",
  "eventVersion":"1.1",
  "eventSource":"aws:dynamodb",
  "awsRegion":"us-east-1",
  "dynamodb":{
    "ApproximateCreationDateTime":123,
    "Keys":{
      "object_id":{
        "S":"123"
      }
    },
    "NewImage":{
      "name":{
        "S":"NAME"
      },
      "hash_value":{
        "M":{
          "sub_hash_value":{
            "M":{
              "second_name":{
                "S":"name"
              },

```

And I would like to parse the `NewImage` to:

``` json
{ 
  "object_id": "123", 
  "name": "name", 
  "hash_value": { 
    "sub_hash_value":  { "second_name": "name" } 
  }
}
```

Is there any helper in the SDK that parses it?

mullermp:
Closes #1212 

Adds a utility that can be used in Lambda functions (`Aws::DynamoDBStreams::AttributeTranslator`) for parsing events.

Adds a plugin `:simple_attributes` similar to the one in DynamoDB to return stream records with simple attributes.

The implementation of the utility is slower than a switch case but not terrible. It provides consistency.

### DIFF
--- a/gems/aws-sdk-dynamodbstreams/lib/aws-sdk-dynamodbstreams/attribute_translator.rb
+++ b/gems/aws-sdk-dynamodbstreams/lib/aws-sdk-dynamodbstreams/attribute_translator.rb
@@ -5,7 +5,7 @@ module Aws
     class AttributeTranslator
       # Parse a DynamoDBStream event hash from Lambda that contains 1 or more
       # records. When using the SDK to retrieve DynamoDB stream records, use the
-      # `simple_attributes: true` option instead.
+      # `simple_attributes: true` client option instead.
       #
       # @param [Hash] event A DynamoDBStream event.
       #
@@ -32,7 +32,7 @@ module Aws
           shape_ref, :unmarshal
         )
         parser = Aws::Json::Parser.new(shape_ref)
-        input = parser.parse(JSON.dump(event))
+        input = parser.parse(Aws::Json.dump(event))
         translator.apply(input).records
       end
     end
--- a/gems/aws-sdk-dynamodbstreams/lib/aws-sdk-dynamodbstreams/attribute_translator.rb
+++ b/gems/aws-sdk-dynamodbstreams/lib/aws-sdk-dynamodbstreams/attribute_translator.rb
@@ -1,0 +1,40 @@
+module Aws
+  module DynamoDBStreams
+    # A utility class that translates DynamoDBStream events from Lambda
+    # functions into their simple attribute equivalents.
+    class AttributeTranslator
+      # Parse a DynamoDBStream event hash from Lambda that contains 1 or more
+      # records. When using the SDK to retrieve DynamoDB stream records, use the
+      # `simple_attributes: true` option instead.
+      #
+      # @param [Hash] event A DynamoDBStream event.
+      #
+      # @example Parse a DynamoDB stream event from Lambda
+      #   def lambda_handler(event:, context:)
+      #     records = Aws::DynamoDBStreams::AttributeTranslator.from_event(event)
+      #     records.each do |record|
+      #       puts record.dynamodb.new_image
+      #       # => { "size" => 123, "enabled" => true, ... }
+      #     end
+      #   end
+      def self.from_event(event)
+        return unless event.is_a?(Hash)
+
+        # TODO: This implementation is slow and inefficient. It would be
+        # better to write a switch-case that has similar logic to AttributeValue
+        # used in the ValueTranslator. This implementation however provides
+        # consistency between both DynamoDB and DynamoDBStreams. The translator
+        # only works with shapes and structs and not a regular Ruby hash.
+        shape_ref = ClientApi::Shapes::ShapeRef.new(
+          shape: ClientApi::GetRecordsOutput
+        )
+        translator = Plugins::SimpleAttributes::ValueTranslator.new(
+          shape_ref, :unmarshal
+        )
+        parser = Aws::Json::Parser.new(shape_ref)
+        input = parser.parse(JSON.dump(event))
+        translator.apply(input).records
+      end
+    end
+  end
+end
--- a/gems/aws-sdk-dynamodbstreams/lib/aws-sdk-dynamodbstreams/attribute_value.rb
+++ b/gems/aws-sdk-dynamodbstreams/lib/aws-sdk-dynamodbstreams/attribute_value.rb
@@ -10,67 +10,11 @@ module Aws
     class AttributeValue
 
       def initialize
-        @marshaler = Marshaler.new
         @unmarshaler = Unmarshaler.new
-      end
-
-      def marshal(value)
-        @marshaler.format(value)
       end
 
       def unmarshal(value)
         @unmarshaler.format(value)
-      end
-
-      class Marshaler
-        HASHY_TEST = lambda { |val| val.nil? ? false : val.respond_to?(:to_h) }
-        STRINGY_TEST = lambda { |val| val.respond_to?(:to_str) }
-
-        def format(obj)
-          case obj
-          when Hash
-            obj.each.with_object(m:{}) do |(key, value), map|
-              map[:m][key.to_s] = format(value)
-            end
-          when Array
-            obj.each.with_object(l:[]) do |value, list|
-              list[:l] << format(value)
-            end
-          when String then { s: obj }
-          when Symbol then { s: obj.to_s }
-          when STRINGY_TEST then { s: obj.to_str }
-          when Numeric then { n: obj.to_s }
-          when StringIO, IO then { b: obj }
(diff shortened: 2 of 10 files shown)


gems/aws-sdk-dynamodbstreams/lib/aws-sdk-dynamodbstreams/attribute_translator.rb
```diff
@@ -0,0 +1,40 @@
+module Aws
+  module DynamoDBStreams
+    # A utility class that translates DynamoDBStream events from Lambda
+    # functions into their simple attribute equivalents.
+    class AttributeTranslator
+      # Parse a DynamoDBStream event hash from Lambda that contains 1 or more
+      # records. When using the SDK to retrieve DynamoDB stream records, use the
+      # `simple_attributes: true` option instead.
```
alextwoods, line 8: I'd rephrase to "use the simple_attributes:true when creating the client" to make it clear that thats an option that applies to client rather than this class.

mullermp:
@watmin @saluzafa

gems/aws-sdk-dynamodbstreams/lib/aws-sdk-dynamodbstreams/attribute_translator.rb
```diff
@@ -0,0 +1,40 @@
...
+        # consistency between both DynamoDB and DynamoDBStreams. The translator
+        # only works with shapes and structs and not a regular Ruby hash.
+        shape_ref = ClientApi::Shapes::ShapeRef.new(
+          shape: ClientApi::GetRecordsOutput
+        )
+        translator = Plugins::SimpleAttributes::ValueTranslator.new(
+          shape_ref, :unmarshal
+        )
+        parser = Aws::Json::Parser.new(shape_ref)
+        input = parser.parse(JSON.dump(event))
```
alextwoods, line 35: Use the `Aws::Json.dump` instead to be JSON engine agnostic.

gems/aws-sdk-dynamodbstreams/lib/aws-sdk-dynamodbstreams/client.rb
```diff
@@ -266,6 +268,14 @@ class Client < Seahorse::Client::Base
     #
     #   @option options [String] :session_token
     #
+    #   @option options [Boolean] :simple_attributes (false)
+    #     Enables working with DynamoDB attribute values using
+    #     hashes, arrays, sets, integers, floats, booleans, and nil.
+    #
+    #     Disabling this option requires that all attribute values have
```
alextwoods: This really only applies to inputs.

gems/aws-sdk-dynamodbstreams/lib/aws-sdk-dynamodbstreams/plugins/simple_attributes.rb
```diff
@@ -0,0 +1,153 @@
+# frozen_string_literal: true
+
+module Aws
+  module DynamoDBStreams
+    module Plugins
+      # Simplifies working with Amazon DynamoDBStream attribute values.
+      # Translates attribute values for requests and responses to sensible
```
alextwoods, line 7: Responses only

gems/aws-sdk-dynamodbstreams/lib/aws-sdk-dynamodbstreams/attribute_value.rb
```diff
@@ -0,0 +1,117 @@
...
+
+      def marshal(value)
+        @marshaler.format(value)
+      end
+
+      def unmarshal(value)
+        @unmarshaler.format(value)
+      end
+
+      class Marshaler
```
alextwoods, line 25: In theory we could do away with this here since its unused.  But for the purposes of keeping it the same as the DDB one, I think its better to leave.

gems/aws-sdk-dynamodbstreams/lib/aws-sdk-dynamodbstreams/client.rb
mullermp: Yeah, I fixed the docs but didn't rebuild.

gems/aws-sdk-dynamodbstreams/lib/aws-sdk-dynamodbstreams/attribute_value.rb
mullermp: I thought it was fine to keep. Maybe it makes sense to remove.

gems/aws-sdk-dynamodbstreams/lib/aws-sdk-dynamodbstreams/attribute_translator.rb
watmin: If you wanted to drop re-encoding the event to JSON, maybe remove the private qualifier here

https://github.com/aws/aws-sdk-ruby/blob/203aaba7242357884fe182be8f2d36f77428b8a7/gems/aws-sdk-core/lib/aws-sdk-core/json/parser.rb#L22

or copy this from the JSON parser:

https://github.com/aws/aws-sdk-ruby/blob/203aaba7242357884fe182be8f2d36f77428b8a7/gems/aws-sdk-core/lib/aws-sdk-core/json/parser.rb#L24-L72

Calling `parse_ref` on the event Hash looks like it'll resolve the Hash -> JSON -> Struct sequence to just Hash -> Struct.

Quick and dirty demo on the existing sdk's ddb item

```ruby
module Demo
  include Seahorse::Model::Shapes
  
  module_function

  def translate(image)
    shape_ref = Aws::DynamoDB::ClientApi::Shapes::ShapeRef.new(shape: Aws::DynamoDB::ClientApi::GetItemOutput)
    translator = Aws::DynamoDB::Plugins::SimpleAttributes::ValueTranslator.new(shape_ref, :unmarshal)
    translator.apply(parse_ref(shape_ref, { 'Item' => image })).item
  end

  def parse_ref(ref, value, target = nil)
    if value.nil?
      nil
    else
      case ref.shape
      when StructureShape then structure(ref, value, target)
      when ListShape then list(ref, value, target)
      when MapShape then map(ref, value, target)
      when TimestampShape then time(value)
      when BlobShape then Base64.decode64(value)
      when BooleanShape then value.to_s == 'true'
      else value
      end
    end
  end

  def structure(ref, values, target = nil)
    shape = ref.shape
    target = ref.shape.struct_class.new if target.nil?
    values.each do |key, value|
      member_name, member_ref = shape.member_by_location_name(key)
      if member_ref
        target[member_name] = parse_ref(member_ref, value)
      end
    end
    target
  end

  def list(ref, values, target = nil)
    target = [] if target.nil?
    values.each do |value|
      target << parse_ref(ref.shape.member, value)
    end
    target
  end

  def map(ref, values, target = nil)
    target = {} if target.nil?
    values.each do |key, value|
      target[key] = parse_ref(ref.shape.value, value)
    end
    target
  end

  def time(value)
    value.is_a?(Numeric) ? Time.at(value) : Time.parse(value)
  end
end
```

Using part of Lambda's ddbEvent test input

```ruby
[10] pry(main)> ddb_event['Records'].first['dynamodb']['NewImage']
=> {"Message"=>{"S"=>"New item!"}, "Id"=>{"N"=>"101"}}
[11] pry(main)> Demo.translate(ddb_event['Records'].first['dynamodb']['NewImage'])
=> {"Message"=>"New item!", "Id"=>0.101e3}
[12] pry(main)>
```

watmin (commented):
Address the TODO

gems/aws-sdk-dynamodbstreams/lib/aws-sdk-dynamodbstreams/attribute_translator.rb
mullermp: Thanks for the suggestion. The re-encoding is not actually the slowest part.. it's mainly because of the shape traversal. Some quick benchmarking shows the re-encoding to be about 1/4 or 1/5th the overall speed. Also "slow" is a bit of an exaggeration, it's still plenty fine. I think the best thing to do would be to make a switch case similar to that of the AttributeValue unmarshaler and guarantee that it's compatible. For now I am going to ship this.

alextwoods (approved):

PR #2466, opened 2021-01-18
Require the correct files when using specific gems (aws-sdk-sts & aws-sdk-sso)

stergiom:
#### Context

We recently updated the dependency manifest on one of our projects to be specific about using the `aws-sdk-sts` gem, and found that `aws-sdk-core` fails to load the appropriate `aws-sdk-sts` file.

```
require 'aws-sdk-sts'
=> true

Aws::CORE_GEM_VERSION
 => "3.111.0"

Aws::STS::GEM_VERSION
<..> NameError (uninitialized constant Aws::STS)
```

#### Changes

Require the correct file for `aws-sdk-sts`, and pre-emptively do the same for `aws-sdk-sso`.

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core.rb
@@ -89,10 +89,10 @@ require_relative 'aws-sdk-core/arn_parser'
 require_relative 'aws-sdk-core/ec2_metadata'
 
 # aws-sdk-sts is included to support Aws::AssumeRoleCredentials
-require 'aws-sdk-sts'
+require_relative 'aws-sdk-sts'
 
 # aws-sdk-sso is included to support Aws::SSOCredentials
-require 'aws-sdk-sso'
+require_relative 'aws-sdk-sso'
 
 module Aws
 


mullermp:
Thanks for opening a pull request.

>We recently updated the dependency manifest on one of our projects to be specific about using the aws-sdk-sts gem.

Could you elaborate a bit more on what you mean? The STS gem is not intended to be installed. It was created recently because rubygems.org doesn't reserve the rights to gem names. STS (and SSO) are both included in core. The gems are supposed to act like an alias. That said, I'm not against this change.

orien:
We found installing the `aws-sdk-sts` gem broke our application. Take for instance the following script:

```ruby
#!/usr/bin/env ruby

require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'aws-sdk-sts', '1.0.1' # download and require the STS alias gem
  gem 'aws-sdk-core', '3.111.1'
end

puts Aws::STS # test the module is available
```

Running this fails with the error:

```
./aws-with-sts-gem.rb:11:in `<main>': uninitialized constant Aws::STS (NameError)
```

When we remove the `aws-sdk-sts` alias gem and rely on the `aws-sdk-core` gem:

```ruby
#!/usr/bin/env ruby

require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'aws-sdk-core', '3.111.1' # download and require the core gem, but not the STS alias gem
end

puts Aws::STS # test the module is available
```

The script successfully prints the module name indicating it is available:

```
Aws::STS
```

Our investigation found the cause of the problem to be duplicate Ruby file names.

- A [`lib/aws-sdk-sts.rb`](https://github.com/aws/aws-sdk-ruby/blob/2a4a361ec966ae86fcdfd774ccdf819e8a9eb628/gems/aws-sdk-sts/lib/aws-sdk-sts.rb) in the `aws-sdk-sts` gem
- A [`lib/aws-sdk-sts.rb`](https://github.com/aws/aws-sdk-ruby/blob/2a4a361ec966ae86fcdfd774ccdf819e8a9eb628/gems/aws-sdk-core/lib/aws-sdk-sts.rb) in the `aws-sdk-core` gem.

Ruby seems to load only the `aws-sdk-sts.rb` file in the `aws-sdk-sts` gem. When it encounters this line in core

https://github.com/aws/aws-sdk-ruby/blob/2a4a361ec966ae86fcdfd774ccdf819e8a9eb628/gems/aws-sdk-core/lib/aws-sdk-core.rb#L92

it assumes it does not need to load anything as it has already loaded a file by this name. Hence the STS library in core is not loaded.

Changing this line to use a `require_relative` instruction resolves this problem. With this change, Ruby will know to look inside the `aws-sdk-core` gem to find the file.

alextwoods (approved):

mullermp:
Thanks for your contribution!

PR #2471, opened 2021-02-02
Deduplicate partitions.json

casperisfine:
`partitions.json` contains a lot of duplicated data, but recent `json` gem have a `freeze: true` option to deduplicate data in documents.

Test script:

```ruby
# typed: false
# frozen_string_literal: true

require 'json'
require 'set'
require 'objspace'

module DeepMemsize
  class << self
    def memsize_of(object, seen = Set.new.compare_by_identity)
      return 0 unless seen.add?(object)

      size = ObjectSpace.memsize_of(object)
      object.instance_variables.each { |v| size += memsize_of(object.instance_variable_get(v), seen) }

      case object
      when Hash
        object.each { |k, v| size += memsize_of(k, seen) + memsize_of(v, seen) }
      when Array
        object.each { |i| size += memsize_of(i, seen) }
      end
      size
    end

    UNIT_PREFIXES = {
      0 => 'B',
      3 => 'kB',
      6 => 'MB',
      9 => 'GB',
      12 => 'TB',
      15 => 'PB',
      18 => 'EB',
      21 => 'ZB',
      24 => 'YB',
    }.freeze

    def scale_bytes(bytes)
      return "0 B" if bytes.zero?

      scale = Math.log10(bytes).div(3) * 3
      scale = 24 if scale > 24
      format("%.2f #{UNIT_PREFIXES[scale]}", (bytes / 10.0**scale))
    end
  end
end

json_path = ARGV.first
puts "base: #{DeepMemsize.scale_bytes(DeepMemsize.memsize_of(JSON.load(File.read(json_path))))}"
puts "freeze: #{DeepMemsize.scale_bytes(DeepMemsize.memsize_of(JSON.parse(File.read(json_path), freeze: true)))}"
``` 

```bash
$ ruby /tmp/size_of.rb gems/aws-partitions/partitions.json 
base: 1.63 MB
freeze: 776.20 kB
```

So it reduces the memory footprint by a bit more than 50%.

### DIFF
--- a/gems/aws-partitions/lib/aws-partitions.rb
+++ b/gems/aws-partitions/lib/aws-partitions.rb
@@ -212,7 +212,12 @@ module Aws
       def defaults
         @defaults ||= begin
           path = File.expand_path('../../partitions.json', __FILE__)
-          JSON.load(File.read(path))
+          defaults = if JSON::VERSION >= '2.4.0'
+            JSON.load(File.read(path), freeze: true)
+          else
+            JSON.parse(File.read(path))
+          end
+          defaults.merge('partitions' => defaults['partitions'].dup)
         end
       end
 


casperisfine:
It's weird, I can't seem to reproduce the test failures locally.

casperisfine:
Ok, so it's caused by some test leak:

```
rspec ./gems/aws-partitions/spec/aws_partitions_spec.rb[1:7:1:2,1:7:1:3:1,1:7:2:1] ./gems/aws-sdk-core/spec/aws/process_credentials_spec.rb[1:5] ./gems/aws-sdk-s3/spec/plugins/streaming_retry_spec.rb[1:6:2] --seed 1
```

mullermp:
Hey, what's the use case for this? The partitions JSON file comes directly from an upstream source so we don't modify it. We have "internal only" APIs and "regions" where we add different partitions and what not, so those interfaces need to exist.

casperisfine:
> Hey, what's the use case for this (other than a memory improvement)?

No use case other than a footprint reduction.

> We have "internal only" APIs and "regions" where we add different partitions and what not, so those interfaces need to exist.

From what I gathered these APIs were only used from the test suite, and were easily replaced by stubs. I removed them and the test suite still pass. Am I missing something?

BTW, I finally figured the test leak, CI should go green now.

mullermp:
>From what I gathered these APIs were only used from the test suite, and were easily replaced by stubs. I removed them and the test suite still pass. Am I missing something?

Amazon has an internal GitFarm. I do know that there are some unmodeled regions and partitions internally. MVP regions (DCA/LCK) used to be that way, but now those are modeled. It's not an easy task to check all of the packages because some are hidden/private under LDAP/POSIX permissions and what not. If we were to remove these operations without checking and build this package internally, those consumers might be broken at runtime. When I get some time, I can start investigating that. If there's a way to reduce the memory footprint without removing these operations, I am also open to that.

casperisfine:
> If there's a way to reduce the memory footprint without removing these operations, I am also open to that.

We could only deduplicate the strings I suppose. I can open an alternative PR tomorrow.

casperisfine:
@mullermp I've changed the PR to be only the minimal change. We still use `freeze: true`, but we then dup the `partitions` Array so that it remains mutable.

alextwoods:
It looks like tests for older rubies are failing on this (See the travis checks).  If the partitions array is duped does it still achieve a reduction in memory?  Also, do you know when `freeze` was introduced to `JSON.load`, I haven't been able to find good documentation - I'm still a bit worried about unintended impact, esp to users with older rubies, from this change.

casperisfine:
> It looks like tests for older rubies are failing on this (See the travis checks)

Interesting. I can add some easy feature testing, that's no problem to handle.

> If the partitions array is duped does it still achieve a reduction in memory?

Yes.

> Also, do you know when freeze was introduced to JSON.load

It was released as part of `json` `2.4.0` back in last december: https://github.com/flori/json/blob/master/CHANGES.md

But `JSON.load` takes an option hash, so if the library is older it's just ignored (well except for the very old versions of JSON, but I'll add feature checking for that).

casperisfine:
721638e should take care of older rubies/json.

gems/aws-partitions/lib/aws-partitions.rb
```diff
@@ -209,11 +209,27 @@ def default_partition_list
...
       # @return [Hash]
       # @api private
-      def defaults
-        @defaults ||= begin
-          path = File.expand_path('../../partitions.json', __FILE__)
-          defaults = JSON.load(File.read(path), freeze: true)
-          defaults.merge('partitions' => defaults['partitions'].dup)
+      json_freeze = begin
+        JSON.load('true', freeze: true)
+      rescue
```
mullermp, line 214: Apologies for additional churn. I'd rather not use exceptions for logic control flow - and also not use dynamic method definition..

You mentioned that JSON load takes an options hash even for early Ruby versions and it would ignore the `:freeze` option. The SDK at a minimum supports 1.9.3 and I see the options hash you are talking about here: https://ruby-doc.org/stdlib-1.9.3/libdoc/json/rdoc/JSON.html#method-i-load . I agree with Alex that the documentation isn't very clear here.. I guess the Ruby version wouldn't matter here, unless that Ruby version has an upper bound on the JSON version? My understanding is that it was built into stdlib but that could just be wrong. These errors were definitely strange as I agree with you that the option should have been ignored https://github.com/aws/aws-sdk-ruby/runs/1829320013.

Do you think a `JSON::VERSION` (or worst case a `RUBY_VERSION`) check would work instead? 

Perhaps the method would look something like:

```
 def defaults
   @defaults ||= begin
     path = File.expand_path('../../partitions.json', __FILE__)
     options = {}
     options[:freeze] = true if JSON::VERSION >= '2.4.0'
     defaults = JSON.load(File.read(path), options)
     defaults.merge('partitions' => defaults['partitions'].dup)
   end
 end
```

Or worst case, use `JSON.load(File.read(path), freeze: true)` when `JSON::VERSION >= '2.4.0'` and `JSON.load(File.read(path)` otherwise.

casperisfine:
`RUBY=2.0 KITCHEN_SINK=1` failed on travis, but I suspect it's a red herring. All other jobs passed.

gems/aws-partitions/lib/aws-partitions.rb
casperisfine: > Do you think a JSON::VERSION (or worst case a RUBY_VERSION) check would work instead?

In my experience feature testing is better because it also handle future breakage. But sure no problem.

casperisfine: I pushed a new commit that does a version check.

mullermp (approved):
Looks good! Thanks for making that change.

gems/aws-partitions/lib/aws-partitions.rb
mullermp: Thanks! In the SDK we have many checks against RUBY_VERSION and some other version constants so we find it better to be consistent. I'm personally not a fan of rescue logic for control flow - it's a bit opinionated, but I get where you're coming from. In theory gems follow semantic versioning, so 2.4.0+ should have the feature, but then again, a gem could have a regression, so I see both points.

mullermp:
I'll get this merged Monday morning.

mullermp:
Thanks for your contribution. This should get released on the next aws-partitions update, or we can make a changelog and have it Monday.

mullermp:
FYI - this change broke older Ruby using newer JSON - #2474 . We have a new change #2475 which is similar to your testing script - to use `JSON.parse` instead of `JSON.load`. `JSON.parse` will handle `freeze: true` for older ruby.

We did some memory testing using `MemoryProfiler`

```
require 'memory_profiler'
r = ::MemoryProfiler.report { require ‘aws-partitions’; Aws::Partitions.defaults }
```

Original code (prior to this PR)
```
 @total_allocated=14649,
 @total_allocated_memsize=1378472,
 @total_retained=14637,
 @total_retained_memsize=1092408>
```

Released code (after this PR), no change unfortunately
```
 @total_allocated=14656,
 @total_allocated_memsize=1379296,
 @total_retained=14640,
 @total_retained_memsize=1092648>
```

Proposed change (`JSON.parse` with `freeze: true`)
```
 @total_allocated=14652,
 @total_allocated_memsize=1378856,
 @total_retained=6450,
 @total_retained_memsize=762600>
```

casperisfine:
Argh, really sorry. Thanks for fixing it.

PR #2483, opened 2021-02-24
🐛  fix: Handle IAD regional endpoint plugin correctly for S3 PrivateLink endpoints

harihara-n:
Description
------------

Currently there is a bug when the `aws-sdk-s3` client is used with a private link endpoint and the region is set to `us-east-1`. The IAD regional endpoint plugin drops the `us-east-1` portion. 

So, I get the following issue (using `aws-sdk-s3` gem version `1.88.0`) when I use a PrivateLink endpoint in `us-east-1` but not in other regions like `us-west-2`

```
$ client = Aws::S3::Client.new(region: 'us-east-1', endpoint: 'https://bucket.vpce-xxxx-yyyy.s3.us-east-1.vpce.amazonaws.com')
$ client.list_buckets
Seahorse::Client::NetworkingError: Failed to open TCP connection to bucket.vpce-xxxx-yyyy.s3.vpce.amazonaws.com:443 (getaddrinfo: Name or service not known)
```

So in this PR, we modify the IAD regional endpoint plugin to additionally not drop `us-east-1` if the specified endpoint is a PrivateLink endpoint. 

Testing
--------

- Added unit tests to cover the `us-east-1` case with PrivateLink endpoints. 

Agreement (Generated)
-----------

### DIFF
--- a/gems/aws-sdk-s3/CHANGELOG.md
+++ b/gems/aws-sdk-s3/CHANGELOG.md
@@ -1,7 +1,7 @@
 Unreleased Changes
 ------------------
 
-* Issue - Fix an issue with the IAD regional endpoint plugin dropping the region portion of the PrivateLink endpoint.
+* Issue - Fix an issue with the IAD regional endpoint plugin dropping the region(us-east-1) portion of the PrivateLink endpoint.
 
 1.88.1 (2021-02-12)
 ------------------
--- a/gems/aws-sdk-s3/CHANGELOG.md
+++ b/gems/aws-sdk-s3/CHANGELOG.md
@@ -1,7 +1,7 @@
 Unreleased Changes
 ------------------
 
-* Issue - Fix an issue with the IAD regional endpoint plugin dropping the region(us-east-1) portion of the PrivateLink endpoint.
+* Issue - Fix an issue with the IAD regional endpoint plugin removing `us-east-1` from a PrivateLink endpoint.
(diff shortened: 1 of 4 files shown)


mullermp:
Thanks for opening a pull request. For a workaround (and I recommend personally) you should disable this plugin, either when creating a client or through shared config, setting the value to 'regional' instead of 'legacy'.

harihara-n:
Thanks @mullermp - yes currently we are setting the `AWS_S3_US_EAST_1_REGIONAL_ENDPOINT` environment variable to `legacy` to get around this 👍

mullermp:
Did you mean that you set `AWS_S3_US_EAST_1_REGIONAL_ENDPOINT` to `regional` and not `legacy`? The plugin is skipped if it is `regional`. If that's not happening then something weird is happening.

I'm reviewing your PR now - I am making some changes to it, I hope you don't mind.

mullermp:
I moved the tests to the IAD regional plugin where we also check ARN behavior.

mullermp (approved):

mullermp:
Actually, I think this needs a different implementation. The IAD plugin should not be used for any customer provided endpoint. That solves this case and future cases as well.

Kind of related, I think we'd also need to check a case where "us-east-1" is in the bucket name.

alextwoods (approved):

PR #2488, opened 2021-03-01
Gemspec updates and include license/changelog

### LINKED ISSUE #2487: Ship LICENSE file with gem
https://github.com/aws/aws-sdk-ruby/blob/master/LICENSE.txt should be shipped with the gem, otherwise it's impossible to have the gem properly used with compliance tooling.
Which makes it harder to use it in a corporate environment - this gem is just picked as an example - the missing LICENSE file applies to **all** aws gems provided via rubygems.org

mullermp:
Closes #2487

This change:
* Adds a copy of the current LICENSE.txt to each gem
* Add LICENSE.txt, CHANGELOG.md, and VERSION as files for each gemspec
* Remove personal emails from gemspec

Open questions to investigate
- ~~Whether we need a changelog entry for each gem or will this get flagged as a "build tool" update?~~ I think this is handled.
- ~~Special gems (aws-sigv4, aws-partitions, etc) might need changelog entries.~~


alextwoods (approved):

alextwoods (changes requested):
Hold off on merging - need to address new services / API Gateway generation.

alextwoods (approved):

PR #2489, opened 2021-03-01
Fix global fips endpoint resolution

### LINKED ISSUE #2486: Specifying region `fips-aws-global` does not provide a fips endpoint.

## Summary

Specifying the `fips-aws-global` region for clients like Route53 causes client to be created with non-FIPS endpoint.

## Expected Behavior

```ruby
client = Aws::Route53::Client.new(region: 'fips-aws-global')
client.config.endpoint # => #<URI::HTTPS https://route53-fips.amazonaws.com>
```

## Actual Behavior

```ruby
client = Aws::Route53::Client.new(region: 'fips-aws-global')
client.config.endpoint # => #<URI::HTTPS https://route53.amazonaws.com>
```

## Explanation / Root-Cause

I believe the bug is caused by `Aws::Partitions::EndpointProvider#endpoint_for` [where it checks if the service is regionalized and changes the region if it is not](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-partitions/lib/aws-partitions/endpoint_provider.rb#L80). This causes `fips-aws-global` to be replaced with `aws-global`, thereby [causing the non-FIPS endpoint to be returned.](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-partitions/lib/aws-partitions/endpoint_provider.rb#L84)

This _does_ work correctly for regional endpoints:

```ruby
client = Aws::CloudFormation::Client.new(region: 'us-west-2-fips')
client.config.endpoint # => #<URI::HTTPS https://cloudformation-fips.us-west-2.amazonaws.com>
```

## Additional Thoughts

Specifying FIPS endpoints using an alternate region is _very_ cumbersome. I tried to write a method to tell me if a FIPS endpoint is even available (and what the corresponding region is). Maybe something like this could be included in `Aws::Partitions`.  All of the `Aws::Partitions::Partition` objects strip FIPS region details away and never show up in lists or lookups.

```ruby
    def fips_region_if_exists(client, region)
      client_id = client.identifier.to_s
      %W[
        #{region}-fips
        fips-#{region}
      ].find do |fips_region|
        # This works because every FIPS endpoint has a corresponding `credentialScope` that specifies the 
        # non-FIPS version of the region. Otherwise, the original value is returned.
        Aws::Partitions::EndpointProvider.signing_region(fips_region, client_id) != fips_region 
      end || region
    end

    fips_region_if_exists(Aws::Route53::Client, 'aws-global') # => "fips-aws-global"
    fips_region_if_exists(Aws::CloudFormation::Client, 'us-west-2') # => "us-west-2-fips"
    fips_region_if_exists(Aws::EBS::Client, 'us-west-2') # => "fips-us-west-2"
```

mullermp:
Fixes #2486

The current behavior of SDKs is to use the partition endpoint when the service is not regionalized. Now that FIPS is being modeled, it is breaking current behavior. This is a proposed solution, where we simply check if the customer provided region is a modeled region. If the region is not modeled, then fall back to the normal global endpoint behavior. This change is pending verification and approach from other SDK teams since most SDKs are following this same behavior but others are not.


gems/aws-partitions/lib/aws-partitions/endpoint_provider.rb
```diff
@@ -70,20 +70,22 @@ def endpoint_for(region, service, sts_regional_endpoints)
...
           region = service_cfg.fetch('partitionEndpoint', region)
         end
 
         # Check for service/region level endpoint.
-        endpoint = service_cfg
-                   .fetch('endpoints', {})
-                   .fetch(region, {}).fetch('hostname', endpoint)
+        endpoint = endpoints
+                   .fetch(region, {})
+                   .fetch('hostname', endpoint)
```
alextwoods, line 88: Not really related to this change, but changing `endpoint` on line 69 to `default_endpoint` I think would help the readability of this code...

alextwoods (approved):
LGTM.

alextwoods (approved):

PR #2492, opened 2021-03-17
Retry when waiting for stopped task state if starting or running

plukevdh:
Currently, the waiter raises an immediate failure unless it is already in the stopped state. It does not wait for the task to finish, making this waiter virtually useless.

There may be other wait states in this config that might be worth considering, these are the two states I am concerned about retrying. I  made this change in the gem and the API, unsure of where it belongs. The contrib guidelines state waiters are API-backed changes but I'm not sure what that means in terms of proposing changes. 

Please advise (first time contrib to this project). Let me know if there is anything I can do to resolve.

### DIFF
--- a/apis/dms/2016-01-01/waiters-2.json
+++ b/apis/dms/2016-01-01/waiters-2.json
@@ -189,13 +189,13 @@
                     "argument":"ReplicationTasks[].Status",
                     "expected":"starting",
                     "matcher":"pathAny",
-                    "state":"failure"
+                    "state":"retry"
                 },
                 {
                     "argument":"ReplicationTasks[].Status",
                     "expected":"running",
                     "matcher":"pathAny",
-                    "state":"failure"
+                    "state":"retry"
                 },
                 {
                     "argument":"ReplicationTasks[].Status",
--- a/gems/aws-sdk-databasemigrationservice/lib/aws-sdk-databasemigrationservice/waiters.rb
+++ b/gems/aws-sdk-databasemigrationservice/lib/aws-sdk-databasemigrationservice/waiters.rb
@@ -518,13 +518,13 @@ module Aws::DatabaseMigrationService
                 "argument" => "replication_tasks[].status",
                 "expected" => "starting",
                 "matcher" => "pathAny",
-                "state" => "failure"
+                "state" => "retry"
               },
               {
                 "argument" => "replication_tasks[].status",
                 "expected" => "running",
                 "matcher" => "pathAny",
-                "state" => "failure"
+                "state" => "retry"
               },
               {
                 "argument" => "replication_tasks[].status",


mullermp:
Thank you for opening a pull request. These changes need to be made to an upstream package in our internal git farm. When you made these changes locally with the Ruby SDK, did they work as expected?

mullermp:
The [Status documentation](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/DatabaseMigrationService/Types/ReplicationTask.html#status-instance_method) suggests that you should never see "starting" or "running" when waiting for the task to be stopped. Do you mind sharing how you are using the waiter? Are you perhaps waiting for the task to be stopped shortly after you had created one?

plukevdh:
Yes, right now, I want to be able to kick off a task and immediately begin waiting for it to complete. I have a couple of synchronous tasks which depend on the task completing before moving on. Attempting to wait immediately after the job starts immediately fails because it sees the "starting" state. I believe that a task, once started, moves from starting → running → stopping → stopped. And I want to be able to wait immediately after starting a task until it enters the final stopped state.

plukevdh:
Also, in response to your question "did they work as expected," yes. With a local build of this package with the changes as given in this PR, I can kick the task off, call a waiter, and it will release once the task is completed (i.e., enters the stopped state).

mullermp:
Based on your use case.. I think you're using the wrong waiter? The ReplicationStoppedWaiter is intended to be used after calling the [stop_replication_task](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/DatabaseMigrationService/Client.html#stop_replication_task-instance_method) method. The STOPPED status says: "The task has stopped in response to running the StopReplicationTask operation". So the ideal usage for this waiter is:

```
dms_client.stop_replication_task(...)
dms_client.wait_until(:replication_task_stopped, ...)
```
I'm not sure why you're seeing "stopping" and "stopped" after running what I assume is `start_replication_task`. Can you share any code samples? The docs indicate that you should see STARTING, RUNNING, FAILED, or READY status after running that.

I would need to verify but I am thinking that READY is the "finish" state that should be expected. I think you should actually be using both the `replication_task_ready` and `replication_task_running` waiters. Something like:

```
dms_client.start_replication_task(...)
# waits until task is running
dms_client.wait_until(:replication_task_running, ...)
# at this point, wait to see if the task is done, because it would be ready for another
dms_client.wait_until(:replication_task_ready, ...)
```

mullermp:
Based on usage, I think we can relax some of these waiter failures and change them to be retry.. but I'm fairly confident that it's not supposed to be on StopReplicationTask..

plukevdh:
@mullermp thanks for engaging on this. To clarify, the task I'm running is a one time full-reload. It is not attempting to start an ongoing replication task which is why it transitions through starting → running → stopped. We never need to call `stop_replication_task` because it will stop once it completes. That may be why the confusion here. We trigger a full reload periodically and need to run secondary tasks once that load completes.

plukevdh:
Also I'm not sure if the proposed:

```ruby
dms_client.wait_until(:replication_task_running, ...)
dms_client.wait_until(:replication_task_ready, ...)
```

will work in this case. It might (and could be the solution) but I'm not sure what occurs because this task doesn't reach "ready" state in our use case.

plukevdh:
Actually, given the case where ongoing replication is in question, this waiter would hang indefinitely, which is probably why the waiter is configured to listen only in the case of an explicit stop request. Likely this change could break other use cases for this waiter. However, I'm still not sure how we would detect for my use case because there is no waiter currently that will retry during a `running` state when waiting for `stopped` as opposed to `ready`

mullermp:
Can you clarify what you mean by "one time full-reload"? Are you referring to the DMS Migration Type which is set to "Migrate existing data"? Personally I've never used DMS but I'm happy to setup a minimal setup for my understanding and to help debug. Any additional reproduction steps would be helpful here.

plukevdh:
I'll try to provide all the details I can here.

Describing the task from the API (and filtering for relevant details): `aws dms describe-replication-tasks --filter Name=replication-task-arn,Values=myTaskARN | jq ".ReplicationTasks[0] | {MigrationType, Settings: (.ReplicationTaskSettings | fromjson), Status, StopReason}"`

```json
{
  "MigrationType": "full-load",
  "Settings": {
    "TargetMetadata": {
      "TargetSchema": "",
      "SupportLobs": true,
      "FullLobMode": false,
      "LobChunkSize": 64,
      "LimitedSizeLobMode": true,
      "LobMaxSize": 32,
      "InlineLobMaxSize": 0,
      "LoadMaxFileSize": 0,
      "ParallelLoadThreads": 0,
      "ParallelLoadBufferSize": 0,
      "BatchApplyEnabled": false,
      "TaskRecoveryTableEnabled": false,
      "ParallelLoadQueuesPerThread": 0,
      "ParallelApplyThreads": 0,
      "ParallelApplyBufferSize": 0,
      "ParallelApplyQueuesPerThread": 0
    },
    "FullLoadSettings": {
      "TargetTablePrepMode": "DROP_AND_CREATE",
      "CreatePkAfterFullLoad": false,
      "StopTaskCachedChangesApplied": false,
      "StopTaskCachedChangesNotApplied": false,
      "MaxFullLoadSubTasks": 8,
      "TransactionConsistencyTimeout": 600,
      "CommitRate": 10000
    },
    "Logging": {
      "EnableLogging": false,
      "LogComponents": [
        {
          "Id": "DATA_STRUCTURE",
          "Severity": "LOGGER_SEVERITY_DEFAULT"
        },
        {
          "Id": "COMMUNICATION",
          "Severity": "LOGGER_SEVERITY_DEFAULT"
        },
        {
          "Id": "IO",
          "Severity": "LOGGER_SEVERITY_DEFAULT"
        },
        {
          "Id": "COMMON",
          "Severity": "LOGGER_SEVERITY_DEFAULT"
        },
        {
          "Id": "FILE_FACTORY",
          "Severity": "LOGGER_SEVERITY_DEFAULT"
        },
        {
          "Id": "FILE_TRANSFER",
          "Severity": "LOGGER_SEVERITY_DEFAULT"
        },
        {
          "Id": "REST_SERVER",
          "Severity": "LOGGER_SEVERITY_DEFAULT"
        },
        {
          "Id": "ADDONS",
          "Severity": "LOGGER_SEVERITY_DEFAULT"
        },
        {
          "Id": "TARGET_LOAD",
          "Severity": "LOGGER_SEVERITY_DEFAULT"
        },
        {
          "Id": "TARGET_APPLY",
          "Severity": "LOGGER_SEVERITY_DEFAULT"
        },
        {
          "Id": "SOURCE_UNLOAD",
          "Severity": "LOGGER_SEVERITY_DEFAULT"
        },
        {
          "Id": "SOURCE_CAPTURE",
          "Severity": "LOGGER_SEVERITY_DEFAULT"
        },
        {
          "Id": "TRANSFORMATION",
          "Severity": "LOGGER_SEVERITY_DEFAULT"
        },
        {
          "Id": "SORTER",
          "Severity": "LOGGER_SEVERITY_DEFAULT"
        },
        {
          "Id": "TASK_MANAGER",
          "Severity": "LOGGER_SEVERITY_DEFAULT"
        },
        {
          "Id": "TABLES_MANAGER",
          "Severity": "LOGGER_SEVERITY_DEFAULT"
        },
        {
          "Id": "METADATA_MANAGER",
          "Severity": "LOGGER_SEVERITY_DEFAULT"
        },
        {
          "Id": "PERFORMANCE",
          "Severity": "LOGGER_SEVERITY_DEFAULT"
        },
        {
          "Id": "VALIDATOR_EXT",
          "Severity": "LOGGER_SEVERITY_DEFAULT"
        }
      ],
      "CloudWatchLogGroup": null,
      "CloudWatchLogStream": null
    },
    "ControlTablesSettings": {
      "historyTimeslotInMinutes": 5,
      "ControlSchema": "",
      "HistoryTimeslotInMinutes": 5,
      "HistoryTableEnabled": false,
      "SuspendedTablesTableEnabled": false,
      "StatusTableEnabled": false
    },
    "StreamBufferSettings": {
      "StreamBufferCount": 3,
      "StreamBufferSizeInMB": 8,
      "CtrlStreamBufferSizeInMB": 5
    },
    "ChangeProcessingDdlHandlingPolicy": {
      "HandleSourceTableDropped": true,
      "HandleSourceTableTruncated": true,
      "HandleSourceTableAltered": true
    },
    "ErrorBehavior": {
      "DataErrorPolicy": "LOG_ERROR",
      "DataTruncationErrorPolicy": "LOG_ERROR",
      "DataErrorEscalationPolicy": "SUSPEND_TABLE",
      "DataErrorEscalationCount": 0,
      "TableErrorPolicy": "SUSPEND_TABLE",
      "TableErrorEscalationPolicy": "STOP_TASK",
      "TableErrorEscalationCount": 0,
      "RecoverableErrorCount": -1,
      "RecoverableErrorInterval": 5,
      "RecoverableErrorThrottling": true,
      "RecoverableErrorThrottlingMax": 1800,
      "RecoverableErrorStopRetryAfterThrottlingMax": false,
      "ApplyErrorDeletePolicy": "IGNORE_RECORD",
      "ApplyErrorInsertPolicy": "LOG_ERROR",
      "ApplyErrorUpdatePolicy": "LOG_ERROR",
      "ApplyErrorEscalationPolicy": "LOG_ERROR",
      "ApplyErrorEscalationCount": 0,
      "ApplyErrorFailOnTruncationDdl": false,
      "FullLoadIgnoreConflicts": true,
      "FailOnTransactionConsistencyBreached": false,
      "FailOnNoTablesCaptured": true
    },
    "ChangeProcessingTuning": {
      "BatchApplyPreserveTransaction": true,
      "BatchApplyTimeoutMin": 1,
      "BatchApplyTimeoutMax": 30,
      "BatchApplyMemoryLimit": 500,
      "BatchSplitSize": 0,
      "MinTransactionSize": 1000,
      "CommitTimeout": 1,
      "MemoryLimitTotal": 1024,
      "MemoryKeepTime": 60,
      "StatementCacheSize": 50
    },
    "PostProcessingRules": null,
    "CharacterSetSettings": null,
    "LoopbackPreventionSettings": null,
    "BeforeImageSettings": null
  },
  "Status": "stopped",
  "StopReason": "Stop Reason FULL_LOAD_ONLY_FINISHED"
}
```

This job is configured to run a FULL_LOAD_ONLY, and it drops the tables and reloads from a remote source. I trigger this function through the Ruby SDK as follows:

```ruby
DMS_SYNC_ARN = "myTaskARN"

search_params = {filters: [{name: "replication-task-arn", values: [DMS_SYNC_ARN]}]}
client = Aws::DatabaseMigrationService::Client.new(region: "my-aws-region")

client.start_replication_task({
  replication_task_arn: DMS_SYNC_ARN,
  start_replication_task_type: "reload-target"
})

client.wait_until(:replication_task_stopped, search_params) do |w|
  w.max_attempts = 10
  w.delay = 5
end
```

Once the waiter completes, we run a number of other tasks against the database to tweak some of the data to meet specs. The mix of settings `MigrationType: "full-load"` and the trigger option `start_replication_task_type: "reload-target"` mean this task fires up, runs a single time, then stops.

mullermp:
Thanks.. Let me see if I can engage the open source support team, and maybe even the service team, to help take a look.

mullermp:
So I tried reproducing this. I created two MySQL RDS instances, and labeled them as source and destination in RDS. I created a replication instance as well. All three were in the same VPC and I granted all traffic access to my inbound rules in the security group. The connection tests succeeded. I then created a test task and copied the same settings that you provided. I ran the task and repeatedly called the `describe_replication_tasks` operation using the Ruby SDK. The statuses I saw were: starting -> running -> stopped.

Can you check your console to see if it reports as an error? The Events section will give you that information. My console reports an error, and it has the same Stop Reason as you: `Stop Reason FULL_LOAD_ONLY_FINISHED`

mullermp:
I got it to work now. I made a fake database and table in my source instance, and modified my task to only copy that database/table. In console it says "Load complete" but it also gives the same Stop Reason. That's weird. In any case, I still see the statuses moving from starting -> running -> stopped.

mullermp:
I agree the existing waiters are not sufficient here. After starting the task, I am able to wait for `ReplicationTaskRunning` to return. Then, if I wanted to chain `ReplicationTaskStopped`, I wouldn't be able to do so, because the current status "running" is considered a failure. So most definitely a change might be needed but I think it will be breaking. I need to consider the tradeoff with that.

mullermp:
Ok, so I think the solution to this would be to modify the `ReplicationTaskStopped` waiter to remove this condition entirely:

```
                {
                    "argument":"ReplicationTasks[].Status",
                    "expected":"running",
                    "matcher":"pathAny",
                    "state":"failure"
                },
```

So when using this waiter, if the task is currently running, we will keep waiting until the status is "stopped" even when the status is still "running". The "stopped" state is achieved through a completed full load OR when a deliberate call is made to `stop_replication_task`.

So, after removing the above condition, the following code works for me with your use case:

```
# start the task
databasemigrationservice.start_replication_task(
  replication_task_arn: 'my-task-arn',
  start_replication_task_type: 'reload-target'
)
# wait until it's running. if your migration type wasn't a full load, you'd only want
# to wait for it to be running
databasemigrationservice.wait_until(:replication_task_running) { |w| w.delay = 5 }; 
# since your migration type is full load, wait for it to be "stopped" which is a success
databasemigrationservice.wait_until(:replication_task_stopped) { |w| w.delay = 5 }; 
```

There are other migration types that are continuous replication that we need to support. Someone from DMS confirmed that they would stay in the "running" state, which is why there is a `ReplicationTaskRunning` waiter defined. In your case, I think you would use that waiter AND the stopped waiter.

mullermp:
Could you try removing that entire condition and see if it works for you too? To get this change into the Ruby SDK (and other SDKs) I need to make it to an upstream, internal only package, so I can't merge this PR as is.

plukevdh:
This is great. I appreciate your work on this. I can close this PR if you'd like or let you close it after the upstream change is in place? 

When you say "Could you try removing that entire condition and see if it works," do you mean drop my custom code and test with the standard sdk gem?

mullermp:
We can keep this open until the issue is resolved. I guess what I mean is, apply the following git diff to a local install of your gem and test the changes, the same way you tested the original changes you made as part of this pull request.

Diff:

```
diff --git a/apis/dms/2016-01-01/waiters-2.json b/apis/dms/2016-01-01/waiters-2.json
index ebfc736ec..928b781c7 100644
--- a/apis/dms/2016-01-01/waiters-2.json
+++ b/apis/dms/2016-01-01/waiters-2.json
@@ -1,4 +1,4 @@
-{ 
+{
     "version":2,
     "waiters":{
         "TestConnectionSucceeds":{
@@ -191,12 +191,6 @@
                     "matcher":"pathAny",
                     "state":"failure"
                 },
-                {
-                    "argument":"ReplicationTasks[].Status",
-                    "expected":"running",
-                    "matcher":"pathAny",
-                    "state":"failure"
-                },
                 {
                     "argument":"ReplicationTasks[].Status",
                     "expected":"failed",
diff --git a/gems/aws-sdk-databasemigrationservice/lib/aws-sdk-databasemigrationservice/waiters.rb b/gems/aws-sdk-databasemigrationservice/lib/aws-sdk-databasemigrationservice/waiters.rb
index ec9843c95..7b9e27b1c 100644
--- a/gems/aws-sdk-databasemigrationservice/lib/aws-sdk-databasemigrationservice/waiters.rb
+++ b/gems/aws-sdk-databasemigrationservice/lib/aws-sdk-databasemigrationservice/waiters.rb
@@ -520,12 +520,6 @@ module Aws::DatabaseMigrationService
                 "matcher" => "pathAny",
                 "state" => "failure"
               },
-              {
-                "argument" => "replication_tasks[].status",
-                "expected" => "running",
-                "matcher" => "pathAny",
-                "state" => "failure"
-              },
               {
                 "argument" => "replication_tasks[].status",
                 "expected" => "failed",
```

mullermp:
DMS is looking to lump this change into a future feature release targeted for the next couple weeks. If you can verify that using the two waiters with this change works for you, that would be helpful.

mullermp:
DMS is targeting 4/12 for a new feature release + this fix. I will close this out. Feel free to re-open if needed.

plukevdh:
Sorry I'd not gotten back to this. I appreciate you following this through!

mullermp:
No problem. Seems the feature is delayed..

PR #2495, opened 2021-03-18
Fix FIPS and global endpoint behavior for S3 ARNs

mullermp:
Correctly resolve FIPS and global endpoints for S3 ARNs. If `s3_use_arn_region` is false, we do not want to use s3-external-1 or aws-global regions because ARNs are regionalized and not global. For FIPS, regardless of `s3_use_arn_region`, if the pseudo region matches the ARN's region, and the client region is FIPS, we will use a FIPS endpoint.


mullermp:
This still needs a CHANGELOG entry when approved.

PR #2500, opened 2021-03-29
Update Readme.md

rapito:
# Patching readme.md

This change reflects the new gem being required `aws-sdk` -> `aws-sdk-core` on the example to use config hashes when setting up the gem.


mullermp:
Thanks

PR #2506, opened 2021-04-05
Increase multipart_threshold default from 15 megabytes to 100 megabytes

### LINKED ISSUE #2505: Higher default multipart threshold for S3 uploads
**Is your feature request related to a problem? Please describe.**

I recently went through a frustrating exercise with an, arguably dated [`1.8.0`](https://rubygems.org/gems/aws-sdk/versions/1.8.0), version of the sdk involving the [`upload_file`](https://github.com/aws/aws-sdk-ruby/blob/a82c8981c95a8296ffb6269c3c06a4f551d87f7d/gems/aws-sdk-s3/spec/object/upload_file_spec.rb) function. This particular function was used as part of an ETL pipeline where batched file exports were previously less than 15MB, and became larger than 15 MB (but still less than 100 MB) through some refactoring. Due to the underlying default here - we experienced hung jobs when uploads became multiparted. Fortunately, the support for the options hash to set this value directly allowed us to squash these particular hangups.

**Describe the solution you'd like**

From the [AWS documentation](https://aws.amazon.com/s3/faqs/#:~:text=Individual%20Amazon%20S3%20objects%20can,using%20the%20Multipart%20Upload%20capability.):
> Q: How much data can I store in Amazon S3?
> The total volume of data and number of objects you can store are unlimited. Individual Amazon S3 objects can range in size from a minimum of 0 bytes to a maximum of 5 terabytes. The largest object that can be uploaded in a single PUT is 5 gigabytes. For objects larger than 100 megabytes, customers should consider using the Multipart Upload capability.

Given this guidance, I think a more sensible default for the multipart threshold would be 100 MB instead of 15 MB.

**Describe alternatives you've considered**

There is a workaround, to [set this value directly in the options hash](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/Object.html#upload_file-instance_method) - but I think keeping the library in parity with the guidance from the documentation could save future developers the frustration that I experienced.

- [X] :wave: I may be able to implement this feature request

**Additional context**

This seems like a pretty trivial change, and with the maintainer's blessing - I'd be happy to make this particular change to bump the default to 100 MB.

dbelling:
Fixes #2505 

Increases the default file size threshold for multipart uploads from 15MB to 100MB, per the [guidance from the S3 FAQs](https://aws.amazon.com/s3/faqs/#:~:text=Individual%20Amazon%20S3%20objects%20can,using%20the%20Multipart%20Upload%20capability.)

> Individual Amazon S3 objects can range in size from a minimum of 0 bytes to a maximum of 5 terabytes. The largest object that can be uploaded in a single PUT is 5 gigabytes. For objects larger than 100 megabytes, customers should consider using the Multipart Upload capability.

### DIFF
--- a/gems/aws-sdk-s3/spec/object/upload_file_spec.rb
+++ b/gems/aws-sdk-s3/spec/object/upload_file_spec.rb
@@ -44,9 +44,9 @@ module Aws
           end
         end
 
-        let(:seventeen_meg_file) do
+        let(:one_hundred_seventeen_meg_file) do
           Tempfile.new('one-meg-file').tap do |f|
-            17.times { f.write(one_mb) }
+            117.times { f.write(one_mb) }
             f.rewind
           end
         end
@@ -98,11 +98,11 @@ module Aws
             expect(client).to receive(:put_object).with(
               bucket: 'bucket',
               key: 'key',
-              body: seventeen_meg_file
+              body: one_hundred_seventeen_meg_file
             )
             object.upload_file(
-              seventeen_meg_file,
-              multipart_threshold: 20 * one_meg
+              one_hundred_seventeen_meg_file,
+              multipart_threshold: 200 * one_meg
             )
           end
 
@@ -120,7 +120,7 @@ module Aws
         end
 
         context 'large objects' do
-          it 'uses multipart APIs for objects >= 15MB' do
+          it 'uses multipart APIs for objects >= 100MB' do
             client.stub_responses(:create_multipart_upload, upload_id: 'id')
             client.stub_responses(:upload_part, etag: 'etag')
             expect(client).to receive(:complete_multipart_upload).with(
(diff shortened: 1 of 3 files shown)


dbelling:
It looks like the [travis build](https://github.com/aws/aws-sdk-ruby/runs/2271806408) on the `version-3` branch is currently red, which I've picked up here. Are addressing these errors critical for merging PR candidates?

alextwoods:
The travis build failures for ruby 2.0 are caused by issues installing gems - fixing that isn't in scope for this PR (I'm looking at that separately).  

Overall this looks reasonable - we're having an internal team discussion on the risk/possible impact of this change and should have an answer by end of week.

gems/aws-sdk-s3/spec/object/upload_file_spec.rb
```diff
@@ -44,9 +44,9 @@ module S3
           end
         end
 
-        let(:seventeen_meg_file) do
+        let(:one_hundred_seventeen_meg_file) do
           Tempfile.new('one-meg-file').tap do |f|
```
mullermp: Tempfile name should probably be changed.

mullermp (commented):
Overall the changes make sense, assuming we want to do this. Applications such as Rails [handle their own default](https://github.com/rails/rails/blob/main/activestorage/lib/active_storage/service/s3_service.rb#L19) but I'm worried about other applications that don't do this.

gems/aws-sdk-s3/spec/object/upload_file_spec.rb
dbelling: Noticed the 10 MB file also had a stale identifier, so I've changed that as well.

mullermp:
Thanks for the contribution. We have decided to merge this. I think that applications with proper testing shouldn't be broken, and I believe there aren't any changes to actual use cases.

Issue #2511, opened 2021-04-25
Support for S3:UploadPart presigned URL


josegomezr:
**Is your feature request related to a problem? Please describe.**
I'm trying to implement multipart uploads on rails using direct storage. However I've found it very difficult to sign the `S3:UploadPart` URL endpoints. I can get `S3:PutObject` signed urls, but if I try the same method aka:

```
# obj is an S3::Object
> obj.presigned_url(:upload_part)
ArgumentError (unknown operation "upload_part_object")
```

**Describe the solution you'd like**
Can the `S3::Object` class expose `Aws::S3::Presigner#presigned_request` so I can pass `upload_part` and get signed urls for multipart?

Or Maybe support for `upload_part` for `S3::Object#presigned_url` method.

**Describe alternatives you've considered**
As a quick patch for my experiments, I've injected this in my WIP code (shamelessly copying it from `presigner.rb`):

```ruby
module Aws
  module S3
    class Object
      def presigned_request(http_method, params = {})
        presigner = Presigner.new(client: client)
        presigner.presigned_request(
          http_method,
          params.merge(bucket: bucket_name, key: key)
        )
      end
    end
  end
end
```

and made it work, but it doesn't feel right to be monkeypatching classes around ;)

- [x] :wave: I may be able to implement this feature request

**Additional context**

mullermp:
Thanks for opening an issue. The Resource models contain many issues and aren't really maintained much - There are many, many instances of monkey patching for S3 resource models in the SDK. We're open to pull requests. However, you should be able to do this easily by constructing a `Presigner` object directly and avoiding resource models entirely (recommended).

```
presigner = Aws::S3::Presigner.new(client: obj.client) # you can get the client from a resource model, or pass your own
presigner.presigned_url(:upload_part, ... params ... )
```

josegomezr:
I'll follow your advice!

Thanks for the reply!

mullermp:
No problem. Please re-open if there is concern or if you wish to open a PR.

josegomezr:
I've made a small PR for this case, thought it was gonna be more complicated 😅

PR #2512, opened 2021-04-26
Allow non HTTP verbs actions into S3 Presigner

josegomezr:
By applying the `_object` suffix only if the operation is an HTTP Verb is
possible to use the rest of the S3 like:

- upload_part
- abort_multipart_upload
- complete_multipart_upload

See aws/aws-sdk-ruby#2511

### DIFF
--- a/gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
+++ b/gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
@@ -153,21 +153,35 @@ module Aws
       #     obj.presigned_url(:put, acl: 'public-read')
       #     #=> "https://bucket-name.s3.amazonaws.com/object-key?..."
       #
-      # @param [Symbol] http_method
-      #   The HTTP method to generate a presigned URL for. Valid values
-      #   are `:get`, `:put`, `:head`, and `:delete`.
+      # @example Pre-signed UploadPart PUT
+      #
+      #     # the object uploaded using this URL will be publicly accessible
+      #     obj.presigned_url(:upload_part, part_number: 1, upload_id: 'uploadIdToken')
+      #     #=> "https://bucket-name.s3.amazonaws.com/object-key?..."
+      #
+      # @param [Symbol] method
+      #   The S3 operation to generate a presigned URL for. Valid values
+      #   are `:get`, `:put`, `:head`, `:delete`, `:create_multipart_upload`, 
+      #   `:list_multipart_uploads`, `:complete_multipart_upload`,
+      #   `:abort_multipart_upload`, `:list_parts`, and `:upload_part`.
       #
       # @param [Hash] params
       #   Additional request parameters to use when generating the pre-signed
       #   URL. See the related documentation in {Client} for accepted
       #   params.
       #
-      #   | HTTP Method   | Client Method          |
-      #   |---------------|------------------------|
-      #   | `:get`        | {Client#get_object}    |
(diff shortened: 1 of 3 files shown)


gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
```diff
@@ -188,10 +188,11 @@ def presigned_post(options = {})
       #
       # @return [String]
       #
-      def presigned_url(http_method, params = {})
+      def presigned_url(method, params = {})
         presigner = Presigner.new(client: client)
+        method = "#{http_method}_object" if %w(delete head get put).include?(method)
```
mullermp: These will be symbols, not strings. However `%i` will not work because we support Ruby 1.9.3 for now.

mullermp (changes requested):
The documentation for the method will need to be updated, namely the http_method part as well as the params piece.

Also some unit tests would be appreciated.

gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
josegomezr: I came up with a rather explicit solution, but it goes well in ruby 1.9.3 :sweat_smile:

josegomezr:
> Also some unit tests would be appreciated.

Delivered :rocket: Tested the old behavior (renaming HTTP Verbs to `${verb}_object`) and the new behavior as well.

alextwoods (approved):

mullermp (approved):
Thanks for your contribution!

josegomezr:
My pleasure :+1:

cbeer:
This change also made the `method` case-sensitive. Our client code is pretty clearly in the wrong, and called `presigned_url` with:

```
aws_file.file.presigned_url('GET')
```

Prior to #2512, this worked, but now fails with:

```
ArgumentError: unknown operation "get"

[GEM_ROOT]/gems/aws-sdk-core-3.114.0/lib/seahorse/model/api.rb:39 :in `operation`
[GEM_ROOT]/gems/aws-sdk-s3-1.94.1/lib/aws-sdk-s3/client.rb:13908 :in `build_request`
[GEM_ROOT]/gems/aws-sdk-s3-1.94.1/lib/aws-sdk-s3/presigner.rb:139 :in `_presigned_request`
[GEM_ROOT]/gems/aws-sdk-s3-1.94.1/lib/aws-sdk-s3/presigner.rb:78 :in `presigned_url`
[GEM_ROOT]/gems/aws-sdk-s3-1.94.1/lib/aws-sdk-s3/customizations/object.rb:212 :in `presigned_url`
```

I don't know whether you'd consider fixing this, but I'm offering this here mainly for anyone else that happens to run into it because it is a funny error to track down.

mullermp:
I'm so sorry about that. I think we can fix that. That said, you've probably already changed your code..

cbeer:
Yeah, the previous documentation clearly states we should have used lower-case symbols and we were just lucky it worked before 🤷‍♂️ .

mullermp:
I've created the PR anyway. That said, I'd rather not merge it (because the docs were explicit) but if I see other reports about this I will go ahead and do it. Is that fair?

cbeer:
👍 works for me, thanks. We were clearly in the wrong, but as long as we just made it up and got lucky, there's no reason to muddy the waters more.

josegomezr:
@cbeer I'm sorry this change broke your code 😕 I was about to offer my help but I see a working PR is already done!

PR #2513, opened 2021-05-03
Expose presigned request status to the request handler stack

marcotc:
While implementing a custom request handler for the S3 client, I noticed that requests to generate a presigned URL and real `get_object` requests looked identical to my handler.
Because my handler generates telemetry side-effects that should only account for real S3 remote requests, the these presign requests were generating miscounted telemetry data for my case.

I noticed that there is a signal that this request was a presign request (`context[:presigned_url] = true`), but it was only set _after_ the request stack was executed.

This PR moves that context marking (`context[:presigned_url] = true`) to before the request stack is executed, to allow it to make an informed decision on how to process a presigned url request.

I also noticed that the manual removal of these three handlers is likely due to a similar need as in my case: https://github.com/aws/aws-sdk-ruby/blob/656691b470bb970627286cf1df2bd84049b6787d/gems/aws-sdk-s3/lib/aws-sdk-s3/presigner.rb#L194-L196

### DIFF
--- a/gems/aws-sdk-s3/CHANGELOG.md
+++ b/gems/aws-sdk-s3/CHANGELOG.md
@@ -1,7 +1,7 @@
 Unreleased Changes
 ------------------
 
-* Feature - Expose presigned request status to the request handler stack #2513
+* Issue - Expose presigned request status to the request handler stack #2513
 
 1.94.0 (2021-04-27)
 ------------------
--- a/gems/aws-sdk-s3/CHANGELOG.md
+++ b/gems/aws-sdk-s3/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Feature - Expose presigned request status to the request handler stack #2513
+
 1.94.0 (2021-04-27)
 ------------------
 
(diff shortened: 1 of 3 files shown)


gems/aws-sdk-s3/lib/aws-sdk-s3/presigner.rb
```diff
@@ -195,6 +195,15 @@ def sign_but_dont_send(
...
 
+        # Used for excluding presigned_urls from API request count.
+        #
+        # Store context information as early as possible, to allow
+        # handlers to perform decisions based on this flag if need.
+        req.handle(step: :initialize, priority: 98) do |context|
+          context[:presigned_url] = true
+          @handler.call(context)
+        end
+
```
mullermp: Could you move this to its own method, similar to `use_bucket_as_hostname` and call it from `_presigned_request`?

gems/aws-sdk-s3/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Feature - Expose presigned request status to the request handler stack #2513
```
mullermp: I think it makes sense to classify this as an issue instead - the intended feature has a gap.

mullermp (changes requested):
Thanks for your contribution. The approach seem fine to us. Please make the changes and we can get this merged today.

marcotc:
Thank you for the quick review, @mullermp.
I've made changes as requested, let me know if there's any more feedback. Appreciate it 🙇

mullermp:
Thanks for your contribution!

PR #2515, opened 2021-05-04
Fix finspace clients

mullermp:

### DIFF
--- a/gems/aws-sdk-finspace/CHANGELOG.md
+++ b/gems/aws-sdk-finspace/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Add a `Content-Type` header to mitigate a service side issue.
+
 1.1.0 (2021-05-03)
 ------------------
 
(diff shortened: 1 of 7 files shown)


gems/aws-sdk-finspace/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Add a `Content-Type` header to mitigate a service side issue.
```
joguSD, line 4: Let's note what the impacted services were.

mullermp: Each service has their own change log..

trevorrowe (approved):
The change looks good.

Curious question, have we tried to apply the content-type header to all rest-json services to see if this will work and could potentially be removed as a customization? This would need to be investigated.

mullermp:
>Curious question, have we tried to apply the content-type header to all rest-json services to see if this will work and could potentially be removed as a customization? This would need to be investigated.

Unfortunately the rest json handler inherits from the rest handler, which applies to other protocols other than JSON.. I think it may be possible if we duplicate code.

PR #2517, opened 2021-05-05
Fix content-type for rest json services

mullermp:
Set the `Content-Type` header for REST JSON services and remove plugins. New SDK builds will depend on this new `aws-sdk-core` version.

Finspace ListEnvironments
```
[Aws::Finspace::Client 200 0.294998 0 retries] list_environments()  
=> #<struct Aws::Finspace::Types::ListEnvironmentsResponse environments=[], next_token=nil>
```

NimbleStudio CreateStudio
```
[Aws::NimbleStudio::Client 200 0.532184 0 retries] create_studio(studio_name:"rubysdktest",display_name:"rubysdktest",user_role_arn:"<redact>",admin_role_arn:"<redact>",client_token:"<redact>")  
=> #<struct Aws::NimbleStudio::Types::CreateStudioResponse
``` 


mullermp:
Test failure because of change log

gems/aws-sdk-core/lib/aws-sdk-core/plugins/protocols/rest_json.rb
```diff
@@ -5,7 +5,23 @@ module Plugins
     module Protocols
       class RestJson < Seahorse::Client::Plugin
 
+        class ContentTypeHandler < Seahorse::Client::Handler
+          def call(context)
+            if modeled_body?(context)
```
mullermp: This is currently wrong. I think it needs to be for payloads, with either the payload trait, or any members that aren't bound to URI or header.

gems/aws-sdk-core/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Fix an issue where Rest JSON services do not have a `Content-Type` header.
```
alextwoods, line 4: I'd argue this should be a minor version bump rather than a patch version.

gems/aws-sdk-core/lib/aws-sdk-core/rest/request/body.rb
```diff
@@ -17,7 +17,12 @@ def initialize(serializer_class, rules)
         # @param [Seahorse::Client::Http::Request] http_req
         # @param [Hash] params
         def apply(http_req, params)
-          http_req.body = build_body(params)
+          body = build_body(params)
+          # for rest-json, ensure we send at least an empty object for
+          if body.nil? && @serializer_class == Json::Builder && @rules[:payload]
```
mullermp: I think the `payload` part here is wrong. But I think this is the correct place to put the hack.. This class is shared so we need to check the serializer class.

mullermp (commented):
Not sure if we can remove the blank `content-type` in the net http handler..

gems/aws-sdk-core/lib/aws-sdk-core/rest/request/body.rb
```diff
@@ -17,7 +17,12 @@ def initialize(serializer_class, rules)
...
+            body = Json.dump({})
```
alextwoods: We can just hard code the empty body string (avoid extra serialization overhead).  Though I'd also argue this should be in the `build_body` method instead of here as we're already checking on whether there are params and payload.

gems/aws-sdk-core/CHANGELOG.md
mullermp: Why? It's not a feature. This is what patch versions are for.

gems/aws-sdk-core/lib/aws-sdk-core/rest/request/body.rb
mullermp: Agree on hard coded body string. I don't think it belongs in `build_body` because it applies in all three cases I think. The body must default to `{}` when modeled with the `payload` trait or when there are structures on the body (may not be marked with `payload`).

mullermp:
We need to verify services in which we removed customizations.

mullermp:
I verified:
```
finspacedata.get_working_location
lexmodelsv2.list_bots
lookoutmetrics.list_alerts
nimblestudio.list_studios
sso.list_accounts
```

I also verified S3 after removing the blank content-type.

gems/aws-sdk-core/lib/aws-sdk-core/rest/request/body.rb
```diff
@@ -17,11 +17,27 @@ def initialize(serializer_class, rules)
         # @param [Seahorse::Client::Http::Request] http_req
         # @param [Hash] params
         def apply(http_req, params)
-          http_req.body = build_body(params)
+          body = build_body(params)
+          # for rest-json, ensure we send at least an empty object
+          if body.nil? && @serializer_class == Json::Builder && modeled_body?
```
alextwoods: Is there a reason this can't go in the `Json::Builder#serialize` (aka `to_json`) method.

`Json::Builder#serialize` is called only on the body - and the same runtime inspection of the model for body parameters is already done before calling it - so if its called with a non-empty set of params - we know a body is expected.

Does the empty body also apply to api-gateway?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/protocols/rest_json.rb
```diff
@@ -5,7 +5,31 @@ module Plugins
     module Protocols
       class RestJson < Seahorse::Client::Plugin
 
+        class ContentTypeHandler < Seahorse::Client::Handler
+          def call(context)
```
alextwoods, line 9: Rather than checking for `modeled_body?` here (and having to duplicate the search essentially) - could we ensure that this handler runs after the Rest::Handler, and then just check if a body is set - that we we reduce the number of places that need to understand/inspect the model.

mullermp: Yeah, that can be done.

gems/aws-sdk-core/lib/aws-sdk-core/rest/request/body.rb
mullermp: Talked offline - I don't think so unfortunately. The runtime inspection is an intersection of user params and body params, and is no guarantee that the operation can take a body.

Yes it applies to API Gateway. I've added the handler for it.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/protocols/api_gateway.rb
```diff
@@ -4,9 +4,24 @@ module Aws
   module Plugins
     module Protocols
       class ApiGateway < Seahorse::Client::Plugin
+
+        class ContentTypeHandler < Seahorse::Client::Handler
+          def call(context)
+            body = context.http_request.body
+            # Rest::Handler will set a default JSON body, so size can be checked
+            # if this handler is run after serialization.
+            if body.respond_to?(:size) && body.size > 0
```
alextwoods: Needs the same logic as in the http handler - if body doesn't respond to size, we need to assume there may be streaming content.

mullermp:
Verified APIG
```
<- "POST /test/pets HTTP/1.1\r\nAccept-Encoding: \r\nUser-Agent: aws-apig-ruby/3.120.1 ruby/2.7.4 x86_64-darwin19\r\nContent-Type: application/json\r\nContent-Length: 29\r\nAccept: */*\r\nHost: dqesyacar2.execute-api.us-west-2.amazonaws.com\r\n\r\n"
```

mullermp:
Test failures are because of premature merge from main (changelog verifier)

nateprewitt (approved):

mullermp:
For content-length, may need to do something like:
```
Net::HTTP.constants.map { |c| Net::HTTP.const_get(c) }.select { |c| c.is_a?(Class) }.first.const_get(:REQUEST_HAS_BODY)
```

For content-type, can't avoid the default when there's an empty payload..

build_tools/aws-sdk-code-generator/spec/protocols/input/rest-json.json
```diff
@@ -1486,5 +1486,309 @@
...
+          }
+        },
+        "params": {},
+        "serialized": {
+          "uri": "/blob-payload",
+          "body": "",
+          "headers": {
+            "Content-Length": 0
+          },
+          "excludeHeaders": {
```
alextwoods: Does the excludeHeaders part of the test actually work? I dont see code for it. (Also, does it need to be a map instead of an array?)
Edit: nvm, I see its added w. this change.  But still question the map vs array here - these should never have a value.

gems/aws-sdk-core/CHANGELOG.md
alextwoods: I still am skeptical of this being just a patch version.  I agree its not a feature exactly - but this is a large, high impact change.

gems/aws-sdk-core/lib/seahorse/client/net_http/handler.rb
```diff
@@ -74,7 +74,9 @@ def error_message(req, error)
         # @return [void]
         def transmit(config, req, resp)
           session(config, req) do |http|
+            Thread.current[:net_http_skip_default_content_type] = true
             http.request(build_net_request(req)) do |net_resp|
+              Thread.current[:net_http_skip_default_content_type] = nil
```
alextwoods: I think this unset should also be done in an `ensure` so that its still unset if there are errors.

build_tools/aws-sdk-code-generator/spec/protocols/input/rest-json.json
mullermp: Yeah. We're using an array in the shared testing service, so I switched to an array here.

mullermp: It did work before

gems/aws-sdk-core/lib/seahorse/client/net_http/patches.rb
```diff
@@ -11,10 +11,21 @@ module NetHttp
...
         def self.apply!
+          Net::HTTPGenericRequest.prepend(PatchDefaultContentType)
           return unless RUBY_VERSION < '2.5'
+
           Net::HTTP::IDEMPOTENT_METHODS_.clear
         end
 
+
+        module PatchDefaultContentType
+          def supply_default_content_type
```
alextwoods, line 27: We should document why we patch this here

gems/aws-sdk-core/lib/seahorse/client/plugins/content_length.rb
```diff
@@ -7,16 +7,22 @@ class ContentLength < Plugin
 
         # @api private
         class Handler < Client::Handler
+          # https://github.com/ruby/net-http/blob/master/lib/net/http/requests.rb
+          # Methods without body is forwards compatible, because content-length
```
alextwoods: nit is => are

alextwoods (approved):

PR #2523, opened 2021-05-18
Support Document Types

alextwoods:
Add support for document types in models.  Documents are a union of types supported by json and serialized/deserialized directly onto the wire as is without additional encoding/escaping.  Document types in C2J models use a new property "document" on structure types.:
```json
{
    "shapes": {
        "MyDocument": {
            "type": "structure",
            "document": true
        }
    }
}
```

Documentation added to client (output of generated sdk w/ a test api):
```
    # @option params [Hash,Array,String,Numeric,Boolean] :document_value
    #   Document type which accepts Open Content - unescaped/unencoded JSON
    #   data
    #
    # @return [Types::CreateConnectionResponse] Returns a {Seahorse::Client::Response response} object which responds to the following methods:
    #
    #   * {Types::CreateConnectionResponse#document_value #document_value} => Hash,Array,String,Numeric,Boolean
    #
```

### Usage
Given the MyDocument member definition above, usage of Document types as an input:
```ruby
# send a hash
client.document_input_operation(my_document: { greeting: 'hello' } )

# send an array
client.document_input_operation(my_document: [1,2,3,4] )

# send a string
client.document_input_operation(my_document: "my document string" )
```

And usage on output:
```ruby
resp = client.document_output_operation
# if document is a map/hash (greeting example above)
puts resp.my_document # outputs {"greeting"=>"hello"} - note string keys.
puts resp.my_document['greeting'] # outputs "hello"
```


mullermp:
>Do we need additional documentation added to inputs or outputs to make them more usable?

I think so. We should just mention all of the JSON types.

>Our param validation currently checks against all supported types, but we could be more permissable and allow any type.

Makes sense to only check the types I think

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/client_api_module.rb
```diff
@@ -259,7 +260,9 @@ def shape_class_name(shape)
         if @service.protocol == 'api-gateway' && type == 'timestamp'
           shape['timestampFormat'] = 'iso8601'
         end
-        if SHAPE_CLASSES.key?(type)
+        if type == 'structure' && shape['document']
+          ["Shapes::DocumentShape", shape]
+        elsif SHAPE_CLASSES.key?(type)
```
mullermp: Why is this a special case? Also you can use `document_struct?(shape)` here.

build_tools/aws-sdk-code-generator/spec/protocols/input/json.json
```diff
@@ -491,6 +491,145 @@
       }
     ]
   },
+  {
+    "description": "Document Type",
```
mullermp, line 495: Are these ported from Smithy?

mullermp (approved):

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/client_api_module.rb
alextwoods: Good call, `document_struct?(shape)` works here.  This is a special case because document types have "structure" as the type, but do not behave at all like structures - specifically they will never have any members.  This breaks a number of assumptions we have for our `Shapes::StructureShape` and is why I ended up creating a `Shapes::DocumentShape` instead of adding a property onto StructureShape.  All of the other types in `SHAPE_CLASSES` map neatly from type to shape class, but Document doesn't.

build_tools/aws-sdk-code-generator/spec/protocols/input/json.json
alextwoods: They are loosely based on the smithy protocol tests.

alextwoods:
Need to bump core version dependency in gemspec for all service gems before release.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/client_operation_documentation.rb
```diff
@@ -100,6 +101,9 @@ def option_tags(operation, api)
           if member_ref['jsonvalue']
             docstring = docstring.to_s + "<p><b>SDK automatically handles json encoding and base64 encoding for you when the required value (Hash, Array, etc.) is provided according to the description.</b></p>"
           end
+          if member_shape['document']
+            docstring = docstring.to_s + "<p>Document type which accepts Open Content - unescaped/unencoded JSON data</p>"
```
mtdowling: Nit: probably better to enumerate the accepted types here and avoid the term JSON. Document types aren't limited to JSON, they just use the JSON data model (and in a protocol-agnostic way).

mtdowling (commented):
Where's the code that actually serializes document types? How will your XML serializers and query serializers react if they encounter  a document type (e.g., they should fail rather than try to serialize or deserialize something so that we avoid accidentally introducing some behavior).

alextwoods:
The actual serialization code is in the json builder - it translates input shapes into appropriate Ruby types and then passes that ruby object to our Json "engine". Document types require no additional translation because the union of types we support is a subset of the types supported by the JSON engine.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/gem_builder.rb
```diff
@@ -8,11 +8,16 @@ class GemBuilder
     def initialize(options)
       @options = options
       @service = options.fetch(:service)
+      validate_model!
     end
 
     # @return [Hash]
     attr_reader :options
 
+    def validate_model!
```
mullermp, line 17: Kinda redundant..

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/client_operation_documentation.rb
mullermp: I think address this

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/client_api_module.rb
```diff
@@ -268,8 +269,9 @@ def shape_class_name(shape)
         if @service.protocol == 'api-gateway' && type == 'timestamp'
           shape['timestampFormat'] = 'iso8601'
         end
-
-        if shape['union']
+        if document_struct?(shape)
+          ["Shapes::DocumentShape", shape]
+	elsif shape['union']
```
mullermp: formatting

gems/aws-sdk-core/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Feature - Support Document Types. (#2523)
```
mullermp: Perhaps expand out what they are

mullermp (commented):
Nice!

PR #2527, opened 2021-05-26
aws-sdk-core/xml/builder: no default indentation

jeremy-clerc:
XML document indentation increases the body size. It can
be useful for debugging purposes, but under normal operations,
it is not helpful.

The result of a larger body size makes operations like
`complete_multipart_upload` fail for certain S3 gateways
(Ceph Object Gateway at least) when having 10k parts,
because of the body being too large.

Compared to boto3, boto3 does not indent, which makes
the body smaller than the fatal limit.

This commit removes the default indentation to have the
same behavior as boto3 and to be able to use multipart
uploads with 10k parts with Ceph Object Gateway.

### DIFF
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Change XML Builder to not indent by default
+
 3.114.0 (2021-04-13)
 ------------------
 
--- a/gems/aws-sdk-core/lib/aws-sdk-core/xml/builder.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/xml/builder.rb
@@ -11,7 +11,7 @@ module Aws
       def initialize(rules, options = {})
         @rules = rules
         @xml = options[:target] || []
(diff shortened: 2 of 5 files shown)


alextwoods:
I generally agree with this PR - we should be sending XML that is as short as possible.  I'm a little nervous about changing this  core serialization behavior thats been in place for a long time - however other SDKs do send more compact XML and the services certainly accept it.  

When you push updates, can you push new commits instead of amending/force pushing?  It makes it easier to review/see CI history.

jeremy-clerc:
> When you push updates, can you push new commits instead of amending/force pushing? It makes it easier to review/see CI history.

Sure, sorry about that I was trying to get things working while keeping a clean history. I'll make new commits if changes are needed.

jeremy-clerc:
On your first part of the comment, I can understand.
So maybe if the part size was made configurable, we could kind of workaround it by doing a pseudo calculation of part size on our side to stay below 9900 parts for example.

alextwoods:
Sorry for the delay - this change looks good overall - we should be serializing to more minimized xml.  I'll merge and release it early next week.

alextwoods (approved):
LGTM - Thanks for submitting this!

PR #2533, opened 2021-06-08
Fix nil pagination tokens not merging into next params

mullermp:
Sometimes an API returns an intentionally nil token. Merging this nil token over a previous token will not work. This approach deletes all previous tokens from the original param context.

Needs unit tests. Manual testing works on `route53.list_resource_record_sets`.

### DIFF
--- a/gems/aws-sdk-core/spec/aws/pageable_response_spec.rb
+++ b/gems/aws-sdk-core/spec/aws/pageable_response_spec.rb
@@ -79,8 +79,8 @@ module Aws
 
       it 'removes previous tokens on subsequent #next_page calls' do
         client = double('client')
-        new_request = double('new-request')
-        new_new_request = double('new-new-request')
+        request_1 = double('new-request')
+        request_2 = double('new-new-request')
 
         resp.data = { 'next_token' => 'OFFSET', 'other_token' => 'TOKEN' }
         resp.context.client = client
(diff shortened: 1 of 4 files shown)


gems/aws-sdk-core/spec/aws/pageable_response_spec.rb
```diff
@@ -76,6 +77,38 @@ def pageable(resp, pager)
         resp.next_page
       end
 
+      it 'removes previous tokens on subsequent #next_page calls' do
+        client = double('client')
+        new_request = double('new-request')
```
alextwoods: maybe change these names to be `request_page_1` and `request_page_2`

gems/aws-sdk-core/spec/aws/pageable_response_spec.rb
```diff
@@ -76,6 +77,38 @@ def pageable(resp, pager)
...
+        next_page = resp.next_page
+
+        expect(client).to receive(:build_request).
+          with('operation-name', { :offset => 'OFFSET' }).
+          and_return(new_new_request)
+
+        expect(new_new_request).to receive(:send_request).
+          and_return(Seahorse::Client::Response.new)
+
+        next_page.next_page
```
alextwoods: What is this checking?  the `next_page.next_page` is a bit confusing

alextwoods (approved):

PR #2537, opened 2021-06-10
Fix stubbing for XML by using indent

mullermp:
Fixes #2536

PR #2538, opened 2021-06-10
Fix xml parsing of S3 bucket location constraint

### LINKED ISSUE #2536: AWS S3 GetBucketLocation XML Parser Fails with Stubbed Responses
  
**Describe the bug**
When using stubbed responses for get_bucket_location location_constraint the XML parsing fails to correctly parse the region.

`aws-sdk-s3` version `1.96.0`
`aws-sdk-core` version `3.114.2`

This issue was introduced in `aws-sdk-core` version `3.114.1` with https://github.com/aws/aws-sdk-ruby/pull/2527

```
$ ruby -v
ruby 2.7.3p183 (2021-04-05 revision 6847ee089d) [x86_64-linux-gnu]
```

```
>> require 'aws-sdk-s3'
=> true
>> Aws.config[:s3] = {stub_responses: {get_bucket_location: {location_constraint: "us-east-1"}}}
=> {:stub_responses=>{:get_bucket_location=>{:location_constraint=>"us-east...
>> Aws::S3::Client.new.get_bucket_location(:bucket => "test")
=> #<struct Aws::S3::Types::GetBucketLocationOutput location_constraint="<LocationConstraint>us-east-1">
>> Aws::S3::Client.new.get_bucket_location(:bucket => "test").location_constraint
=> "<LocationConstraint>us-east-1"
```

**Expected behavior**
`get_bucket_location(bucket: "test").location_constraint` should return `"us-east-1"` not `"<LocationConstraint>us-east-1"`

**Additional Context**
Reverting https://github.com/aws/aws-sdk-ruby/pull/2527 or updating the regex [here](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/get_bucket_location_fix.rb#L14) both resolve the issue.

agrare:
Just wanted to throw out a possible fix for the issue described in https://github.com/aws/aws-sdk-ruby/issues/2536

The XML that is returned when using stubbed responses is: `<GetBucketLocationResult xmlns=\"\"><LocationConstraint>EU</LocationConstraint></GetBucketLocationResult>` which fails with the regex `/>(.+?)<\/LocationConstraint>/`
Fix an issue with parsing of an S3 bucket location constraint introduced by removal of default xml indentation.

Fixes https://github.com/aws/aws-sdk-ruby/issues/2536

### DIFF
--- a/gems/aws-sdk-s3/CHANGELOG.md
+++ b/gems/aws-sdk-s3/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - fix GetBucketLocation location_constraint XML parsing (#2536)
+
 1.96.0 (2021-06-03)
 ------------------
 
--- a/gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/get_bucket_location_fix.rb
+++ b/gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/get_bucket_location_fix.rb
@@ -11,7 +11,7 @@ module Aws
             @handler.call(context).on(200) do |response|
               response.data = S3::Types::GetBucketLocationOutput.new
               xml = context.http_response.body_contents
-              matches = xml.match(/>(.+?)<\/LocationConstraint>/)
+              matches = xml.match(/<LocationConstraint.*?>(.+?)<\/LocationConstraint>/)
               response.data[:location_constraint] = matches ? matches[1] : ''
             end
           end


mullermp:
Interesting. I have out this change here: https://github.com/aws/aws-sdk-ruby/pull/2537 but your change uses no indenting. I actually did not notice we had a plugin *specifically* for location constraint. I suppose yours might make more sense than mine which prefers previous behavior.

mullermp:
I think this problem is specifically for `GetBucketLocation`, right? Because current mainline changes seem to work for stubbing something like `list_buckets` and we would have heard serious backlash if it broke every Rest XML stubbing response.

agrare:
Yeah I wanted to throw out a possible solution but I'm not as familiar with this gem's internals as you guys so I'll defer to you on which approach you want to take.

I agree that this is specific to the `GetBucketLocation` method.

mullermp:
The SDK is so big, I sometimes see something new every day :D

I think this change is preferred over mine.

mullermp (approved):

alextwoods (approved):

mullermp:
Thanks for reporting and contributing.

agrare:
:bow: thanks @mullermp @alextwoods

PR #2540, opened 2021-06-15
Don't modify original_params in PageableResponse

driv3r:
as frozen hashes will cause it to fail. Followup to #2533


mullermp:
Thanks for a pull request. Sorry that I broke this. I did not consider frozen hashes.

Issue #1725, opened 2018-03-06
s3_client.get_bucket_acl doesn't follow 301 redirect


vbichov:
### Issue description

when trying the following:
bucket location: "us-west-1"

    s3 = Aws::S3::Client.new ({:region => "us-east-1"})
    s3.get_bucket_acl ({:bucket => "some-bucket-name"})

you will get Aws::S3::Errors::PermanentRedirect exception

the exception will contain the correct region.

python api follows that redirect and retrieves the acl

vbichov:
boilerplate workaround:
def getBucketACL(bucketName)
  begin
    return @s3.get_bucket_acl ( {:bucket => bucketName})
  rescue Aws::S3::Errors::PermanentRedirect => e
    Aws.config.update ({region: e.context.http_response.headers['x-amz-bucket-region']})
    @s3 = Aws::S3::Client.new
    return getBucketACL(bucketName)
    
  end
end

cjyclaire:
@vbichov Interestingly, I cannot reproduce this. I tried the script and it worked for me.
Some quick questions, is `some-bucket-name` in the `us-east-1` region for you? Did you provide `region` from other places? like Aws.Config, `AWS_REGION` parameters? For credentials, are you using `/.aws/credentials` file or `/.aws/config` file?

awood45:
Which version of the SDK are you using, also?

vbichov:
aws-sdk-s3 (1.8.2)

Out of 143 buckets I'm getting this error only on 2.

Those 1 is located in us-west-1 and the other is in us-east-1

credentials are passed as a hash.

I will try to understand what's different in those buckets

cjyclaire:
@vbichov Does this still happening for you? SDK does support fetching region from `'x-amz-bucket-region'` [here](https://github.com/aws/aws-sdk-ruby/blob/9ea447b7e9e57e27bb7924aca40a5bf0c3265a94/gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/s3_signer.rb#L124).

I also tried with a client in `us-east-1` and fetching `us-west-1` bucket, it succeed without an issue for me, also, it sounds weird when you mentioned a `us-east-1` client failed to fetch `us-east-1` bucket with this error. Could you provide more debugging info by setting `http_wire_trace` to true for the client? This would provide the wire trace, then you can identify which endpoint/region it's trying to hitting and whether SDK is functioning properly.

cjyclaire:
[...]

PR #2546, opened 2021-06-28
Correct override for null LocationConstraint in us-east-1.

dvogel:

Related to issue #1725, this changes the synthetic value returns by the SDK for a bucket's location constraint to match the API docs.

### DIFF
--- a/gems/aws-sdk-s3/CHANGELOG.md
+++ b/gems/aws-sdk-s3/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Replace null LocationConstraint with us-east-1 instead of an empty string.
+
 1.96.1 (2021-06-10)
 ------------------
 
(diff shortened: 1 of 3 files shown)


mullermp:
Thanks for opening a pull request. Unfortunately, I believe fixing this is a breaking change. This behavior has existed for too long. I imagine other applications are checking this value and comparing it to an empty string.

mullermp:
Thanks for opening a pull request. I think this change on its own is breaking behavior. It modifies the response value, and other applications may explicitly be looking for this value as an empty string.

Could you tell me more about your use case?

alextwoods:
Yeah - I unfortunately agree with @mullermp that this is a breaking change and existing applications are likely relying on the nil value.

dvogel:
My use case isn't complicated. I'm merely trying to enumerate all buckets in an account and then fetch the tag set for each.

This is technically a breaking change. In my review of the history of this plugin it seemed to be introduced as a breaking change. Upon closer inspection this morning I see I missed a flurry of file renames and the history goes back much further. I'll close this PR.

FWIW I think it would be best in version-4 for this to allow users to rely on a null value. The SDK currently returns an empty string. The service's API returns null, the awscli shows a bare `null` in the JSON response, and the docs in this gem's docs repeat the service's docs re: null. However the value returned is a string.

alextwoods:
I agree this should be changed in a new major version.  If you create an issue, we can mark it as `needs-major-version` and add it to our backlog of breaking changes for version 4 of the SDK.

mullermp:
I think that future behavior could/should return nil instead of an empty string.

mullermp:
The original issue you commented on was about following redirects. Is that not the case for you? It would not be breaking to fix following redirects AFAIK.

PR #2555, opened 2021-07-13
Fix global service config for multiple services

### LINKED ISSUE #2554: ArgumentError (invalid configuration option `:dynamodb') thrown when using custom DynamoDB endpoint
  
**Describe the bug**
ArgumentError (invalid configuration option `:dynamodb') error is thrown when using AWS SDK with a custom endpoint for DynamoDB.

From Gemfile.lock
```
aws-sdk (~> 3)
aws-sdk-s3 (~> 1)
aws-sdk-dynamodb (1.60.0)
```

ruby 2.5.3p105 (2018-10-18 revision 65156)

Start Docker instances of S3, SQS and DynamoDB for testing. In the Ruby app set the custom Dynamo DB, S3 and SQS endpoint like this:
```
Aws.config.update({:s3 => {
  :endpoint => 'http://localhost:41001', :force_path_style => true
}}) 

Aws.config.update({:sqs => {
  :endpoint =>'http://localhost:41002'
}}) 
Aws.config.update({:dynamodb => {
  :endpoint => 'http://localhost:41000'
}}) 
```

Then when calling a S3 operation, I see the error.
```
F, [2021-07-09T18:48:15.195773 #13176] FATAL -- :   
F, [2021-07-09T18:48:15.195924 #13176] FATAL -- : ArgumentError (invalid configuration option `:dynamodb'):
F, [2021-07-09T18:48:15.196004 #13176] FATAL -- : 
```

**Expected behavior**
This error should not happen. The custom DynamoDB configuration (endpoint) should be accepted.

**Screenshots**

**Additional context**

mullermp:
Fixes #2554

Needs a test.. The test I wrote originally would pass even before and after..

### DIFF
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Fix an issue where global `Aws.config` could not be used with more than one service.
+
 3.117.0 (2021-07-12)
 ------------------
 
--- a/gems/aws-sdk-core/lib/aws-sdk-core/plugins/global_configuration.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/plugins/global_configuration.rb
@@ -44,6 +44,7 @@ module Aws
     class GlobalConfiguration < Seahorse::Client::Plugin
 
       @identifiers = Set.new()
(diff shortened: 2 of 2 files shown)


mullermp:
Implementation is not ideal - relies on partitions endpoint data which may be a subset of actual code generated client.

mullermp:
This is already supported..

PR #2557, opened 2021-07-13
Tagged unions

alextwoods:
**Summary**: Add support for the union trait following a "sealed" classes like approach while remaining backwards compatible with existing Structure.  Union types have generated subclasses for each member as well as unknown.
This is loosely based on the original PR: https://github.com/aws/aws-sdk-ruby/pull/2427

There are two types of behavior modified:
1. Unions as inputs - when used as an input, unions validate that exactly one member is set.
2. Unions as outputs - during parsing, the generated subclass for the set member is used and the single member is set on it.  While all of the structure member accessors still exist to preserve backwards compatibility, the single set member's value can be accessed through the `value` method.

The subclasses make it easy to use union output types in case statements:
```ruby
item = `ddb.get_item(params).item['value']
case item
when Types::AttributeValue::S then handle_string_member(item.value)
when :Types::AttributeValue::SS then handle_ss_member(item.value)
when Types::AttributeValue::Unknown then handle_unknown_value(item)
end
```

**NOTE**:  (do not merge as is) This PR contains a modification to dynamodb api to add `union` to the dynamodb `AttributeValue` type to demonstrate the changes from code gen.


build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/client_api_module.rb
```diff
@@ -31,7 +31,8 @@ class ClientApiModule < View
         'xmlNamespace' => true,
         'streaming' => true, # transfer-encoding
         'requiresLength' => true, # transder-encoding
-        # event stream modeling
+        'union' => false,
+      # event stream modeling
```
mullermp: format

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/client_api_module.rb
```diff
@@ -122,6 +122,7 @@ def shapes
       end
 
       def shape_definitions
+
```
mullermp: format

mullermp (approved):
Approach and changes are all good. Nice. Minor formatting.

PR #2559, opened 2021-07-13
Add retry on InvocationDoesNotExist to SSM SendCommand waiter

### LINKED ISSUE #2556: SSM: `wait_until :command_executed` fails with `InvocationDoesNotExist` error
  
**Describe the bug**

Calling `Aws::SSM::Client#wait_until :command_executed` immediately after `#send_command` often raises `Aws::SSM::Errors::InvocationDoesNotExist`, presumably due to eventual consistency on the API side. Inserting a sleep between `send_command` and the waiter makes it less likely, but a robust solution essentially requires manually implementing a waiter for the command invocation to exist before waiting for it to be executed.

aws-sdk-ssm 1.111.0

ruby 2.7.3p183 (2021-04-05 revision 6847ee089d) [x86_64-darwin20]

```ruby
ssm = Aws::SSM::Client.new

instance_id = "..."

send_command_result = ssm.send_command(document_name: "...", instance_ids: [instance_id])
command_id = send_command_result.command.command_id
p ssm.wait_until(:command_executed, command_id: command_id, instance_id: instance_id)
# => ~/.gem/ruby/2.7.3/gems/aws-sdk-core-3.114.2/lib/aws-sdk-core/waiters/waiter.rb:114:in `block in poll': stopped waiting due to an unexpected error: Aws::SSM::Errors::InvocationDoesNotExist (Aws::Waiters::Errors::UnexpectedError)

```

**Expected behavior**

The `command_executed` waiter waits until the command has been executed.

**Additional context**

Related: https://github.com/aws/aws-sdk-js/issues/3457

alextwoods:
Fixes #2556 

[Do not merge] - This change needs to be merged upstream - this PR is created to demo/test the changes.

There is an eventually consistency issue with SendCommand - the command may not exist the first time the waiter attempts to check the status, resulting in an `InvocationDoesNotExist` error.   

The risk here is that a waiter given a command_id that indeed does not exist, will go through all of its retries before failing. 

Related JS issue: https://github.com/aws/aws-sdk-js/issues/3457

### DIFF
--- a/apis/ssm/2014-11-06/waiters-2.json
+++ b/apis/ssm/2014-11-06/waiters-2.json
@@ -53,6 +53,11 @@
           "matcher": "path",
           "state": "failure",
           "argument": "Status"
+        },
+        {
+          "state": "retry",
+          "matcher": "error",
+          "expected": "InvocationDoesNotExist"
         }
       ]
     }
--- a/gems/aws-sdk-ssm/lib/aws-sdk-ssm/waiters.rb
+++ b/gems/aws-sdk-ssm/lib/aws-sdk-ssm/waiters.rb
@@ -136,6 +136,11 @@ module Aws::SSM
                 "matcher" => "path",
                 "state" => "failure",
                 "argument" => "status"
+              },
+              {
+                "state" => "retry",
+                "matcher" => "error",
+                "expected" => "InvocationDoesNotExist"
               }
             ]
           )


mullermp:
I think a second waiter might be necessary here.. One that checks for the command existing. Otherwise this waiter can wait too long (if command ID is not correct).

alextwoods:
> I think a second waiter might be necessary here.. One that checks for the command existing. Otherwise this waiter can wait too long (if command ID is not correct).

There is a risk of the waiter here waiting too long for a command ID that is not correct.  However, introducing a new waiter that waits for the Invocation to exist has the same risk and then requires a user to use 2 waiter sequentially.

mullermp:
How does it have the same risk? It would retry on 404 and return success when it's available.

alextwoods:
Waiter changes were released by the service today and are available in `aws-sdk-ssm` version `1.118.0`.

PR #2553, opened 2021-07-09
Support modeled dualstack and fips variant endpoints; Support S3 Outpost FIPS

mullermp:
Resolves endpoint variants and adds configuration for them

Core changes:

* Feature - Support modeled dualstack endpoints. It can be configured with shared configuration (`use_dualstack_endpoint`), an ENV variable (`AWS_USE_DUALSTACK_ENDPOINT`), and a constructor option (`:use_dualstack_endpoint`). Requests made to services without a dualstack endpoint will fail.

* Feature - Support modeled fips endpoints. It can be configured with shared configuration (`use_fips_endpoint`), an ENV variable (`AWS_USE_FIPS_ENDPOINT`), and a constructor option (`:use_fips_endpoint`). Requests made to services without a fips endpoint will fail.

Partitions changes:
* Feature - Support modeled dualstack and fips endpoints in `Aws::Partitions::EndpointProvider`.

PR #2562, opened 2021-07-19
Support S3 Outposts FIPS using S3Control

mullermp:
Support FIPS endpoints for S3 Outposts


mullermp:
Merged into https://github.com/aws/aws-sdk-ruby/pull/2553/files

PR #2563, opened 2021-07-19
Fix to prevent downloading 1 byte ranges

### LINKED ISSUE #2561: aws-sdk-s3 fails when downloading file of exactly 5242881 bytes from S3 bucket
  
**Describe the bug**
When obtaining an object of exactly 5242881 bytes from an S3 bucket, the download will fail with the error `Aws::S3::Errors::InvalidRange: The requested range is not satisfiable`.

Gem: **`aws-sdk-s3` 1.96.1**

```
ruby 2.7.4p191 (2021-07-07 revision a21a3b7d23) [x86_64-darwin20]
```
But this is also reproducible under Linux with the same Ruby version and Gem version.

 1. Upload a file of exactly 5242881 bytes to an S3 bucket.
 2. Using the gem, build and S3 object and download the file:
 ```
s3_obj = s3_bucket.object(object_name)
s3_obj.download_file(path)
```
 3. See the error.

**Expected behavior**
I expect the S3 Gem to be able to download the file completely.

**Additional context**
I did some debugging and it seems related to how the request for a file download is processed by the gem. By default, the gem will try to download the file in multiple ranges simultaneously for a fast download, however, the chunks the file is devided into seem a bit odd:
```
["bytes=0-5242880", "bytes=5242881-5242881"]
```
In other words, the first 5242880 bytes are downloaded as 1 part, and the final byte as a separate part. The calculation of these chunks happens in [file_downloader.rb](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb#L81):
```ruby
      def construct_chunks(file_size)
        offset = 0
        default_chunk_size = compute_chunk(file_size)
        chunks = []
        while offset <= file_size
          progress = offset + default_chunk_size
          chunks << "bytes=#{offset}-#{progress < file_size ? progress : file_size}"
          offset = progress + 1
        end
        chunks
      end
```
The `default_chunk_size` is set to `5242880` causing the function to return `["bytes=0-5242880", "bytes=5242881-5242881"]`.

The `default_chunk_size` is calculated by `compute_chunk` in [file_downloader.rb](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb#L93):
```
      def compute_chunk(file_size)
        if @chunk_size && @chunk_size > file_size
          raise ArgumentError, ":chunk_size shouldn't exceed total file size."
        else
          @chunk_size || [(file_size.to_f / MAX_PARTS).ceil, MIN_CHUNK_SIZE].max.to_i
        end
      end
```
In our scenario, `@chunk_size` equals `nil`, causing the `chunk_size` to be calculated using `[(file_size.to_f / MAX_PARTS).ceil, MIN_CHUNK_SIZE].max.to_i`. Since the file is 5242881 bytes and `MAX_PARTS` > 1, the `MIN_CHUNK_SIZE` is used, which is exactly 5242880 bytes.

The download eventually fails when doing the request for the last single byte, chunk `"bytes=5242881-5242881"`. This results in de S3 gem returning `The requested range is not satisfiable (Aws::S3::Errors::InvalidRange)`.

It seems that a single byte range is invalid, but since this is all calculated within the code of the gem, this seems to be a bug.

mullermp:
Fixes #2561. If part size would leave one byte in a part, we subtract one byte from the chunk size.

### DIFF
--- a/gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
+++ b/gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
@@ -94,12 +94,12 @@ module Aws
         if @chunk_size && @chunk_size > file_size
           raise ArgumentError, ":chunk_size shouldn't exceed total file size."
         else
-          @chunk_size ||= [
+          chunk_size = @chunk_size || [
             (file_size.to_f / MAX_PARTS).ceil,
             MIN_CHUNK_SIZE
           ].max.to_i
-          @chunk_size -= 1 if file_size % @chunk_size == 1
-          @chunk_size
+          chunk_size -= 1 if file_size % chunk_size == 1
+          chunk_size
         end
       end
 
--- a/gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
+++ b/gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
@@ -94,7 +94,11 @@ module Aws
         if @chunk_size && @chunk_size > file_size
           raise ArgumentError, ":chunk_size shouldn't exceed total file size."
         else
-          @chunk_size || [(file_size.to_f / MAX_PARTS).ceil, MIN_CHUNK_SIZE].max.to_i
+          @chunk_size ||= [
+            (file_size.to_f / MAX_PARTS).ceil,
+            MIN_CHUNK_SIZE
+          ].max.to_i
+          @chunk_size -= 1 if file_size % @chunk_size == 1
         end
       end
 
--- a/gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
+++ b/gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
@@ -99,6 +99,7 @@ module Aws
             MIN_CHUNK_SIZE
           ].max.to_i
           @chunk_size -= 1 if file_size % @chunk_size == 1
+          @chunk_size
         end
       end
 
--- a/gems/aws-sdk-s3/spec/object/download_file_spec.rb
+++ b/gems/aws-sdk-s3/spec/object/download_file_spec.rb
@@ -36,10 +36,19 @@ module Aws
           )
         end
 
+        let(:boundary_obj) do
+          S3::Object.new(
+            bucket_name: 'bucket',
+            key: 'boundary',
+            client: client
+          )
+        end
+
         let(:one_meg) { 1024 * 1024 }
         let(:small_file) { Tempfile.new('small-file') }
         let(:large_file) { Tempfile.new('large-file') }
         let(:single_part_file) { Tempfile.new('single-part-file') }
+        let(:boundary_file) { Tempfile.new('range-boundary-file') }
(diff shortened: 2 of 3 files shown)


gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -94,7 +94,12 @@ def compute_chunk(file_size)
         if @chunk_size && @chunk_size > file_size
           raise ArgumentError, ":chunk_size shouldn't exceed total file size."
         else
-          @chunk_size || [(file_size.to_f / MAX_PARTS).ceil, MIN_CHUNK_SIZE].max.to_i
+          @chunk_size ||= [
```
alextwoods: I don't think we want to actually set the `@chunk_size` as the object may be re-used.  Create a temp `computed_size`?

gems/aws-sdk-s3/spec/object/download_file_spec.rb
```diff
@@ -174,6 +183,53 @@ module S3
               ":chunk_size shouldn't exceed total file size."
             )
         end
+
+        context 'file size is chunk size boundary' do
+          before :each do
+            allow(client).to receive(:head_object).with(
```
alextwoods, line 188: Why not use stub_responses for this? or does it need argument matching?

gems/aws-sdk-s3/spec/object/download_file_spec.rb
```diff
@@ -174,6 +183,53 @@ module S3
...
+              bucket: 'bucket',
+              key: 'boundary',
+              part_number: 1
+            ).and_return(
+              client.stub_data(
+                :head_object,
+                content_length: (one_meg * 5) + 1
```
alextwoods: would be good to have that as a let

gems/aws-sdk-s3/spec/object/download_file_spec.rb
```diff
@@ -174,6 +183,53 @@ module S3
...
+          end
+
+          it 'downloads the file with provided chunk size' do
+            expect(client).to receive(:head_object).with(
+              bucket: 'bucket',
+              key: 'boundary',
+              part_number: 1
+            ).exactly(1).times
+
+            boundary_obj.upload_file(boundary_file)
```
alextwoods: I know all of the other tests have this pattern as well - but why is it nessecary to call upload_file before download_file.  These aren't integration tests

gems/aws-sdk-s3/spec/object/download_file_spec.rb
mullermp: Other tests needed it because the test files were re-used. I'm just following existing pattern.

gems/aws-sdk-s3/spec/object/download_file_spec.rb
mullermp: It's not. Just following pattern and easy to reason about.

alextwoods (approved):

PR #2573, opened 2021-08-23
Drop old Ruby runtimes

mullermp:
Drops old Ruby runtimes per announcement https://aws.amazon.com/blogs/developer/announcing-the-end-of-support-for-ruby-runtimes-1-9-2-0-2-1-and-2-2-for-the-aws-sdk-for-ruby/


mullermp:
Adding gemspecs for sigv2, sts, and sso gems, even though it may not be necessary.

gems/aws-sdk-s3/spec/encryption/client_spec.rb
```diff
@@ -632,31 +632,29 @@ def stub_encrypted_get_with_instruction_file(sfx = '.instruction')
             "\x8E\x0E\xC0\xD5\x1A\x88\xAF2\xB1\xEEg#\x15"
           end
 
-          if RUBY_VERSION > '1.9.3' && OpenSSL::Cipher.ciphers.include?('aes-256-gcm')
-            it 'supports decryption via KMS w/ GCM' do
```
alextwoods, line 636: We still need to include the second check here AFAIK - there are instances where the openssl library locally does not support that this this that are outside of ruby version.

alextwoods (approved):

michaelwittig:
This is a breaking change? We install aws-sdk-ec2 version 1.200.0 which seems to depend on an unpinned version of aws-partitions which now breaks because aws-partitions required ruby 2.3.

and BTW we install ruby on AL2 which is still at Version 2.0

mullermp:
Thanks for reporting an issue. This change was intended in accordance with our blog post (see top comment on this PR). The changes from the PR were considered safe as most applications use Bundler.

I think I may have failed to consider a subset of customers. The `aws-sdk-ec2` gem depends on `aws-sdk-core`, which depends on `aws-partitions >= 1.239.0`. Since service gems such as EC2 depend on new core versions, the core version should depend on `aws-partitions >= 1.494.0`. I can investigate whether we should bump that or if it's too late at this point..

Have you looked into suitable workarounds? On AL2, you can use the [Extras library](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/amazon-linux-ami-basics.html#extras-library) to install newer versions of Ruby. Alternatively, you can pin your `aws-partitions` version to `1.493.0`.

michaelwittig:
I would prefer to follow https://github.com/aws/aws-sdk-ruby#versioning

numberten:
We experienced an incident relating to this as well on 9/1. Specifically, we began depending on `aws-eventstream:1.2.0` and `aws-sigv4:1.3.0` which dropped support for our old version of Ruby.

Totally makes sense that y'all would drop old versions of Ruby (and thanks for making the announcement so many months in advance), but I'm curious if there's a reason for not making the changes a major version bump?

At this point we've pinned necessary transitive dependencies and are looking at more systemic fixes to prevent these sorts of things in the future, but it would still be nice to know if it's possible to have future backwards incompatible changes correspond with major version bumps.

mullermp:
@numberten I'm sorry for any troubles this change may have caused. We tried to make this as painless as possible. This change was made in line with the [SDK maintenance policy](https://docs.aws.amazon.com/sdkref/latest/guide/maint-policy.html#dep-life-cycle) which states that AWS reserves the right to stop support for underlying dependencies (including Language runtimes) without increasing the major version. I believe this change does not technically break semantic versioning (no public interface changes) though I admit it is a bit of a grey area when it comes to dependencies. We took precautions (blog post, usage metrics) and relied on tooling (bundler gem version selection) to make this decision. Dropping old versions allowed for us to remove old, pinned dependencies, remove old automation around testing and releasing, and remove various hacks throughout the SDK, so that we can focus on new development efforts.

michaelwittig:
@mullermp I tried to install ruby 2.6. We install the ruby gems using [cfn-init](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-init.html) (preinstalled on Amazon Linux 1 & 2). It doesn't work with ruby 2.6. I get the following error:Gem failed. Output: ERROR:  While executing gem ... (OptionParser::InvalidOption)
    invalid option: --no-ri
There must be a reason why AL2 uses ruby 2.0? 

Maybe it's time to acknowledge that your change is indeed a breaking change and you better bump the major version of the lib?

PR #2584, opened 2021-09-13
chore: add command to generate API reference

trivikr:
### Description

Added rake script to test documentation.

### Testing

Verified that the documentation is created in `api-docs` folder on running `bundle exec rake docs`

### DIFF
--- a/tasks/docs.rake
+++ b/tasks/docs.rake
@@ -1,0 +1,8 @@
+desc 'Generate the API documentation.'
+task(:docs) do
+  env = {}
+  ENV['DOCSTRINGS'] = '1'
+  ENV['SITEMAP_BASEURL'] = 'http://docs.aws.amazon.com/sdk-for-ruby/v3/api/'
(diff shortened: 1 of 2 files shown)


tasks/docs.rake
```diff
@@ -0,0 +1,8 @@
+desc 'Generate the API documentation.'
+task(:docs) do
+  env = {}
+  ENV['DOCSTRINGS'] = '1'
+  ENV['SITEMAP_BASEURL'] = 'http://docs.aws.amazon.com/sdk-for-ruby/v3/api/'
```
alextwoods, line 5: I think we want to avoid putting the internals of the docs into the SDK repo.  
We do provide docs tasks for some of our other (non-sdk) libraries, but expect those ENV vars to be set.  Eg: [aws-sdk-ruby-release-tools: docs.rake](https://github.com/aws/aws-sdk-ruby-release-tools/blob/6475953e6c531cc2c494a7dc9987a1feff199414/docs.rake)

PR #2602, opened 2021-10-27
Use credential scope for signing region/name

alextwoods:


gems/aws-partitions/spec/test_partition.json
```diff
@@ -9,7 +9,7 @@
       "dnsSuffix" : "amazonaws.com",
       "partition" : "test",
       "partitionName" : "AWS SDK For Ruby Test Partition",
-      "regionRegex" : "^(us|eu|ap|sa|ca|me)\\-\\w+\\-\\d+$",
+      "regionRegex" : "^(us|eu|ap|sa|ca|me|peccy)\\-\\w+\\-\\d+$",
```
mullermp, line 12: shouldn't need to change this right? `us-peccy` starts with `us`

mullermp (approved):
Seems fine to me

gems/aws-partitions/spec/test_partition.json
alextwoods: us-peccy does, but it matches the `aws` partition first making testing some of these cases difficult, so I added this so we can use regions like `peccy-west-1`.

PR #2610, opened 2021-11-04
Raise when accelerate and fips are provided in s3

mullermp:
There are no fips/accelerate endpoints, and this might break compliance if non-fips endpoint is used. We must raise an error (hopefully too breaking)


gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/accelerate.rb
```diff
@@ -46,6 +46,12 @@ def call(context)
               raise ArgumentError,
                     'Cannot use both :use_accelerate_endpoint and :endpoint'
             end
+            # Raise if :endpoint and accelerate are both provided
```
alextwoods: comment doesnt match

alextwoods (approved):

gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/accelerate.rb
mullermp: copy paste....

Issue #2547, opened 2021-07-01
Expose `Aws::S3::Presigner#presigned_request` to `Aws::S3::Object`


josegomezr:
**Is your feature request related to a problem? Please describe.**
Following #2511, in our use case, `Aws::S3::Object#presigned_url` worked, but it's still a bit short, although I can now pre-sign any S3 Method, I also require the headers (For example, when using SSE: `x-amz-server-side-encryption`, `x-amz-server-side-encryption-aws-kms-key-id`, etc.).

When I created my resources (bucket, policies & extra) following [this guide](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingServerSideEncryption.html) I constantly got a 403 error when enforcing the `DenyIncorrectEncryptionHeader` & `DenyUnencryptedObjectUploads` statements that the guide provides.

Turns out that according to my experimentation, the policy checks only for headers present, although the parameters are passed by query-string they're ignored by the verification on the bucket side and hence renders a 403.

**Describe the solution you'd like**
Can the `S3::Object` class expose `Aws::S3::Presigner#presigned_request` so I can get the presigned_url & the headers?

**Describe alternatives you've considered**
As @mullermp [pointed out here](https://github.com/aws/aws-sdk-ruby/issues/2511#issuecomment-827005466) 
>  The Resource models contain many issues and aren't really maintained much

I could just use the presigner directly, although I have a working PR for exposing the `Aws::S3::Presigner#presigned_request` to `Aws::S3::Object` that makes it look a bit more elegant IMO 😅. I'm willing to contribute & maintain this pieces of code if needed.

- [x] :wave: I may be able to implement this feature request.

**Additional context**

PR #2613, opened 2021-11-16
Add presigned request to S3 object

mullermp:
Adds #2547

### DIFF
--- a/gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
+++ b/gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
@@ -161,7 +161,7 @@ module Aws
       #
       # @param [Symbol] method
       #   The S3 operation to generate a presigned URL for. Valid values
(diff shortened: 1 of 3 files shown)


gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
```diff
@@ -215,6 +215,79 @@ def presigned_url(method, params = {})
...
+      # @raise [ArgumentError] Raised if `:expires_in` exceeds one week
+      #   (604800 seconds).
+      #
+      # @return [String, Hash] A tuple with a presigned URL and headers that
+      #   should be included with the request.
+      #
+      def presigned_request(method, params = {})
+        presigner = Presigner.new(client: client)
+
+        if %w(delete head get put).include?(method.to_s)
```
alextwoods, line 281: Why are we translating these instead of just requiring the method name?  Do we do that elsewhere?

mullermp: Yes, in presigned_url

alextwoods (approved):

PR #2616, opened 2021-11-24
Support list of integers in rest query params

mullermp:
Support list of integers in rest query params

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core/rest/request/querystring_builder.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/rest/request/querystring_builder.rb
@@ -49,8 +49,9 @@ module Aws
               raise NotImplementedError, msg
             end
           when ListShape
-            if StringShape === shape_ref.shape.member.shape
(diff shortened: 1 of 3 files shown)


gems/aws-sdk-core/lib/aws-sdk-core/rest/request/querystring_builder.rb
```diff
@@ -49,8 +49,9 @@ def build_part(shape_ref, param_value)
               raise NotImplementedError, msg
             end
           when ListShape
-            if StringShape === shape_ref.shape.member.shape
-              list_of_strings(shape_ref.location_name, param_value)
+            if StringShape === shape_ref.shape.member.shape ||
+               IntegerShape === shape_ref.shape.member.shape
```
alextwoods, line 53: Should we support all of the primitive shapes here as Smithy does (eg, boolean, float, ect).

mullermp: Probably, depends on what the ask is (awaiting more info)

PR #2627, opened 2022-01-02
Parse a response with consecutive spaces correctly when ox is used as the XML parser

abicky:
Ox.sax_parse collapses consecutive white spaces into a single space if the option "skip" is `:skip_white`. As a result, responses are parsed incorrectly.
For example, assume that you have an S3 object whose key is "foo&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;bar\n", which includes five consecutive spaces and a LF. The response of the following code includes the object with the key "foo bar", not "foo&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;bar\n":

```ruby
Aws::S3::Client.new.list_objects_v2(bucket: bucket)
```

This PR fixes the issue.

Here is reproducible code:

```ruby
require 'ox'
require 'aws-sdk-s3'

BUCKET = ENV.fetch('BUCKET') do
  raise 'Environment variable "BUCKET" is required'
end

client = Aws::S3::Client.new
key = "foo     bar\n"
client.put_object(bucket: BUCKET, key: key)
client.list_objects_v2(bucket: BUCKET, prefix: key).contents.each do |o|
  puts "key:"
  p o.key
end
client.delete_object(bucket: BUCKET, key: key)
```

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core/xml/parser/engines/ox.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/xml/parser/engines/ox.rb
@@ -16,7 +16,7 @@ module Aws
           Ox.sax_parse(
             @stack, StringIO.new(xml),
             :convert_special => true,
-            :skip => :skip_white
+            :skip => :skip_return
           )
         end
 
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Parse a response with consecutive spaces correctly when ox is used as the XML parser.
+
 3.125.0 (2021-12-21)
 ------------------
 
--- a/gems/aws-sdk-core/spec/api_helper.rb
+++ b/gems/aws-sdk-core/spec/api_helper.rb
@@ -67,6 +67,8 @@ module ApiHelper
             'Integer' => { 'shape' => 'IntegerShape' },
             'Long' => { 'shape' => 'LongShape' },
             'String' => { 'shape' => 'StringShape' },
+            'StringWithConsecutiveSpaces' => { 'shape' => 'StringShape' },
+            'StringWithLF' => { 'shape' => 'StringShape' },
             'Timestamp' => { 'shape' => 'TimestampShape' },
             'EventStream' => { 'shape' => 'EventStream' },
             'DocumentType' => { 'shape' => 'DocumentShape' }
--- a/gems/aws-sdk-core/spec/aws/stubbing/empty_stub_spec.rb
+++ b/gems/aws-sdk-core/spec/aws/stubbing/empty_stub_spec.rb
@@ -25,6 +25,8 @@ module Aws
           integer: 0,
           long: 0,
           string: "StringShape",
+          string_with_consecutive_spaces: "StringShape",
+          string_with_lf: "StringShape",
           sensitive_string: 'SensitiveStringShape',
           timestamp: now,
         })
--- a/gems/aws-sdk-core/spec/aws/xml/parser_spec.rb
+++ b/gems/aws-sdk-core/spec/aws/xml/parser_spec.rb
@@ -9,10 +9,15 @@ module Aws
       [:ox, :oga, :nokogiri, :libxml, :rexml].each do |engine|
         describe("ENGINE: #{engine}") do
 
-          begin
-            Parser.engine = engine
-          rescue LoadError
-            next
+          around do |example|
+            previous_engine = Parser.engine
+            begin
+              Parser.engine = engine
+              example.run
+            rescue LoadError
+            ensure
+              Parser.engine = previous_engine
+            end
           end
 
           let(:shapes) { ApiHelper.sample_shapes }
--- a/gems/aws-sdk-core/spec/aws/xml/parser_spec.rb
+++ b/gems/aws-sdk-core/spec/aws/xml/parser_spec.rb
@@ -110,6 +110,8 @@ module Aws
               <Integer>123</Integer>
               <Long>321</Long>
               <String>Hello</String>
+              <StringWithConsecutiveSpaces>foo  bar</StringWithConsecutiveSpaces>
+              <StringWithLF>foo\nbar</StringWithLF>
               <Timestamp>123456789</Timestamp>
             </xml>
             XML
@@ -145,6 +147,8 @@ module Aws
               long: 321,
               string: "Hello",
               timestamp: Time.at(123456789),
+              string_with_consecutive_spaces: "foo  bar",
+              string_with_lf: "foo\nbar",
             })
           end
 


gems/aws-sdk-core/lib/aws-sdk-core/xml/parser/engines/ox.rb
```diff
@@ -16,7 +16,7 @@ def parse(xml)
           Ox.sax_parse(
             @stack, StringIO.new(xml),
             :convert_special => true,
-            :skip => :skip_white
+            :skip => :skip_return
```
abicky, line 19: Here is example code:

```ruby
require 'ox'

class Sample < ::Ox::Sax
  def text(value); p value; end
end

xml = <<~XML
<?xml version="1.0" encoding="UTF-8"?><list>
  <value>
 with linebreak
</value>
  <value>with  consecutive spaces  </value>
</list>

XML

handler = Sample.new

puts '== skip_white =='
Ox.sax_parse(handler, StringIO.new(xml), { skip: :skip_white })
puts '== skip_return =='
Ox.sax_parse(handler, StringIO.new(xml), { skip: :skip_return })
puts '== skip_off =='
Ox.sax_parse(handler, StringIO.new(xml), { skip: :skip_off })
puts '== skip_none =='
Ox.sax_parse(handler, StringIO.new(xml), { skip: :skip_none })
```

The output like below:

```
== skip_white ==
" with linebreak "
"with consecutive spaces "
== skip_return ==
"\n with linebreak\n"
"with  consecutive spaces  "
== skip_off ==
"\n  "
"\n with linebreak\n"
"\n  "
"with  consecutive spaces  "
"\n"
== skip_none ==
"\n  "
"\n with linebreak\n"
"\n  "
"with  consecutive spaces  "
```

As you can see above, `:skip_return` works as expected.

mullermp:
This seems safe at a glance. But what about keys/objects with newlines in them?

mullermp:
I'm not certain, but we may need `skip_none` because `skip_white` looks to be removing line breaks and `skip_return` doesn't, which might be backwards incompatible?

alextwoods:
Yeah - I share the concern that `skip_return` changes the behavior of line breaks and could potentially be breaking - but `skip_none` has the same behavior (ie, it leaves the `\n` on the "with linebreaks" values).

abicky:
@mullermp @alextwoods Thank you for your quick response!
As you can see in the comment https://github.com/aws/aws-sdk-ruby/pull/2627#discussion_r777185838, `skip_return` never removes `\n`. It just removes `\r` from `\r\n`.
I've changed the reproducible code a little to make the key include `\n` and added a test for it in the commit https://github.com/aws/aws-sdk-ruby/pull/2627/commits/ccbd26a537fa67846a24ef5d9e0625339272ee57.

As for `\r\n` or `\r`, XML parsers should convert it to `\n` as described on https://www.w3.org/TR/2008/REC-xml-20081126/#sec-line-ends (neither ox nor oga does so, though), that is, keeping `\r` and `\r\n` requires changes in both of the API response and some XML libraries.

Actually, if I change the tests as follow, the tests for ox and oga fail:

```diff
diff --git a/gems/aws-sdk-core/spec/api_helper.rb b/gems/aws-sdk-core/spec/api_helper.rb
index abba409ee..ea0cc793e 100644
--- a/gems/aws-sdk-core/spec/api_helper.rb
+++ b/gems/aws-sdk-core/spec/api_helper.rb
@@ -68,6 +68,8 @@ module ApiHelper
             'Long' => { 'shape' => 'LongShape' },
             'String' => { 'shape' => 'StringShape' },
             'StringWithConsecutiveSpaces' => { 'shape' => 'StringShape' },
+            'StringWithCRLF' => { 'shape' => 'StringShape' },
+            'StringWithSingleCR' => { 'shape' => 'StringShape' },
             'Timestamp' => { 'shape' => 'TimestampShape' },
             'EventStream' => { 'shape' => 'EventStream' },
             'DocumentType' => { 'shape' => 'DocumentShape' }
diff --git a/gems/aws-sdk-core/spec/aws/xml/parser_spec.rb b/gems/aws-sdk-core/spec/aws/xml/parser_spec.rb
index c121633dd..aa7faa7fe 100644
--- a/gems/aws-sdk-core/spec/aws/xml/parser_spec.rb
+++ b/gems/aws-sdk-core/spec/aws/xml/parser_spec.rb
@@ -116,6 +116,8 @@ module Aws
               <Long>321</Long>
               <String>Hello</String>
               <StringWithConsecutiveSpaces>foo  bar</StringWithConsecutiveSpaces>
+              <StringWithCRLF>foo\r\nbar</StringWithCRLF>
+              <StringWithSingleCR>foo\rbar</StringWithSingleCR>
               <Timestamp>123456789</Timestamp>
             </xml>
             XML
@@ -152,6 +154,8 @@ module Aws
               string: "Hello",
               timestamp: Time.at(123456789),
               string_with_consecutive_spaces: "foo  bar",
+              string_with_crlf: "foo\nbar",
+              string_with_single_cr: "foo\nbar",
             })
           end

```

And if I change the tests as follow, all the tests except for those for ox fail:

```diff
diff --git a/gems/aws-sdk-core/spec/api_helper.rb b/gems/aws-sdk-core/spec/api_helper.rb
index abba409ee..ea0cc793e 100644
--- a/gems/aws-sdk-core/spec/api_helper.rb
+++ b/gems/aws-sdk-core/spec/api_helper.rb
@@ -68,6 +68,8 @@ module ApiHelper
             'Long' => { 'shape' => 'LongShape' },
             'String' => { 'shape' => 'StringShape' },
             'StringWithConsecutiveSpaces' => { 'shape' => 'StringShape' },
+            'StringWithCRLF' => { 'shape' => 'StringShape' },
+            'StringWithSingleCR' => { 'shape' => 'StringShape' },
             'Timestamp' => { 'shape' => 'TimestampShape' },
             'EventStream' => { 'shape' => 'EventStream' },
             'DocumentType' => { 'shape' => 'DocumentShape' }
diff --git a/gems/aws-sdk-core/spec/aws/xml/parser_spec.rb b/gems/aws-sdk-core/spec/aws/xml/parser_spec.rb
index c121633dd..93aa0ffdc 100644
--- a/gems/aws-sdk-core/spec/aws/xml/parser_spec.rb
+++ b/gems/aws-sdk-core/spec/aws/xml/parser_spec.rb
@@ -116,6 +116,8 @@ module Aws
               <Long>321</Long>
               <String>Hello</String>
               <StringWithConsecutiveSpaces>foo  bar</StringWithConsecutiveSpaces>
+              <StringWithCRLF>foo\r\nbar</StringWithCRLF>
+              <StringWithSingleCR>foo\rbar</StringWithSingleCR>
               <Timestamp>123456789</Timestamp>
             </xml>
             XML
@@ -152,6 +154,10 @@ module Aws
               string: "Hello",
               timestamp: Time.at(123456789),
               string_with_consecutive_spaces: "foo  bar",
+              string_with_crlf: "foo\nbar",
+              # Note that the expected result doesn't comply with the specification
+              # https://www.w3.org/TR/REC-xml/#sec-line-ends.
+              string_with_single_cr: "foo\rbar",
             })
           end

```

So I want to focus on consecutive spaces in this PR.

abicky:
I've created the issue https://github.com/ohler55/ox/issues/282 on how ox processes a carriage return.
As for oga, I can't create any issue. I think this is because the repository is no longer active.

alextwoods:
To address @mullermp  and I's concerns with the change in behavior in the `" with linebreak "` test, I checked the behavior in our libxml engine:
```
require 'libxml'

class LibxmlParser
  include LibXML::XML::SaxParser::Callbacks
  def on_characters(chars)
    p chars
  end

end

xml = <<~XML
<?xml version="1.0" encoding="UTF-8"?><list>
  <value>
 with linebreak
</value>
  <value>with  consecutive spaces  </value>
</list>

XML

handler = LibxmlParser.new

parser = ::LibXML::XML::SaxParser.string(xml)
parser.callbacks = handler
parser.parse
```

This prints:
```
"\n  "
"\n with linebreak\n"
"\n  "
"with  consecutive spaces  "
"\n"
```

So - critically, it yields `"\n with linebreak\n"` which is the same behavior as `skip_return` in ox - so I believe that addresses our concerns there.

mullermp (approved):

alextwoods (approved):

razum2um:
Thanks guys, this also fixes an issue for `Aws::RDS::Client#download_db_log_file_portion` when using `Ox`: previously it collapsed all log lines inside `LogFileData` node into 1 long line

PR #2662, opened 2022-02-14
Raise an error when credentials or config path does not exist

### LINKED ISSUE #2654: Attempting to use `Aws::SharedCredentials` with no `~/.aws/credentials` file fails in an unexpected way
  
**Describe the bug**

If I give a non-existent profile to `Aws::SharedCredentials.new`, I normally get a `Aws::Errors::NoSuchProfileError`. However, if the `~/.aws/credentials` and `~/.aws/config` files don't exist, then instead of an error I get a `Aws::SharedCredentials` instance that returns `nil` for `credentials`.

This causes somewhat cryptic failures later when trying to use the `Aws::SharedCredentials` instance, for example

```
gems/aws-sigv4-1.4.0/lib/aws-sigv4/signer.rb:690:in `credentials_set?': undefined method `access_key_id' for nil:NilClass (NoMethodError)
```

aws-sdk-core 3.126.0

ruby 2.7.5p203 (2021-11-24 revision f69aeb8314) [x86_64-darwin21]

(with existing `~/.aws/credentials` and `~/.aws/config` files)

```console
$ ruby -r aws-sdk-core -e 'p Aws::SharedCredentials.new(profile_name: "wat").credentials'
	4: from -e:1:in `<main>'
	3: from -e:1:in `new'
	2: from /Users/andrew/.gem/ruby/2.7.5/gems/aws-sdk-core-3.126.0/lib/aws-sdk-core/shared_credentials.rb:42:in `initialize'
	1: from /Users/andrew/.gem/ruby/2.7.5/gems/aws-sdk-core-3.126.0/lib/aws-sdk-core/shared_config.rb:103:in `credentials'
/Users/andrew/.gem/ruby/2.7.5/gems/aws-sdk-core-3.126.0/lib/aws-sdk-core/shared_config.rb:372:in `validate_profile_exists': Profile `wat' not found in /Users/andrew/.aws/credentials or /Users/andrew/.aws/config (Aws::Errors::NoSuchProfileError)
```

(without existing `~/.aws/credentials` and `~/.aws/config` files)

```console
$ ruby -r aws-sdk-core -e 'p Aws::SharedCredentials.new(profile_name: "wat").credentials'
nil
```

**Expected behavior**

`Aws::Errors::NoSuchProfileError` is also raised when the profile does not exist because the shared config files do not exist.

mullermp:
Fixes #2654

It does change behavior in the shared config class, but I don't believe this to be a breaking change for legitimate use cases.


gems/aws-sdk-core/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Raise a `NoSuchProfileError` when config or credentials files don't exist.
```
alextwoods: nit: or -> and

alextwoods (approved):

ojab:
A bit breaking, we have something like
```
aws_credentials = Aws::SharedCredentials.new(profile_name: 'company_name').credentials # for local development
aws_credentials ||= Aws::Credentials.new(Settings.aws.access_key, Settings.aws.secret_key).credentials # for non-local
```
in the codebase and now it fails in CI.

mullermp:
Ah! Sorry about that.. Do you have a path forward or need assistance?

ojab:
Didn't looked into that yet, but I guess we'll fix it somehow.
```ruby
credentials = begin
  profile: 
rescue 
  accces_key, secret_key
end
```
should work anyway.

PR #2663, opened 2022-02-14
Add a before_refresh callback to AssumeRoleCredentials

### LINKED ISSUE #2529: Mechanism/Hook to Provide MFA Token Code For AssumeRole Authentication
Today, providing a `token_code` for assuming a role with an MFA requirement has to be done manually during the client creation. It would be great if a mechanism could be added to the Aws config to support a way to dynamically fetch the token. Perhaps the ability to provide a lambda that gets called?

This would allow users to either invoke a library, executable, or even prompt for STDIN in order to retrieve the appropriate token code.

alextwoods:
Fixes: #2529 

This provides a `before_refresh` hook that will get called before new credentials are fetched allowing users to refresh the token used (or update any other parameters as required).  Applications likely have a wide range of strategies for updating MFA tokens, so rather than attempt to provide default implementations this just provides a hook for application code to update the token when its needed.

Note - the TokenProvider interface/implementation as suggested in #2529 could be confused with [Bearer Token Auth](https://awslabs.github.io/smithy/1.0/spec/core/auth-traits.html#httpbearerauth-trait) so I've gone with a generic callback hook instead.

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core/assume_role_credentials.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/assume_role_credentials.rb
@@ -17,6 +17,11 @@ module Aws
   #
   # If you omit `:client` option, a new {STS::Client} object will be
   # constructed.
+  #
+  # The AssumeRoleCredentials also provides a `before_refresh` callback
+  # that can be used to help manage refreshing tokens.
+  # `before_refresh` is called when AWS credentials are required and need
+  # to be refreshed and it is called with the AssumeRoleCredentials object.
   class AssumeRoleCredentials
 
     include CredentialProvider
@@ -49,8 +54,6 @@ module Aws
         end
       end
       @client = client_opts[:client] || STS::Client.new(client_opts)
-      @before_refresh = options.delete(:before_refresh)
-
       super
     end
 
@@ -63,9 +66,6 @@ module Aws
     private
 
     def refresh
-
-      @before_refresh.call(self) if @before_refresh
-
       c = @client.assume_role(@assume_role_params).credentials
       @credentials = Credentials.new(
         c.access_key_id,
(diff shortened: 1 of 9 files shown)


gems/aws-sdk-core/lib/aws-sdk-core/assume_role_credentials.rb
```diff
@@ -28,6 +28,16 @@ class AssumeRoleCredentials
     # @option options [Integer] :duration_seconds
     # @option options [String] :external_id
     # @option options [STS::Client] :client
+    # @option options [Callable] before_refresh Proc called before
```
mullermp, line 36: Does it make sense to push this to refreshable credentials and have all refreshable credentials support this?

alextwoods: Yeah - I was trying to think through that.  I think the key part is the documentation, which likely should be per class.... and to be useful, each credential provider may need to provide other accessors.  But I think the actual call to before_refresh could be done in the RefreshingCredentials `refresh!` method.

gems/aws-sdk-core/lib/aws-sdk-core/assume_role_credentials.rb
```diff
@@ -17,6 +17,11 @@ module Aws
   #
   # If you omit `:client` option, a new {STS::Client} object will be
   # constructed.
+  #
+  # The AssumeRoleCredentials also provides a `before_refresh` callback
```
mullermp, line 21: Misleading doc

alextwoods: What part is misleading?

mullermp: Should this be that RefreshingCredentials provides the callback? Will each credential provider have a comment like this?

alextwoods: The use of RefreshingCredentials is, to me, a behind the scenes implementation detail.  Someone reading the documentation here shouldn't care how its done.

I dont think each refreshing credential provider should have a comment until we know what use cases they may want to solve with them - the documentation is useful here I think because it gives an idea what you would actually use this for.

mullermp (approved):

PR #2670, opened 2022-02-28
Refactor PageableResponse to avoid busting Ruby's constant cache

casperisfine:
Ruby has a global constant cache, which is basically a counter
that is incremented anytime a constant is set or modified.

Whenever it is bumped, the vast majority of the Ruby VM inline caches
are invalidated, as well as JITed code. So busting this cache
at "runtime" is very detrimental to performance.

`Object#extend` doesn't normally bust the cache, except if the
module contains constants of it's own:

```ruby
module A; end
obj.extend(A) # The global cache is preserved

module B
  FOO = 1
end
obj.extend(B) # The global cache is invalidated
```

In this case, the constant causing the invalidation is
`Aws::PageableResponse::LastPageError`.

So to avoid regularly flushing the cache, this PR move all
pagination methods to a submodule, and include this one.

### DIFF
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Fixed `Aws::PageableResponse` invalidating Ruby's global constant cache.
+
 3.128.0 (2022-03-04)
 ------------------
 
--- a/gems/aws-sdk-core/lib/aws-sdk-core/pageable_response.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/pageable_response.rb
@@ -48,11 +48,11 @@ module Aws
   #
   module PageableResponse
 
-    def self.extended(base)
-      base.extend Enumerable
-      base.extend UnsafeEnumerableMethods
-      base.instance_variable_set("@last_page", nil)
-      base.instance_variable_set("@more_results", nil)
+    def self.apply(base)
+      base.extend Extension
+      base.instance_variable_set(:@last_page, nil)
+      base.instance_variable_set(:@more_results, nil)
+      base
     end
 
     # @return [Paging::Pager]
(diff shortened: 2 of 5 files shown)


casperisfine:
The test somehow fail on CI, I'm not sure why, as it passes locally when I run the `aws-core` suite independently.

casperisfine:
Ok, so the reason why it was failing on CI, it because https://github.com/aws/aws-sdk-ruby/pull/1739  defined `Enumerable::Enumerator` in the top of another spec file.

I don't think it's necessary to test this though.

mullermp:
Thanks for opening a pull request for this. I can review this today/tomorrow. I think there is also an open issue that can be resolved from this.

mullermp (approved):
Needs @alextwoods stamp of approval

casperisfine:
Thanks!

On a sidenote, [DynamicErrors](https://github.com/aws/aws-sdk-ruby/blob/1409d5c2cc1e499963c296e711d3e79cc6e1912c/gems/aws-sdk-core/lib/aws-sdk-core/errors.rb#L324) pretty much expose the same problem, except it's bounded as it only flush the cache on the first time an error is referenced.

I haven't dug into it yet, but there may be a way to eager load the most common errors at least? Is there a way all possible errors could be listed on boot?

mullermp:
Usually the service will model their errors and list which errors are returned by each operation, but that list is never complete, and doesn't handle generic cases which dynamic errors handles.

casperisfine:
> but that list is never complete, and doesn't handle generic cases which dynamic errors handles.

I don't think it would be a big deal, as long as the most common ones are eager generated, it would be a win IMHO.

alextwoods (approved):
Looks good to me!

casperisfine:
I rebased to fix the merge conflict. Let me know if there's anything else I can do to get this merged.

mullermp:
It looks good, thanks for the contribution! In a new major version, pagination will be handled via abstraction classes and not by response objects with state, and should avoid this problem too.

PR #2679, opened 2022-03-09
Make stubs thread safe by adding a mutex per stubbed request

### LINKED ISSUE #2675: Stubbed responses are subject to race conditions when parallel requests use the same stub.
  
**Describe the bug**
When stubbing responses, the S3 SDK can sometimes hit a race condition when multiple threads use the same stub at once. Multi-part uploads are an example of where multiple threads might be using the same stubs, as each part of the multi-part upload is uploaded in parallel.

When this happens, you receive the error `Aws::S3::MultipartUploadError: multipart upload failed: Extra content at the end of the document`, due to the same response body being returned multiple times.

i.e. The client receives

```
<UploadPartResult xmlns="">
</UploadPartResult>
<UploadPartResult xmlns="">
</UploadPartResult>
```

when it should have just received a single `UploadPartResult` in the response body.

I believe this is likely due to this code not being thread-safe.

https://github.com/aws/aws-sdk-ruby/blob/a82c8981c95a8296ffb6269c3c06a4f551d87f7d/gems/aws-sdk-core/lib/aws-sdk-core/plugins/stub_responses.rb#L90-L93

If one thread is currently in the while loop, and another thread simultaneously calls `stub.body.rewind` then the while loop in the first thread will run another time even though it's already reached the end of the content.

`aws-sdk-core` `3.112.0`

```
ruby 3.1.0p13 (2022-02-02 revision 42c9ef769f) [x86_64-linux]
last_commit=merge revision(s) 7ff1bf317887c0d7b21e91ad548d07b9f05c540c,e89d80702bd98a8276243a7fcaa2a158b3bfb659: [Backport #18516]
```

The easiest way to reproduce this is with the following code:

```ruby
#!/usr/bin/env ruby
# frozen_string_literal: true

require "aws-sdk-s3"

s3_client = Aws::S3::Client.new(stub_responses: true)
s3_client.stub_responses(:create_multipart_upload, upload_id: "id")
s3_client.stub_responses(:upload_part, etag: "etag")
s3_resource = Aws::S3::Resource.new(client: s3_client)
s3_bucket = s3_resource.bucket("my-bucket")
s3_object = s3_bucket.object("my-key")

while true
	result = s3_object.upload_file(
		"./random-data.bin",
		multipart_threshold: 1024 * 1024,
	)
end
```

Where `./random-data.bin` is a large file, for example generated with `head -c 15M </dev/urandom >random-data.bin`.

Even with this tight loop, on some machines it can take a long time for the problem to manifest, so to make it more reliably reproducible you can locally modify the while loop in `stub_responses.rb` (linked above) to include a sleep after each chunk. After making this change the issue is very reliably reproducible.

**Expected behavior**
The stubbed responses returned to the SDK should always be valid and not duplicated.

alextwoods:
Fixes #2675. 

Previously we were creating 1 Response object for each stub added (at the time it was added to the client), each operation call would then reference the same response object, with the same body object, causing race conditions.  This change moves the conversion of stub data into response objects to operation call time - thus ensuring that each call gets its own response.  

Todo: Still need to evaluate other possible impacts/backwards compatibility issues.

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb
@@ -301,9 +301,7 @@ module Aws
 
     def validate_stub(operation_name, data)
       if Hash === data && data.keys.sort != [:body, :headers, :status_code]
-        api = config.api
-        operation = api.operation(operation_name)
-        ParamValidator.new(operation.output, input: false).validate!(data)
+        data_to_http_resp(operation_name, data)
       end
     end
 
--- a/gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb
@@ -235,8 +235,8 @@ module Aws
         stubs = @stubs[operation_name] || []
         case stubs.length
         when 0 then default_stub(operation_name)
-        when 1 then convert_stub(operation_name, stubs.first)
-        else convert_stub(operation_name, stubs.shift)
+        when 1 then stubs.first
+        else stubs.shift
         end
       end
       Proc === stub ? convert_stub(operation_name, stub.call(context)) : stub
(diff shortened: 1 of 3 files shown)


gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb
```diff
@@ -262,13 +262,17 @@ def apply_stubs(operation_name, stubs)
...
+      stub = case stub
       when Proc then stub
       when Exception, Class then { error: stub }
       when String then service_error_stub(stub)
       when Hash then http_response_stub(operation_name, stub)
       else { data: stub }
       end
+      if Hash === stub
+        stub[:mutex] = Mutex.new
+      end
```
mullermp, line 274: Why does this need to be separate from the Hash case?

alextwoods: Other cases (everything except the Proc) produce a hash, which we then need to add the mutex to.
With a proc, it gets called at operation time to produce a new hash, which doesn't need the mutex.

mullermp (approved):

PR #2688, opened 2022-04-15
Add CRT Signers to aws-sigv4

alextwoods:

### DIFF
--- a/gems/aws-sdk-core/spec/aws/plugins/signature_v4_spec.rb
+++ b/gems/aws-sdk-core/spec/aws/plugins/signature_v4_spec.rb
@@ -224,7 +224,7 @@ module Aws
         }
 
         let(:datetime) { '20120101T10:11:12Z' }
-        let(:now) { Time.now }
+        let(:now) { Time.parse(datetime) }
         let(:utc) { now.utc }
 
         before(:each) {
@@ -234,17 +234,16 @@ module Aws
         }
 
         it "unsigns payload for operations has 'v4-unsigned-payload' for 'authtype'" do
+          allow(Time).to receive(:now).and_return(now)
           resp = client.streaming_foo(foo_name: 'foo')
           req = resp.context.http_request
           expect(req.headers['x-amz-content-sha256']).to eq('UNSIGNED-PAYLOAD')
-          expect(req.headers['authorization']).to eq('AWS4-HMAC-SHA256 Credential=stubbed-akid/20120101/region/svc-name/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=6d86306ea8dcf03db133cd35bcf626b0a440daefadc8ddde3304517126edb1bb')
         end
 
         it "signs payload for operations without 'v4-unsigned-payload' for 'authtype'" do
           resp = client.non_streaming_bar(bar_name: 'bar')
           req = resp.context.http_request
           expect(req.headers['x-amz-content-sha256']).not_to eq('UNSIGNED-PAYLOAD')
-          expect(req.headers['authorization']).to eq('AWS4-HMAC-SHA256 Credential=stubbed-akid/20120101/region/svc-name/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=c6394995838d4b4a1ec9b19229d92bac3f11441308f953d89306663230e95713')
         end
 
         it "signs payload for HTTP request even when 'v4-unsigned-payload' is set" do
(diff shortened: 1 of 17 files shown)


gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/s3_signer.rb
```diff
@@ -175,7 +175,7 @@ def custom_endpoint?(resp)
           end
 
           def wrong_sigv4_region?(resp)
-            resp.context.http_response.status_code == 400 &&
+            [400, 301].include?(resp.context.http_response.status_code) &&
```
mullermp: Is this intended for this change? (I typically agree we should be doing this... but maybe not as part of this)

gems/aws-sdk-core/spec/aws/plugins/signature_v4_spec.rb
```diff
@@ -234,17 +234,16 @@ module Plugins
         }
 
         it "unsigns payload for operations has 'v4-unsigned-payload' for 'authtype'" do
+          allow(Time).to receive(:now).and_return(now)
           resp = client.streaming_foo(foo_name: 'foo')
           req = resp.context.http_request
           expect(req.headers['x-amz-content-sha256']).to eq('UNSIGNED-PAYLOAD')
-          expect(req.headers['authorization']).to eq('AWS4-HMAC-SHA256 Credential=stubbed-akid/20120101/region/svc-name/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=6d86306ea8dcf03db133cd35bcf626b0a440daefadc8ddde3304517126edb1bb')
```
mullermp, line 240: These tests are assuming the payload is signed, so don't we want to check for signatures?

gems/aws-sdk-s3/spec/object/presigned_url_spec.rb
```diff
@@ -159,14 +159,14 @@ module S3
           allow(Time).to receive(:now).and_return(now)
           url, headers = obj.presigned_request(
             :get, expires_in: 86_400, request_payer: 'peccy')
-          expect(url).to eq(
+          expect(CGI.parse(url)).to eq(CGI.parse(
```
mullermp, line 162: Why do we need CGI parsing now?

gems/aws-sdk-s3/spec/presigner_spec.rb
```diff
@@ -75,19 +75,22 @@ module S3
             bucket: 'examplebucket',
             key: 'test.txt',
             expires_in: 86_400,
-            time: Time.utc(1969, 4, 20)
```
mullermp, line 78: !

gems/aws-sigv4/lib/aws-sigv4/signer.rb
```diff
@@ -693,8 +712,135 @@ def credentials_set?(credentials)
           !credentials.secret_access_key.empty?
       end
 
+      ### CRT Code
+
+      # the credentials used by CRT must be a
+      # CRT StaticCredentialsProvider object
+      def fetch_credentials_crt
```
mullermp: I think crt_ should be a prefix and not a suffix (at least when inspecting methods it's more obvious)

gems/aws-sdk-s3/spec/object/presigned_url_spec.rb
alextwoods: The order of the query parameters may differ.

gems/aws-sdk-core/spec/aws/plugins/signature_v4_spec.rb
alextwoods: The signing is tested extensively in the `aws-sigv4` gem and there are issues with mocking here that are difficult to fix and unrelated to the functionality of unsigned-payload so I dont feel like these tests need to re-test that.

gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/s3_signer.rb
alextwoods: Good catch - removed from this PR.

gems/aws-sdk-s3/spec/presigner_spec.rb
alextwoods: That actually does not work in our CRT FFI bindings (dates pre-1970 will not be represented correctly).

gems/aws-sigv4/lib/aws-sigv4/signer.rb
alextwoods: Sure - I don't have a preference as these are private methods.

mullermp: (even for public ones too)

PR #2694, opened 2022-04-25
Improve credential documentation and expired token error raising

mullermp:
Improve the documentation for credential providers.

With an expired token, the ExpiredToken error is raised instead of a cryptic OpenSSL socket error.


gems/aws-sdk-core/lib/aws-sdk-core/assume_role_credentials.rb
```diff
@@ -3,25 +3,28 @@
...
   #     )
-  #
   #     ec2 = Aws::EC2::Client.new(credentials: role_credentials)
   #
-  # If you omit `:client` option, a new {STS::Client} object will be
-  # constructed.
+  # If you omit `:client` option, a new {Aws::STS::Client} object will be
+  # constructed with additional options that were provided.
+  #
+  # Automatically handles refreshing credentials if an Expiration time is
```
alextwoods: What does if an expiration time is provided in the payload mean?  This to me sounds like something a user needs to do to get this behavior.

gems/aws-sdk-core/lib/aws-sdk-core/assume_role_web_identity_credentials.rb
```diff
@@ -16,12 +15,20 @@ module Aws
...
-  #     @see Aws::STS::Client#assume_role_with_web_identity
+  #     ec2 = Aws::EC2::Client.new(credentials: role_credentials)
+  #
+  # If you omit `:client` option, a new {Aws::STS::Client} object will be
+  # constructed with additional options that were provided.
+  #
+  # Automatically handles refreshing credentials if an Expiration time is
+  # provided in the credentials payload.
   #
+  # This credential provider also provides a `before_refresh` callback
```
alextwoods: I'm not sure we need to document this here - its already documented on the client initialize options.

gems/aws-sdk-core/lib/aws-sdk-core/assume_role_credentials.rb
```diff
@@ -3,25 +3,28 @@
 require 'set'
 
 module Aws
-
-  # An auto-refreshing credential provider that works by assuming
-  # a role via {Aws::STS::Client#assume_role}.
+  # A credential provider that assumes a role via
```
alextwoods: Nit: Why remove the auto-refreshing from all of these descriptions?

gems/aws-sdk-core/lib/aws-sdk-core/assume_role_credentials.rb
mullermp: If AssumeRole returns Expiration in the JSON body, it's used here.

gems/aws-sdk-core/lib/aws-sdk-core/assume_role_web_identity_credentials.rb
mullermp: Assume Role Credentials had it already in the Class documentation (but that was the only instance). I can remove all of them, or keep all of them added here. Which do you prefer?

gems/aws-sdk-core/lib/aws-sdk-core/assume_role_credentials.rb
mullermp: It was already mentioned multiple times (I made it generically below, too)

gems/aws-sdk-core/lib/aws-sdk-core/assume_role_web_identity_credentials.rb
alextwoods: I think it likely makes more sense to remove from class documentation and keep in initialize options.

gems/aws-sdk-core/lib/aws-sdk-core/assume_role_credentials.rb
alextwoods: How do I as a customer know if my credentials will be refreshing or if the expiration will be in the json body of the response?  
I'd lean towards leaving the "Auto refreshing" in the description.  Or for credentials like process credentials where the user's custom process has control over the behavior, then document it like this.

alextwoods (approved):

PR #2727, opened 2022-07-13
Add support for serializing shapes on the body with jsonvalue members.

alextwoods:


gems/aws-sdk-core/lib/aws-sdk-core/param_validator.rb
```diff
@@ -152,7 +152,7 @@ def shape(ref, value, errors, context)
       when MapShape then map(ref, value, errors, context)
       when DocumentShape then document(ref, value, errors, context)
       when StringShape
-        unless value.is_a?(String)
+        unless value.is_a?(String) || ref['jsonvalue']
```
mullermp: I don't think this one is needed. The jsonvalue converter should run before this.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/jsonvalue_converter.rb
```diff
@@ -11,15 +11,31 @@ class Handler < Seahorse::Client::Handler
...
-              param_value = context.params[m]
-              unless param_value.respond_to?(:to_json)
-                raise ArgumentError, "The value of params[#{m}] is not JSON serializable."
+            convert_jsonvalue(m, ref, context.params)
+          end
+          @handler.call(context)
+        end
+
+        def convert_jsonvalue(m, ref, params)
+          return if params.nil? || params[m].nil?
```
mullermp: We want to allow for serializing nil if it's provided explicitly. I think we will want to check params.key?(m) and if present, serialize it. Also can params ever be nil? It should always be at least a hash.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/jsonvalue_converter.rb
```diff
@@ -11,15 +11,31 @@ class Handler < Seahorse::Client::Handler
...
+
+          if ref['jsonvalue']
+            param_value = params[m]
+            unless param_value.respond_to?(:to_json)
+              raise ArgumentError, "The value of params[#{m}] is not JSON serializable."
+            end
+            params[m] = param_value.to_json
+          elsif ref.shape.is_a?(Seahorse::Model::Shapes::StructureShape)
```
mullermp: I think we need to handle other shape types, like Map and Union?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/jsonvalue_converter.rb
alextwoods: The params can be nil here only because its getting called recursively.  But I think I agree that we should use .key? and serialize a nil that is explicitly set.

gems/aws-sdk-core/lib/aws-sdk-core/param_validator.rb
alextwoods: Yup - removed.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/jsonvalue_converter.rb
alextwoods: I'm not 100% positive which shapes we need to support - but I think you're right, I think anything with members that can reference a string shape.  I'll update.

mullermp: Better yet, is there a solution that doesn't involve shape checks here?

alextwoods: The only other option (which is what I started with) was to check `respond_to?(:members)` and `respond_to?(:member)` and `respond_to?(:value)`.  But I think thats worse than just being shape based.

mullermp: Yeah, agreed, then shape is fine, assuming we cover them all.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/jsonvalue_converter.rb
mullermp: Yes. And we need a test for not serializing "null" and doing so when deliberate.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/jsonvalue_converter.rb
```diff
@@ -11,15 +11,43 @@ class Handler < Seahorse::Client::Handler
 
         def call(context)
           context.operation.input.shape.members.each do |m, ref|
-            if ref['jsonvalue']
-              param_value = context.params[m]
-              unless param_value.respond_to?(:to_json)
-                raise ArgumentError, "The value of params[#{m}] is not JSON serializable."
+            convert_jsonvalue(m, ref, context.params, 'params')
```
mullermp, line 14: I think we can just call this method and let the "structure shape" case handle the top level input?

alextwoods: I think so - but will require a refactoring to remove the `m` parameter.  But should be possible.

alextwoods: Edit: This doesn't quite work - because we need to set values back on the params it gets tricky.  The code ends up being a bit cleaner here if we keep the top level iteration of input members.

mullermp (approved):
LGTM apart from formatting / line lengths in specs.

PR #2733, opened 2022-07-26
Fix to_json usage in Rails

### LINKED ISSUE #2732: `to_json` causing `SystemStackError` for SES response in v3
### Describe the bug

`send_raw_email` method on `Aws::SES::Client` object returns a `Aws::SES::Types::SendRawEmailResponse` which should be supporting `to_json` but on **v3** it is causing `SystemStackError: stack level too deep`. 
It was working fine in **v2**.

### Expected Behavior

It should return a valid JSON string.

### Current Behavior

Currently it is causing `SystemStackError: stack level too deep`,

### Reproduction Steps

call `send_raw_email` method on a valid `Aws::SES::Client` object and call `to_json` method for the response.

### Possible Solution

I tried calling the `to_s` method for both v2 & v3 versions.

On V2, `to_s` returns `"#<struct Aws::SES::Types::SendRawEmailResponse message_id=\"01010182363d52db-5bcb1ffd-f7aa-4863-b383-8f984cac1e6a-000000\">"`

On V3, `to_s` returns `"{:message_id=>\"0111118236469049-ec2d2a68-9eb1-411f-9f70-e50ae14de671-1110000\"}"`

This might help in fixing it.

I think this behavior has changed because of https://github.com/aws/aws-sdk-ruby/blame/e969f29e23f4930ff58728c5215a2e22fb2578cd/gems/aws-sdk-core/lib/aws-sdk-core/structure.rb#L51

'aws-sdk-ses'

Any version

mullermp:
Fixes #2732 When Rails is loaded, to_json will infinitely recurse because PageableResponse is enumerable. We instead define our own to_json that applies it to the returned data.

### DIFF
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,10 +1,12 @@
 Unreleased Changes
 ------------------
 
+* Issue - Fix `to_json` usage on pageable responses when using Rails.
+
 3.131.3 (2022-07-18)
 ------------------
 
-* Issue - Add support for serializing shapes on the body with `jsonvalue` members. 
+* Issue - Add support for serializing shapes on the body with `jsonvalue` members.
 
 3.131.2 (2022-06-20)
 ------------------
--- a/gems/aws-sdk-core/lib/aws-sdk-core/pageable_response.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/pageable_response.rb
@@ -146,6 +146,9 @@ module Aws
         data.to_h
       end
 
+      def to_json(options = {})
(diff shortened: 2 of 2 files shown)


gems/aws-sdk-core/CHANGELOG.md
```diff
@@ -1,10 +1,12 @@
 Unreleased Changes
 ------------------
 
+* Issue - Fix `to_json` usage on pageable responses when using Rails.
```
alextwoods, line 4: reference the GH issue here?

alextwoods (approved):

rishabhsairawat (approved):

PR #2737, opened 2022-07-28
Fixes `to_json` usage on pageable responses in custom objects

### LINKED ISSUE #2732: `to_json` causing `SystemStackError` for SES response in v3

rishabhsairawat:
This PR aims to resolve #2732. 

### DIFF
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Fix `to_json` usage on pageable responses in custom objects (#2733).
+
 3.131.4 (2022-07-27)
 ------------------
 
--- a/gems/aws-sdk-core/lib/aws-sdk-core/pageable_response.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/pageable_response.rb
@@ -146,8 +146,8 @@ module Aws
         data.to_h
       end
 
-      def to_json(options = {})
-        to_h.to_json(options)
+      def as_json(options = {})
+        to_h.as_json(options)
       end
     end
 


mullermp (approved):
Thanks for the bug fix! This seems fine.

mullermp:
Actually.. I think we need to revert this. Does this work without rails usage?

alextwoods:
Yeah, I think this will fail as is without rails, since the `as_json` method won't be defined on hash.

rishabhsairawat:
Oops.. Sorry :tired_face:  
Yes, It won't work without rails. I should have put the check for that.

PR #2751, opened 2022-09-08
fix typo "us-eas-1"

mattghali:
Hello, we discovered this typo coincidentally while tracking down the same typo somewhere else. Since it's in the spec, I'm not sure if this is _supposed_ to fail or not, but on the off chance that it isn't; here is a single-character fix.


mullermp:
Thanks. Though it's not really of concern since these tests don't care about the endpoint.

PR #2766, opened 2022-10-13
Support awsQueryCompatible trait: Parse error code when present in HTTP response header

haocheng-aws:
Support awsQueryCompatible trait: Parse error code when present in HTTP response header.

## Motivation

As we decided to parse AwsQuery Compatible Error in the Http response header, the required changes need to take in place on both SDK client and Service.
This will be a pre-requisite for migrating services from AwsQuery wire protocol to AwsJson.

## What changed

Error code can be fetched from response content (for example `__type#fooException`). This PR takes another scenario into consideration, of which when `x-amzn-query-error` header with valid value is found from the Http response.

The parsed error code remains unchanged if this header is missing, or contains null or invalid value.

---


gems/aws-sdk-core/lib/aws-sdk-core/json/error_handler.rb
```diff
@@ -26,7 +26,11 @@ def extract_error(body, context)
       end
 
       def error_code(json, context)
-        code = json['__type']
+        code = if context.config.api.metadata['awsQueryCompatible'] and context.http_response.headers['x-amzn-query-error']
```
mullermp: I think we should wrap this check in a "aws_query_error?(context)" method.

gems/aws-sdk-core/spec/api_helper.rb
```diff
@@ -13,6 +13,14 @@ def sample_json # dynamodb
       end
     end
 
+    def sample_json_aws_query_compatible # sqs
```
mullermp: I think we should remove this - it won't get much re-use. I would prefer if, in error_handler_spec, we define a minimal API. There are some examples (endpoint_discovery_spec or all in that plugins folder).

gems/aws-sdk-core/spec/aws/json/error_handler_spec.rb
```diff
@@ -98,6 +98,45 @@ module Json
...
+          stub_request(:post, "https://sqs.us-west-2.amazonaws.com/").
+            to_return(:status => 400, headers: {
+              'x-amzn-query-error': 'AwsQueryError;Sender'
+            }, :body => error_resp)
+          expect {
+            client_aws_query_compatible.send_message(
+              queue_url: 'https://queue.url',
+              message_body: 'message body'
+            )
+          }.to raise_error {|e|
```
mullermp: For multiline ruby blocks, don't use `{}` but instead use `do end`.

gems/aws-sdk-core/spec/fixtures/apis/sqs.json
```diff
@@ -0,0 +1,1147 @@
+{
```
mullermp: Can be deleted if following above recommendation

gems/aws-sdk-core/spec/api_helper.rb
haocheng-aws: Thanks a lot for advice! Updated.

mullermp (approved):
Nice!

alextwoods (approved):

PR #2785, opened 2022-11-17
CredentialProviderChain: Fallback to client region resolution for AssumeRoleCredentials

pier-oliviert:
The current implementation ignores any data that is retrieve through user configuration schema. By calling Aws.shared_config.region, it allows the call to attempt to retrieve the region from any configuration elsewhere in the system that wasn't provided in the configuration object.

As this is a fallback, this should have no implication on working system that already rely on the config object.

### DIFF
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,5 +1,6 @@
 Unreleased Changes
 ------------------
+* Issue - Use Aws.shared_config when deciding on region in CredentialProviderChain.
 
 3.167.0 (2022-11-09)
 ------------------
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,6 @@
 Unreleased Changes
 ------------------
-* Issue - Allow region resolution in `AssumeRoleCredentials` from `CredentailProviderChain`.
+* Issue - Allow region resolution in `AssumeRoleCredentials` from `CredentialProviderChain`.
 
 3.167.0 (2022-11-09)
 ------------------
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,6 @@
 Unreleased Changes
 ------------------
-* Issue - Use Aws.shared_config when deciding on region in CredentialProviderChain.
+* Issue - Allow region resolution in `AssumeRoleCredentials` from `CredentailProviderChain`.
 
 3.167.0 (2022-11-09)
 ------------------
--- a/gems/aws-sdk-core/lib/aws-sdk-core/credential_provider_chain.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/credential_provider_chain.rb
@@ -141,7 +141,7 @@ module Aws
     end
 
     def assume_role_web_identity_credentials(options)
-      region = (options[:config] && options[:config].region || Aws.shared_config.region)
+      region = options[:config].region if options[:config]
       if (role_arn = ENV['AWS_ROLE_ARN']) && (token_file = ENV['AWS_WEB_IDENTITY_TOKEN_FILE'])
         cfg = {
           role_arn: role_arn,
@@ -169,12 +169,14 @@ module Aws
     end
 
     def assume_role_with_profile(options, profile_name)
-      region = (options[:config] && options[:config].region || Aws.shared_config.region)
-      Aws.shared_config.assume_role_credentials_from_config(
+      assume_opts = {
         profile: profile_name,
-        region: region,
         chain_config: @config
-      )
+      }
+      if options[:config] && options[:config].region
+        assume_opts[:region] = options[:config].region
+      end
+      Aws.shared_config.assume_role_credentials_from_config(assume_opts)
     end
   end
 end
--- a/gems/aws-sdk-core/lib/aws-sdk-core/credential_provider_chain.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/credential_provider_chain.rb
@@ -141,7 +141,7 @@ module Aws
     end
 
     def assume_role_web_identity_credentials(options)
-      region = options[:config].region if options[:config]
+      region = (options[:config] && options[:config].region || Aws.shared_config.region)
       if (role_arn = ENV['AWS_ROLE_ARN']) && (token_file = ENV['AWS_WEB_IDENTITY_TOKEN_FILE'])
         cfg = {
           role_arn: role_arn,
@@ -169,7 +169,7 @@ module Aws
     end
 
     def assume_role_with_profile(options, profile_name)
-      region = (options[:config] && options[:config].region)
+      region = (options[:config] && options[:config].region || Aws.shared_config.region)
       Aws.shared_config.assume_role_credentials_from_config(
         profile: profile_name,
         region: region,
--- a/gems/aws-sdk-core/spec/aws/credential_provider_chain_spec.rb
+++ b/gems/aws-sdk-core/spec/aws/credential_provider_chain_spec.rb
@@ -154,6 +154,17 @@ CREDS
         ENV['AWS_DEFAULT_PROFILE'] = 'BAD_PROFILE'
         validate_credentials(expected_creds)
       end
+
+      describe 'assume role' do
+        it 'uses the shared config region if present' do
+          allow(Aws.shared_config).to receive(:region).and_return('us-east-2')
+          expect(Aws.shared_config).to receive(:region).twice.and_return('us-east-2') #twice because each call below counts for 1
+
+
+          chain.send(:assume_role_with_profile, {}, 'default')
+          chain.send(:assume_role_web_identity_credentials, {})
+        end
+      end
     end
 
     describe 'with multiple sources of credentials' do
--- a/gems/aws-sdk-core/spec/aws/credential_provider_chain_spec.rb
+++ b/gems/aws-sdk-core/spec/aws/credential_provider_chain_spec.rb
@@ -154,17 +154,6 @@ CREDS
         ENV['AWS_DEFAULT_PROFILE'] = 'BAD_PROFILE'
         validate_credentials(expected_creds)
       end
-
-      describe 'assume role' do
-        it 'uses the shared config region if present' do
-          allow(Aws.shared_config).to receive(:region).and_return('us-east-2')
-          expect(Aws.shared_config).to receive(:region).twice.and_return('us-east-2') #twice because each call below counts for 1
-
-
-          chain.send(:assume_role_with_profile, {}, 'default')
-          chain.send(:assume_role_web_identity_credentials, {})
-        end
-      end
     end
 
     describe 'with multiple sources of credentials' do
--- a/gems/aws-sdk-core/spec/aws/credential_resolution_chain_spec.rb
+++ b/gems/aws-sdk-core/spec/aws/credential_resolution_chain_spec.rb
@@ -529,6 +529,26 @@ module Aws
             client.config.credentials.credentials.access_key_id
           ).to eq('AR_AKID')
         end
+
+        it 'allows region to be resolved when unspecified' do
+          assume_role_stub(
+            'arn:aws:iam:123456789012:role/bar',
+            'ACCESS_KEY_ARPC',
+            'AR_AKID',
+            'AR_SECRET',
+            'AR_TOKEN'
+          )
+
+          allow(ENV).to receive(:[])
+          allow(ENV).to receive(:[]).with('AWS_PROFILE').and_return('ar_from_self')
+          allow(ENV).to receive(:values_at).and_return(['us-east-1'])
+
+          credentials = CredentialProviderChain.new.resolve
+
+          expect(
+            credentials.credentials.access_key_id
+          ).to eq('AR_AKID')
+        end
       end
 
       it 'can assume a role with EC2 Instance Metadata as a source' do


pier-oliviert:
The test is not a great test, but this is my first contribution here and I wasn't sure how to approach it, I tried to do it similar to how the other tests were written. Let me know if you want me to change them, or add new ones. I'd be more than happy to do it.

alextwoods:
I'm a little uncertain on what the issue is that this PR is trying to address.   When the `CredentialProviderChain` is called, its passed the client config object which has region resolved here: https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/lib/aws-sdk-core/plugins/regional_endpoint.rb#L99 - this chain includes `Aws.shared_config.region(profile: cfg.profile)`.

pier-oliviert:
@alextwoods The config is an optional field, and in the way we're using it, we're not passing any configuration objects to it. The reason for us not passing configuration in this specific scenario is that we don't have a configuration available at that time, `Aws.shared_config` is not a valid object to pass to the `CredentialProviderChain.new()`. If that was the case, then possibly this problem would not exist.

On the other hand, both the method I patched don't handle regions properly either. So, I believe the current implementation is leaky, at the very least. This seemed like the smallest changeset that followed what already existed.

mullermp:
@pier-oliviert Could you provide a snippet for how you're using the SDK and running into that issue?

pier-oliviert:
Our platform can connect to many different AWS account and so we have different credentials we're using depending on which account we want to connect to. Some accounts uses different type of credentials and so we need to support what the customer has configured.

Moreover, our devs run the platform locally and we want to use Assume Role policies to connect to some other accounts.  In the simplest scenario, here's how we do it:

```
      if Rails.env.development?
          if release?
            return ::Aws::CredentialProviderChain.new.resolve
          else
            return ::Aws::AssumeRoleCredentials.new(role_arn: role, role_session_name: ENV["USER"])
          end
        end
```

mullermp:
Ah ok.. `CredentialProviderChain` is definitely API private - it's not intended to be initialized by our customers, that is what caused confusion. It's run automatically when creating a Client. I would not depend on the order of the credential provider chain either. If you have different types of credentials that you expect customers your customers to use, you should manage (accept as argument) and pass them into Client classes directly.

Could your code not be simplified to something like:

```
if Rails.env.development? && !release?
  credentials = Aws::AssumeRoleCredentials.new(...)
end
config = {}
config[:credentials] if credentials
# if no credentials are set, chain is used
client = Aws::S3::Client.new(**config)
...
```

pier-oliviert:
@mullermp It's not a confusion, I had to use it because we don't know if environment has AssumeRoleCredentials or Credentials setup locally, we can't make that decision in the code like that. We would need to inspect the system to figure out which type of Credential to use. The only place in the code that I found that was making the distinction was in CredentialProviderChain. We can't create a logic in our app to check for what CredentialProviderChain provides as that's an implementation details of AWS's gem.

This PR isn't blocking us. We monkey patched your library and we'll maintain the fix if we have to. I think it's arguable that CredentialProviderChain shouldn't be private, or at least some kind of default mechanism should be in place where I can just load credentials the same way any of your client does.

For example, if I call `Aws::S3::Client.new` without argument, it will, under the hood call CredentialProviderChain and inhibit the proper behaviour. However, since all the available services of AWS uses their own Client type, it's not a reasonable solution to generate a client and use `Object#send()` to retrieve the credentials for all other clients we need to use.

For us, it won't change much, we'll support a monkey patch and we'll maintain it against upstream if we have to. We don't want to because I feel this can be properly fixed upstream, either through the fix I provided, or extending some logic to `Aws.config.credentials`, or something else.

alextwoods:
While @mullermp is correct that the interface to the `CredentialProviderChain` is not considered part of the SDK (meaning that we may make breaking changes to the interface or behavior), you could get the behavior you want without maintaining a monkey patch by using the existing interface, which excepts a configuration object to be passed:
```
CredentialProviderChain.new(config: Struct.new(:region).new(Aws.shared_config.region))
```

pier-oliviert:
@alextwoods that won't work as the `config` object needs to responds to a few method: https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/lib/aws-sdk-core/credential_provider_chain.rb#L36-L38

I could add those attributes to the struct, but again that feels like the implementation details would leak into our codebase.  Why can't this PR be merged if using Aws.shared_config is used elsewhere in that file?

https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/lib/aws-sdk-core/credential_provider_chain.rb#L63-L64
https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/lib/aws-sdk-core/credential_provider_chain.rb#L54-L55
https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/lib/aws-sdk-core/credential_provider_chain.rb#L153

mullermp:
I agree with you that providing your own struct to CredentialProviderChain is leaky and not recommended. The chain is intentionally private because we reserve the right to change the order and also its interface as we add new credential types. The chain itself is a fallback for not providing credentials explicitly. I am trying to understand your use case more.

>We would need to inspect the system to figure out which type of Credential to use. The only place in the code that I found that was making the distinction was in CredentialProviderChain. We can't create a logic in our app to check for what CredentialProviderChain provides as that's an implementation details of AWS's gem.

What do you mean by inspecting the system to figure out what type of credential to use? When you create a client, i.e. `client = Aws::S3::Client.new`, you can check what credentials were sourced by the chain by calling `client.config.credentials` (a public interface) and it should return the type of credentials (`Aws::Credentials`, `Aws::SharedCredentials`, etc).

pier-oliviert:
Oh, I had forgotten about `client.config.credentials`. So what you recommend is that I initialize a _dummy_ client and retrieve the credentials from there?

I understand that you want to keep the CredentialProviderChain a private entity and there are no guarantee with how credentials are parsed or how this class behaves. When I decided to use it, I looked at the git history and nothing really changed in the last 6 years so I felt confident that if there was going to be a breaking change, it wouldn't be hard for us to track and fix.

We only use CredentialsProviderChain in `Rails.env.development?` as we have explicit credentials in our other environment. So even if a change was to break our implementation, it would only affect Dev.

From the get go, the reason why I made this PR is because I got `Aws::Errors::MissingRegionError` when trying use my AssumedRole. The error was raised even when I had `AWS_REGION=us-west-2 bundle exec main.rb` which didn't make much sense to me as the region was being set. So I think there's an underlying issue with the CredentialProviderChain even if it's private and should not be used directly.

Now, I don't mind using a different API to get what I need, but I don't want to over complicate things and make things more brittle for nothing.

mullermp:
> So what you recommend is that I initialize a dummy client and retrieve the credentials from there?

Well personally .. I'd recommend creating and controlling the credentials objects you expect to use, and pass them into your clients. The client creation approach does work but I understand it's not exactly ideal for your use case..

> I got Aws::Errors::MissingRegionError when trying use my AssumedRole

AWS_REGION environment variable is loaded during the config resolution step, prior to credentials resolution. CredentialProviderChain requires a reference to the config. Some credentials objects care about region/config, others don't. `Aws.shared_config.region` will not work on its own even as a fallback because it will assume the `default` profile, but someone may want to use an explicit profile. The `client.config.credentials` approach would work in this case because AWS_REGION will be used on config resolution during client creation. However - AssumeRoleCredentials should be creating an STS client by default, and it would honor AWS_REGION if you were initializing it. What specifically caused that error to raise?

mullermp:
Could you please provide a stack trace for the missing region error and some context around your main.rb file? There may be a different fix hidden in there that I'd like to investigate.

pier-oliviert:
I took some time this morning to build a more complete snippet that will hopefully cover the problem space as well as convey  enough context to help you understand what we're trying to do here.

```ruby
#encoding: utf-8
ENV['BUNDLE_GEMFILE'] ||= File.expand_path('Gemfile', __dir__)

require 'bundler/setup' # Set up gems listed in the Gemfile.

require 'aws-sdk'
require 'aws-sdk-eks'

# User executing this script is already logged in as the assumed role 'crossaccount_administrator'
# As a result, Aws::STS::Client.new will return a configured client with the crossaccount_administrator credentials
role = "arn:aws:sts::0000000000:assumed-role/crossaccount_administrator"
begin
  credentials = ::Aws::AssumeRoleCredentials.new(role_arn: role, role_session_name: "Employee")
rescue => e
  puts e
  puts e.backtrace
end

god = Aws::SharedCredentials.new(path: "~/.aws/credentials.bak", profile_name: "god")
client = ::Aws::EKS::Client.new(credentials: god)
client.describe_cluster(name: "preprod")&.cluster
puts "SUCCESS"

# The use is now 'god' and the policies on the assume role allows god to assume this role:
credentials = ::Aws::AssumeRoleCredentials.new(role_arn: role, role_session_name: "Employee")
client = ::Aws::EKS::Client.new(credentials: credentials)
client.describe_cluster(name: "preprod")&.cluster
puts "SUCCESS"

# Trying to retrieve the system credentials, in this case, this would be the 'crossaccount_administrator'
# at the top of the script.
credentials = ::Aws::CredentialProviderChain.new.resolve
client = ::Aws::EKS::Client.new(credentials: credentials)
client.describe_cluster(name: "preprod")&.cluster
```

And the output:
```bash
AWS_REGION='us-west-2' ruby ./main.rb
User: arn:aws:sts::000000000000:assumed-role/crossaccount_administrator/Employee is not authorized to perform: sts:AssumeRole on resource: arn:aws:sts::000000000000:assumed-role/crossaccount_administrator
/Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/seahorse/client/plugins/raise_response_errors.rb:17:in `call'
/Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/aws-sdk-core/plugins/checksum_algorithm.rb:111:in `call'
/Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/aws-sdk-core/plugins/jsonvalue_converter.rb:16:in `call'
/Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/aws-sdk-core/plugins/idempotency_token.rb:19:in `call'
/Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/aws-sdk-core/plugins/param_converter.rb:26:in `call'
/Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/seahorse/client/plugins/request_callback.rb:71:in `call'
/Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/aws-sdk-core/plugins/response_paging.rb:12:in `call'
/Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/seahorse/client/plugins/response_target.rb:24:in `call'
/Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/seahorse/client/request.rb:72:in `send_request'
/Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/aws-sdk-sts/client.rb:842:in `assume_role'
/Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/aws-sdk-core/assume_role_credentials.rb:65:in `refresh'
/Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/aws-sdk-core/refreshing_credentials.rb:30:in `initialize'
/Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/aws-sdk-core/assume_role_credentials.rb:53:in `initialize'
./main.rb:13:in `new'
./main.rb:13:in `<main>'
SUCCESS
SUCCESS
/Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/aws-sdk-core/plugins/regional_endpoint.rb:92:in `after_initialize': No region was provided. Configure the `:region` option or export the region name to ENV['AWS_REGION'] (Aws::Errors::MissingRegionError)
	from /Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/seahorse/client/base.rb:81:in `block in after_initialize'
	from /Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/seahorse/client/base.rb:80:in `each'
	from /Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/seahorse/client/base.rb:80:in `after_initialize'
	from /Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/seahorse/client/base.rb:24:in `initialize'
	from /Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/aws-sdk-sts/client.rb:373:in `initialize'
	from /Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/seahorse/client/base.rb:102:in `new'
	from /Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/aws-sdk-core/assume_role_credentials.rb:51:in `initialize'
	from /Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/aws-sdk-core/shared_config.rb:238:in `new'
	from /Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/aws-sdk-core/shared_config.rb:238:in `assume_role_from_profile'
	from /Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/aws-sdk-core/shared_config.rb:125:in `assume_role_credentials_from_config'
	from /Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/aws-sdk-core/credential_provider_chain.rb:173:in `assume_role_with_profile'
	from /Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/aws-sdk-core/credential_provider_chain.rb:139:in `assume_role_credentials'
	from /Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/aws-sdk-core/credential_provider_chain.rb:13:in `block in resolve'
	from /Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/aws-sdk-core/credential_provider_chain.rb:12:in `each'
	from /Users/pier-olivier/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/aws-sdk-core-3.167.0/lib/aws-sdk-core/credential_provider_chain.rb:12:in `resolve'
	from ./main.rb:33:in `<main>'
```

Notice that I ran the script with the ENV variable set. I know I used the private method `Aws::CredentialProviderChain.new.resolve` it's just to illustrate the issue.

Honestly, if Aws::STS::Client could assume *itself* (the stack trace), I would have never needed to dig this rabbit hole to begin with.

Please let me know if you need more info, I'll try to fill in the blank as much as I can.

alextwoods:
I've updated the approach - now rather than explicitly setting the `region` to `nil` when missing, the `assume_role_with_profile` will leave it unset in the options, which will then allow the `STS::Client` created by `AssumeRoleCredentials` to follow normal resolution for `region` (which will, amongst other sources, check `AWS_REGION`).

gems/aws-sdk-core/CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 Unreleased Changes
 ------------------
-* Issue - Use Aws.shared_config when deciding on region in CredentialProviderChain.
+* Issue - Allow region resolution in `AssumeRoleCredentials` from `CredentailProviderChain`.
```
mullermp: nit: Credentail

mullermp (approved):
New approach seems fine

pier-oliviert:
Thank you @alextwoods @mullermp for your patience and help. I really appreciate it!

PR #2789, opened 2022-11-21
Validate that max_number_of_messages is a positive integer

AlexMooney:

- `* Issue - Raise ArgumentError when QueuePoller's :max_number_of_messages is not a positive integer.`

- I do not believe the `QueuePoller` is generated.

- You're welcome! :)

Reason for this PR is that the docs [here](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/SQS/QueuePoller.html) gave me some unexpected behavior:

> You can specify a maximum number of messages to receive with each
> polling attempt via `:max_number_of_messages`. When this is set to a
> positive value, greater than 1, the block will receive an array of
> messages, instead of a single message.

From the docs I expect `nil` and `0` to yield one message at a time, but it actually does `== 1` and yields an array of messages.  Let's make the unexpected behavior impossible to hit by validating the param is a positive integer.  The default behavior is unchanged: yield 1 message at a time.

### DIFF
--- a/gems/aws-sdk-sqs/CHANGELOG.md
+++ b/gems/aws-sdk-sqs/CHANGELOG.md
@@ -1,7 +1,7 @@
 Unreleased Changes
 ------------------
 
-* Issue - Raise ArgumentError when QueuePoller's :max_number_of_messages is nil or less than 1.
+* Issue - Raise ArgumentError when QueuePoller's :max_number_of_messages is not a positive integer.
 
 1.52.0 (2022-10-25)
 ------------------
--- a/gems/aws-sdk-sqs/CHANGELOG.md
+++ b/gems/aws-sdk-sqs/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Raise ArgumentError when QueuePoller's :max_number_of_messages is nil or less than 1.
+
 1.52.0 (2022-10-25)
 ------------------
 
--- a/gems/aws-sdk-sqs/lib/aws-sdk-sqs/queue_poller.rb
+++ b/gems/aws-sdk-sqs/lib/aws-sdk-sqs/queue_poller.rb
@@ -498,10 +498,12 @@ module Aws
               raise ArgumentError, "invalid option #{opt_name.inspect}"
             end
           end
-          unless @request_params[:max_number_of_messages].to_i >= 1
-            raise ArgumentError,
-                  ':max_number_of_messages must be a positive integer'
+
+          max_number_of_messages = @request_params[:max_number_of_messages]
+          unless max_number_of_messages.is_a?(Integer) && max_number_of_messages.positive?
+            raise ArgumentError, ':max_number_of_messages must be a positive integer'
           end
+
           @request_params.freeze
           freeze
         end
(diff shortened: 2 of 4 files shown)


wholien (approved):

mullermp:
Thanks for opening a pull request. The documentation for `max_number_of_messages` says:

`When this is set to a positive value, greater than 1, the block will receive an array of messages, instead of a single message.`

I don't think this option was ever intended to receive 0 or nil, so I believe the current behavior is correct and not a bug. The poller will yield 1 message at a time by default, i.e., calling `poller.poll` without options. The SQS client operation `receive_message` expects this same behavior:

```
    # @option params [Integer] :max_number_of_messages
    #   The maximum number of messages to return. Amazon SQS never returns
    #   more messages than this value (however, fewer messages might be
    #   returned). Valid values: 1 to 10. Default: 1.
```

Could you explain more of your use case and how you ran into this? If you need 1 message at a time, you could simply not provide the option to the poller.

AlexMooney:
I sent a default value of `nil` to one of my poller instances instead of omitting the param entirely, believing that _`nil` is not a positive value greater than 1_ and expected to get one message instead of an array.  Omitting a named param and sending it with a `nil` value seem like they should produce the same behavior to me.

> The SQS client operation `receive_message` expects this same behavior:

If the valid values are only 1 through 10, then the `initialize` should `raise ArgumentError` on other inputs instead of silently doing something unexpected.  How about that change?

mullermp:
That makes sense. I suppose a better default value could be `1` since that's what our poller's default is too. Raising an error could make sense, but I don't think we'd want to validate between 1-10 because then the client is not forwards compatible if SQS decides to raise their limit of 10. I think we'd possibly want to validate that the value is not nil and a positive integer.

> Omitting a named param and sending it with a nil value seem like they should produce the same behavior to me.

It really depends on the API. Sometimes nil values are supposed to be used and/or serialized in requests. If we don't do this, there's no way to specify null. We've typically used presence of key for that.

mullermp:
I'm not against this current proposal either. My point was that it was already intended behavior but it could be opened up.

mullermp:
As a sanity test, I passed both 0 and nil to `receive_message`. When 0, I get a validation error from the service. When nil, the call succeeds. Checking the request body, we are not serializing the value - This IS correct for "[aws query](https://smithy.io/2.0/aws/protocols/aws-query-protocol.html)" protocol which SQS uses: "Members with null values are not serialized."

mullermp:
So I think there are a few options here:
* Since nil is not serialized, I think it could make sense if we check `config.request_params[:max_number_of_messages].nil?` along with checking `== 1`, and yield 1 message. This is more in-line with the API's protocol that omits nil and uses the service default of 1.
* Raise error when nil or not a positive integer.
* Do nothing - update documentation to be more clear that this option should only exist when it is a value > 1

mullermp:
@AlexMooney Any input on this?

AlexMooney:
Sorry, I've been out of town and away from GH for Thanksgiving.  Given all of this information, I think we should make `max_number_of_messages: nil` yield 1 message and we should also raise an error when not `nil` or a positive integer.  I'll update this PR today.  Thanks for all of the input @mullermp!

mullermp:
No worries. The PR has been updated to raise an error when nil or not positive. We can optionally have nil yield 1 message (or just raise, too).

AlexMooney:
I've changed my mind to come around to what you wrote, @mullermp, but I changed the validation to be more strict so that no one else is surprised later by implicit casting.

mullermp:
Looks good. Thanks for the report and your contribution!

PR #2804, opened 2022-12-08
Fix Sign to not sign sigv2 cases

mullermp:
Fixes issue in https://github.com/aws/aws-sdk-ruby/discussions/2802

Custom endpoints will still flow through endpoint resolution, but for S3, it also returns Sigv4 signing. That endpoint is merged with service defaults (also Sigv4) and then eventually signed by the generic signer. Legacy sigv2 signing, when enabled, would sign on top of generic signing, which overrode Authorization. This worked for S3 but not for 3rd party solutions coincidentally because of Date Time parsing. This change fixes the behavior so that they are mutually exclusive.

### DIFF
--- a/gems/aws-sdk-core/spec/aws/plugins/sign_spec.rb
+++ b/gems/aws-sdk-core/spec/aws/plugins/sign_spec.rb
@@ -43,8 +43,8 @@ module Aws
         }
       end
 
-
       let(:client) { TestClient.new(client_options) }
+
       context 'sigv4' do
         let(:auth_scheme) do
           {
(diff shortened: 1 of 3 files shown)


gems/aws-sdk-core/spec/aws/plugins/sign_spec.rb
```diff
@@ -222,10 +222,23 @@ module Plugins
...
-          expect {
+          client = TestClient.new(client_options.merge(token_provider: nil))
+          expect do
             client.operation
-          }.to raise_error(Errors::MissingBearerTokenError)
+          end.to raise_error(Errors::MissingBearerTokenError)
+        end
+      end
+
+      context 'sigv2' do
```
alextwoods, line 232: This generic test is good - but it seems like we should have a test in S3 specifically for this as well

alextwoods (approved):

gems/aws-sdk-core/spec/aws/plugins/sign_spec.rb
mullermp: The skipping is in Sign - this Client does not have sigv2 plugin added.

PR #2837, opened 2023-03-21
Add generalized ECS credential provider support

mullermp:
Generalized ECS credential provider uses the `AWS_CONTAINER_CREDENTIALS_FULL_URI` environment variable only if `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` or `credential_path` option is not set. Also supports an Authorization header set by `AWS_CONTAINER_AUTHORIZATION_TOKEN`.


gems/aws-sdk-core/lib/aws-sdk-core/ecs_credentials.rb
```diff
@@ -77,80 +91,114 @@ def initialize options = {}
...
+      when Socket::AF_INET
+        ip_address & 0xff000000 == 0x7f000000
+      when Socket::AF_INET6
+        ip_address == 1
+      else
+        false
+      end
+    end
+
     def backoff(backoff)
```
alextwoods, line 142: We have this logic duplicated in a few places (ec2 metadata and imds credentials) - should we have a common utility for it?

I can see an argument for keeping them separate as we may want to evolve the backoff strategy separately....

gems/aws-sdk-core/lib/aws-sdk-core/ecs_credentials.rb
```diff
@@ -77,80 +91,114 @@ def initialize options = {}
...
     end
 
-    def http_get(connection, path)
-      response = connection.request(Net::HTTP::Get.new(path))
-      if response.code.to_i == 200
-        response.body
-      else
-        raise Non200Response
-      end
+    def http_get(connection)
```
alextwoods: naming nit - since we're using @credential_path rather than passing it - this really isn't a generalized "http_get".  Maybe just "http_get_credentials"?

gems/aws-sdk-core/lib/aws-sdk-core/ecs_credentials.rb
```diff
@@ -52,17 +56,27 @@ class Non200Response < RuntimeError; end
     #   credentials are refreshed. `before_refresh` is called
     #   with an instance of this object when
     #   AWS credentials are required and need to be refreshed.
-    def initialize options = {}
+    def initialize(options = {})
       @retries = options[:retries] || 5
-      @ip_address = options[:ip_address] || '169.254.170.2'
+      @host = options[:ip_address] || '169.254.170.2'
```
alextwoods: I think the ordering of initialization is pretty confusing since these get overridden by the `endpoint` option.   It may increase the number of lines - but I'd prefer to set them once, rather than overwrite.

Also side question - if `credential_path`  and `endpoint` are both unset - does the `URI.parse(endpoint)` on line 95 fail?

gems/aws-sdk-core/lib/aws-sdk-core/ecs_credentials.rb
mullermp: I think a twice duplication is fine and not needing a utility.

gems/aws-sdk-core/lib/aws-sdk-core/ecs_credentials.rb
mullermp: I'll revert back to it taking a path.

gems/aws-sdk-core/lib/aws-sdk-core/ecs_credentials.rb
mullermp: It should raise an error before line 95 right? How do you recommend setting these once? I think what I have is least confusing because we still need to check full uri is given anyway, so I do it once after defaults.

alextwoods: Theres basically only two (non-error) cases right?  Either a credential_path is configured, in which case we use host and port from the options OR endpoint is configured, in which case we set the host and port from that so something like:
```
if @credential_path
  @port = options[:port] || 80
  ...
elsif endpoint
   override_full_uri
else
  raise ArgumentError
end
```

mullermp: I think that could work. If they're both configured, full URI is supposed to be ignored. That way does set them multiple times like I already am though.

alextwoods: How does that set them multiple times? 

But anyway - just a suggestion - I'm fine with either way, just found the current initialization difficult to follow.

gems/aws-sdk-core/lib/aws-sdk-core/ecs_credentials.rb
```diff
@@ -39,6 +39,10 @@ class Non200Response < RuntimeError; end
     # @option options [Integer] :port (80)
     # @option options [String] :credential_path By default, the value of the
     #   AWS_CONTAINER_CREDENTIALS_RELATIVE_URI environment variable.
+    # @option options [String] :endpoint The ECS credential endpoint.
+    #   By default, this is the value of the AWS_CONTAINER_CREDENTIALS_FULL_URI
+    #   environment variable. This value is ignored if `credential_path` or
+    #   ENV['AWS_CONTAINER_CREDENTIALS_RELATIVE_URI'] is set.
```
alextwoods, line 47: May want to update docs to mention that port and ip address are ignored when endpoint is used (ie, when it is set and when the credential path is not).

gems/aws-sdk-core/lib/aws-sdk-core/ecs_credentials.rb
mullermp: Your snippet uses override_full_uri. But I see what you wanted to convey

alextwoods (approved):
nice - LGTM!

PR #2838, opened 2023-03-21
Enhance permanent redirect error with endpoint and bucket information

### LINKED ISSUE #2623: Please make the Endpoint accessible on Aws::S3::Errors::PermanentRedirect
**Is your feature request related to a problem? Please describe.**

When accessing buckets in bulk, they may be in different regions. A client with region X can access buckets in that region but when accessing a bucket in region Y, an `Aws::S3::Errors::PermanentRedirect` is raised. The error message returned by the API includes the `Endpoint`:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>PermanentRedirect</Code><Message>The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.</Message><Endpoint>this.is.not.available.in.Ruby</Endpoint><Bucket>[redacted]</Bucket><RequestId>[redacted]</RequestId><HostId>[redacted]</HostId></Error>
```

but this is not accessible in Ruby.

**Describe the solution you'd like**

When rescuing `Aws::S3::Errors::PermanentRedirect`, I'd like to access the endpoint so we can reconfigure the client for this bucket in subsequent requests. This is what `awscli` does internally.

**Describe alternatives you've considered**

We are currently rescuing `Aws::S3::Errors::PermanentRedirect`, followed by a call to `get_bucket_location` so we can determine which region is the right one. This requires an additional IAM permission which would not be necessary if the SDK returned all information from the API.

[ ] :wave: I may be able to implement this feature request

**Additional context**

mullermp:
Adds endpoint and bucket to PermanentRedirect error.

Closes #2623

```
[1] pry(Aws)> begin
[1] pry(Aws)*   s3.list_objects_v2(bucket: 'mamuller-us-east-1')  
[1] pry(Aws)* rescue Aws::S3::Errors::PermanentRedirect => e  
[1] pry(Aws)*   puts e.data  
[1] pry(Aws)*   puts e.message
[1] pry(Aws)* end  
...
{:endpoint=>"s3.amazonaws.com", :bucket=>"mamuller-us-east-1", :message=>"The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint."}
The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.
```

### DIFF
--- a/gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/types/permanent_redirect.rb
+++ b/gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/types/permanent_redirect.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+module Aws
+  module S3
+    module Types
+      # This error is not modeled.
+      #
+      # The bucket you are attempting to access must be addressed using the
+      # specified endpoint. Please send all future requests to this endpoint.
+      #
+      # @!attribute [rw] endpoint
+      #   @return [String]
+      #
+      # @!attribute [rw] bucket
+      #   @return [String]
+      #
+      # @!attribute [rw] message
+      #   @return [String]
+      #
+      class PermanentRedirect < Struct.new(:endpoint, :bucket, :message)
+        SENSITIVE = []
+        include Aws::Structure
+      end
+    end
+  end
+end
--- a/gems/aws-sdk-s3/CHANGELOG.md
+++ b/gems/aws-sdk-s3/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Provide `endpoint` and `bucket` attributes on `Aws::S3::Errors::PermanentRedirect` error objects.
+
 1.119.1 (2023-02-13)
 ------------------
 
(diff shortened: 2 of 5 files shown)


gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/types/permanent_redirect.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module Aws
+  module S3
+    module Types
+      # This error is not modeled.
```
alextwoods, line 6: What happens if it ever is modeled?  

Would it work to instead add the error w/ the correct members to the API (if none exist) during code generation? (ie: transform the model rather than add custom code)?

mullermp: Yeah I thought about transforming the model. I would prefer to do that, but I don't know all of the cases where PermanentRedirect would be thrown.

alextwoods (approved):

PR #2842, opened 2023-03-30
Skip bucket region cache if custom enpoint provided

0x2C6:

There is a little issue when using an s3 client's `#head_bucket` with `use_accelerate_endpoint: true` and then on a new s3 instance as `Client.new(endpoint: 'https://customendpoint.com', ...).head_bucket(bucket: 'bucket')` it redirects us to the wrong endpoint and not the one we set.

When we initiated a new s3 client with incorrect credentials and pass `use_accelerate_endpoint: true` to `head_bucket`, it makes a request to AWS and then AWS returns 400 from `s3-accelerate.amazonaws.com`. and sometimes the header returns regions as well even if the credentials are wrong. but when doing it also caches the region (where the region is not correct and actually it was credentialed for custom s3 service)

Then if we initialize a new s3 client with the correct credentials and params `Client.new(endpoint: 'https://customendpoint.com', ...)` and call `head_bucket(bucket: 'bucket')` it redirects me to AWS  endpoint instead of the custom endpoint. This happens if you do with an order. 

So here is a code example:
```ruby
# 1. making request to correct endpoint
Aws::S3::Client.new(
  credentials: credentials, 
  region: region, 
  endpoint: 'https://customendpoint.com'
).head_bucket(bucket: 'bucketname') # => bucketname.customendpoint.com:443

# 2. Make a request to AWS accelerate with the wrong credential but the same bucket name
Aws::S3::Client.new(
  credentials: credentials, 
  region: region,
).head_bucket(bucket: 'bucketname', use_accelerate_endpoint: true) # => 
# bucketname.s3-accelerate.amazonaws.com:443 
# and caches the region by using only bucket name return back from s3-accelerate (wrong region)


# 3. Make the first request again
Aws::S3::Client.new(
  credentials: credentials, 
  region: region, 
  endpoint: 'https://customendpoint.com'
).head_bucket(bucket: 'bucketname') # => bucketname.s3.us-east-2.amazonaws.com:443 which return 403
```

So after trying to access the bucket with the wrong credentials and then with the correct ones to the correct server, it checks if the bucket exists on cache `S3::BUCKET_REGIONS[bucket]` and if it is cached it tries to construct a new endpoint, and set the custom endpoint to nil `s3_signer.rb:166` and because custom endpoint set to nil it merges AWS default to it.

As I see there are only two places in `s3_signer.rb` where we need to construct a new hostname, and one of them already skipped if the client using custom endpoint `s3_signer.rb:80`


gems/aws-sdk-s3/VERSION
```diff
@@ -1 +1 @@
-1.119.2
```
alextwoods: You don't need to bump this manually - our release system will do so automatically.

gems/aws-sdk-s3/lib/aws-sdk-s3.rb
```diff
@@ -73,6 +73,6 @@
 # @!group service
 module Aws::S3
 
-  GEM_VERSION = '1.119.2'
+  GEM_VERSION = '1.119.3'
```
alextwoods: Ditto here - the release system will do this

gems/aws-sdk-s3/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 Unreleased Changes
 ------------------
 
+1.119.3 (2023-03-30)
```
alextwoods: Can remove this line - the `* issue - ` line should just be under "Unreleased Changes".  The release system will take all unreleased features/issues and create a release + make the appropriate updates to Changelog/Version

alextwoods (commented):
Thanks for submitting this - couple of minor comments!

gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/s3_signer.rb
```diff
@@ -52,7 +52,10 @@ def call(context)
 
           def check_for_cached_region(context, bucket)
             cached_region = S3::BUCKET_REGIONS[bucket]
-            if cached_region && cached_region != context.config.region
+
+            if cached_region &&
+               cached_region != context.config.region &&
+               context[:endpoint_params].endpoint.nil?
```
alextwoods: This is a little bit of a confusing internal detail - but we should use `context.config.regional_endpoint` instead of `context[:endpoint_params].endpoint.nil?`

`regional_endpoint` is set to `false` when a custom endpoint has been configured.

gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/s3_signer.rb
```diff
@@ -163,7 +166,6 @@ class << self
           def new_hostname(context, region)
             endpoint_params = context[:endpoint_params].dup
             endpoint_params.region = region
-            endpoint_params.endpoint = nil
```
0x2C6: @alextwoods do you think after this change we should remove the regional_endpoint check? because if we don't assign it nil it will be the value of metadata endpoint and will construct the host with the correct endpoint. I honestly didn't touch it becuase i couldn't found corresponding specs and got afraid to break something

alextwoods: I was honestly on the fence about this - in the case you described, you are first getting a bucket redirect from an actual S3 bucket, then using a client with a custom endpoint, but using that same bucket key - in this case, you don't really want to assume anything about the bucket's region (since it may not be an actual S3 bucket).  I'd lean towards keeping the check for `regional_endpoint` as well as this fix.

mullermp:
Please add a test!

mullermp (approved):

gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/s3_signer.rb
mullermp: I think this line should be added back. Checking `regional_endpoint` config is correct. In `new_hostname`, we should assume it is a regional endpoint, so the endpoint provider should not be passed an endpoint.

mullermp: It might already default to nil here, but I think it makes sense to just set it to be nil to be sure, since we know we want this case.

mullermp:
I had to revert this, there is an issue with the s3 service when running `Scenario: Follow 307 redirect on new buckets`.

PR #2844, opened 2023-04-04
Skip bucket region cache if custom endpoint is provided

mullermp:
#2842 was reverted because of breaking tests. This approach seems to be correct. It uses a custom endpoint method that checks dns suffix rather than whether endpoint was set or not (an unfortunate quirk for v3)

### DIFF
--- a/gems/aws-sdk-s3/CHANGELOG.md
+++ b/gems/aws-sdk-s3/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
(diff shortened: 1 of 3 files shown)


mullermp:
@0x2C6

alextwoods (approved):

0x2C6:
and what was the issue with regional_endpoint?

mullermp:
Our integration tests were setting a regional endpoint as a custom endpoint. Either approach is technically correct. This one checks custom endpoint's dns suffix.

PR #2854, opened 2023-05-11
User Agent 2.0

mullermp:
Needs some other form of testing, specifically around whether resources, paginators, etc have user agent info.


gems/aws-sdk-core/lib/aws-sdk-core/plugins/user_agent.rb
```diff
@@ -9,38 +9,109 @@ class UserAgent < Seahorse::Client::Plugin
       # @api private
       class Handler < Seahorse::Client::Handler
         def call(context)
-          set_user_agent(context)
+          context.http_request.headers['User-Agent'] = UserAgent.new(context)
```
alextwoods: I know the `to_s` will get implicitly called...somewhere down the stack, but it does seem a little weird to set the header to an object.  Maybe explicitly call `to_s`?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/user_agent.rb
```diff
@@ -9,38 +9,109 @@ class UserAgent < Seahorse::Client::Plugin
...
-        def set_user_agent(context)
-          ua = "aws-sdk-ruby3/#{CORE_GEM_VERSION}"
+        class UserAgent
+          def initialize(context)
+            @context = context
+          end
+
+          def to_s
+            ua = sdk_metadata
+            ua += " #{api_metadata}" if api_metadata
```
alextwoods, line 52: I think performance wise, it may be better to avoid using "+=" here since it will create a bunch of objects.  I think you can either use `<<` on the string or use a StringIO

gems/aws-sdk-core/lib/aws-sdk-core/plugins/user_agent.rb
```diff
@@ -9,38 +9,109 @@ class UserAgent < Seahorse::Client::Plugin
...
+          end
+
+          private
+
+          def sdk_metadata
+            "aws-sdk-ruby3/#{CORE_GEM_VERSION}"
+          end
+
+          def api_metadata
+            # TODO: transform serviceId to something?
```
alextwoods: I think the transform should be: lower case all letter and replace space with "-"

gems/aws-sdk-core/lib/aws-sdk-core/plugins/user_agent.rb
```diff
@@ -9,38 +9,109 @@ class UserAgent < Seahorse::Client::Plugin
...
+            return unless (execution_env = ENV['AWS_EXECUTION_ENV'])
 
-          if context[:gem_name] && context[:gem_version]
-            ua += " #{context[:gem_name]}/#{context[:gem_version]}"
+            "exec-env/#{execution_env}"
           end
 
-          if (execution_env = ENV['AWS_EXECUTION_ENV'])
-            ua += " exec-env/#{execution_env}"
+          # TODO: somehow set this for paginators, waiters, resources, s3 transfer, s3 encryption, etc
```
alextwoods: Yeah - this gets a bit trickier. For cases where we are creating the client in the HLL, we can set a config option.  However, when passed in... we shouldn't modify the config.  

We may need to add an option to operation methods to take additional options for this.  IE, per operation config?

gems/aws-sdk-core/spec/aws/plugins/user_agent_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  module Plugins
+    describe UserAgent do
+      let(:svc) { ApiHelper.sample_service }
+      let(:client) { svc::Client.new(stub_responses: true) }
+
+      context 'sets user agent' do
+        it 'sets the user agent' do
+          resp = client.example_operation
+          header = resp.context.http_request.headers['User-Agent']
+          puts header
```
alextwoods: remove

alextwoods (commented):
Approach makes sense and looks good overall.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/user_agent.rb
mullermp: Yeah it's looking like we might need per operation options. Otherwise some features require a config option or a context, but never both.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/user_agent.rb
mullermp: I think it's a specific transform. Dashes seem fine. Otherwise gem name would be easiest.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/user_agent.rb
mullermp: I don't think << works with frozen string literal. I would have to use string io. I think the performance is probably negligible. I think I may want metadata objects too.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/user_agent.rb
mullermp: I believe it's called by Headers. I might remove to_s and instead make a module with a method.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/user_agent.rb
```diff
@@ -6,6 +6,35 @@ module Plugins
     class UserAgent < Seahorse::Client::Plugin
       option(:user_agent_suffix)
 
+      option(
```
mullermp, line 9: TODO: sanitize this

build_tools/aws-sdk-code-generator/spec/spec_helper.rb
```diff
@@ -96,6 +96,7 @@ def generate_service(module_names, options = {})
       apig_opts = shared_opts.merge({
         api: model_path(:service, api_dir),
         default_endpoint: "https://foobar.us-west-2.amazonaws.com/test",
+        remove_plugins: ['Aws::Plugins::UserAgent']
```
alextwoods, line 99: Why are we doing this in the spec_helper?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/user_agent.rb
```diff
@@ -14,33 +38,113 @@ def call(context)
...
+
+              frameworks[match[:name]] = match[:version]
+            end
+            frameworks.map { |name, version| "lib/#{name}##{version}" }.join(' ')
+          end
         end
       end
 
-      handler(Handler)
+      handler(Handler, priority: 1)
```
alextwoods, line 146: Why this ordering?

gems/aws-sdk-s3/aws-sdk-s3.gemspec
```diff
@@ -27,7 +27,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency('aws-sdk-kms', '~> 1')
   spec.add_dependency('aws-sigv4', '~> 1.4')
-  spec.add_dependency('aws-sdk-core', '~> 3', '>= 3.165.0')
+  spec.add_dependency('aws-sdk-core', '~> 3', '>= 3.173.0')
```
alextwoods: I thought this would be overwritten by codegen?

alextwoods (approved):
Overall approach looks good - few minor comments.

gems/aws-sdk-s3/aws-sdk-s3.gemspec
mullermp: It would. This is a relic from testing (manual build). I can remove

gems/aws-sdk-core/lib/aws-sdk-core/plugins/user_agent.rb
mullermp: I figure we should be building the string as late as possible. I had this when we were doing a context approach. I can revert back or we can keep doing it as late as possible.

build_tools/aws-sdk-code-generator/spec/spec_helper.rb
mullermp: APIG has its own user agent. The tests previously had both, and we would overwrite it (because I moved priority)

gems/aws-sdk-core/lib/aws-sdk-core/plugins/user_agent.rb
```diff
@@ -14,33 +38,113 @@ def call(context)
...
+            return unless service_id
+
+            service_id = service_id.gsub(' ', '_').downcase
+            gem_version = @context[:gem_version]
+            "api/#{service_id}##{gem_version}"
+          end
+
+          # Used to be RUBY_PLATFORM
+          def os_metadata
+            local = Gem::Platform.local
```
alextwoods: Does not work well on JRuby I think - see the aws-ruby-crt code for ways to get the OS and cpu.

For me on Jruby I get: `"os/other#18 md/universal"`

mullermp: Yeah. I thought I tested this, but may have been only for engine/version.

PR #2857, opened 2023-05-23
Performance benchmark

alextwoods:
Overview:
- We are not able to simply add rake tasks for benchmarking - one of the important benchmarks we want to measure is the time/memory of requiring each gem, but our rake tasks already require the SDK.  Instead, we introduce a `rake benchmark` that will start the benchmarking with `sh`.
- Benchmarking that requires a "clean" state (sdk code not loaded at all) is all run in forked processes and metrics are communicated back to the parent through IO.pipes
- Benchmarking is broken down into: 
  - running the actual benchmarking, which produces a `benchmark_report.json` file.
  - Reading the report and putting metrics to cloudwatch.

Future improvements:
- Currently all benchmarking code is in a single file.  As we expand the set of benchmarks done this will become larger.  I'd like an easy way to source benchmarking for each service from separate files.


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,57 @@
...
+  id-token: write # required for OIDC
+  contents: read
+
+jobs:
+  benchmark:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby: [2.6, 2.7, 3.1, 3.2, jruby-9.2]
```
mullermp: We support 2.3+ and jruby 9.1+

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+
+      - uses: actions/checkout@v3
+
+      - name: Install gems
+        run: |
+          bundle install --without docs repl development --with benchmark
```
mullermp: `with benchmark` shouldn't be necessary

bin/benchmark_performance.rb
```diff
@@ -0,0 +1,425 @@
+require 'tmpdir'
+require 'memory_profiler' # MemoryProfiler does not work for JRuby
```
mullermp: CI has jruby 9.2. Should it not if this does not work?

bin/benchmark_performance.rb
```diff
@@ -0,0 +1,425 @@
...
+  else
+    "generic string"
+  end
+end
+
+def test_hash(n_keys=5, seed=0)
+  n_keys.times.map { |i| ["key#{i}", random_value(i, seed)]}.to_h
+end
+
+benchmark_gems = {
```
mullermp: It'd be nice to define this somewhere else, like in a yaml

bin/benchmark_performance.rb
```diff
@@ -0,0 +1,425 @@
+require 'tmpdir'
```
mullermp: This file is fairly long. Can it be broken down into modules and objects?

tasks/benchmark.rake
```diff
@@ -0,0 +1,91 @@
...
+
+
+  puts "Archiving benchmark report from GH with: #{ENV['GH_REF']}, #{ENV['GH_EVENT']}"
+  ref = ENV['GH_REF'] || "version-3"
+  event = ENV['GH_EVENT'] || 'push'
+  client = Aws::S3::Client.new
+
+  key = "#{event == 'pull_request' ? 'pr/' + ref : 'release'}/#{Time.now.strftime('%Y-%m-%d')}/benchmark_#{SecureRandom.uuid}.json"
+  puts "Uploading report to: #{key}"
+  client.put_object(bucket: 'aws-sdk-ruby-performance-benchmark-archive', key: key, body: File.read('benchmark_report.json'))
```
mullermp, line 24: nit - remove "-archive" from bucket name.. implies that they're cold storage.

tasks/benchmark.rake
```diff
@@ -0,0 +1,91 @@
...
+  # common dimensions
+  report_dims = {
+    event: event,
+    target: target,
+    os: report['os'],
+    cpu: report['cpu'],
+    env: report['execution_env']
+  }
+
+  def put_metric(client, dims, timestamp, k, v)
```
mullermp: nit - shouldn't define methods in rake tasks.. should likely be a module or a class to do reporting.

.github/workflows/benchmark.yml
alextwoods: This was based on top usage (plus adding a few more recent rubies) - the goal was not to produce too many metrics overall.

bin/benchmark_performance.rb
alextwoods: We still want to collect timing metrics for JRuby

bin/benchmark_performance.rb
alextwoods: Yeah - I agree - I'd like a nice way to define this and split it up.  I don't think yaml would work well since its a lot of actual code (ie, blocks - and those blocks need functions defined, ect). Any other thoughts on how to break this down and organize it better?

alextwoods: Ended up moving each service to its own file and coming up with a way to have them automatically get added.

bin/benchmark_performance.rb
alextwoods: Did a big refactoring into various files.

tasks/benchmark.rake
alextwoods: Moved to a utility file that is required.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,57 @@
...
+  id-token: write # required for OIDC
+  contents: read
+
+jobs:
+  benchmark:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby: [2.7, 2.6, jruby-9.2, 3.2]
```
mullermp: nit: ordering ..

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,57 @@
...
+        run: EXECUTION_ENV=github-action bundle exec rake benchmark
+
+      - name: configure aws credentials
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
+          role-to-assume: arn:aws:iam::469596866844:role/aws-sdk-ruby-performance-reporter
+          role-session-name: benchmark-report
+          aws-region: us-west-2
+
+      - name: Archive benchmark report
```
mullermp, line 51: I personally like "upload" instead of "archive" - up to you. Archive sounds like cold storage and we'd never see it again. I think these results would be processed regularly.

benchmark/metrics.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
```
mullermp, line 1: Should these methods be wrapped in a class or object, instead of just in a separate file?

benchmark/run_benchmark.rb
```diff
@@ -0,0 +1,83 @@
+# frozen_string_literal: true
```
mullermp, line 1: We could consider putting a bin/benchmark in aws-sdk-resources (where our aws-v3 lives) such that anyone can run benchmarking tools. It may be useful for debugging or data collection. What do you think? If so, we'd probably want to put the uploading/reporting stuff in a rake task instead.

benchmark/run_benchmark.rb
```diff
@@ -0,0 +1,83 @@
...
+Benchmark::Service.constants
+                   .map {|const| Benchmark::Service.const_get(const) }
+                   .select {|const| const.is_a? Module }.each do |m|
+  if (gem = m.const_get(:GEM_NAME)) && (benchmarks = m.const_get(:BENCHMARKS))
+    puts "Adding benchmarks from #{m} (gem: #{gem})"
+    benchmark_gems[gem] = benchmarks
+  end
+end
+
+report_data = {'version' => '1.0'}
```
mullermp: I think most of this stuff can be split out into smaller methods. Like a method to generate the report_data, one to setup benchmark, run the benchmark, etc

benchmark/services/cloudwatch.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Benchmark
```
mullermp: I do like this approach, but will this ever collide with anything? I suppose it only would if this was ever shipped with anything.

benchmark/services/cloudwatch.rb
```diff
@@ -0,0 +1,34 @@
...
+  module Service
+    module CloudWatch
+      GEM_NAME = 'aws-sdk-cloudwatch'.freeze
```
mullermp: Are constants a good idea for the interface? I would think maybe methods, like gem_name, module_name (should probably be its own thing like gem_name), and benchmarks. It could also inherit a base benchmark that knew how to execute.

benchmark/test_data.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+# polyfill random_number for older versions of Ruby.
+unless Random.new.respond_to?(:random_number)
+  class Random
```
mullermp: Must we do this? Would it make sense for random_number to just be a utility method similar to random_value?

Also should this file be wrapped in a module or object?

benchmark/util.rb
```diff
@@ -0,0 +1,120 @@
+# frozen_string_literal: true
```
mullermp: Ditto on wrapping in a module or object?

tasks/benchmark.rake
```diff
@@ -0,0 +1,59 @@
...
+
+desc 'Upload benchmarking data to cloudwatch'
+task 'benchmark:put-metrics' do
+  $:.unshift("#{$GEMS_DIR}/aws-sdk-cloudwatch/lib")
+  require 'aws-sdk-cloudwatch'
+  require_relative '../benchmark/metrics'
+
+  event = ENV['GH_EVENT'] == 'pull_request' ? 'pr' : 'release'
+  report = JSON.parse(File.read('benchmark_report.json'))
+  target = report['ruby_engine'] + "-" + report['ruby_version'].split('.').first(2).join('.') # TODO: How do we want to capture the environment
```
mullermp: What does this target end up being? This looks fragile

mullermp (commented):
Nice!

.github/workflows/benchmark.yml
alextwoods: Ordering is based on usage

.github/workflows/benchmark.yml
alextwoods: They aren't intended to be processed regularly.  Its just archive data - we can come back to it if we need to, but I don't expect to regularly.

benchmark/run_benchmark.rb
alextwoods: I dont think we want it in resources - its designed to run on a commit in a GH repo (including how it sets up paths).  When we eventually move past an initial POC, we can look at where parts of this should live (likely not in the SDK at all - probably separate packages)

benchmark/services/cloudwatch.rb
alextwoods: Its never shipped with anything.  Its only run as an isolated task - it shouldn't conflict with anything.

When we move past the POC and create separate packages, we may want different name spaces.

benchmark/test_data.rb
alextwoods: it doesn't work as a method since it works on the state of an object.  I guess I could create a new Random class instead.

tasks/benchmark.rake
alextwoods: It's the major and minor version.  It drops the third/patch version to prevent metrics churn on minor version updates.

Its fairly robust because the split/first/join method all work on a range of inputs.  Eg, if ruby version is "beta", the target will be "ruby-beta".  If its "1.2.3.4.5.beta", it will be "ruby-1.2".

benchmark/util.rb
alextwoods: I guess I can wrap everything in a module.  These are all isolated usage, this is never distributed anywhere and is intended to eventually be refactored and run very differently, so I just wanted to keep it flat and simple.  But I'll move it into a module.

benchmark/test_data.rb
mullermp: Monkey patch is fine if there's never going to be a collision or it's not distributed anywhere

benchmark/services/cloudwatch.rb
mullermp: What separate packages would we need? And why would it ever be more than one?

.github/workflows/benchmark.yml
mullermp: That order can change at any time ..

tasks/benchmark.rake
mullermp: I know with rails they have constants for major/minor. Maybe there's a ruby constant for those too. But this is otherwise fine.

benchmark/util.rb
mullermp: I think a Aws::Benchmark module would be fine.

mullermp: Or just Benchmark is fine, too.

alextwoods: I think Aws::Benchmark is more likely to overlap (with a possible Benchmark service) - this is not intended for distribution or use as a library, so I think Benchmark is fine.

benchmark/test_data.rb
alextwoods: Yeah - its not a library - it only ever runs in an isolated environment so I think this monkeypatch approach is reasonable.

benchmark/metrics.rb
alextwoods: Everything has been moved to a module now.

benchmark/run_benchmark.rb
alextwoods: Moved everything to methods.

benchmark/services/cloudwatch.rb
alextwoods: Yeah, a base benchmark class was a good idea - I've created a class that has most of the benchmarking logic and each gem we want to benchmark just needs to define a few simple methods.

tasks/benchmark.rake
mullermp: I think the TODO can be removed, then, seems like this is the best way.

benchmark/metrics.rb
```diff
@@ -0,0 +1,37 @@
...
+
+module Benchmark
```
mullermp, line 3: Did you mean for this to be in Benchmark::Metrics? Otherwise I think it's fine to stuff it under Benchmark in benchmark.rb

mullermp (approved):

PR #2858, opened 2023-05-24
Fix Checksum algo spec to pass JRuby CI

jterapin:
Checksum algorithm's unit test relating to unsigned-payload auth has been failing JRuby CI. At the time of the plugin release, we quickly patched a fix to match with other SDKs' release timing.  This PR addresses a permanent fix so that JRuby CI passes.

### DIFF
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,7 +1,7 @@
 Unreleased Changes
 ------------------
 
-* Issue - Updated `checksum_algorithm` plugin with a permanent fix to pass JRuby CI.
+* Issue - Updated `checksum_algorithm` plugin to use IO.copy_stream for JRuby.
 
 3.173.0 (2023-05-18)
 ------------------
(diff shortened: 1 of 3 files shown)


gems/aws-sdk-core/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Updated `checksum_algorithm` plugin with a permanent fix to pass JRuby CI.
```
mullermp:
```suggestion
* Issue - Updated `checksum_algorithm` plugin to use IO.copy_stream for JRuby.
```

mullermp (approved):

alextwoods (approved):
Great fix!! Thanks for updating this

gems/aws-sdk-core/CHANGELOG.md
alextwoods: Thats not really what its doing though - its just fixing a default argument that was missing in the plugin so that the way our specs work don't need special casing for JRuby (the actual JRuby net::http client worked fine with this code, so I don't believe this was an actual issue, just an artifact of the way these specs were written).

PR #2860, opened 2023-05-31
Fix s3 download ranges for chunks

### LINKED ISSUE #2859: aws-sdk-s3 fails when downloading file of exactly 20971524 bytes from S3 bucket bug
### Describe the bug

Hi,

Using FileDownloader's multithreaded download, downloading a file that is exactly 20971524 bytes long results in an `Aws::S3::Errors::InvalidRange` (The requested range is not satisfiable) exception.

The issue comes from what looks like a off-by-one error, resulting in that case in an invalid range for that last chunk.

### Expected Behavior

File should obviously be downloadable without an exception due to incorrect math/logic :)

### Current Behavior

```
[Aws::S3::Client 206 12.824679 0 retries] get_object(bucket:"xxx",key:"yyy.jpeg",range:"bytes=0-5242880")  
[Aws::S3::Client 206 12.635049 0 retries] get_object(bucket:"xxx",key:"yyy.jpeg",range:"bytes=5242881-10485761")  
[Aws::S3::Client 206 15.445052 0 retries] get_object(bucket:"xxx",key:"yyy.jpeg",range:"bytes=10485762-15728642")  
[Aws::S3::Client 206 11.991756 0 retries] get_object(bucket:"xxx",key:"yyy.jpeg",range:"bytes=15728643-20971523")  
[Aws::S3::Client 416 0.804207 0 retries] get_object(bucket:"xxx",key:"yyy.jpeg",range:"bytes=20971524-20971524") Aws::S3::Errors::InvalidRange The requested range is not satisfiable
```

The last chunk is empty/non existant, resulting in the exception.

### Reproduction Steps

Any 20971524 bytes long file should show the error, using default `download_file` options.

### Possible Solution

From my understanding, instead of downloading chunks that are MIN_CHUNK_SIZE (= 5242880 bytes) long, the code currently downloads chunks that are MIN_CHUNK_SIZE+1 long.

A file that is 20971524 bytes long should be downloaded as 4 5242880-bytes parts + a last 4-bytes part (20971524 = 5242880 * 4 + 4): 5 parts in total.
However, as the code downloads 5242880+1 bytes parts, we currently have 4*5242881 parts plus a final... 0-byte part (which is invalid).

Possible solutions:
- avoid building empty chunks
- fix `construct_chunks` to build `default_chunk_size` long parts instead of `default_chunk_size+1` (then, also check that a file that is exactly 4*5242880 bytes long gets downloaded in 4 parts w/o a final empty part)

See also https://github.com/aws/aws-sdk-ruby/issues/2561

aws-sdk-s3

Ruby 3.1, 3.2, Linux

mullermp:
Fixes #2859
Related to #2563

### DIFF
--- a/gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
+++ b/gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
@@ -58,7 +58,7 @@ module Aws
         resp = @client.head_object(@params.merge(part_number: 1))
         count = resp.parts_count
         if count.nil? || count <= 1
-          if resp.content_length <= MIN_CHUNK_SIZE
+          if resp.content_length < MIN_CHUNK_SIZE
             single_request
           else
             multithreaded_get_by_ranges(construct_chunks(resp.content_length))
@@ -66,7 +66,7 @@ module Aws
         else
           # partNumber is an option
           resp = @client.head_object(@params)
-          if resp.content_length <= MIN_CHUNK_SIZE
+          if resp.content_length < MIN_CHUNK_SIZE
             single_request
           else
             compute_mode(resp.content_length, count)
@@ -88,10 +88,11 @@ module Aws
         offset = 0
         default_chunk_size = compute_chunk(file_size)
         chunks = []
-        while offset <= file_size
-          progress = offset + default_chunk_size - 1
-          chunks << "bytes=#{offset}-#{progress < file_size ? progress : file_size}"
-          offset = progress + 1
+        while offset < file_size
+          progress = offset + default_chunk_size
+          progress = file_size if progress > file_size
+          chunks << "bytes=#{offset}-#{progress - 1}"
+          offset = progress
         end
         chunks
       end
@@ -100,11 +101,9 @@ module Aws
         if @chunk_size && @chunk_size > file_size
           raise ArgumentError, ":chunk_size shouldn't exceed total file size."
         else
-          chunk_size = @chunk_size || [
+          @chunk_size || [
             (file_size.to_f / MAX_PARTS).ceil, MIN_CHUNK_SIZE
           ].max.to_i
-          chunk_size -= 1 if file_size % chunk_size == 0
-          chunk_size
         end
       end
 
--- a/gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
+++ b/gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
@@ -58,7 +58,7 @@ module Aws
         resp = @client.head_object(@params.merge(part_number: 1))
         count = resp.parts_count
         if count.nil? || count <= 1
-          if resp.content_length < MIN_CHUNK_SIZE
+          if resp.content_length <= MIN_CHUNK_SIZE
             single_request
           else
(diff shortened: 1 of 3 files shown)


gems/aws-sdk-s3/spec/object/download_file_spec.rb
```diff
@@ -179,53 +182,6 @@ module S3
               ":chunk_size shouldn't exceed total file size."
             )
         end
-
-        context 'file size is chunk size boundary' do
```
alextwoods, line 183: It seems like we should still keep a similar test.

I think we should likely also have a test that size=min_chunk_size uses single_request.

alextwoods (commented):
Looks good overall - but I do think its worth expanding on tests a little.

gems/aws-sdk-s3/spec/object/download_file_spec.rb
mullermp: I think the mock that I added will work in all cases. It verifies that there's no range where the bytes are equal. There's already a case here where it's on the boundary, too.

mullermp: 1 byte ranges apparently work..

alextwoods (approved):
LGTM!

cmaion:
:+1:

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -58,15 +58,15 @@ def multipart_download
         resp = @client.head_object(@params.merge(part_number: 1))
         count = resp.parts_count
         if count.nil? || count <= 1
-          if resp.content_length <= MIN_CHUNK_SIZE
+          if resp.content_length < MIN_CHUNK_SIZE
```
cmaion: I believe you can keep the `<=` operator: no need to fire a single thread for A MIN_CHUNK_SIZE bytes long file, as it will be downloaded in a single chunk anyway.

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -58,15 +58,15 @@ def multipart_download
...
             single_request
           else
             multithreaded_get_by_ranges(construct_chunks(resp.content_length))
           end
         else
           # partNumber is an option
           resp = @client.head_object(@params)
-          if resp.content_length <= MIN_CHUNK_SIZE
```
cmaion: Same here

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
mullermp: I didn't want to change behavior if I didn't have to. I think changing this could break some test expectations for some customers. This code is old as dirt.

mullermp: I think we can roll with it anyway. The optimization is worth it.

cmaion: Oh ok sorry, I misread a commit. I didn't see that it's currently <, but you changed it to <= then back to < :smiley: 
I agree, <= should be harmless :+1:

PR #2861, opened 2023-06-02
Accessibility fixes for docs

mullermp:
First wave of accessibility fixes.


doc-src/templates/default/fulldoc/html/css/common.css
```diff
@@ -0,0 +1,13 @@
+/* Default color is 888, change to 666 for at least a 4.5 contrast ratio */
+li .item { color: #666; cursor: pointer; }
+#search { width: 170px; position: static; margin: 3px; margin-left: 10px; font-size: 0.9em; color: #666; padding-left: 0; padding-right: 24px; }
+
+/* Override opacity for at least a 4.5 contrast ratio */
+li.collapsed a.toggle {
+    opacity: 1;
+    cursor: default;
```
alextwoods, line 8: why cursor: default?  What is it otherwise?

doc-src/templates/default/fulldoc/html/css/common.css
```diff
@@ -0,0 +1,13 @@
+/* Default color is 888, change to 666 for at least a 4.5 contrast ratio */
+li .item { color: #666; cursor: pointer; }
+#search { width: 170px; position: static; margin: 3px; margin-left: 10px; font-size: 0.9em; color: #666; padding-left: 0; padding-right: 24px; }
```
alextwoods, line 3: Is specifying width in px going to cause a problem with different screen sizes/layouts?

alextwoods (commented):
Looks good overall - just minor comments.

.yardopts
```diff
@@ -6,4 +6,5 @@
 --hide-api private
 --plugin sitemap
 --exclude \/client_api.rb$
-gems/**/lib/**/*.rb
+gems/aws-sdk-core/lib/**/*.rb
```
alextwoods: Assume this is just for testing and should be reverted before merge?

doc-src/templates/default/fulldoc/html/css/common.css
mullermp: It's copied from the existing styling, just change opacity.

doc-src/templates/default/fulldoc/html/css/common.css
mullermp: Probably, but this is also a direct copy of the existing styling, but with a color change.

.yardopts
mullermp: Yeah. It would be reverted. You can use it to build s3 docs and see the changes for yourself (please do so)

doc-src/templates/default/fulldoc/html/css/common.css
mullermp: I think it's just for the search pop up anyway. So even smaller layouts will probably have search that's this wide.

alextwoods (approved):

PR #2872, opened 2023-06-22
Minimum of expiration time and expires_in for presigned urls

mullermp:
Behavior change where `expiration` does not refresh. In doing so, we can check expiration on credentials freely and use it to calculate the expiration time of a presigned url. If expiration of credentials is shorter than the provided expires_in, use the credentials expiration.

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core/refreshing_credentials.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/refreshing_credentials.rb
@@ -36,12 +36,6 @@ module Aws
       @credentials
     end
 
-    # @return [Time,nil]
-    def expiration
-      refresh_if_near_expiration!
-      @expiration
-    end
-
     # Refresh credentials.
     # @return [void]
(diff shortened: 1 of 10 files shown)


gems/aws-sdk-core/lib/aws-sdk-core/refreshing_credentials.rb
```diff
@@ -36,12 +36,6 @@ def credentials
       @credentials
     end
 
-    # @return [Time,nil]
-    def expiration
-      refresh_if_near_expiration!
```
alextwoods, line 41: I'm a little worried about changing this behavior - I suspect we might rely on that in some places?

mullermp: I don't see any cases in the SDK but certainly can in the wild. My concern is that by fetching expiration (after credentials) they will get refreshed, then the request will be invalid

alextwoods (approved):

PR #2878, opened 2023-07-05
More accessibility fixes

mullermp:
* Adds support for 320 px viewing (except tables)
* Support keyboard navigation of full_list
* Adds aria labels
* Bug fix for searching (introduced when upgrading jquery locally)
* Reads out search results

### DIFF
--- a/.yardopts
+++ b/.yardopts
@@ -6,5 +6,4 @@
 --hide-api private
 --plugin sitemap
 --exclude \/client_api.rb$
-gems/aws-sdk-core/lib/**/*.rb
-gems/aws-sdk-s3/lib/**/*.rb
+gems/**/lib/**/*.rb
--- a/.yardopts
+++ b/.yardopts
@@ -6,4 +6,5 @@
 --hide-api private
 --plugin sitemap
 --exclude \/client_api.rb$
-gems/**/lib/**/*.rb
+gems/aws-sdk-core/lib/**/*.rb
+gems/aws-sdk-s3/lib/**/*.rb
--- a/doc-src/templates/default/fulldoc/html/full_list.erb
+++ b/doc-src/templates/default/fulldoc/html/full_list.erb
@@ -30,7 +30,7 @@
         <!-- Add label for search -->
         <div id="search">
           <label for="search-class">Search:</label>
-          <input id="search-class" type="text">
+          <input id="search-class" type="text" />
         </div>
       </div>
 
--- a/doc-src/templates/default/fulldoc/html/full_list.erb
+++ b/doc-src/templates/default/fulldoc/html/full_list.erb
@@ -28,7 +28,7 @@
         </div>
 
         <!-- Add label for search -->
-        <div id="search" >
+        <div id="search">
           <label for="search-class">Search:</label>
           <input id="search-class" type="text">
         </div>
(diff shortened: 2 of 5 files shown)


doc-src/templates/default/fulldoc/html/js/full_list.js
```diff
@@ -0,0 +1,246 @@
+(function() {
```
mullermp, line 3: This file comes from https://github.com/lsegal/yard/blob/main/templates/default/fulldoc/html/js/full_list.js but has some modifications.

Here is the diff:

```
64a65,68
>     // custom - flip aria attribute
>     $(this).attr('aria-expanded', function (i, attr) {
>         return attr == 'true' ? 'false' : 'true'
>     });
66a71,84
> 
>   // custom - needed for accessibility
>   $('#full_list a.toggle').on('keypress',function(evt) {
>     // enter key
>     if (evt.which == 13) {
>       evt.stopPropagation();
>       evt.preventDefault();
>       $(this).parent().parent().toggleClass('collapsed');
>       $(this).attr('aria-expanded', function (i, attr) {
>           return attr == 'true' ? 'false' : 'true'
>       });
>       highlight();
>     }
>   });
94c112,113
<   $('#full_list').after("<div id='noresults' style='display:none'></div>");
---
>   // custom - add role status
>   $('#full_list').after("<div id='noresults' role='status' style='display:none'></div>");
144a164
> 
157c177,179
<   if ($('#full_list li:visible').size() === 0) {
---
>   // bug fix
>   var found = $('#full_list li > div.found').length;
>   if (found === 0) {
160c182,183
<     $('#noresults').text('').hide();
---
>     // accessibility - read out results
>     $('#noresults').text('There are ' + found + ' results.').hide().fadeIn();
190a214,220
> 
>   // custom - flip aria attribute
>   $target.find('a.toggle').attr('aria-expanded', 'true')
>   $target.parentsUntil('#full_list', 'li').each(function(i, el) {
>     $(el).find('> div > a.toggle').attr('aria-expanded', 'true');
>   });
> 
216c246
< })();
---
> })();
\ No newline at end of file
```

.yardopts
```diff
@@ -6,4 +6,5 @@
 --hide-api private
 --plugin sitemap
 --exclude \/client_api.rb$
-gems/**/lib/**/*.rb
+gems/aws-sdk-core/lib/**/*.rb
+gems/aws-sdk-s3/lib/**/*.rb
```
alextwoods: temporary?

doc-src/templates/default/fulldoc/html/full_list.erb
```diff
@@ -28,7 +28,7 @@
         </div>
 
         <!-- Add label for search -->
-        <div id="search">
+        <div id="search" >
```
alextwoods: Is the space significant?

doc-src/templates/default/fulldoc/html/js/full_list.js
alextwoods: Maybe add a comment at the top of the file about the source.

alextwoods (approved):
LGTM!

.yardopts
mullermp: Yeah, for you to test.

doc-src/templates/default/fulldoc/html/full_list.erb
mullermp: No. I was doing something here before. I've removed it.

doc-src/templates/default/fulldoc/html/js/full_list.js
mullermp: Done

PR #2879, opened 2023-07-06
Checksum support for multipart copies and support for source part sizes

benklop:


gems/aws-sdk-s3/lib/aws-sdk-s3/object_multipart_copier.rb
```diff
@@ -78,10 +78,15 @@ def copy_part_thread(queue)
...
         {
-          etag: @client.upload_part_copy(part).copy_part_result.etag,
+          etag: result.etag,
+          checksum_crc32: result.checksum_crc32,
+          checksum_crc32c: result.checksum_crc32c,
+          checksum_sha1: result.checksum_sha1,
+          checksum_sha256: result.checksum_sha256,
           part_number: part[:part_number],
-        }
+        }.compact
```
alextwoods: I believe `compact` was introduced in Ruby 2.4, and we still need to support Ruby 2.3.  

Additionally - I think this should be a bit more generic (ie, if additional checksum algorithms are added in the future), something more like: https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb#L164

I know the API doesn't provide the actual checksum here, but maybe we can check for any `checksum_*` fields that are set.

gems/aws-sdk-s3/lib/aws-sdk-s3/object_multipart_copier.rb
```diff
@@ -104,24 +109,33 @@ def compute_parts(size, default_part_size, options)
...
+      def byte_range(offset, part_size, size)
+        if offset + part_size < size
+          "bytes=#{offset}-#{offset + part_size - 1}"
         else
           "bytes=#{offset}-#{size - 1}"
         end
       end
 
+      def calculate_part_size(part_number, default_part_size, options)
+        if @use_source_parts && source_metadata(options.merge({ part_number: 1 }))[:parts_count]
```
alextwoods: using `source_metadata` here will make an additional call to `head_object` for each part we need to copy, but I believe this call should always return the same value.

gems/aws-sdk-s3/lib/aws-sdk-s3/object_multipart_copier.rb
```diff
@@ -20,13 +20,13 @@ class ObjectMultipartCopier
       #   will be constructed from the given `options' hash.
       # @option [Integer] :thread_count (10) Number of concurrent threads to
       #   use for copying parts.
+      # @option [Boolean] :use_source_parts (false) Use part sizes defined on
```
alextwoods: Whats the reason for adding this option?

benklop: If copying or moving an object that is already multipart, this does not re-part the object, instead re-using the part definitions on the original. That means the etag and any checksums will not change.

benklop: It's possible however that some may be using this to intentionally re-part an object that has part sizes that are not convenient, so I thought it made sense to make that behavior optional

gems/aws-sdk-s3/lib/aws-sdk-s3/object_multipart_copier.rb
benklop: I've updated to avoid specifying specific algorithms and compact.

gems/aws-sdk-s3/lib/aws-sdk-s3/object_multipart_copier.rb
benklop: yes, this will always return the same value. I've extracted it to a memoized method so it only gets called once.

gems/aws-sdk-s3/lib/aws-sdk-s3/object_multipart_copier.rb
benklop: this is especially useful if the source object has parts with varied sizes, like the output from `EC2::Client#create_store_image_task`. In that case it's not possible to simply specify a max part size that matches that of the source object.

mullermp: I think this is important context to put in the documentation for the option.

benklop: I'll add a summary, sure.

benklop:
Is there anything I can do to help this along?

alextwoods (approved):
Thanks for this contribution - LGTM!

alextwoods:
Note - credential fetching failures in benchmark CI are expected and we can ignore.

mullermp:
The docs need to be placed on the public API. I'm trying to push a change but it's being rejected. Can you make sure you allow contributions to the PR?

mullermp:
`gh pr checkout` is having issues because of the source of this (fork of crittercism). Please just apply this diff to the PR:

```
diff --git a/gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb b/gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
index 07956f49a9..490837dd40 100644
--- a/gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
+++ b/gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
@@ -46,6 +46,14 @@ module Aws
       #   different region. You do not need to specify this option
       #   if you have provided a `:source_client` or a `:content_length`.
       #
+      # @option options [Boolean] :use_source_parts (false) Only used when
+      #   `:multipart_copy` is `true`. Use part sizes defined on the source
+      #   object if any exist. If copying or moving an object that
+      #   is already multipart, this does not re-part the object, instead
+      #   re-using the part definitions on the original. That means the etag
+      #   and any checksums will not change. This is especially useful if the
+      #   source object has parts with varied sizes.
+      #
       # @example Basic object copy
       #
       #   bucket = Aws::S3::Bucket.new('target-bucket')
diff --git a/gems/aws-sdk-s3/lib/aws-sdk-s3/object_multipart_copier.rb b/gems/aws-sdk-s3/lib/aws-sdk-s3/object_multipart_copier.rb
index 9ebcca2b76..54adbcd528 100644
--- a/gems/aws-sdk-s3/lib/aws-sdk-s3/object_multipart_copier.rb
+++ b/gems/aws-sdk-s3/lib/aws-sdk-s3/object_multipart_copier.rb
@@ -15,17 +15,16 @@ module Aws
       MAX_PARTS = 10_000
 
       # @option options [Client] :client
-      # @option [Integer] :min_part_size (52428800) Size of copied parts.
-      #   Defaults to 50MB.
-      #   will be constructed from the given `options' hash.
-      # @option [Integer] :thread_count (10) Number of concurrent threads to
-      #   use for copying parts.
-      # @option [Boolean] :use_source_parts (false) Use part sizes defined on
-      #   the source object if any exist. If copying or moving an object that
-      #   is already multipart, this does not re-part the object, instead
-      #   re-using the part definitions on the original. That means the etag
-      #   and any checksums will not change. this is especially useful if the
-      #   source object has parts with varied sizes.
+      # @option options [Integer] :min_part_size (52428800)
+      #   Size of copied parts. Defaults to 50MB.
+      # @option options [Integer] :thread_count (10) Number of concurrent
+      #   threads to use for copying parts.
+      # @option options [Boolean] :use_source_parts (false) Use part sizes
+      #   defined on the source object if any exist. If copying or moving an
+      #   object that is already multipart, this does not re-part the object,
+      #   instead re-using the part definitions on the original. That means
+      #   the etag and any checksums will not change. This is especially
+      #   useful if the source object has parts with varied sizes.
       def initialize(options = {})
         @use_source_parts = options.delete(:use_source_parts) || false
         @thread_count = options.delete(:thread_count) || 10
```

benklop:
@mullermp sorry for the trouble. I've applied the diff.

mullermp:

PR #2882, opened 2023-07-10
Generalized ECS provider v2

mullermp:
Main differences are: token file support and loosening of requirements for full_uri.


gems/aws-sdk-core/lib/aws-sdk-core/ecs_credentials.rb
```diff
@@ -174,10 +203,28 @@ def get_credentials
...
         end
       end
+    rescue TokenFileReadError
+      raise
     rescue StandardError
       '{}'
     end
 
+    def fetch_authorization_token
+      if (path = ENV['AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE'])
```
alextwoods, line 215: Is the ENV the only way to set this path? ie, no way to initialze these credentials with a path?

gems/aws-sdk-core/lib/aws-sdk-core/ecs_credentials.rb
```diff
@@ -202,6 +250,8 @@ def retry_errors(error_classes, options = {})
       retries = 0
       begin
         yield
+      rescue TokenFileReadError
```
alextwoods: Why do we need to do this explicitly here? It's not in the list of exceptions to retry right?

gems/aws-sdk-core/spec/aws/ecs_credentials_spec.rb
```diff
@@ -165,33 +118,79 @@ module Aws
...
+        it 'defaults to 0' do
+          stub_request(:get, "http://169.254.170.2#{path}").to_raise(SocketError)
+          expect(ECSCredentials.new(backoff: 0).retries).to be(5)
+        end
 
-      it 'defaults to 0' do
-        stub_request(:get, "http://169.254.170.2#{path}").to_raise(SocketError)
-        expect(ECSCredentials.new(backoff: 0).retries).to be(5)
-      end
+        it 'keeps trying "retries" times, with exponential backoff' do
```
alextwoods: Not: can you use backticks here rather than double quotes?

alextwoods (commented):
Looks good!

gems/aws-sdk-core/lib/aws-sdk-core/ecs_credentials.rb
mullermp: It is otherwise a standard error and gets retried. I assume if the file doesn't exist then retrying with back off will likely fail.

gems/aws-sdk-core/lib/aws-sdk-core/ecs_credentials.rb
mullermp: Only env was called out in the design. I'm not sure about a client option. We didn't have one with the token itself. I assume this value can change during runtime.

gems/aws-sdk-core/spec/aws/ecs_credentials_spec.rb
mullermp: I'll just rename the test

alextwoods (approved):

PR #2883, opened 2023-07-12
Add store image task complete waiter

benklop:

### DIFF
--- a/apis/ec2/2016-11-15/waiters-2.json
+++ b/apis/ec2/2016-11-15/waiters-2.json
@@ -521,13 +521,13 @@
           "expected": "Failed",
           "matcher": "pathAny",
           "state": "failure",
-          "argument": "store_image_task_results[].store_task_state"
+          "argument": "StoreImageTaskResults[].StoreTaskState"
         },
         {
           "expected": "InProgress",
           "matcher": "pathAny",
           "state": "retry",
-          "argument": "store_image_task_results[].store_task_state"
+          "argument": "StoreImageTaskResults[].StoreTaskState"
         }
       ]
     },
--- a/apis/ec2/2016-11-15/waiters-2.json
+++ b/apis/ec2/2016-11-15/waiters-2.json
@@ -507,7 +507,7 @@
       ]
     },
     "StoreImageTaskComplete": {
-      "Delay": 5,
+      "delay": 5,
       "operation": "DescribeStoreImageTasks",
       "maxAttempts": 40,
       "acceptors": [
@@ -528,7 +528,7 @@
           "matcher": "pathAny",
           "state": "retry",
           "argument": "store_image_task_results[].store_task_state"
-        },
+        }
       ]
     },
     "SubnetAvailable": {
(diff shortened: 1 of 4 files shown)


mullermp:
Thanks. Unfortunately this has to be merged into an internal package upstream. I will have to test the changes before doing so. The turn around time is long because it involves someone from ec2.

benklop:
no problems. I've monkey patched this in my codebase so i'm in no huge hurry; I'm just trying to drain as many of my patches as possible to simplify things. As long as it gets through eventually I'll be happy.

apis/ec2/2016-11-15/waiters-2.json
```diff
@@ -506,6 +506,31 @@
...
+          "expected": "Completed",
+          "matcher": "pathAll",
+          "state": "success",
+          "argument": "StoreImageTaskResults[].StoreTaskState"
+        },
+        {
+          "expected": "Failed",
+          "matcher": "pathAny",
+          "state": "failure",
+          "argument": "store_image_task_results[].store_task_state"
```
mullermp: Are you sure this branch worked? I think this needs to be StoreImageTaskResults[].StoreTaskState

mullermp:
No need to update PR, but it would be nice if you can help verify the change for failure and retry states.

apis/ec2/2016-11-15/waiters-2.json
benklop: you're definitely right. I built the resultant waiter by hand when i was testing (before i was aware of how the code generation worked), and converted it into the yaml format later. I clearly screwed that up on these last two.

mullermp: No worries. Change both of those, rebuild, and verify the output you expect (if you can force a failure and/or use a small polling delay to test in progress).

I've found someone on Ec2 willing to take the change alongside a new service feature, ETA 8/1-ish.

benklop: That change doesn't appear to alter the generated code, and it's identical to the hand-built one i've tested aside from using double instead of single quotes.

I haven't tested the failure state - i'm not sure how to force a failure of an image store task - but retry certainly works

mullermp: Depending on how quick it functions, you can possibly cancel it on another thread/terminal or from the console.

benklop: I don't actually see an API to cancel it - only create and describe.

benklop: I was able to test it by cancelling the multipart upload the task initiated. I received `block in poll': stopped waiting, encountered a failure state (Aws::Waiters::Errors::FailureStateError)` which seems like probably the correct behavior.

benklop:
As mentioned in the code comment, retry and failure states have been verified.

mullermp:
I concur. Thank you for double checking. I'll close this out when it's gone out.

mullermp:
This has been released here: https://github.com/aws/aws-sdk-ruby/commit/8c183c89cbeb462dcb6791467e2eab39a89fb9a6#diff-7c8f8f47fac8535cbf861ed8c854ab9e5ad87a785f9d21c87f66df698f2044f8

PR #2885, opened 2023-07-17
Minor accessibility fixes

mullermp:
Cannot use display: none and need to add role='button' to arrows.


doc-src/templates/default/fulldoc/html/css/common.css
```diff
@@ -3,17 +3,24 @@ li .item { color: #666; cursor: pointer; }
 #search { width: 170px; position: static; margin: 3px; margin-left: 10px; font-size: 0.9em; color: #666; padding-left: 0; padding-right: 24px; }
 
 /* Override opacity for at least a 4.5 contrast ratio */
```
alextwoods, line 5: Nit: with opacity removed from this, is it overriding?

alextwoods (approved):
LGTM

doc-src/templates/default/fulldoc/html/css/common.css
mullermp: It defaults to 1 normally. But actually, I think I need it..

PR #2886, opened 2023-07-19
Add support for verifying checksums in FileDownloader

alextwoods:
Add support for verifying checksums in `FileDownloader`/`object.download_file` by adding the `checksum_mode` option.  This defaults it to `ENABLED`, as it is considered a best practice to enable in S3 TransferManagers.  However, validation will ONLY happen if the object has checksum data associated with it.    

When validation occurs and fails, an error is raised.   However, is is difficult to determine if validation was successful or was skipped (because the object did not have checksum data stored) - a normal `get_object` request returns metadata in `response.context[:http_checksum][:validated]`, but this response (or in the case of multipart, multiple responses) is not available in the return from `object.download_file`.  Customers can use the `on_checksum_validated` option to provide a callback which can be used to verify that validation occurred and which algorithm was used to validate. 


gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
```diff
@@ -492,6 +492,18 @@ def upload_file(source, options = {})
       #   retrieve the object. For more about object versioning, see:
       #   https://docs.aws.amazon.com/AmazonS3/latest/dev/ObjectVersioning.html
       #
+      # @option options [String] checksum_mode (ENABLED) When `ENABLED` and
```
mullermp, line 495: What are other values here and why is it a String instead of a boolean?

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -142,9 +161,17 @@ def write(resp)
       end
 
       def single_request
-        @client.get_object(
+        resp = @client.get_object(
           @params.merge(response_target: @path)
         )
+        if @on_checksum_validated &&
```
mullermp: guard clause

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -129,6 +140,14 @@ def thread_batches(chunks, param)
                 @params.merge(param.to_sym => chunk)
               )
               write(resp)
+              if @on_checksum_validated &&
+                resp.context[:http_checksum] &&
+                resp.context[:http_checksum][:validated]
```
mullermp: Checking context in the response here is kinda fragile. We also were talking about not having context on responses in the future. Is there a better way to determine this information? Maybe pushing logic to the checksum plugin?

gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
alextwoods: ENABLED is actually the only value for it. It matches what is modeled on the api and leaves open the possibility of additional values in the future - I think we'd want to match the underlying api here in case this is added to in the future.

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
alextwoods: This value is set by the checksum plug-in and we decided response context was how we were going to make this information available in V3 when we implemented flexible checksums. 

I guess we could add an additional config option on the client for a similar callback? Any other ideas?

gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
mullermp: Didn't know this was on the API. Makes sense

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -31,6 +31,17 @@ def download(destination, options = {})
...
+        @on_checksum_validated = options[:on_checksum_validated]
+
+        if @on_checksum_validated && @params[:checksum_mode] != 'ENABLED'
+          raise ArgumentError, "You must set checksum_mode: 'ENABLED' " +
+            "when providing a on_checksum_validated callback"
+        end
+
+        if @on_checksum_validated && !@on_checksum_validated.respond_to?(:call)
+          raise ArgumentError, 'on_checksum_validated must be callable'
+        end
```
mullermp: Maybe validation can be a separate method.

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
mullermp: Does it make sense to have an API private (ruby public) method on the Checksum plugin that our plugins/implementations can use to check if http checksum is validated?

I guess the alternative is that we have a config option for on_checksum_validated that is called by the checksum middleware - then it wouldn't exist in high level abstraction and be called on each get_object (or part).

mullermp (approved):

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
alextwoods: After more discussion - added a `checksum_validated` method to the `Response` - this makes it easy for users to access (and is now better documented/discoverable as well).

alextwoods:
Note - I've marked as draft - do not merge yet.  Adding the `checksum_validated` method to `Response` requires bumping the min core version (or checking `responds_to`).

alextwoods:
I've reverted changes from: https://github.com/aws/aws-sdk-ruby/pull/2414/files#diff-1f4566a292174ef034f378ecad42c5f33a59db0a2f795be6de866e5cc9f44996 
That change removed the s3 specific min core version.  This change now sets the s3 min core version.

mullermp (approved):

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -31,6 +31,10 @@ def download(destination, options = {})
           key: options[:key],
         }
         @params[:version_id] = options[:version_id] if options[:version_id]
+        @params[:checksum_mode] = options[:checksum_mode] || 'ENABLED'
```
mjankowski, line 34: I think this line has the side effect of making every request from `FileDownloader` default to enabling the checksum_mode=ENABLED feature at the client/api level, which was previously not enabled, even if the new feature here (on_checksum_validated callbacks) is not being used.

Was that an intentional change here?

It also doesn't seem possible to disable this, because setting this to falsey values (nil, false, etc) will have it get set by this line; and trying to set it to something like `DISABLED` or `FALSE` will throw up an error from elsewhere, because `ENABLED` is the only valid value at the api level.

alextwoods: This was an intentional change here to set the default to ENABLED as a best practice.  Checksums will only be validated when they exist on the object (if the object was uploaded without checksums, the download will succeed and nothing will be verified).

However - you are right - given the lack of support on the API for a `disabled` or `false` value, its impossible to disable this.  I'll release a patch today.

PR #2890, opened 2023-08-07
Update checksums.rb

PatoSoft:
Hi!

Thank you for your great job!

Headers are missing :complete_multipart_upload so I need to calculate it outside the plugin. I don't know if this is intended but it is the oposite of what I understand by ready this file comments.

### DIFF
--- a/gems/aws-sdk-glacier/CHANGELOG.md
+++ b/gems/aws-sdk-glacier/CHANGELOG.md
@@ -1,13 +1,15 @@
 Unreleased Changes
 ------------------
 
+* Issue - Fix `complete_multipart_upload` to calculate a Tree Hash checksum. (#2890)
+
 1.54.0 (2023-07-11)
 ------------------
 
 * Feature - Code Generated Changes, see `./build_tools` or `aws-sdk-core`'s CHANGELOG.md for details.
 
 1.53.0 (2023-07-06)
-------------------
+------------------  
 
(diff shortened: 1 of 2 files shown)


mullermp:
Yep, that looks to be a mistake that's been around for a long time quite possibly.. Were you able to confirm that this change succeeds?

mullermp:
I'm not too sure about this change working. The plugin looks to be computing checksums on the request body. Compete Multipart Upload does not take a body here, so the calculated TreeHash is always the same digest.

mullermp:
Unfortunately I think you'll need to manage the TreeHash yourself - this looks to have been intended. I think you could either construct a TreeHash and read the entire file in 1 MB chunks like the documentation suggests and stick it at the end of CompleteMultipartUpload:

```
    #    tree_hash = TreeHash.new
    #    tree_hash.update(file.read(1024 * 1024)) until file.eof?
    #    tree_hash.digest
```

.. or for each UploadMultipartPart output's checksum, append it to an array and assign it to TreeHash's `hashes` accessor and then call `digest`.

I think we could possibly consider this as a feature request for creating an abstraction class for Glacier multipart uploads, similar to how we manage S3 uploads and download file through `Aws::S3::Object.upload_file(multipart_copy: true)`.

PatoSoft:
Thank you for the information!

I am using `Threehash::caculate_tree_hash(body)` at the moment.

It would be nice to have that` (multipart_copy: true) `feature for Glacier.

Thank you!

mullermp:
Created an issue in our repo to track this (closing this PR)

PR #2892, opened 2023-08-07
Fix INI parsing of blank properties

alextwoods:
Fixes an issue introduced in #2880

Nested sections *may* have whitespace in their first line.  However, if there is not a nested property on the next non-blank line, then the blank value should be treated as a property.

Example:
```
[profile blank-property]
aws_session_token =  # has a space, this should be treated as a blank property
s3 =  # also has a space - but is followed by a subsection
   region = ap-southeast-1
```

### DIFF
--- a/gems/aws-sdk-core/spec/aws/ini_parser_spec.rb
+++ b/gems/aws-sdk-core/spec/aws/ini_parser_spec.rb
@@ -38,6 +38,8 @@ s3 =
 aws_session_token = 
 s3 = 
    region = ap-southeast-1
+   blank_sub_property = 
+
       FILE
     }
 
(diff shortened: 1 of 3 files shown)


jterapin (approved):

gems/aws-sdk-core/spec/aws/ini_parser_spec.rb
```diff
@@ -34,34 +34,40 @@ module Aws
 s3 =
    endpoint_url = https://localhost:8000
 
+[profile blank-property]
+aws_session_token = 
+s3 = 
+   region = ap-southeast-1
```
mullermp, line 40: Does this also work with a blank property inside s3? Maybe we should add one here.

mullermp (approved):

gems/aws-sdk-core/spec/aws/ini_parser_spec.rb
alextwoods: It does!  Added one here + expect in the test.

Issue #2901, opened 2023-08-17
Object#download_file Add Progress Callback


foxtrod:
### Describe the feature

Progress Callback for Object#download_file

Introduce a built-in mechanism within the Object#download_file that provides real-time feedback on the download progress just like in a [Object#upload_file](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/Object.html#upload_file-instance_method)

### Use Case

Seeing is Believing.

Currently, to get any sense of progress, I was left either diving deep into the AWS client logger or constantly pinging the file size, I've picked the second option and wondering how users of the gem are handling the progress (let me know in the comments). The Object#upload_file method has this awesome progress feedback, so why shouldn't Object#download_file have the same?

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### SDK version used

3.180.3

### Environment details (OS name and version, etc.)

MacOs Ventura 13.3

foxtrod:
Would be glad if i missed existing functionality that covers download progress, but if someone wonders I'm using next workaround:

```
download_thread = Thread.new do
  s3_object.download_file(destination, { thread_count: threads_count, mode: "get_range", chunk_size: chunk_size })
end

progress_thread = Thread.new do
  begin
    while download_thread.alive? # ping file size until download processing
      current_size = File.exist?(destination) ? File.size(destination) : 0
      progress_percentage = (current_size.to_f / expected_size) * 100

      puts "Downloaded: #{current_size} / #{expected_size} (#{progress_percentage.round(2)}%)"

      break if current_size >= expected_size

      sleep 1 # wait for a second before checking again
    end
  rescue => e
    logger.error(event: :progress_thread, error: e)
  end
end

download_thread.join 
progress_thread.join
```

jterapin:
Resolved in #2902.

PR #2902, opened 2023-08-18
Add support for progress_callback in Object#download_file

### LINKED ISSUE #2681: Missing documentation for some S3 Object methods
**Describe the issue with documentation**
There are some methods that are missing documentation for their optional parameters. I haven't used all methods in the `Aws::S3::Object` class so I can't talk for all of them, but for example `upload_file` only reflects 3 optional parameters `multipart_threshold`, `thread_count`, and `progress_callback` but the reality is that it uses multipart upload code and for that reason it inherits all of the options available in that method too. The same thing happens with the  `copy_to` and `copy_from` methods.

I hope I made the point clear if not I can elaborate further.

Thanks,
Alberto

alextwoods:
Implements #2901 
Closes #2681 

This adds support for a `progress_callback` option in `Object#download_file`, similar to what is supported in `upload_file`.   

This also improves the performance of `download_file` when multiple threads are used - previously it was done in batches, requiring all threads in the batch to complete before moving onto another batch (and additionally, creating a new set of threads each time instead of re-using one set of worker threads).  This follows the approach used in the MultipartUploader.

### DIFF
--- a/gems/aws-sdk-core/lib/seahorse/client/plugins/request_callback.rb
+++ b/gems/aws-sdk-core/lib/seahorse/client/plugins/request_callback.rb
@@ -60,6 +60,16 @@ the number of bytes read from the body, and the total number of
 bytes in the body.
           DOCS
 
+        option(:on_chunk_received,
+          default: nil,
+          doc_type: 'Proc',
+          docstring: <<-DOCS)
+When a Proc object is provided, it will be used as callback when each chunk 
+of the response body is received. It provides three arguments: the chunk,
+the number of bytes received, and the total number of 
+bytes in the response (or nil if the server did not send a `content-length`).
+        DOCS
+
         # @api private
         class OptionHandler < Client::Handler
           def call(context)
@@ -68,7 +78,28 @@ bytes in the body.
             end
             on_chunk_sent = context.config.on_chunk_sent if on_chunk_sent.nil?
             context[:on_chunk_sent] = on_chunk_sent if on_chunk_sent
+
+            if context.params.is_a?(Hash) && context.params[:on_chunk_received]
+              on_chunk_received = context.params.delete(:on_chunk_received)
+            end
+            on_chunk_received = context.config.on_chunk_received if on_chunk_received.nil?
+
+            add_response_events(on_chunk_received, context) if on_chunk_received
+
             @handler.call(context)
+          end
+
+          def add_response_events(on_chunk_received, context)
+            shared_data = {bytes_received: 0}
+
+            context.http_response.on_headers do |_status, headers|
+              shared_data[:content_length] = headers['content-length']&.to_i
+            end
+
+            context.http_response.on_data do |chunk|
+              shared_data[:bytes_received] += chunk.bytesize if chunk && chunk.respond_to?(:bytesize)
+              on_chunk_received.call(chunk, shared_data[:bytes_received], shared_data[:content_length])
+            end
           end
         end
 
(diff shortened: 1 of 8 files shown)


gems/aws-sdk-core/lib/seahorse/client/plugins/request_callback.rb
```diff
@@ -60,6 +60,16 @@ class RequestCallback < Plugin
 bytes in the body.
           DOCS
 
+        option(:on_chunk_received,
```
mullermp, line 63: In the future I think maybe these shouldn't be on config and definitely should not be part of params (request options instead). I know you did it here to follow pattern.

gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
```diff
@@ -505,6 +514,13 @@ def upload_file(source, options = {})
       #   response.  For multipart downloads, this will be called for each
       #   part that is downloaded and validated.
       #
+      # @option options [Proc] :progress_callback
```
mullermp, line 540: I think we have an open issue in the repo for adding extra documentation to these s3 custom methods. Do you think we could just take care of that while we are adding options here? I think there's a yard trait for expanding options. `@param thing (see some_ref)`. We should check that.

gems/aws-sdk-s3/spec/object/download_file_spec.rb
```diff
@@ -172,8 +172,10 @@ module S3
         end
 
         it 'raises an error when checksum validation fails on multipart' do
```
mullermp, line 220: I think this or a different test in this file is flakey with jruby. We should investigate..

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -178,6 +215,59 @@ def single_request
...
+          @progress_callback.call([bytes_read], [total_size])
+        end
+      end
+
+      class Part < Struct.new(:part_number, :size, :params)
+        include Aws::Structure
+      end
+
+      # @api private
+      class PartList
```
mullermp, line 230: Does it make sense to reuse this and multipart progress for upload instead of duplication?

mullermp (approved):
Seems fine

gems/aws-sdk-core/lib/seahorse/client/plugins/request_callback.rb
alextwoods: Yeah - in the future I think we would have these as request options.

gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
alextwoods: I assume you mean: https://github.com/aws/aws-sdk-ruby/issues/2681
I'll take a look.

gems/aws-sdk-s3/spec/object/download_file_spec.rb
alextwoods: I think it may be in the upload stream spec and not this one, at least based on this failure: https://github.com/aws/aws-sdk-ruby/actions/runs/5754207155/job/15599029463

We should take a separate task to update that.

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
alextwoods: I had tried to refactor out and re-use the same classes, but overall, there are a few key differences (most driven by the different Part types/needs.... which mostly come from the fact that we may not know each part size ahead of time) and those differences made it messy and complicated to make more generic.

mullermp: That's fine, they can be duplicated

mullermp (approved):

PR #2917, opened 2023-09-26
Simple attributes support for conditional check errors

mullermp:
Simple attributes support for conditional check errors. When configured, it will return item data, preventing another get item call. The item data should be simplified if configured to do so. This is a breaking change for existing users however the feature is fairly new.

Will investigate test failures


gems/aws-sdk-dynamodb/lib/aws-sdk-dynamodb/plugins/simple_attributes.rb
```diff
@@ -116,9 +116,16 @@ class Handler < Seahorse::Client::Handler
 
           def call(context)
             context.params = translate_input(context)
-            @handler.call(context).on(200) do |response|
+            resp = @handler.call(context)
+            resp.on(200) do |response|
               response.data = translate_output(response)
             end
+            resp.on(400) do |response|
+              if response.error.is_a?(Errors::ConditionalCheckFailedException)
```
alextwoods: What about other future error cases? Maybe we should check if the response has an item or item list?

gems/aws-sdk-dynamodb/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Feature - (Breaking change) Simple attributes support for item data returned in `ConditionalCheckFailedException`.
```
alextwoods: I'd potentially consider this a big fix?

gems/aws-sdk-core/lib/seahorse/client/plugins/raise_response_errors.rb
```diff
@@ -21,7 +21,7 @@ def call(context)
 
         def add_handlers(handlers, config)
           if config.raise_response_errors
-            handlers.add(Handler, step: :validate, priority: 95)
+            handlers.add(Handler, step: :initialize, priority: 95)
```
alextwoods: I do think this is the right thing to do - but what other handlers were getting skipped before? Any that might cause issues?

alextwoods (approved):
Looks good!

gems/aws-sdk-dynamodb/CHANGELOG.md
mullermp: I am not sure I entirely agree but what should I do with it? I think feature (minor bump) is preferred here.

alextwoods: Yeah - nvm, I think minor bump is probably correct.

gems/aws-sdk-core/lib/seahorse/client/plugins/raise_response_errors.rb
mullermp: There is an issue with client side metrics because it rescues and re-raises errors, and it relies on this being where it's at. I found a solution that doesn't involve changing this, though I agree, we should ideally have this as late as possible.

gems/aws-sdk-dynamodb/lib/aws-sdk-dynamodb/plugins/simple_attributes.rb
mullermp: I found a general solution.

alextwoods (approved):
Nice - LGTM!

Issue #2913, opened 2023-09-22
SQS QueuePoller sometimes returns dupe messages, raises BatchEntryIdsNotDistinct


alextwoods:
### Discussed in https://github.com/aws/aws-sdk-ruby/discussions/2908

<div type='discussions-op-text'>

<sup>Originally posted by **jdelStrother** September  6, 2023</sup>
We use Aws::SQS::QueuePoller like so:

```ruby
poller = Aws::SQS::QueuePoller.new(queue_url, idle_timeout: 0, wait_time_seconds: 1)
poller.poll(max_number_of_messages: 10) do |batch|
  do_work(batch)
end
```

Very rarely - say, every 100k requests - `batch` contains duplicate messages, with the same message_id & body, but different receipt_handles.

When QueuePoller tries to clean up these messages (after `do_work` is complete), it calls delete_message_batch which raises `Aws::SQS::Errors::BatchEntryIdsNotDistinct`.

I've not discovered this behaviour documented anywhere (that SQS might occasionally return duplicate messages when calling ReceiveMessage). It's presumably not considered a bug in SQS itself (..?) so should the SDK fix it by de-duping in `Client#receive_message` or `QueuePoller#receive_messages` ? Or is this considered an end-user problem to deal with?</div>

PR #2918, opened 2023-09-28
Deduplicate message batches in SQS Poller

jterapin:
Fixes an issue mentioned in [#2908](https://github.com/aws/aws-sdk-ruby/issues/2913)

This PR ensures that there's no duplicate messages before yield. To successfully delete a message, it must have the "most recently receipt handle" as mentioned [here](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-queue-message-identifiers.html) under Receipt handle.

Also, ran rubocop to keep the files up to today's standards.

---

To make sure we include your contribution in the release notes, please make sure to add description entry for your changes in the "unreleased changes" section of the CHANGELOG.md file (at corresponding gem). For the description entry, please make sure it lives in one line and starts with Feature or Issue in the correct format.

For generated code changes, please checkout below instructions first:

### DIFF
--- a/gems/aws-sdk-sqs/CHANGELOG.md
+++ b/gems/aws-sdk-sqs/CHANGELOG.md
@@ -1,7 +1,8 @@
 Unreleased Changes
 ------------------
 
-* Issue - Update QueuePoller to handle duplicate messages so that these can be deleted successfully.
+* Feature - Update QueuePoller to handle duplicate messages before yielding. 
+* Issue - Fix message batch deletion issue in (#2913) by deduping the messages.
 
 
 1.63.0 (2023-09-27)
--- a/gems/aws-sdk-sqs/CHANGELOG.md
+++ b/gems/aws-sdk-sqs/CHANGELOG.md
@@ -1,6 +1,9 @@
 Unreleased Changes
 ------------------
 
+* Issue - Update QueuePoller to handle deduplicate messages so that those messages can be deleted successfully.
+
+
 1.63.0 (2023-09-27)
 ------------------
 
--- a/gems/aws-sdk-sqs/CHANGELOG.md
+++ b/gems/aws-sdk-sqs/CHANGELOG.md
@@ -1,9 +1,7 @@
 Unreleased Changes
 ------------------
 
-* Feature - Update QueuePoller to handle duplicate messages before yielding. 
-* Issue - Fix message batch deletion issue in (#2913) by deduping the messages.
-
+* Feature - Update QueuePoller to handle duplicate messages before yielding. Fixes bug in (#2913).
 
 1.63.0 (2023-09-27)
 ------------------
--- a/gems/aws-sdk-sqs/CHANGELOG.md
+++ b/gems/aws-sdk-sqs/CHANGELOG.md
@@ -1,7 +1,7 @@
 Unreleased Changes
 ------------------
 
-* Issue - Update QueuePoller to handle deduplicate messages so that those messages can be deleted successfully.
+* Issue - Update QueuePoller to handle duplicate messages so that these can be deleted successfully.
 
 
 1.63.0 (2023-09-27)
--- a/gems/aws-sdk-sqs/lib/aws-sdk-sqs/queue_poller.rb
+++ b/gems/aws-sdk-sqs/lib/aws-sdk-sqs/queue_poller.rb
@@ -4,7 +4,6 @@ require 'set'
 
 module Aws
   module SQS
-
     # A utility class for long polling messages in a loop. **Messages are
     # automatically deleted from the queue at the end of the given block.**
     #
@@ -203,7 +202,6 @@ module Aws
     #   ```
     #
     class QueuePoller
-
       # @param [String] queue_url
       # @option options [Client] :client
       # @option (see #poll)
@@ -349,10 +347,10 @@ module Aws
       # @param [Integer] seconds
       def change_message_visibility_timeout(message, seconds)
         @client.change_message_visibility({
-          queue_url: @queue_url,
-          receipt_handle: message.receipt_handle,
-          visibility_timeout: seconds,
-        })
+                                            queue_url: @queue_url,
+                                            receipt_handle: message.receipt_handle,
+                                            visibility_timeout: seconds
+                                          })
       end
 
       # @note This method should be called from inside a {#poll} block.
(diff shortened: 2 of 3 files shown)


gems/aws-sdk-sqs/CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 Unreleased Changes
 ------------------
 
+* Issue - Update QueuePoller to handle duplicate messages so that these can be deleted successfully.
```
alextwoods: I think we may want to consider this a feature and minor version bump.  Also reference the issue it fixes like `(#2913)`.

gems/aws-sdk-sqs/lib/aws-sdk-sqs/queue_poller.rb
```diff
@@ -349,20 +347,20 @@ def poll(options = {}, &block)
       # @param [Integer] seconds
       def change_message_visibility_timeout(message, seconds)
         @client.change_message_visibility({
-          queue_url: @queue_url,
-          receipt_handle: message.receipt_handle,
-          visibility_timeout: seconds,
-        })
+                                            queue_url: @queue_url,
```
alextwoods: I know rubocop auto fixes these - but I think its more readable (and the style we use elsewhere in the V3 sdk) to keep the previous indention in these cases.

gems/aws-sdk-sqs/lib/aws-sdk-sqs/queue_poller.rb
```diff
@@ -392,17 +390,18 @@ def send_request(config)
         @client.receive_message(params)
       end
 
-      def check_idle_timeout(config, stats, messages)
-        if config.idle_timeout
-          since = stats.last_message_received_at || stats.polling_started_at
-          idle_time = Time.now - since
-          throw :stop_polling if idle_time > config.idle_timeout
-        end
+      def check_idle_timeout(config, stats, _messages)
```
alextwoods: Since this is a private method - I think we can just remove this parameter

gems/aws-sdk-sqs/lib/aws-sdk-sqs/queue_poller.rb
```diff
@@ -419,9 +418,19 @@ def yield_messages(config, messages, stats, &block)
...
 
+      def dedupe_messages(messages)
+        u_messages = {}
+        messages.each do |msg|
+          # duplicate messages will have a different receipt handle
+          # so must provide the most recently receipt handle
+          # when deleting the message
+          u_messages[msg.message_id] = msg
+        end
+        u_messages
```
alextwoods: I'd move the `values` call to here.

gems/aws-sdk-sqs/lib/aws-sdk-sqs/queue_poller.rb
```diff
@@ -419,9 +418,19 @@ def yield_messages(config, messages, stats, &block)
         end
       end
 
+      def dedupe_messages(messages)
+        u_messages = {}
+        messages.each do |msg|
```
alextwoods: Not to play too much golf, but I think you could shorten this slightly by using `inject({})`.

gems/aws-sdk-sqs/lib/aws-sdk-sqs/queue_poller.rb
```diff
@@ -444,27 +453,25 @@ def initialize
...
       class PollerConfig
-
         # @api private
-        CONFIG_OPTIONS = Set.new([
-          :idle_timeout,
-          :skip_delete,
-          :before_request,
-        ])
+        CONFIG_OPTIONS = Set.new(%i[
+                                   idle_timeout
```
alextwoods: Ditto on spacing here - I do like the %i, but in general the SDK has used the less indented style

alextwoods (commented):
Nice - approach looks good!

gems/aws-sdk-sqs/lib/aws-sdk-sqs/queue_poller.rb
jterapin: I'll update this and other ones that I see that looks really out of place.

mullermp: I think rubocop is doing this because of the curly braces. If those are removed, auto formatting should work.

gems/aws-sdk-sqs/CHANGELOG.md
mullermp: Either is fine IMO. Arguably it's a bug.

gems/aws-sdk-sqs/lib/aws-sdk-sqs/queue_poller.rb
mullermp: Double check it's not called anywhere else...

gems/aws-sdk-sqs/lib/aws-sdk-sqs/queue_poller.rb
mullermp: I think the recommendation is `each_with_object` and not inject (at least rubocop might say that..)

gems/aws-sdk-sqs/lib/aws-sdk-sqs/queue_poller.rb
mullermp: I've found rubocop will preserve this spacing not just with curly brace but also the parenthesis. So I believe it will respect something like

```
CONFIG_OPTIONS = Set.new [
  thing
  thing
]
```

mullermp (commented):
Approach looks good... looks like rubocop screwed up most things.

gems/aws-sdk-sqs/lib/aws-sdk-sqs/queue_poller.rb
jterapin: I played around using `reverse.uniq.reverse!` on the messages. 
It unfortunately won't work since the duplicated messages with have different `receipt_handles` and the `uniq` will classifies these duplicates as separate messages. 

Going with `each_with_object`

jterapin: Scratch my above comment - it does work with block being passed in with .uniq - thanks Matt!

gems/aws-sdk-sqs/lib/aws-sdk-sqs/queue_poller.rb
```diff
@@ -392,17 +390,22 @@ def send_request(config)
...
       end
 
       def process_messages(config, stats, messages, &block)
         stats.received_message_count += messages.count
         stats.last_message_received_at = Time.now
+
+        # duplicated messages will have a different receipt handle
+        # so we need to provide the most recent receipt to
+        # delete a batch - thus, filtering below by message_id
+        messages = messages.reverse.uniq(&:message_id).reverse!
```
alextwoods, line 408: Nice!

alextwoods (approved):
LGTM!

mullermp (approved):

PR #2920, opened 2023-09-30
Links DAX acronym in aws-sdk-dax documentation to product page

alvincrespo:
This PR links the yardoc reference of `DAX` to the product offering to help clarify which service this gem is associated to. I was originally trying to look at the source for `aws-sdk-ecs` but landed here by mistake and was curious was DAX meant, so I looked it up and thought this might help others in the future.

### DIFF
--- a/gems/aws-sdk-dax/CHANGELOG.md
+++ b/gems/aws-sdk-dax/CHANGELOG.md
@@ -1,6 +1,11 @@
 Unreleased Changes
 ------------------
 
+1.48.1 (2023-09-30)
+------------------
+
+* Documentation - Added product offering link in the client.
+
 1.48.0 (2023-09-27)
 ------------------
 
--- a/gems/aws-sdk-dax/lib/aws-sdk-dax/client.rb
+++ b/gems/aws-sdk-dax/lib/aws-sdk-dax/client.rb
@@ -37,7 +37,7 @@ require 'aws-sdk-core/plugins/protocols/json_rpc.rb'
 Aws::Plugins::GlobalConfiguration.add_identifier(:dax)
 
 module Aws::DAX
-  # An API client for DAX.  To construct a client, you need to configure a `:region` and `:credentials`.
+  # An API client for [DAX](https://aws.amazon.com/dynamodb/dax/).  To construct a client, you need to configure a `:region` and `:credentials`.
   #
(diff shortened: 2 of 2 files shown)


gems/aws-sdk-dax/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 Unreleased Changes
 ------------------
 
+1.48.1 (2023-09-30)
+------------------
+
+* Documentation - Added product offering link in the client.
+
```
alvincrespo, line 8: Not sure if this is right, but the template requested I fill this out. Let me know if this isn't the right place for this and I'll update. Thank you!

mullermp:
This change will not work because it edits a code generated file. I can take this as a feature request to improve documentation for the service.

alvincrespo:
@mullermp I'm more than happy to help out here. What file requires the change? I looked at the json for this gem but I was unsure what should be updated.

alvincrespo:
Ok. Spent some time this morning going over how the client class gets generated and found that `{{service_name}}` in `build_tools/aws-sdk-code-generator/templates/client_class.mustache` is derived by the manifest, or `services.json`. It uses the key name of the service, in this case "DAX", to populate the template because `service_name` is set to the key passed in at `BuildTools::ServiceEnumerator#build_service`. 

@mullermp So this would basically require a possible new key, or using an existing key, in `apis/dax/2017-04-19/api-2.json` that would have a link to the service in order to populate the yardoc correctly. I can take a crack at this, any thoughts on approach here? Use an existing key? Add a new key?

mullermp:
The APIs and documentation are distributed to SDK teams to code generate against. I just checked and there is no such JSON metadata key for adding service level documentation. In a future version of this SDK, the clients will be generated using Smithy (a new modeling language) and would include top level service documentation ([JS v3 example for DAX](https://github.com/aws/aws-sdk-js-v3/blob/fa961c2350b650a1c29cc09306f0e2e0b9eb6c4e/codegen/sdk-codegen/aws-models/dax.json#L112)). I appreciate the sentiment but I don't think we can improve this here without special casing documentation for each service.

alvincrespo:
@mullermp That makes a ton of sense. Thanks for your input here! TIL about Smithy so I'll check that out. Going to close this.

PR #2921, opened 2023-10-02
Smoke tests v2

mullermp:
Actual smoke test models are not available yet, but this generates a sample smoke test on version 2. 

Refactor smoke tests to use shared definitions in aws-sdk-core.

Also improves the testing scripts (smoke tests are moved to a separate rake task, outside of integrations).

Only changed aws-sdk-acm for now.


mullermp:
Reminder for myself to add core changelog and bump minimum core version

build_tools/services.rb
```diff
@@ -10,10 +10,10 @@ class ServiceEnumerator
     MANIFEST_PATH = File.expand_path('../../services.json', __FILE__)
 
     # Minimum `aws-sdk-core` version for new gem builds
-    MINIMUM_CORE_VERSION = "3.188.0"
+    MINIMUM_CORE_VERSION = "3.190.1"
```
alextwoods: Do we need to bump this since smoke tests/features are not part of the built gems?

mullermp: No, but they will bump anyway....

tasks/test.rake
```diff
@@ -42,9 +42,51 @@ task 'test:spec' do
...
+  tags = "-t 'not @smoke' -t 'not @veryslow'"
+  sh("bundle exec cucumber --retry 3 #{tags} -r #{dir} #{dir} --publish-quiet")
+end
+
+desc 'Executes all integration tests'
+task 'test:integration' do
+  failures = []
+  Dir.glob('gems/*/features').each do |dir|
+    gem_name = dir.match(%r{gems/(.*)/features})[1]
+    sh("bundle exec rake test:integration:#{gem_name}") do |ok, _|
```
alextwoods, line 86: I'm wondering if we should change the format used when executing all of the tests.... It tends to produce so much output that it can be hard to search for the failures.

Maybe we shoudl run with `--format summary` or something like that? 

Or - long term, most test runners (eg: codebuild) have some junit integration and give nice displays - maybe we should be outputing junit format.  

Nothing we need to do now....

tasks/test.rake
```diff
@@ -42,9 +42,51 @@ task 'test:spec' do
...
+  ENV.fetch('AWS_SMOKE_TEST_SKIP_TAGS', '').split(',').each do |tag|
+    tags += " -t 'not @#{tag}'"
+  end
+  sh("bundle exec cucumber --retry 3 #{tags} -r #{dir} #{dir} --publish-quiet")
+end
+
+desc 'Executes all smoke tests'
+rule 'test:smoke' do
+  failures = []
+  Dir.glob('gems/*/features').each do |dir|
```
alextwoods, line 61: I think we should skip the gem if the features directory doesn't contain any *.feature files.  It ends up adding a lot of time to the total to spin up a new process for every service just to run 0 tests for it.

mullermp: Yeah, that's a good idea, done

tasks/test.rake
mullermp: I don't think we should do summary because it relies on color and some build logs like in ToD do not support that. I think the current format we have is fine and I think with the other suggestion you had, it is a lot more manageable.

alextwoods (approved):

PR #2923, opened 2023-10-06
Support Auth trait for Sigv4a

mullermp:
Implement `auth` trait support, which also supports sigv4a.


mullermp:
Reminder to bump minimum core version when merging!

build_tools/services.rb
```diff
@@ -151,6 +151,13 @@ def gem_dependencies(api, dependencies)
       core_version_string = "', '>= #{min_core}"
       dependencies['aws-sdk-core'] = "~> #{version_file.split('.')[0]}#{core_version_string}"
 
+      api['metadata'].fetch('auth', []).each do |auth|
+        if %w[aws.auth#sigv4 aws.auth#sigv4a].include?(auth)
+          dependencies['aws-sigv4'] = '~> 1.1'
```
alextwoods: For sigv4a, should we set the min dependency to ~> 1.5? (I think thats where we added sigva support).

gems/aws-sdk-core/lib/aws-sdk-core/endpoints.rb
```diff
@@ -33,8 +33,60 @@ def resolve_auth_scheme(context, endpoint)
...
+
+      def sigv4_name(config)
+        config.api.metadata['signingName'] ||
+          config.api.metadata['endpointPrefix']
+      end
+
       def default_auth_scheme(context)
-        case default_api_authtype(context)
+        if (auth_list = default_api_auth(context))
+          auth = auth_list.first
```
alextwoods: Should this loop through the auth to check for supported ones as we do w/ `resolve_auth_scheme`

gems/aws-sdk-core/lib/aws-sdk-core/plugins/regional_endpoint.rb
```diff
@@ -81,6 +99,13 @@ def resolve_region(cfg)
           env_region || cfg_region
         end
 
+        def resolve_sigv4a_signing_region_set(cfg)
+          value = ENV['AWS_SIGV4A_SIGNING_REGION_SET']
+          value ||= Aws.shared_config.sigv4a_signing_region_set(profile: cfg.profile)
+          value ||= cfg.region || ''
```
alextwoods: I think this shouldn't default to cfg.region - I think the config should be an override and most of the time should be unset.

gems/aws-sdk-core/lib/aws-sdk-core/endpoints.rb
```diff
@@ -33,8 +33,60 @@ def resolve_auth_scheme(context, endpoint)
 
       private
 
+      def merge_signing_defaults(auth_scheme, config)
+        if %w[sigv4 sigv4a].include?(auth_scheme['name'])
+          auth_scheme['signingName'] ||= sigv4_name(config)
+          if auth_scheme['name'] == 'sigv4a'
+            auth_scheme['signingRegionSet'] ||= config.sigv4a_signing_region_set
```
alextwoods: I think this order should be different and the configured signing_region_set should act as an override.  Someething like `config.sigv4a_signing_region_set || auth_scheme['signingRegionSet'] || cfg.region`

gems/aws-sdk-core/lib/aws-sdk-core/plugins/sign.rb
```diff
@@ -97,7 +97,7 @@ def initialize(auth_scheme, config, region_override = nil)
           end
 
           region = if scheme_name == 'sigv4a'
-                     auth_scheme['signingRegionSet'].first
+                     auth_scheme['signingRegionSet'].join(',')
```
alextwoods, line 105: This should work with the sigv4 signer, but its a little confusing.  Maybe we should at least update the docs in the sigv4 signer to specify that for sigv4a region is the regionSet as a comma seperated list.

gems/aws-sdk-core/spec/fixtures/schema/validates_using_the_schema/api.json
```diff
@@ -1 +0,0 @@
-{}
```
alextwoods, line 1: Why are these being deleted, do we not use them anymore?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/regional_endpoint.rb
mullermp: Yes. This is old code.

gems/aws-sdk-core/lib/aws-sdk-core/endpoints.rb
mullermp: Yes. This is old code.

gems/aws-sdk-core/spec/fixtures/schema/validates_using_the_schema/api.json
mullermp: They're not.

build_tools/services.rb
mullermp: Yes

gems/aws-sdk-core/lib/aws-sdk-core/endpoints.rb
mullermp: Yeah, I think so - for the sigv4a first / fallback case

alextwoods (approved):

PR #2924, opened 2023-10-09
Support disabling IMDS v1

mullermp:
Allows disabling of IMDS v1 using config.

Also removes refreshing from set? - IMDS is called more than 3 times when checking set credentials.


mullermp:
Pending manual test on EC2

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
```diff
@@ -227,6 +222,35 @@ def get_credentials
...
+        end
+      end
+    rescue *NETWORK_ERRORS
+      # token attempt failed, reset token
+      # fallback to non-token mode
+      @token = nil
+      @imds_v1_fallback = true
+    end
+
+    def _get_credentials(conn, token)
```
alextwoods, line 242: Maybe add comment that token is optional, and if missing, will use insecure v1 flow.

alextwoods (approved):

PR #2929, opened 2023-10-13
Adding callback for after_empty_receive

bmalinconico:
I would like to add a callback for when an empty receive is encountered. I'm looking to use this callback to indicate that the worker node is preferred for termination during a scaling event.  Right now I'm detecting this using the before_request callback but it is convoluted.

I've added the code, and after I get a tentative 👍 for the feature I'll add documentation, and correct any changes you would like me to make.

I've named the callback after the SQS EmptyReceives metric measuring the same thing.


mullermp:
Thanks for opening a pull request. Can you please share how you're doing this with `before_request`?

bmalinconico:
@mullermp Thanks for the review.

Pesudocode for what I'm using right now would be:

```

poller.before_request do 
  empty_request_detected unless received_work

  received_work = false
end

poller.poll do |msgs|
  received_work = true 
end
```

It takes two empty receives to detect it using this method. First empty receive marks the received_work as false and a second empty receive will actually trigger it. Using the default 20s long polling configuration, it takes 40s to detect the worker is idle which is less than ideal.

mullermp:
I think this makes sense to do. If you could add a CHANGELOG entry to aws-sdk-sqs and add documentation, that would be great.

mullermp (approved):

bmalinconico:
Yep, I'll do that and add some documentation when I'm back in town next week

bmalinconico:
@mullermp Documentation updated & CHANGELOG entry added

gems/aws-sdk-sqs/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Feature - Update QueuePoller allow for the definition of an after_empty_receive callback which is triggered when a polling request returns no records.
```
mullermp: Please fix with:

```
* Feature - Update `QueuePoller` to allow for the definition of an `after_empty_receive` callback which is triggered when a polling request returns no records.
```

gems/aws-sdk-sqs/lib/aws-sdk-sqs/queue_poller.rb
```diff
@@ -180,6 +180,16 @@ module SQS
     #   end
     #   ```
     #
+    # * Configure a {#after_empty_receive} callback.
```
mullermp: a => an

mullermp (approved):
Thank you!

mullermp:

PR #2932, opened 2023-10-17
Credential scope support

mullermp:
Adds a credential scope config option that is passed to endpoint rules (no rules use it yet).

Pending endpoint tests

Pending plumbing through credential providers when present (like STS, SSO, IAM, IMDS, etc)

Test cases are loosely covered already through existing testing mechanisms.


build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/endpoints_module.rb
```diff
@@ -134,6 +134,8 @@ def built_in_client_context_param_value(param_data)
           else
             'context.config.use_dualstack_endpoint'
           end
+        when 'AWS::Auth::CredentialScope'
```
alextwoods, line 133: I think we may also need updates to the endpoint_spec_class (eg: https://github.com/aws/aws-sdk-ruby/blob/version-3/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/spec/endpoint_provider_spec_class.rb#L149)

mullermp: Yeah. I was going to wait for the actual rules but I can add it now.

mullermp: This change was premature

gems/aws-sdk-core/lib/aws-sdk-core/plugins/credentials_configuration.rb
```diff
@@ -57,13 +59,15 @@ class CredentialsConfiguration < Seahorse::Client::Plugin
 locations will be searched for credentials:
 
 * `Aws.config[:credentials]`
-* The `:access_key_id`, `:secret_access_key`, and `:session_token` options.
-* ENV['AWS_ACCESS_KEY_ID'], ENV['AWS_SECRET_ACCESS_KEY']
+* The `:access_key_id`, `:secret_access_key`, `:session_token`, and
+  `credential_scope` options.
```
mullermp: `:credential_scope`

mullermp (commented):
Reminder to self - this is blocked until providers can be updated.

PR #2933, opened 2023-10-23
(Static Stability) use provided expires_in in presigned url when credentials are expired

alextwoods:
See #2673 for original change.

### DIFF
--- a/gems/aws-sigv4/lib/aws-sigv4/signer.rb
+++ b/gems/aws-sigv4/lib/aws-sigv4/signer.rb
@@ -423,7 +423,7 @@ module Aws
         params['X-Amz-Algorithm'] = 'AWS4-HMAC-SHA256'
         params['X-Amz-Credential'] = credential(creds, date)
         params['X-Amz-Date'] = datetime
-        params['X-Amz-Expires'] = presigned_url_expiration(options, expiration).to_s
+        params['X-Amz-Expires'] = presigned_url_expiration(options, expiration, Time.strptime(datetime, "%Y%m%dT%H%M%S%Z")).to_s
         params['X-Amz-Security-Token'] = creds.session_token if creds.session_token
         params['X-Amz-SignedHeaders'] = signed_headers(headers)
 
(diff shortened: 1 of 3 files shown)


gems/aws-sigv4/lib/aws-sigv4/signer.rb
```diff
@@ -423,7 +423,7 @@ def presign_url(options)
         params['X-Amz-Algorithm'] = 'AWS4-HMAC-SHA256'
         params['X-Amz-Credential'] = credential(creds, date)
         params['X-Amz-Date'] = datetime
-        params['X-Amz-Expires'] = presigned_url_expiration(options, expiration).to_s
+        params['X-Amz-Expires'] = presigned_url_expiration(options, expiration, Time.strptime(datetime, "%Y%m%dT%H%M%S%Z")).to_s
```
mullermp, line 426: I can see passing in datetime, but why does it need to be formatted? It's used for arithmetic.

alextwoods: Thats was a bit confusing and I believe actually broken previously.  The datetime above is use for the `x-amz-date` header and must be the formatted string.  But it can be sourced from a few places, either the passed in header or options[:time] or fallback to Time.now (but in all of those cases it must be formatted as a string).  I'm now passing in that datetime so that the expire_at vs expiration is computed consistently when either the header or options[:time] is provided.

mullermp: I see. I think you can just pass datetime to Time.parse? You can possibly do that parsing inside the private method when calculating expiration seconds, and just pass datetime into the method here.

mullermp (approved):

gems/aws-sigv4/lib/aws-sigv4/signer.rb
alextwoods: I was trying to avoid branching in the private method on the class of datetime.  in CRT it is a Time.  In non-CRT it is a string.  The header needs to be the string.  And you can provide the header as an input and we will use that which complicates this code a bit.

I could use Time.parse but wanted to be explicit about the format.

PR #2936, opened 2023-10-24
Account id endpoint support

mullermp:
Support AccountId on credentials
Support AccountId specific endpoints and a new config option account_id_endpoint_mode.

Blocked on service rules (should be passed into rules)


build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/spec/endpoint_provider_spec_class.rb
```diff
@@ -154,6 +155,12 @@ def built_in_to_param(built_in, value)
               Param.new('use_fips_endpoint', value)
             when 'AWS::UseDualStack'
               Param.new('use_dualstack_endpoint', value)
+            when 'AWS::Auth::AccountId'
+              Param.new(
```
alextwoods, line 163: We probably practically don't need to care about this - but what if multiple Auth builtins are set (ie, accountId + credentialScope)

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/endpoints_module.rb
```diff
@@ -134,6 +130,8 @@ def built_in_client_context_param_value(param_data)
           else
             'context.config.use_dualstack_endpoint'
           end
+        when 'AWS::Auth::AccountId'
+          "context.config.account_id_endpoint_mode == 'disabled' ? nil : context.config.credentials.credentials.account_id"
```
alextwoods: Since this logic may end up pushed into endpoint rules anyway, this probably doesn't matter - but I assume here is where we would need to warn/raise for preferred/required when account id is not set. (or rather, generate the code to do so)

alextwoods (commented):
Nice - looks good.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/endpoints_module.rb
mullermp: If the logic is pushed to rules, yeah, all we do is pass in the endpoint mode. I don't think we care about warning (can discuss), and raising would be done by rules, right?

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/spec/endpoint_provider_spec_class.rb
mullermp: What do you mean? If both are set, they are both set as parameters. Do you mean passing in "credentials" and having something delegate to either of those?

alextwoods: If both are set then we would have two credentials params which would end up not working correctly.  At least as of right now, I think they would generally be mutually exclusive, so shouldn't matter...

mullermp: Talked offline, keeping as TODO when either feature is merged.

mullermp:
Moved to staging repo

PR #2937, opened 2023-10-27
Fix Querystring Builder

jterapin:
* Fixes an issue where Querystring Builder SHOULD support different types of shapes.
* Updated `rest-json.json` to cover most supported use-cases.
* Updated `querystring_builder.rb` file to today's rubocop standards

---

### DIFF
--- a/build_tools/aws-sdk-code-generator/spec/protocols/input/rest-json.json
+++ b/build_tools/aws-sdk-code-generator/spec/protocols/input/rest-json.json
@@ -218,7 +218,7 @@
     ]
   },
   {
-    "description": "Querystring of most supported types",
+    "description": "Querystring of supported scalar types",
     "metadata": {
       "protocol": "rest-json",
       "apiVersion": "2014-01-01"
--- a/build_tools/aws-sdk-code-generator/spec/protocols/input/rest-json.json
+++ b/build_tools/aws-sdk-code-generator/spec/protocols/input/rest-json.json
@@ -218,6 +218,95 @@
     ]
   },
   {
+    "description": "Querystring of most supported types",
+    "metadata": {
+      "protocol": "rest-json",
+      "apiVersion": "2014-01-01"
+    },
+    "shapes": {
+      "InputShape": {
+        "type": "structure",
+        "members": {
+          "StringQuery": {
+            "shape": "StringType",
+            "location": "querystring",
+            "locationName": "str"
+          },
+          "ByteQuery": {
+            "shape": "ByteType",
+            "location": "querystring",
+            "locationName": "byt"
+          },
+          "IntegerQuery": {
+            "shape": "IntegerType",
+            "location": "querystring",
+            "locationName": "int"
+          },
+          "LongQuery": {
+            "shape": "LongType",
+            "location": "querystring",
+            "locationName": "lon"
+          },
+          "FloatQuery": {
+            "shape": "FloatType",
+            "location": "querystring",
+            "locationName": "flo"
+          },
+          "DoubleQuery": {
+            "shape": "DoubleType",
+            "location": "querystring",
+            "locationName": "dou"
+          }
+        }
+      },
+      "StringType": {
+        "type": "string"
+      },
+      "ByteType": {
+        "type": "byte"
+      },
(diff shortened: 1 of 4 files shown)


jterapin:
Taking a final lookover tomorrow prior to submitting for review

build_tools/aws-sdk-code-generator/spec/protocols_spec.rb
```diff
@@ -106,7 +106,11 @@ def format_data(ref, src)
       params[key] = format_data(ref.shape.value, value)
     end
   when Seahorse::Model::Shapes::TimestampShape
-    Time.at(src)
+    if src.is_a?(String)
```
alextwoods: I'd lean towards using `case` here - and to preserve the previous behavior, doing `Time.at` on the else.  Something like:
```
case src
when String
  Time.parse(src)
else
  Time.at(src)
end  
```

gems/aws-sdk-core/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Fix QuerystringBuilder to handle different types.
```
alextwoods: May want to reference the protocol + mention list.

gems/aws-sdk-core/lib/aws-sdk-core/rest/request/querystring_builder.rb
```diff
@@ -71,6 +62,53 @@ def timestamp(ref, value)
...
+          else
+            raise NotImplementedError
+          end
+          "#{ref.location_name}=#{escape(value)}"
+        end
+
+        def generate_query_list(ref, values)
+          case ref.shape.member.shape
+          when TimestampShape
+            list_of_timestamps(ref, values)
```
alextwoods: I think rather than having a specific method for timestamps, you can use the generic `generate_query`.

I think if you do:
```
values.map { |value| generate_query(ref.shape.member, value) }
```

alextwoods: Edit: NVM, this doesn't quite work since the current `generate_query` does rely on the location_name.

As a different suggestion, how about a generic `query_value` method which takes the shape and value and returns just the correctly serialized value, eg:

```
def query_value(ref, value)
  case ref.shape
  when TimestampShape
     escape(timestamp(ref, value))
   when *SUPPORTED_TYPES
      escape(value.to_s)
    else
      raise NotImplementedError
     end
end
```

And then both single values and lists can use it I think.

build_tools/aws-sdk-code-generator/spec/protocols/input/rest-json.json
```diff
@@ -217,6 +217,95 @@
       }
     ]
   },
+  {
+    "description": "Querystring of most supported types",
```
alextwoods: maybe query string of supported scalar types

alextwoods (commented):
Nice - looks good!

gems/aws-sdk-core/lib/aws-sdk-core/rest/request/querystring_builder.rb
jterapin: Looks so much cleaner, thanks :)

gems/aws-sdk-core/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Fix QuerystringBuilder to support querystring lists of different types per rest-json protocol.
```
mullermp: Nit - changelog entries should be meaningful to customers. QuerystringBuilder is a private API, I would word this differently, like "Add query string support for lists of booleans, floats, integers, and timestamps in the rest json protocol" or something like that.

mullermp (approved):
Nice!

gems/aws-sdk-core/CHANGELOG.md
jterapin: I always have trouble figuring out how to "word" changelog entries and this is a good tip!!!

PR #2948, opened 2023-11-20
Default empty array for previously flattened lists and maps in query compatible services

### LINKED ISSUE #2947: aws-sdk-sqs raises NoMethodError when messages not found
### Describe the bug

Error occaisonally happens inside `QueuePoller#poll`.

The error message indicates `messages` is nil.

### Expected Behavior

No error happens.

### Current Behavior

```
 undefined method `empty?' for nil:NilClass
  /opt/rubies/ruby-2.7.8/lib/ruby/gems/2.7.0/gems/aws-sdk-sqs-1.67.0/lib/aws-sdk-sqs/queue_poller.rb:358:in `block (2 levels) in poll'
```

### Reproduction Steps

Needs more research

### Possible Solution

_No response_

The error starts on Nov 18, so AWS side might have changed at the timing.

aws-sdk-sqs-1.67.0

ruby-2.7.8 on Amazon Linux 2

alextwoods:
Fixes #2947 


gems/aws-sdk-sqs/lib/aws-sdk-sqs/queue_poller.rb
```diff
@@ -406,7 +406,7 @@ def delete_messages(messages)
 
       def get_messages(config, stats)
         config.before_request&.call(stats)
-        messages = send_request(config).messages
+        messages = send_request(config).messages || []
```
mullermp: I understand this as a temp fix, but shouldn't messages be parsed as an empty array already?

alextwoods: I don't think so - I think in aws json nil array members should be desrialized as nil and not default to `[]`.

mullermp: Then this is a breaking change to the service - `receive_message` API has a different return value, and breaks other usages.

mullermp: I added a plugin as a temporary solution? But I think when we see query compatible trait and also a list, we need to default it to an empty list.

alextwoods (commented):
Lgtm

jterapin (approved):

kstich (approved):

gems/aws-sdk-sqs/lib/aws-sdk-sqs/queue_poller.rb
geeksam: I see that this change was backed out, but nevertheless wish to note for future reference that, if one's style guide / cyborg law enforcement officers permit, the `Kernel#Array` method is a nice way to guarantee that one has an Array :)

PR #2954, opened 2023-11-22
Fix abort on exceptions

### LINKED ISSUE #2946: Errors downloading files from s3 are propagated to incorrect thread
### Describe the bug

While downloading files from S3, any exceptions are propagated to the main thread instead of the calling thread.

This appears to be due to the use of `Thread.abort_on_exception` (see https://github.com/aws/aws-sdk-ruby/blob/e0115affb56b42a08917e14950f730ae89c38183/gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb#L193 ).

### Expected Behavior

Exceptions propagate to the calling context

### Current Behavior

Exceptions are propagated to the main thread, interrupting unrelated code.

### Reproduction Steps

```
workers = []
workers << Thread.new do
  begin
    Aws::S3::Object.new(bucket_name: "test", key: "fails_to_download").download_file("outputfile")
  rescue
    puts "We should get here if the download failed, but do not."
  end
end
workers.each &:join

puts "We should get here regardless of the download failing or succeeding, but do not"
```

Because the error is propagated to the main thread instead of the calling thread, and the main thread has no exception handler, the process aborts unexpectedly.

### Possible Solution

Instead of using `abort_on_exception`, store the calling thread at the start of `download_in_threads` and use `calling_thread.raise(exception)` to propagate errors.

_No response_

aws-sdk-s3

3.2.2 / linux

jterapin:
Fixes #2946 

* Removing `abort_on_exception` for  `file_downloader`, `multipart_stream_uploader` and `multipart_file_uploader`
* Updated existing specs
* After reviews, planning to update outdated styling
---

### DIFF
--- a/gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
+++ b/gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
@@ -175,7 +175,6 @@ module Aws
               error
             end
           end
-          thread.abort_on_exception = true
           threads << thread
         end
         threads.map(&:value).compact
--- a/gems/aws-sdk-s3/spec/object/download_file_spec.rb
+++ b/gems/aws-sdk-s3/spec/object/download_file_spec.rb
@@ -231,24 +231,6 @@ module Aws
           end.to raise_error(Aws::Errors::ChecksumError)
         end
 
-        it 'does not interrupt the main thread when an exception occurs' do
-          client.stub_responses(
-            :get_object,
-            {
-              body: 'body',
-              checksum_sha1: 'invalid'
-            }
-          )
-
-          threads = []
-          threads << Thread.new do
-            expect { large_obj.download_file(path) }
-              .to raise_error(Aws::Errors::ChecksumError)
-          end
-
-          expect { threads.each(&:join) }.not_to raise_error
-        end
-
         it 'calls on_checksum_validated on single part' do
           callback_data = {called: 0}
           mutex = Mutex.new
--- a/gems/aws-sdk-s3/spec/object/download_file_spec.rb
+++ b/gems/aws-sdk-s3/spec/object/download_file_spec.rb
@@ -221,15 +221,32 @@ module Aws
         end
 
         it 'raises an error when checksum validation fails on multipart' do
-          client.stub_responses(:get_object, {body: 'body', checksum_sha1: 'invalid'})
+          client.stub_responses(:get_object, { body: 'body', checksum_sha1: 'invalid' })
 
           thread = double(value: nil)
-          allow(thread).to receive(:abort_on_exception=)
           expect(Thread).to receive(:new).and_yield.and_return(thread)
 
           expect do
             large_obj.download_file(path)
           end.to raise_error(Aws::Errors::ChecksumError)
+        end
+
+        it 'does not interrupt the main thread when an exception occurs' do
+          client.stub_responses(
+            :get_object,
+            {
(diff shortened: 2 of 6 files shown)


gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
```diff
@@ -175,7 +175,6 @@ def upload_in_threads(pending, completed, options)
               error
             end
           end
-          thread.abort_on_exception = true
```
jterapin, line 178: I was tracking down how errors are raised for both `multipart_file_uploader` and `multipart_stream_uploader`. Below is a breakdown for `multipart_file_uploader` but the stream uploader is pretty similar:
* Errors are rescued instead of raised: https://github.com/aws/aws-sdk-ruby/blob/2f0e6f83473c59d6f4875694ae8cb386a9894a61/gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb#L172
* All the errors are collected to `threads_errors` and then passed to abort multipart upload call: 
  * https://github.com/aws/aws-sdk-ruby/blob/2f0e6f83473c59d6f4875694ae8cb386a9894a61/gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb#L79 
  * https://github.com/aws/aws-sdk-ruby/blob/2f0e6f83473c59d6f4875694ae8cb386a9894a61/gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb#L86

Removing `abort_on_exception` should be _ok_ here.

The `file_downloader.rb` is bit different since we raise the error in its own thread. With the removal of `abort_on_exception`, we shouldn't cause any interruption in the main thread.

gems/aws-sdk-s3/spec/object/download_file_spec.rb
```diff
@@ -221,17 +221,34 @@ module S3
...
           thread = double(value: nil)
-          allow(thread).to receive(:abort_on_exception=)
           expect(Thread).to receive(:new).and_yield.and_return(thread)
 
           expect do
             large_obj.download_file(path)
           end.to raise_error(Aws::Errors::ChecksumError)
         end
 
+        it 'does not interrupt the main thread when an exception occurs' do
```
mullermp: Probably not necessary because we're removing functionality here.

mullermp (commented):
Can you investigate whether this code is still needed and if it needs changing? https://github.com/aws/aws-sdk-ruby/blob/e652c549c2b9ae41d7fc7b62ec4733eb1b15a42e/gems/aws-sdk-core/spec/shared_spec_helper.rb#L44-L53

gems/aws-sdk-s3/spec/object/download_file_spec.rb
jterapin: Whoah. The `threat_report_on_exception` flag is cool. I was actually wondering on how I can go about removing the exception clutter left between specs. I found the PR when this was implemented: https://github.com/aws/aws-sdk-ruby/pull/2307

I want to say that it would be nice to keep it since the below spec has it: 
https://github.com/aws/aws-sdk-ruby/blob/ed4f82ecc9e141aafa2fd20bd24719db839510d3/gems/aws-sdk-s3/spec/object/multipart_copy_spec.rb#L371

Without this, this spec prints (7) long exception errors.

Also dropped the spec I wrote.

mullermp (approved):

Issue #2950, opened 2023-11-20
Generate RBS


ksss:
### Describe the feature

Currently, Ruby code generation is done by build_tools, but I propose that RBS files also be generated during this Ruby code generation. RBS refers to Ruby's standard type files and tools, which can be bundled with the gem itself.
I have already been conducting experimental operations for over a year.
In a repository similar to DefinitelyTyped for RBS, I have been generating RBS from `apis/**/*.json`, and it has been used within the RBS community.

https://github.com/ruby/gem_rbs_collection/tree/main/generators/aws-sdk-rbs-generator

### Use Case

Using LSP (Language Server Protocol) in recent years, we can receive coding assistance as follows:

- Automatic suggestion of methods
- Type checking of argument names and method names
- Usage similar to API documentation

Moreover, by being incorporated into the upstream, it becomes possible to accurately manage the Ruby code and corresponding RBS for each version.

### Proposed Solution

I will integrate the existing `aws-sdk-code-generator` and the `aws-sdk-rbs-generator` that I implemented, and organize the workflow so that both Ruby code and RBS code are generated and released simultaneously.
Prepare a `sig` directory in each gem and generate RBS files within this directory. This will not affect the Ruby code at all.

### Other Information

Since I have made numerous commits to RBS, I can provide detailed support regarding RBS.

https://github.com/ruby/rbs/graphs/contributors
https://github.com/ruby/gem_rbs_collection/tree/main/generators/aws-sdk-rbs-generator

### Acknowledgements

- [X] I may be able to implement this feature request
- [X] This feature might incur a breaking change

### SDK version used

3

### Environment details (OS name and version, etc.)

any

mullermp:
Thanks for reaching out! We met with Soutaro at RubyConf and he told us about your project. We would love to get RBS typing into v3 and would appreciate any help doing so. This would certainly be a large project and may take some time. As far as generation goes, v3 code generation uses Mustache templates. Looking at aws-sdk-rbs-generator, I think most of the logic goes away when using the existing aws-sdk-code-generator. I think we should get a simple gem working, like lambda, as a proof of concept. There's no need to be backwards compatible inside aws-sdk-code-generator within reason. You can test builds with `rake build:aws-sdk-lambda` for example. If you start a feature branch, we would be happy to contribute and review when able.

ksss:
@mullermp 
I apologize for the late reply; I was down with a cold 🤧 
[...]

PR #2961, opened 2023-11-30
Support RBS for aws-sdk-core and service gems

ksss:
https://github.com/aws/aws-sdk-ruby/issues/2950

I have added RBS files for aws-sdk-core.
There is no impact on the actual Ruby code.
As a test for RBS, I added CI to run the `rbs validate` command. This test ensures that at a minimum, the RBS can be loaded.


gems/aws-sdk-core/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Feature - Add RBS
```
mullermp: Can we be a little more specific about what files are added?

gems/aws-sdk-core/sig/aws-sdk-core.rbs
```diff
@@ -0,0 +1,57 @@
+module Aws
```
mullermp, line 1: Is it best practice to stuff rbs into one file or to have a directory/file structure similar to lib files?

mullermp (approved):
In general this looks good. Thank you! I would prefer to release this once we get the service RBS generation ready. When we're ready, we can set the minimum version here (https://github.com/aws/aws-sdk-ruby/blob/version-3/build_tools/services.rb#L13)

mullermp:
Looks like RBS fails with Jruby 9.3 and 9.4. Can you investigate?

ksss:
Thank you for reviewing.

Even if this is released, I was dealing with the problem that it could not handle types properly due to conflicts with gem_rbs_collection.
This issue has been fixed on the RBS side, so it will be usable normally from RBS v3.4.
Under RBS v3.4, gem_rbs_collection's signature will be used in preference, but this will not be a problem for a while.
RBS v3.4 is scheduled to be included as a bundled gem in CRuby v3.3.

See also https://github.com/ruby/rbs/pull/1659

> Looks like RBS fails with Jruby 9.3 and 9.4. Can you investigate?

Yes, I'll investigate this issue.

gems/aws-sdk-core/sig/aws-sdk-core.rbs
ksss: There are no specific best practices, and the granularity of files is chosen based on preference.
However, it has been reported that performance issues with Steep may arise if the number of RBS files exceeds 10,000.

https://github.com/soutaro/steep/issues/661

ksss:
RBS includes C extension.
Therefore, I have restricted the platforms for which it will be built.

ksss:
@mullermp
In which code is the automatic release of the gem executed?
RBS can include documentation in itself and can be used by steep, etc.
Also, documentation can be copied from rdoc and given to the RBS.
I have prepared this copy operation as a rake task.
Is it possible to add RBS documentation generation tasks to the automatic release?

alextwoods:
> @mullermp In which code is the automatic release of the gem executed? RBS can include documentation in itself and can be used by steep, etc. Also, documentation can be copied from rdoc and given to the RBS. I have prepared this copy operation as a rake task. Is it possible to add RBS documentation generation tasks to the automatic release?

The code for automated releases uses an internal system, but we can add this task to it.

What does including documentation in RBS do?  It seems like a lot of duplicated information w/ the existing documentation?

mullermp:
Since our documentation is code generated, I agree that we should probably only have one location for it (source code), unless there's good reason.

ksss:
@alextwoods @mullermp 
The benefit of writing documentation in RBS is that, with the current behavior of Steep, for example in VSCode, when you hover over code, RBS documentation is displayed in a tooltip of vscode.
Many methods in aws-sdk-ruby have complex arguments.
Displaying the documentation on the editor can assist in coding and learning.
If the documentation is not written in RBS, only the type information is displayed.

hover with rbs documentation example on vscode:
image

Certainly, adding documentation to RBS leads to double management of documentation and complicates the build tasks. Additionally, the behavior of tools might change in the future.
So, shall we proceed without including documentation in RBS this time?

mullermp:
That's actually a great feature! I use RubyMine as my editor and it does not do that. I think that the code generated documentation (using docs json), the same documentation we use on client methods and params, etc, can be code generated in rbs too. Then management of this becomes easy. We can probably copy docs manually for the few hand written code.

mullermp:
What do you mean by documentation can be provided to rbs? I was assuming that it is just inline documentation in the same rbs file. If we are code generating these files, we can add the same documentation that's in the source code.

gems/aws-sdk-core/sig/aws-sdk-core.rbs
mullermp: Makes sense. I think we should break these out into a similar structure like lib/ then adding more becomes easier to manage and organized.

ksss:
RBS has a feature that allows copying existing rdoc comments into RBS files.
Using this feature, after building the Ruby code, we can copy just the documentation to the RBS files with a CLI command. 
Specifically, we can achieve this by running the following rake task:

```rb
rule /^rbs:doc:aws-sdk-\w+$/ do |task|
  name = task.name.split(':').last
  sh("bundle exec rdoc --ri --output tmp gems/#{name}/lib")
  sh("bundle exec rbs annotate --dir tmp gems/#{name}/sig")
end
```

ksss:
Alternatively, we might consider adding the following to the build task:

```diff
diff --git a/tasks/build.rake b/tasks/build.rake
index 8767374fc3..06bac3aa20 100644
--- a/tasks/build.rake
+++ b/tasks/build.rake
@@ -22,6 +22,12 @@ rule /^build:aws-sdk-\w+$/ do |task|
   )
   writer = BuildTools::FileWriter.new(directory: "#{$GEMS_DIR}/#{service.gem_name}")
   writer.write_files(files)
+
+  if ENV['GEN_RBS'] && ENV['GEN_RBS'].split(',').include?(identifier)
+    name = task.name.split(':').last
+    sh("bundle exec rdoc --ri --output tmp gems/#{name}/lib")
+    sh("bundle exec rbs annotate --dir tmp gems/#{name}/sig")
+  end
 end
```

mullermp:
That makes sense! For doc generation, we use `yard` - will it work similarly to rdoc in this case? Whatever approach we use, I think putting logic in the build task makes the most sense. We probably don't need an ENV variable to turn it off.

I think the biggest concern we have is gem size, especially for S3. The generated documentation is very large. We may want to consider truncating it such that it links to public doc pages instead.

ksss:
As of now, RBS does not have the functionality to read cache of YARD `.yardoc`.
Therefore, we generate temporary files for `ri` using by `rdoc --ri` command, which works fine.

When calculating the gem size, for `aws-sdk-s3`, the current size is 3.0MB.
Adding RBS without documentation increases it by +244K,
while adding RBS with documentation increases it by +1.8MB.

soutaro:
@mullermp 

> We may want to consider truncating it such that it links to public doc pages instead.

It would be the workaround for now.

Does it help if `rubygem` (or `bundler`) has an option to delete installed RBS files after the installation? I assume the gem users will enable the option for their production builds to minimize the footprint.

mullermp:
I'm not sure if an option to delete installed RBS is appropriate or not.. I think the gem size increase is a concern. Could we move forward with not including docs for now and consider an enhancement to add them when we've put more thought into the correct approach and/or when yard is supported? I think at worst case, we can code generate the links to the documentation of each Client operation. i.e. insert operation name `https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/ACM/Client.html#<operation name>-instance_method`

mullermp:
I'm not sure if this is possible, but it would be nice if doc comments on RBS (or RBS tooling) could support a magic tag like `@source` or something that knows to load the exact comments from the source file that it is describing.

soutaro:
> I think the gem size increase is a concern.

Got it. 👍 

> it would be nice if doc comments on RBS (or RBS tooling) could support a magic tag like `@source`

It should be possible. Something like https://shopify.github.io/ruby-lsp/RubyLsp/Requests/DocumentLink.html in Ruby LSP in RBS too would help...

gems/aws-sdk-core/sig/aws-sdk-core.rbs
ksss: OK. Let's divide the files to correspond with the structure under the `lib` directory.

ksss:
@soutaro
Thank you for supporting.

@mullermp @alextwoods 
Based on our discussions so far, I have summarized as follows and also squashed and organized the code:

- For gem RBS documentation, only include links like `https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/ACM/Client.html#<operation name>-instance_method`.
- Divide RBS files to correspond with the structure under `lib`.

gems/aws-sdk-core/sig/seahorse/client/response.rbs
```diff
@@ -0,0 +1,26 @@
+module Seahorse
+  module Client
+    # <!--
+    # RBS does not support Delegator.
+    # the behavior is mimicked `Seahorse::Client::Response` as much as possible.
+    # -->
+    module DelegatorMethodsForResponse[DATA]
```
ksss: This module does not actually exist.

In RBS, it's not possible to represent the behavior of [Delegator](https://docs.ruby-lang.org/en/3.2/Delegator.html), so we can't make the return value of the operations method group `Seahorse::Client::Response`.
Instead, we define a non-existent module called `DelegatorMethodsForResponse` to represent the methods defined in `Seahorse::Client::Response`.
By including `DelegatorMethodsForResponse` in the data class, we can represent the actual behavior.

example:
```rbs
module Aws
  module Lambda
    module Types
      class DestinationConfig
        include Seahorse::Client::DelegatorMethodsForResponse[DestinationConfig]

        attr_accessor on_success: OnSuccess

        attr_accessor on_failure: OnFailure
      end
    end
  end
end
```

Moreover, `DelegatorMethodsForResponse` cannot be an interface. An interface does not allow for duplicate methods. The existence of `Aws::Lambda::DestinationConfig#on_success` would result in a method duplication, which is why it cannot be handled as an interface.

ksss:
In my environment, I have been able to generate valid RBS for the following gems.
It's okay to release `aws-sdk-core` in its current PR.

- aws-sdk-s3
- aws-sdk-sqs
- aws-sdk-lambda
- aws-sdk-ec2
- aws-sdk-cloudfront
- aws-sdk-kms
- aws-sdk-firehose
- aws-sdk-sns
- aws-sdk-ssm

ksss (commented):
In my environment, I have been able to generate valid RBS for the following gems.
It's okay to release `aws-sdk-core` in its current PR.

- aws-sdk-s3
- aws-sdk-sqs
- aws-sdk-lambda
- aws-sdk-ec2
- aws-sdk-cloudfront
- aws-sdk-kms
- aws-sdk-firehose
- aws-sdk-sns
- aws-sdk-ssm

gems/aws-sdk-core/aws-sdk-core.gemspec
```diff
@@ -10,7 +10,7 @@ Gem::Specification.new do |spec|
   spec.homepage      = 'https://github.com/aws/aws-sdk-ruby'
   spec.license       = 'Apache-2.0'
   spec.require_paths = ['lib']
-  spec.files         = Dir['LICENSE.txt', 'CHANGELOG.md', 'VERSION', 'lib/**/*.rb', 'ca-bundle.crt']
+  spec.files         = Dir['LICENSE.txt', 'CHANGELOG.md', 'VERSION', 'lib/**/*.rb', 'ca-bundle.crt', 'sig/**/*.rbs']
```
mullermp: Small nit - place this after lib/

mullermp (approved):
It looks great! Thank you for all the hard work. :D

We should keep this PR until the service gem changes are ready. It is easier for our release system if we do them together.

Rakefile
```diff
@@ -10,6 +10,9 @@ $:.unshift("#{$GEMS_DIR}/aws-sdk-core/lib")
 $:.unshift("#{$GEMS_DIR}/aws-partitions/lib")
 $:.unshift("#{$GEMS_DIR}/aws-eventstream/lib")
 $:.unshift("#{$GEMS_DIR}/aws-sigv4/lib")
+Dir.glob("#{$GEMS_DIR}/aws-sdk-*/lib").each do |lib|
+  $:.unshift(lib)
+end
```
ksss: It is used to know the exact arguments of `Client#initialize`.
Is it acceptable?

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/rbs/dynamic_error_list/sqs.json
```diff
@@ -0,0 +1,59 @@
+{
```
ksss: This file is a dynamically defined error class written by hand.

build_tools/services.rb
```diff
@@ -177,6 +178,10 @@ def eventstream?(api)
       false
     end
 
+    def support_rbs?(identifier)
+      %w[ lambda ].include?(identifier)
```
ksss: Define a gem to support RBS.
Is this acceptable?

ksss (commented):
Additional commits to RBS generator and aws-sdk-lambda output.

## Supported files

- client.rbs
- errors.rbs
- [each resource].rbs
- resource.rbs
- types.rbs
- waiters.rbs

## Type policy

- Provide keyword arguments whenever possible.
- This allows LSP to suggest optional key names and types as needed for each API as they are entered.
- Use `(?Hash[Symbol, untyped])` as another pattern for arguments, so that there are no errors when passing Hash directly or from outside.

Rakefile
mullermp: Seems fine unless there's a better way. Initialize is always a hash of varying (untyped?) options. Will this give accurate options?

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/code_builder.rb
```diff
@@ -109,6 +109,29 @@ def spec_files(options = {})
       end
     end
 
+    # @return [Enumerable<String<path>, String<code>>]
+    def rbs_files(options = {})
+      Enumerator.new do |y|
+        if @service.support_rbs?
```
mullermp: nit: push this check upward (see other comment)

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/gem_builder.rb
```diff
@@ -37,6 +37,9 @@ def each(&block)
         code.spec_files.each do |path, code|
           y.yield("spec/#{path}", code)
         end
+        code.rbs_files.each do |path, code|
```
mullermp, line 39: I think you want to check `@service.support_rbs?` here, similar to `if @service.smoke_tests` above, since `rbs_files` method entirely relies on it being enabled.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/rbs/dynamic_error_list/sqs.json
mullermp: Hmm. Is the intention to generate possible DynamicErrors? I think we'd rather everything be code generated against the model and not anything manual. For cases not in the model, we should reference the `Aws::<Service>::Errors::ServiceError` parent class? We know the methods on this class. It's not a perfect experience but rescue of the parent will succeed for these errors.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/rbs/waiter.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+module AwsSdkCodeGenerator
+  module RBS
+    class Waiter < Struct.new(
+      :name,
+      :class_name,
+      :client_overlod_keyword_argument,
```
mullermp: Is this overload typo?

build_tools/services.rb
mullermp: I think we could assume every AWS service gem is supported, so I'd just flat out pass `true` in. We do have use cases (internally) where this generation code is called, and since it defaults to false, that case is fine.

gems/aws-sdk-lambda/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Feature - Add RBS signature files to support static type checking.
+
```
mullermp: Our release system will detect changes and bump all of the gems, but I think adding these for all gems will be helpful (we can do that step at the end).

gems/aws-sdk-lambda/sig/client.rbs
```diff
@@ -0,0 +1,298 @@
...
+                  | (?Hash[Symbol, untyped]) -> instance
+
+
+      # https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/Lambda/Client.html#add_layer_version_permission-instance_method
+      def add_layer_version_permission: (action: Types::tLayerPermissionAllowedAction, layer_name: Types::tLayerName, principal: Types::tLayerPermissionAllowedPrincipal, statement_id: Types::tStatementId, version_number: Types::tLayerVersionNumber, ?organization_id: Types::tOrganizationId, ?revision_id: Types::tString) -> Types::AddLayerVersionPermissionResponse
```
mullermp: Is this the most useful part for operation signatures? Typically our customers use options with simple types and don't create Types structures. How does RBS resolve this? One improvement we could potentially make is translate types to simples, like "Types::tLayerName" to String. We may be able to do this nested through all API members. What do you think?

gems/aws-sdk-lambda/sig/client.rbs
```diff
@@ -0,0 +1,298 @@
...
+                                      | (Hash[Symbol, untyped] params, ?Hash[Symbol, untyped] options) -> Types::AddLayerVersionPermissionResponse
+
+      # https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/Lambda/Client.html#add_permission-instance_method
+      def add_permission: (action: Types::tAction, function_name: Types::tFunctionName, principal: Types::tPrincipal, statement_id: Types::tStatementId, ?event_source_token: Types::tEventSourceToken, ?function_url_auth_type: Types::tFunctionUrlAuthType, ?principal_org_id: Types::tPrincipalOrgID, ?qualifier: Types::tQualifier, ?revision_id: Types::tString, ?source_account: Types::tSourceOwner, ?source_arn: Types::tArn) -> Types::AddPermissionResponse
```
mullermp: Is prefixing "t" in front of Types like tFunctionName a best practice?

mullermp (commented):
Nice work! I had a few comments that we can discuss.

Rakefile
ksss: It didn't occur to me a year ago, but perhaps this code could be eliminated if we use the same method as in the document. I will give it a try.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/rbs/dynamic_error_list/sqs.json
ksss: This approach was born out of the following PR responses.

If aws-sdk-ruby does not recommend DynamicErrors to be written in code, it should not generate these types of errors.

https://github.com/ruby/gem_rbs_collection/issues/355

build_tools/services.rb
ksss: Oops, I guess I was mistaken.
At first I thought they would release only aws-sdk-lambda with RBS support and gradually increase the number of supported gems.

gems/aws-sdk-lambda/CHANGELOG.md
ksss: Does this PR not require CHANGELOG support?

gems/aws-sdk-lambda/sig/client.rbs
ksss: Items that start with `t` are type aliases. 
For instance, `Types::tString` is actually `String`. 
The aliases not being expanded is due to my inexperience. 
It might be possible to expand aliases by doing something similar to generating documentation.

When a Hash is nested, RBS's record type is used, but it's rarely useful. 
Soutaro also said he had difficulties with the record type. 
The only way to handle it without errors is to use `Hash[Symbol, untyped]`.

Until the type checking system is further developed, the part about keyword arguments is only of documentary significance.

It seems like it would be more readable to expand this record as well in the documentation. 
I'll give it a try.

gems/aws-sdk-lambda/sig/client.rbs
ksss: At first I was treating it with underscore like `function_name`, but then I realized that the names could conflict because of the appearance of the shapes `String` and `string` in aws-sdk-sns.

If type aliases could be expanded, this problem might be solved.

ksss (commented):
I will try to improve it a little more.

gems/aws-sdk-lambda/CHANGELOG.md
mullermp: When we build the gems, if we detect any changes through `git diff`, it becomes 
```
* Feature - Code Generated Changes, see `./build_tools` or `aws-sdk-core`'s CHANGELOG.md for details.
```
and is handled automatically, but this is not descriptive, we probably will want to add manual entries like you did for lambda to all services, but we can do that much later.

gems/aws-sdk-lambda/sig/client.rbs
mullermp: I understand - we can only do the best we can with the current limitation. I defer to you (as a customer) for what you find most useful when developing with the SDK

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/rbs/dynamic_error_list/sqs.json
mullermp: Makes sense.. We have over 300 services, so maintaining a list of un-modeled errors seems like a nightmare and could get stale fast.. I think we should default dynamic errors these to the parent class since they share members (I think this is the useful part). We are pushing on service teams to model errors more extensively.

build_tools/services.rb
mullermp: I was suggesting we use lambda for reviewing, but I think when we've finalized the code, we may as well launch for all services since it's code generated. :)

mullermp: We *could* consider supporting only a few gems as a "dev preview" then open it up to wider. It depends on our confidence. Thoughts?

gems/aws-sdk-lambda/CHANGELOG.md
alextwoods: Depending on the release timing for this, we may want to just improve the automated release system message in those cases (ie: reference the change log from build tools).

tasks/rbs.rake
```diff
@@ -0,0 +1,22 @@
...
+    Dir.glob('gems/*/sig').each do |dir|
+      sdk_gem = dir.sub(%r{gems/(.*)/sig}, '\1')
+      spec = Gem::Specification::load("gems/#{sdk_gem}/#{sdk_gem}.gemspec")
+      deps = spec.dependencies.select do |d|
+        all_gems.include?(d.name) && File.directory?("gems/#{d.name}/sig")
+      end.map { |d| d.name }
+
+      puts "Validating gem `#{sdk_gem}` with deps #{deps}"
+      sh("bundle exec rbs #{deps.map{"-I gems/#{_1}/sig"}.join(' ')} -I #{dir} validate --silent") do |ok, _|
+        sigs << File.basename(File.dirname(dir)) unless ok
```
alextwoods: naming nit.  Maybe `failures` instead of `sigs`

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/rbs/dynamic_error_list/sqs.json
alextwoods: Yeah -  maintaining a hand written list of errors across 300+ services is untenable (and theres no actual source of truth).  We are stuck with the dynamic errors approach within this major version to avoid breaking users, but its far from ideal.  We are pushing service teams to correctly model errors on their APIs.  I think I'd agree with @mullermp  that these should just use the `Aws::<Service>::Errors::ServiceError` parent class, though I know its not an ideal experiance.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/rbs/shape_dictionary/shape.rb
```diff
@@ -0,0 +1,275 @@
...
+          when "integer", "long"
+            "::Integer"
+          when "float", "double"
+            "::Float"
+          when "timestamp"
+            case shape_body["timestampFormat"]
+            when "iso8601", "rfc822", nil
+              "::Time"
+            else
+              raise [name, shape_body].inspect
```
alextwoods: I think it would be reasonable to assume the interface to a timestamp shape is always ::Time regardless of its format on the wire

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/rbs/client_class.rb
```diff
@@ -0,0 +1,133 @@
...
+                "(:none | :equal | :full | ^(Integer) -> Integer)"
+              when :stub_responses, :endpoint, :credentials, :log_formatter, :client_side_monitoring_publisher, :logger, :endpoint_provider, :token_provider, :express_credentials_provider
+                # TODO: Just not supported yet
+                "untyped"
+              else
+                case opt.doc_type
+                when "Boolean"
+                  "bool"
+                when nil
+                  raise opt.inspect
```
alextwoods: Whats the consequenece of using "untyped" here instead of raising?

I'm a little worried about this raising errors on new cases where we would just use "untyped".  (Also, if we are going to raise errors, we should include a more useful message).

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/rbs/dynamic_error_list/sqs.json
ksss: Thanks for the very useful information, I was having trouble finding documentation on how to handle DynamicErrors.
I will modify it so that DynamicErrors are not handled.
By not creating a type for DynamicErrors, we will be able to get users to avoid using them.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/rbs/client_class.rb
ksss: This is a hook I have in place because I wanted to know what the pattern could be, but so far I have not encountered this error. It should be `untyped`.

ksss:
Fixed to specify the type directly.
The return type of client.rbs still needs to be modified.

ksss:
Current status

- [x] Remove require for getting client argument.
- [x] Remove DynamicErrors
- [x] Do not use type aliases, apply types directly
    - Remove `ShapeDictionary`
    - add `KeywordArgumentBuilder`
- [x] Add links to the documentation for major methods
- [x] Pass RBS::Test
	- core
	- s3
- [x] Modify tests & add to CI
- [ ] Types for custom code

mullermp:
Looking. I'm sorry about the wait.

CI / Sig workflow appears to be failing due to some edge cases.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator.rb
```diff
@@ -76,6 +76,25 @@
 require_relative 'aws-sdk-code-generator/code_builder'
 require_relative 'aws-sdk-code-generator/gem_builder'
 
+# RBS
+if RUBY_VERSION >= '3.2'
```
mullermp: Should this be > 3.0? Is this because of dependencies? Please add a more detailed comment about why this needs to be true. We should try to open this up if possible.

gems/aws-sdk-core/spec/seahorse/client/handler_builder_spec.rb
```diff
@@ -48,7 +48,7 @@ def initialize
           expect(handler_class.new.call('context')).to eq('context')
         end
 
-        it 'accepts a step with the block' do
+        it 'accepts a step with the block', rbs_test: :skip do
```
mullermp, line 51: Why do these tests need to be skipped for rbs?

gems/aws-sdk-s3/spec/client_spec.rb
```diff
@@ -332,7 +332,7 @@ module S3
       end
 
       describe '#create_bucket' do
-        it 'omits location constraint for the classic region' do
+        it 'omits location constraint for the classic region', rbs_test: :skip do
```
mullermp, line 335: Ditto on the skipping - what causes these to fail? I could assist in improving the test.

mullermp (commented):
Nice work! I think this is almost ready. Let's get CI to pass and see if test skipping can be fixed. We will need to undo all of the `rake build` changes before merging so that the code generator can pick up a diff for publishing.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/rbs/client_class.rb
```diff
@@ -0,0 +1,176 @@
...
+              [opt.name, "?#{opt.name}: #{type}", opt.doc_type]
+          end
+          # Find duplicated key
+          grouped = buffer.group_by { |name, _| name }
+          grouped.transform_values(&:count).find_all { |_, c| 1 < c }.each do |name,|
+            # :endpoint is duprecated in all client
+            case name
+            when :endpoint, :endpoint_provider
+            else
+              warn("Duprecate client option: `#{grouped[name].map { |g| g.values_at(0, 2) }}`", uplevel: 0)
```
alextwoods: Nit: typo.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/rbs/client_class.rb
```diff
@@ -0,0 +1,176 @@
...
+        def documented_plugin_options(plugins)
+          i = 0
+          plugins.map(&:options).flatten.select(&:documented?).sort_by do |opt|
+            [opt.required ? 'a' : 'b', opt.name, i += 1]
+          end
+        end
+
+        def build_keyword_arguments(plugins)
+          buffer = plugins.map do |opt|
+            type = case opt.name
```
alextwoods: Its great to have these - but I think rather than hard coding in an option -> type here, we can add `rbs_type` to the [PluginOption](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/lib/seahorse/client/plugin.rb#L110) and then in each plugin where these are defined we can specify the rbs.

I know this doesn't work for some of the net http and other  (eg http_proxy, ect) since they aren't specified by plugins, so we can leave those here.  But adding `rbs_type` to the plugins should make it easier for us to keep types up to date when we add new configuration....

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator.rb
alextwoods: Should we just fail/raise an error if we don't have a ruby version that supports rbs generation?

Our release tools / code generator does not need to support the same ruby versions as the general SDK gems - and I'd be worried about the possibility of running code generation on an older ruby and silently NOT generating signatures - and once we release them, I think removing them would be considered a breaking change.

alextwoods (commented):
Nice - this is looking good!

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator.rb
ksss: I would have imagined that the code generation part runs on a specific Ruby version and is run in an automatic release.
As for the dependency, no problem, I will fix it so that the if statement can be removed.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/rbs/client_class.rb
ksss: Great idea! I will give it a try.

gems/aws-sdk-core/spec/seahorse/client/handler_builder_spec.rb
ksss: I have not been able to investigate the details, but it seems to be a bug on the rbs side that prevents testing 😢

```
$ RUBYOPT="-r bundler/setup -r rbs/test/setup" RBS_TEST_RAISE=true RBS_TEST_LOGLEVEL=error RBS_TEST_OPT="-I gems/aws-sdk-core/sig -I gems/aws-sdk-core/sig" RBS_TEST_TARGET="Aws,Aws::*,Seahorse,Seahorse::*" bundle exec rspec gems/aws-sdk-core/spec/seahorse/client/handler_builder_spec.rb
...F..........

Failures:

  1) Seahorse::Client::HandlerBuilder#handler accepts a step with the block
     Failure/Error: super(context)

     NoMethodError:
       super called outside of method
     # ./gems/aws-sdk-core/spec/seahorse/client/handler_builder_spec.rb:54:in `block (4 levels) in <module:Client>'
     # /Users/ksss/src/github.com/ksss/rbs/lib/rbs/test/hook.rb:113:in `instance_exec'
     # /Users/ksss/src/github.com/ksss/rbs/lib/rbs/test/hook.rb:113:in `block in handler__with__RBS_TEST_a0f0b5_878362e7'
     # ./gems/aws-sdk-core/spec/seahorse/client/handler_builder_spec.rb:68:in `block (3 levels) in <module:Client>'

Finished in 0.2305 seconds (files took 5.46 seconds to load)
14 examples, 1 failure

Failed examples:

rspec ./gems/aws-sdk-core/spec/seahorse/client/handler_builder_spec.rb:51 # Seahorse::Client::HandlerBuilder#handler accepts a step with the block
```

gems/aws-sdk-s3/spec/client_spec.rb
ksss: I don't know why, but `resp.data` (and also `resp.error`) is `nil`.

mullermp:
This is still on our radar for deeper review and merge. Thanks for being patient.

gems/aws-sdk-s3/spec/plugins/md5s_spec.rb
```diff
@@ -17,7 +17,7 @@ module Plugins
           expect(client.config.compute_checksums).to be(false)
         end
 
-        it 'retries invalid content MD5s' do
+        it 'retries invalid content MD5s', rbs_test: :skip do
```
mullermp, line 20: When I remove this, the test passes, and rbs:spytest passes. Are these no longer necessary, or am I missing some information?

mullermp (approved):
Nice work! I think we will need to "revert" the generated changes in lambda/s3. When this is released, those gems will regenerate, and our release system will determine that there is a `git diff` and bump the gems automatically.

gems/aws-sdk-s3/spec/plugins/md5s_spec.rb
ksss: It still fails in my environment.
This is because this test is an error pattern.
Are files generated in gems/aws-sdk-s3/sig?

```
$ RUBYOPT="-r bundler/setup -r rbs/test/setup" RBS_TEST_RAISE=true RBS_TEST_LOGLEVEL=error RBS_TEST_OPT="-I gems/aws-sdk-core/sig -I gems/aws-sdk-s3/sig" RBS_TEST_TARGET="Aws,Aws::*,Seahorse,Seahorse::*" bundle exec rspec gems/aws-sdk-s3/spec --tag '~rbs_test:skip'
Run options: exclude {:rbs_test=>"skip"}
........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................F............................................................................................

Failures:

  1) Aws::S3::Plugins::Md5s retries invalid content MD5s
     Failure/Error: resp = s3.put_object(bucket: 'test', key: 'test', content_md5: '456==')

     RBS::Test::Tester::TypeError:
       TypeError: [Aws::S3::Client#put_object] UnresolvedOverloadingError: couldn't find a suitable overloading
     # ./gems/aws-sdk-s3/spec/plugins/md5s_spec.rb:46:in `block (2 levels) in <module:Plugins>'

Finished in 3.3 seconds (files took 1.03 seconds to load)
1141 examples, 1 failure

Failed examples:

rspec ./gems/aws-sdk-s3/spec/plugins/md5s_spec.rb:20 # Aws::S3::Plugins::Md5s retries invalid content MD5s

$ bundle exec rbs --version
rbs 3.4.1
```

ksss (commented):
Yes. I will revert the output of aws-sdk-lambda. It has already served its purpose as a post-generation output sample.

ksss:
@mullermp @alextwoods 
Update completed. Shall I squash the commits previous commits?

alextwoods (approved):

mullermp:
I'm currently seeing errors like:
```
         TypeError: [Aws::S3::Client#get_object] UnresolvedOverloadingError: couldn't find a suitable overloading
```

Let's check CI.

mullermp:
Thank you for all of your hard work. I think this is just about ready for launch. Let's target next week. Please be advised if there are any customer issues or bugs - we will try our best to fix them, but we may need help from the RBS experts :)

ksss:
I'm proud to have been entrusted with such an important work 👍 
Please feel free to reach out anytime. We will continue to support as much as we can.

mullermp:
I am typing up a blog post to be posted alongside merging this change. Could you please assist in writing a minimal case for validating rbs types when using the SDK? I tried the following:

```
# Steepfile
target :test do
  signature "**/*.rbs"
  check "test.rb"
end

# test.rb
# Should fail because region is 123?
client = Aws::S3::Client.new(region: 123)
```

but `steep check` succeeds.

mullermp:
I was also able to get type information in RubyMine but not vs code. Do you have any setup information for displaying type mismatch in vs code?

ksss:
### Steep

The most common way to read RBS with steep is to use `rbs collection`.

https://github.com/ruby/rbs/blob/master/docs/collection.md

The setup in project when using `rbs collection` is as follows.

```
$ bundle exec rbs collection init
$ bundle exec rbs collection install
```

This will automatically read the gem signatures described in the Gemfile.lock.
Thus, you no longer need to use `signature` in the Steepfile as follows.

```ruby
target :test do
  check "test.rb"
end
```

### Sample code

Unfortunately, it is currently difficult to make the following code a type error.

```ruby
client = Aws::S3::Client.new(region: 123)
```

The main reason is that the actual aws-sdk implementation does not use the keyword argument.
Therefore, the following code must be supported.

```ruby
# record pattern
client = Aws::S3::Client.new({region: 123})

# hash pattern
config = {region: 123}
client = Aws::S3::Client.new(config)
```

The record pattern does not currently handle optional keys (implementation is in progress).

The hash pattern only allows key and value types to be specified, so per-key checks are not possible.

Therefore, many method arguments are resolved with `Hash[Symbol, untyped]`. (This may be improved in the future.)


How about an example using Resource instead?

```ruby
resource = Aws::S3::Resource.new
resource.buckets.each do |bucket|
  bucket.objects.limit(50).each do |obj|
    puts obj.get.etag # String
  end
end
```

image

ksss:
As a side note, rbs v3.4 or higher is required to load gem signatures.
As gem_rbs_collection signatures will be loaded first if use rbs v3.3 or lower.

mullermp:
Thanks for your response on this. RubyMine correctly validates this code (and gives an incompatible types error):
```
client = Aws::S3::Client.new(region: 123)
```

It however validates the record and hash pattern as correct.

RE: Steep - I'm looking for a way to have `steep check` fail in some sample code. I am attempting this from the workspace (rake build aws-sdk-s3), so that is why I am using `sig` in the Steepfile. I do not have the aws-sdk-s3 gem in a Gemfile. I am trying to demonstrate how a customer might validate their code.

RE: VS Code - what plugins / extensions are you using to make that work?

ksss:
Sounds good news! I will try RubyMine too.

### For vscode user

If you want to use steep with vscode, you need to install the following extension.

https://marketplace.visualstudio.com/items?itemName=soutaro.steep-vscode

### For user without Gemfile

If you do not use Gemfile, specify the library to load as follows.
(Please make sure to install gems that include `sig` using `gem install`.)

```ruby
# Steepfile
target :lib do
  check "test.rb"

  library "aws-sdk-core"
  library "aws-sdk-s3"
end
```

### For developer

If it is for development, I have prepared sample code. Please check it.

https://github.com/ksss/aws-sdk-ruby/tree/steep-sample/gems/aws-sdk-s3

```
$ bundle install
$ cd gems/aws-sdk-s3
$ code t.rb
```

mullermp:
Thank you for providing some help in this. I'm running into issues with the Steep plugin - 
```
[Info  - 3:07:37 PM] Connection to server got closed. Server will restart.
[Error - 3:07:37 PM] Server initialization failed.
  Message: Pending response rejected since connection got disposed
  Code: -32097 
```
I'm inclined to skip this example for now :( I think the important part of the announcement is that signatures will be available.

mullermp (approved):
I think this is good to ship!

mullermp:
We will monitor our release tomorrow and we can do a sanity check with the published gems.

ksss:
I appreciate the support from the aws-sdk-ruby team ✨

soutaro:
@mullermp 

> I'm running into issues with the Steep plugin -

Is there anything I can help you? Looks like it's an issue of ruby interpreter loading (or maybe `bundle install`...)
Running `bundle exec steep binstub` before reloading VSCode may help.

mullermp:
Blog post here! https://aws.amazon.com/blogs/developer/announcing-rbs-support-for-aws-sdk-for-ruby-v3/

@ksss Thank you for your work

@soutaro Yes, I'll get back to you on trying this, currently working on something else this week.

ksss:
@mullermp Nice work!
Thank you for giving me credit!

PR #2965, opened 2023-12-06
Ensure Expires is a timestamp in s3

mullermp:
Expires is going to be modeled as a String instead for stupid reasons. Ensure backwards compatibility.


alextwoods (approved):
LGTM - needs a changelog?

PR #2966, opened 2023-12-06
Minor Integration test fixes

alextwoods:

### DIFF
--- a/gems/aws-sdk-support/features/smoke_step_definitions.rb
+++ b/gems/aws-sdk-support/features/smoke_step_definitions.rb
@@ -22,7 +22,6 @@ When(/I call the operation '(.*?)' with params:/) do |operation, params|
     @regional_client.send(operation.to_sym, opts)
     @operation_raised_error = false
   rescue StandardError => e
-    puts "Got an error: #{e}"
     @operation_raised_error = true
   end
 end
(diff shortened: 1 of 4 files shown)


gems/aws-sdk-support/features/smoke_step_definitions.rb
```diff
@@ -21,7 +21,8 @@
   begin
     @regional_client.send(operation.to_sym, opts)
     @operation_raised_error = false
-  rescue StandardError
+  rescue StandardError => e
+    puts "Got an error: #{e}"
```
mullermp: Debugging

tasks/test.rake
```diff
@@ -47,15 +47,18 @@ end
 
 rule /^test:features:.+$/ do |task|
   dir = "gems/#{task.name.split(':').last}/features"
-  tags = "-t 'not @veryslow'"
+  # Exclude support smoke tests as these require account settings
+  # the logical or with not results in skipping tests marked with
+  # @support AND @smoke only.
+  tags = "-t 'not @veryslow' -t 'not @support or not @smoke'"
```
mullermp, line 53: Is there another way to meet this condition?

mullermp (approved):

tasks/test.rake
alextwoods: I'm not sure - this was the only way I could find to write this.

gems/aws-sdk-support/features/smoke_step_definitions.rb
alextwoods: Removed

tasks/test.rake
alextwoods: I did test that all other smoke tests are run.  And that other support tests would be run (we don't have any.  but if we wrote them in the future they would be run.  I'm not sure if thats what we want or not.  but I was trying to make a very selective filter).

PR #2970, opened 2023-12-19
Fix nested hash when generating params for operationInput tests

alextwoods:

### DIFF
--- a/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/spec/endpoint_provider_spec_class.rb
+++ b/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/spec/endpoint_provider_spec_class.rb
@@ -141,7 +141,7 @@ module AwsSdkCodeGenerator
             case value
             when Hash
               value.each_with_object({}) do |(k, v), o|
-                o[Underscore.underscore(k)] = transform_operation_values(v)
+                o[Underscore.underscore(k).to_sym] = transform_operation_values(v)
               end
             when Array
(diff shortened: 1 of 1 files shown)


mullermp (approved):

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/spec/endpoint_provider_spec_class.rb
```diff
@@ -141,7 +141,7 @@ def transform_operation_values(value)
             case value
             when Hash
               value.each_with_object({}) do |(k, v), o|
-                o[Underscore.underscore(k)] = transform_operation_values(v)
+                o[Underscore.underscore(k).to_sym] = transform_operation_values(v)
```
mullermp, line 144: Does this need to be applied in the underscores in initialize too?

alextwoods: Confusingly, no (which is why we didn't catch this issue on regular, non nested hashes).  In the view we do: 
```
          client.{{operation_name}}(
            {{#operation_params}}
            {{param}}: {{{value}}},
            {{/operation_params}}
          )
```

So the top level key should be a string since we do `{{param}}:`

mullermp: We should probably refactor this to not iterate operation_params and instead render operation_params as is so it's not so confusing.

PR #2972, opened 2024-01-02
Fix re-initialization of some constants in seahorse client

popematt:
I am using the AWS Ruby SDK indirectly, and my logs are full of messages like this (to the point of making my logs unreadable).
```
<redacted path>/aws-sdk-core-3.190.0/lib/seahorse/client/h2/handler.rb:14: warning: already initialized constant Seahorse::Client::H2::NETWORK_ERRORS
<redacted path>/aws-sdk-core-3.190.0/lib/seahorse/client/h2/handler.rb:14: warning: previous definition of NETWORK_ERRORS was here
```
There are three different constants for which this is happening. This PR adds a guard for the initialization of each of those three constants.

I am aware that it is possible that this _might_ also be fixed by changing the way the other gem is using the AWS SDK, but I have not investigated that because I do not have enough control/influence over that (non-public, proprietary) gem to make any changes to it.

**By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.**

### DIFF
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Fix re-initialization of some constants in seahorse client.
+
 3.190.1 (2023-12-20)
 ------------------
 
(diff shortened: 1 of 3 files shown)


popematt:
Looks like I have some work to do here...

popematt:
Closing since this won't fix my problem anyway because of the non-public dependency I have.

PR #2975, opened 2024-01-10
S3 Access Grants

mullermp:
Adds S3 Access Grants support as a plugin


mullermp:
TODO - remove hard dependency on s3control and make it a soft dependency?

mullermp:
TODO - consider a new gem for this in line with what other languages are doing

gems/aws-sdk-core/lib/aws-sdk-core/lru_cache.rb
```diff
@@ -0,0 +1,72 @@
+# frozen_string_literal: true
+
+module Aws
+  # @api private
+  # A simple thread safe LRU cache
+  class LRUCache
+    def initialize(options = {})
+      @max_entries = options[:max_entries] || 100
+      @expiration = options[:expiration]
```
alextwoods, line 12: I know its api private, but some documentation on the options would still help I think (eg, what is expiration? a time? number of seconds?)

gems/aws-sdk-s3/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Feature - Support S3 Access Grants authentication. Access Grants can be enabled with the `s3_access_grants` option, and custom options can be passed into the `access_grants_credentials_provider` option. This feature is enabled if `aws-sdk-s3control` is installed.
```
alextwoods: The wording is a little confusing - s3_access_grants is always enabled if s3control is installed?  Should it be something like "this feature requires s3control to be installed"?

gems/aws-sdk-s3/lib/aws-sdk-s3/access_grants_credentials_provider.rb
```diff
@@ -0,0 +1,233 @@
+# frozen_string_literal: true
+
+module Aws
+  module S3
+    # @api private
+    ACCESS_GRANTS_CREDENTIALS_CACHE = LRUCache.new
```
alextwoods: It might be better to memoize access to these so that we're not creating caches unless they are actually used.  I know they're not huge, but I think a general best practice would be to not create caches that aren't going to be used.
Also - We may want to specify the size here when creating them instead of relying on the default in core (letting us more easily change the default in core without impacting the value used for access grants in the future).

gems/aws-sdk-s3/lib/aws-sdk-s3/access_grants_credentials_provider.rb
```diff
@@ -0,0 +1,233 @@
...
+      # @option options [Boolean] :fallback (false) When true, if access is
+      #  denied, the provider will fall back to the configured credentials.
+      # @option options [Boolean] :caching (true) When true, credentials and
+      #  bucket account ids will be cached.
+      # @option options [Callable] :before_refresh Proc called before
+      #  credentials are refreshed.
+      def initialize(options = {})
+        @s3_control_options = options.delete(:s3_control_client_options) || {}
+        @s3_client = options.delete(:s3_client)
+        @sts_client = options.delete(:sts_client)
```
alextwoods, line 45: Is this always required or can we construct a reasonable default?

gems/aws-sdk-s3/lib/aws-sdk-s3/access_grants_credentials_provider.rb
```diff
@@ -0,0 +1,233 @@
...
+
+      attr_accessor :s3_client
+
+      private
+
+      def s3_control_client(bucket_region)
+        @s3_control_clients[bucket_region] ||= begin
+          credentials = s3_client.config.credentials
+          config = { credentials: credentials }.merge(@s3_control_options)
+          Aws::S3Control::Client.new(config.merge(
```
alextwoods, line 85: why two separate merges?

gems/aws-sdk-s3/lib/aws-sdk-s3/access_grants_credentials_provider.rb
```diff
@@ -0,0 +1,233 @@
...
+      end
+
+      attr_accessor :s3_client
+
+      private
+
+      def s3_control_client(bucket_region)
+        @s3_control_clients[bucket_region] ||= begin
+          credentials = s3_client.config.credentials
+          config = { credentials: credentials }.merge(@s3_control_options)
```
alextwoods, line 84: this lets the s3control options override these credentials.... Is that intetional?

gems/aws-sdk-s3/lib/aws-sdk-s3/access_grants_credentials_provider.rb
```diff
@@ -0,0 +1,233 @@
...
+            use_dualstack_endpoint: s3_client.config.use_dualstack_endpoint
+          ))
+        end
+      end
+
+      def cached_credentials_for(target, permission, credentials)
+        cached_creds = broad_search_credentials_cache_prefix(target, permission, credentials)
+        return cached_creds if cached_creds
+
+        if %w[READ WRITE].include?(permission)
```
alextwoods, line 97: I may have missed where this happens - but what about just READ?   And would this ever be false? are there ever permission other than READ or WRITE?

gems/aws-sdk-s3/lib/aws-sdk-s3/access_grants_credentials_provider.rb
```diff
@@ -0,0 +1,233 @@
...
+      # returns the account id for the configured credentials
+      def account_id_for_credentials(region, credentials)
+        # use resolved credentials to check for account id
+        if credentials.respond_to?(:account_id) && credentials.account_id &&
+           !credentials.account_id.empty?
+          credentials.account_id
+        else
+          @sts_client ||= Aws::STS::Client.new(
+            credentials: s3_client.config.credentials,
+            region: region,
```
alextwoods, line 212: Do we need to keep a hash of sts client per region as well here?

gems/aws-sdk-s3/lib/aws-sdk-s3/express_credentials_provider.rb
```diff
@@ -2,16 +2,23 @@
...
+      # @option options [Client] :client The S3 client used to create the
+      #  session.
       # @option options [String] :session_mode (see: {Client#create_session})
+      # @option options [Boolean] :caching (true) When true, credentials will
+      #  be cached.
       # @option options [Callable] :before_refresh Proc called before
       #   credentials are refreshed.
       def initialize(options = {})
         @client = options.delete(:client)
+        @caching = options.delete(:caching) || true
```
alextwoods: Is this option used anywhere? I dont see it get checked

gems/aws-sdk-s3/lib/aws-sdk-s3/access_grants_credentials_provider.rb
mullermp: No, I believe not. It's effectively global. It should work with any region. The only consideration here is probably latency..

gems/aws-sdk-s3/lib/aws-sdk-s3/access_grants_credentials_provider.rb
mullermp: In theory there's a READWRITE permission, however, none of the current defaults have it. I believe delete objects will need it eventually. In the API there's:
```
    "Permission":{
      "type":"string",
      "enum":[
        "READ",
        "WRITE",
        "READWRITE"
      ]
    },
```

gems/aws-sdk-s3/lib/aws-sdk-s3/access_grants_credentials_provider.rb
alextwoods: Partition would be a consideration as well... but I assume somethign else prevents us from ever using these cross partition?

gems/aws-sdk-s3/lib/aws-sdk-s3/access_grants_credentials_provider.rb
mullermp: Yeah. It allows the s3 control client to use a separate set of credentials if necessary, but defaults to the same credentials used in the s3 client.

gems/aws-sdk-s3/lib/aws-sdk-s3/access_grants_credentials_provider.rb
mullermp: I recall this being intentional.. it uses any custom options, but then overrides those again with the region and endpoint options of s3.

gems/aws-sdk-s3/lib/aws-sdk-s3/access_grants_credentials_provider.rb
mullermp: There are defaults for all of these, just created lazily. S3 Control and STS are created here, and S3 client is created as self.

gems/aws-sdk-s3/lib/aws-sdk-s3/access_grants_credentials_provider.rb
mullermp: I did this for all 3 caches. I don't think it's breaking but please take a look.

gems/aws-sdk-s3/lib/aws-sdk-s3/express_credentials_provider.rb
```diff
@@ -18,21 +20,35 @@ class ExpressCredentialsProvider
       #   credentials are refreshed.
       def initialize(options = {})
         @client = options.delete(:client)
-        @caching = options.delete(:caching) || true
+        @caching = options.delete(:caching) != false
```
alextwoods, line 23: good catch

alextwoods (approved):
Nice, LGTM!

gems/aws-sdk-s3/lib/aws-sdk-s3/express_credentials_provider.rb
mullermp: I only noticed because you had a different comment ...

gems/aws-sdk-s3/lib/aws-sdk-s3/access_grants_credentials_provider.rb
mullermp: Talked offline - this is what Java does today. I think it's fine.

PR #2986, opened 2024-02-12
Prepare for Ruby 3.4 release - add dependencies on previously default gems (base64/bigdecimal)

### LINKED ISSUE #2984: Gem displays deprecation warning in ruby 3.3
### Describe the bug

When running the gem in ruby 3.3 it displays the following warning

~~~
/home/nathansamson/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/aws-sdk-core-3.191.1/lib/aws-sdk-core/assume_role_web_identity_credentials.rb:5: warning: base64 was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0. Add base64 to your Gemfile or gemspec. Also contact author of aws-sdk-core-3.191.1 to add base64 into its gemspec.
~~~

### Expected Behavior

It should not do this

### Current Behavior

It displays a warning in ruby 3.3

### Reproduction Steps

Use the gem

### Possible Solution

Include base64 in the gemspec

_No response_

aws-sdk-core:3.191.1

Ruby 3.3, linux

alextwoods:
Fixes #2984 

### DIFF
--- a/services.json
+++ b/services.json
@@ -367,7 +367,7 @@
       "Aws::DynamoDB::Plugins::CRC32Validation"
     ],
     "dependencies": {
-      "bigdecimal": "~> 3"
+      "bigdecimal": null
     }
   },
   "DynamoDBStreams": {
@@ -376,7 +376,7 @@
       "Aws::DynamoDBStreams::Plugins::SimpleAttributes"
     ],
     "dependencies": {
-      "bigdecimal": "~> 3"
+      "bigdecimal": null
     }
   },
   "EBS": {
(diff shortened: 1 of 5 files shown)


services.json
```diff
@@ -381,10 +387,7 @@
     "addPlugins": [
       "Aws::EC2::Plugins::CopyEncryptedSnapshot",
       "Aws::EC2::Plugins::RegionValidation"
-    ],
-    "dependencies": {
-      "aws-sigv4": "~> 1.0"
```
jterapin, line 386: I'm thinking that we are getting rid of the `aws-sigv4` dependency here since it is already noted in the core, right? No reason to restate this dependency in both areas and etc.

alextwoods: Not only is it a dependency in core - the codegen already separately adds it as a dependency: https://github.com/aws/aws-sdk-ruby/blob/version-3/build_tools/services.rb#L162

mullermp (approved):

PR #2987, opened 2024-02-19
Remove dependencies on bigdecimal and base64

alextwoods:
Forcing dependencies on bigdecimal may break use cases where dev tools are not available which were previously working.  

### DIFF
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Remove base64 as dependency.
+
 3.191.2 (2024-02-14)
 ------------------
 
@@ -20,7 +22,7 @@ Unreleased Changes
 
 * Feature - Updated Aws::SSO::Client with the latest API changes.
 
-* Feature - Add RBS signature files to support static type checking.
+* Feature - Add RBS signature files to support static type checking
 
 3.190.3 (2024-01-16)
 ------------------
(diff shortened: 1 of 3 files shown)


mullermp (approved):

jterapin:
Should we add a changelog notifying this? Since we have one when we initially added this? https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/CHANGELOG.md

VitaliySerov:
Is this a good idea?

This this again cause warning on `ruby-3.3` - even CI shows it:
```
warning: base64 was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0. Add base64 to your Gemfile or gemspec.
```


 I think this dependency is required for future `ruby-3.4`

Right now if we run sdk code on `ruby-head` (which is not released, yeah)

it will cause

```
Failure/Error: require 'aws-sdk-s3'

LoadError:
  cannot load such file -- base64
  ```
  
  Should this problem be solved little bit more sophisticated?
  
 This change basically conflicts with previous one?
 
 https://github.com/aws/aws-sdk-ruby/issues/2984

mullermp:
We reverted it because it (and bigdecimal) caused issues. We will strategize a plan for Ruby 3.4, but I think the plan unfortunately has to be "bring your own dependency" since we have no mechanism in the gemspec to add dependencies dependent on Ruby versions, unless Ruby 3.4 is the minimum we support.

nathansamson:
I am a bitconfused what the error/problem was

"Forcing dependencies on bigdecimal may break use cases where dev tools are not available which were previously working."

I assume people install the gem with either "gem install" or indirectly with a "bundle install" in both cases the dependencies would be  fetched and installed from the internet?

mullermp:
BigDecimal requires native extension tools on the machine (it is a native extension gem). The gem is fetched but fails to install.

PR #2988, opened 2024-02-23
Update protocol tests

jterapin:
**Objectives**
* Replace existing protocol-tests with Smithy's definition of protocol-tests
* Update the test runner to consume these accordingly (and clean up)
* Resolve test failures
* Implement a mechanism to skip test cases (and have a log of skipped cases with reasons why)
* Carry over protocol-tests that was not covered by Smithy's definition (i.e. eventstreams)
* Update existing protocol-tests (api-gateway, carry overs) to include descriptive format

**Starting total of failures: 349** 
**Current total of failures: 0** but 2 ignored due to compatibility

Current standing:

* All protocol-tests passes but more work required:
  * ~~We have 58 ignored tests~~ We just have 1 ignore test that we need to keep due to limitations on the JSON Engine
  * ~~11 tests cases rely on Smithy fixing the C2J translation (may require additional work after)~~
  * ~~4 test cases cannot be fixed due complex issue~~
  * ~~43 test cases to be resolved when test runner is updated to handle error cases (need to implement)~~
  * ~~There are 6 failures in our gem specs after making changes (these need to be investigated and resolve)~~
  * ~~There are 44 failures in the on build_tools directory (need to investigate and resolve)~~
* ~~Clean up code (optimize approaches, etc)~~
* ~~Create a comprehensive log of changes (intended for changelog)~~
* ~Create a list of services to manually test certain operations that are affected by some SDK core changes~
* Manual testing/~~run integration tests~~ **MANUAL TESTING IN PROGRESS**
* ~~Bring over "extra" test cases (aka eventstreams) and update the formatting~~
* ~~Update api-gateway protocol-tests formatting if needed~~
* ~~CR review and update based on feedback~~

--- 


jterapin:
Current total of failures: 252

jterapin:
**Current status**: 197 failures, 50 ignored

build_tools/aws-sdk-code-generator/spec/protocol-tests-ignore-list.json
```diff
@@ -0,0 +1,98 @@
+{
+  "ec2" :
+    {
+      "input" : [
+        "Ec2QuerySupportsNaNFloatInputs",
```
mullermp: Will we be working through this list or is it set in stone?

jterapin: Half n half. I plan to do an initial run of fixes and trace back to do some in-depth cases (like the `NaN`, `Infinity`, etc) in a separate PR.

jterapin:
Current status: 160 failures, 51 ignored

jterapin:
Current failures: 57

gems/aws-sdk-core/lib/aws-sdk-core/json/builder.rb
```diff
@@ -49,6 +55,7 @@ def format(ref, value)
         when MapShape       then map(ref, value)
         when TimestampShape then timestamp(ref, value)
         when BlobShape      then encode(value)
+        when FloatShape     then Util.serialize_number(value)
```
mullermp, line 58: Is this serialization only specific to JSON? If so I think it's appropriate to put serialize in builder and deserialize in parser instead of putting this in the global util namespace.

gems/aws-sdk-core/lib/aws-sdk-core/param_validator.rb
```diff
@@ -152,11 +156,11 @@ def shape(ref, value, errors, context)
       when MapShape then map(ref, value, errors, context)
       when DocumentShape then document(ref, value, errors, context)
       when StringShape
-        unless value.is_a?(String)
+        unless value.is_a?(String) || value.nil?
```
mullermp: I would avoid `unless` with more than one condition, I think that's also a rubocop recommendation.

gems/aws-sdk-core/lib/aws-sdk-core/param_validator.rb
```diff
@@ -168,7 +172,7 @@ def shape(ref, value, errors, context)
           errors << expected_got(context, "a Time object", value)
         end
       when BooleanShape
-        unless [true, false].include?(value)
+        unless [true, false].include?(value) || value.nil?
```
mullermp: I think nil can be inside this array if you decide to keep unless here.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/jsonvalue_converter.rb
```diff
@@ -43,15 +43,17 @@ def convert_jsonvalue(m, ref, params, context)
           end
         end
 
-        def serialize_jsonvalue(v, context)
-          unless v.respond_to?(:to_json)
+        def serialize_jsonvalue(value, context)
+          return value if value.is_a?(String)
```
mullermp: Is this correct? json values can be strings however I think it still needs to respond to that. What test does this break?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/protocols/rest_xml.rb
```diff
@@ -5,9 +5,21 @@ module Plugins
...
+               (body.respond_to?(:size) && body.size > 0)
+              context.http_request.headers['Content-Type'] ||=
+                'application/xml'
+            end
+            @handler.call(context)
+          end
+        end
+
         handler(Rest::Handler)
+        handler(ContentTypeHandler, priority: 30)
```
mullermp: I'd move this priority closer to the builder. You may be able to just remove the priority as the default might be good enough - if it's out of order, you can register it before the rest handler.

gems/aws-sdk-core/lib/aws-sdk-core/rest/request/endpoint.rb
```diff
@@ -30,30 +30,47 @@ def uri(base_uri, params)
...
           uri.path = path.gsub(/{.+?}/) do |placeholder|
             param_value_for_placeholder(placeholder, params)
           end
         end
 
         def param_value_for_placeholder(placeholder, params)
           name = param_name(placeholder)
-          value = params[name].to_s
+          param_shape = @rules.shape.member(name).shape
+          value = case param_shape.is_a?(Seahorse::Model::Shapes::TimestampShape)
```
mullermp: Your case could be `case param_shape` and your when statement can be TimestampShape.

gems/aws-sdk-core/lib/aws-sdk-core/rest/request/headers.rb
```diff
@@ -20,7 +20,8 @@ def initialize(rules)
         def apply(http_req, params)
           @rules.shape.members.each do |name, ref|
             value = params[name]
-            next if value.nil?
+            next if value.nil? || (ref.shape).is_a?(StringShape) && value.empty?
```
mullermp: Use parenthesis to ensure order here.

gems/aws-sdk-core/lib/aws-sdk-core/json/builder.rb
alextwoods: This is generic to all protocols (see our V4/hearth usage)

gems/aws-sdk-core/lib/aws-sdk-core/plugins/jsonvalue_converter.rb
jterapin: Here's the context: This migrated [test case](https://github.com/smithy-lang/smithy/blob/ac925618457ee9d97da5fd3b63b80e9952ab5cd6/smithy-aws-protocol-tests/model/awsJson1_1/kitchen-sink.smithy#L30) was failing.

Previous to the change above, the output body was escaping twice: `{"JsonValue"=>"\"{\\\"string\\\":\\\"value\\\",\\\"number\\\":1234.5,\\\"boolTrue\\\":true,\\\"boolFa...\\\":false,\\\"array\\\":[1,2,3,4],\\\"object\\\":{\\\"key\\\":\\\"value\\\"},\\\"null\\\":null}\""}`

The first escape happens in the above and second escape during the building the body in the request.

Yeah I could add a bit like, `if value.is_a?(String) && value.responds_to(:to_json)` in the above but shouldn't all ruby strings be able to respond to `to_json`? Lemme know if I'm missing something here.

jterapin:
Current standing: 
- All protocol-tests passes but more work required:
	- We have 58 ignored tests
		- 11 tests cases rely on Smithy fixing the C2J translation (may require additional work after)
		- 4 test cases cannot be fixed due complex issue 
		- 43 test cases to be resolved when test runner is updated to handle error cases (need to implement)
- There are 6 failures in our gem specs after making changes (these need to be investigated and resolve)
- There are 44 failures in the on `build_tools` directory (need to investigate and resolve)
- Clean up code (optimize approaches, etc)
- Create a comprehensive log of changes (intended for changelog)
- Create a list of services to manually test certain operations that are affected by some SDK core changes
- Manual testing/run integration tests (have another team member to run integration tests on their end to verify)
- Bring over "extra" test cases (aka eventstreams) and update the formatting
- Update api-gateway protocol-tests formatting if needed
- CR review and update based on feedback
- After CR approvals, I'm going to rubocop some of the files I touched since I want to leave the file cleaner than when I found it 🧹 ✨

gems/aws-sdk-core/CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
...
+* Issue - Correctly serialize URI label bindings in Rest requests.
+* Issue - Correctly serialize and parse header bindings in Rest services.
+* Issue - Ensure that null and empty headers are not sent in Rest requests.
+* Issue - Ensure keys in query maps do not override modeled keys in Rest requests.
+* Issue - Ensure empty blob payloads are omitted in Rest requests.
+* Issue - Support parsing `NaN`, `Infinity` and `-Infinity` float values.
+* Issue - Apply appropriate `xmlName` for flattened lists and maps in Rest-XML services.
+* Issue - Fix serializing and deserializing `xmlNamespace` on nested structures for Rest-XML services.
+* Issue - Fix deserializing an empty self-closed blob to produce an empty string.
+* Issue - Support parsing of different formats of error data in Rest-XML services.
```
jterapin, line 23: **Note:** Keeping this log here during review but probably will create a ticket to keep this log there instead and link the issue ticket here

mullermp: I think it's fine to dump it all here for posterity

gems/aws-sdk-core/lib/aws-sdk-core/xml/builder.rb
```diff
@@ -121,19 +122,27 @@ def timestamp(ref, value)
...
 
-      def shape_attrs(ref)
-        if xmlns = ref['xmlNamespace']
-          if prefix = xmlns['prefix']
-            { 'xmlns:' + prefix => xmlns['uri'] }
-          else
-            { 'xmlns' => xmlns['uri'] }
+      def shape_attrs(ref, top_level: false)
+        # for non top level shapes, do NOT fall back to metadata key on shape
+        xmlns = top_level ? ref['xmlNamespace'] : ref.metadata['xmlNamespace']
```
jterapin: Ignore reviewing anything relating to `xmlNamespace` since I'm waiting on some clarification from the Smithy so the behavior here may change

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/rbs/keyword_argument_builder.rb
```diff
@@ -79,14 +79,14 @@ def ref_value(ref, i, visited)
...
           else
-            "::String"
+            '::String'
           end
         when 'list' then list(s, i, visited)
         when 'map' then map(s, i, visited)
-        when 'boolean' then "bool"
-        when 'integer', 'long' then '::Integer'
+        when 'boolean' then 'bool'
+        when 'integer', 'long', 'short' then '::Integer'
```
alextwoods: I can't remember, did we decide to support short? I thought it wasn't in c2j?

build_tools/aws-sdk-code-generator/spec/protocol-tests/input/api-gateway.json
```diff
@@ -1,20 +1,22 @@
...
       {
+        "id": "ApiGatewayMarshalEmptyBody",
         "given": {
           "http": {
             "method": "GET",
             "requestUri": "/test/noauth/scalars"
           },
           "name": "OperationName"
         },
+        "description": "No parameters set marshals as empty body",
```
alextwoods, line 19: Yay for descriptions!

build_tools/aws-sdk-code-generator/spec/protocol_tests_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+          end
+
+          input_shape = client.config.api.operation(:operation_name).input
+          request_params = ProtocolTestsHelper.format_data(input_shape, test_case['params'] || {})
+          client.operation_name(request_params)
+
+          matcher.match_req_host(test_case, ctx.http_request, self)
+          matcher.match_req_method(test_case, ctx.http_request, self)
+          matcher.match_req_uri(test_case, ctx.http_request, self)
+          matcher.match_req_headers(suite, test_case, ctx.http_request, self)
```
alextwoods: Rather than using `let` to create an instance of the Matcher class - we could just define these as self methods on a Matcher module.

This would make sense if we had a separate Matcher class for each protocol or something where the behavior would differ I think....

build_tools/aws-sdk-code-generator/spec/protocol_tests_spec_helper.rb
```diff
@@ -0,0 +1,291 @@
...
+PROTOCOL_TESTS_IGNORE_LIST_PATH = File.join(File.dirname(__FILE__), 'protocol-tests-ignore-list.json')
+
+# This module contains helpers relating to protocol-tests
+module ProtocolTestsHelper
+  class << self
+
+    def check_ignore_list(protocol, test_id, test_type)
+      return nil if protocol.include?('extras')
+
+      ignore_list = JSON.parse(File.read(PROTOCOL_TESTS_IGNORE_LIST_PATH))[protocol]
```
alextwoods: The efficiency probably doesn't really matter much here, but this requires us to repeatedly read and parse this file.  We should probably do it once (make a memoized method).

gems/aws-sdk-core/lib/aws-sdk-core/plugins/jsonvalue_converter.rb
alextwoods: Yeah - I'm going back and forth on this one.   The spec says "When marshaling API requests, SDKs must first encode the shape reference's value into a JSON string. ".  So in this case, if we are given a String, we should probably encode it as a JSON string (which would add extra escaping).

However, to get the test to pass - I think we would need to add special case handling when a test case has input for jsonvalue that we first do a json.parse and then pass that as the input.

alextwoods (commented):
Looking good!  This is an incomplete review (made it only about 1/3 of the way through core changes, but leaving comemments now since I need to move on to other tasks for now).

build_tools/aws-sdk-code-generator/spec/protocol_tests_spec_helper.rb
```diff
@@ -0,0 +1,291 @@
...
+        fixtures[protocol] ||= {}
+        fixtures[protocol][input_output] = path
+      end
+      fixtures
+    end
+
+    def each_test_case(context, fixture_path)
+      return unless fixture_path
+
+      Aws::Json.load_file(fixture_path).each do |suite|
```
mullermp: Minor. JSON.load is fine here - Aws::Json is specifically for engine usage in the SDK (and correctly used below). Reading fixtures, I'd use JSON

build_tools/aws-sdk-code-generator/spec/protocol_tests_spec_helper.rb
```diff
@@ -0,0 +1,291 @@
...
+        end
+      end
+    end
+
+    # creates service client for req/resp testing
+    def client_for(suite, test_case, id)
+      name = "Service_#{id}"
+      unless Aws.const_defined?(name)
+        operation = test_case['given']
+        operation['http'] ||= { 'method' => 'POST', 'requestUri' => '/' }
```
mullermp: Is this correct as a default? Some protocols don't always POST to /

build_tools/aws-sdk-code-generator/spec/protocol_tests_spec_helper.rb
```diff
@@ -0,0 +1,291 @@
...
+            'operations' => {
+              'OperationName' => test_case['given']
+            },
+            'shapes' => suite['shapes']
+          },
+          gem_dependencies: { 'aws-sdk-core' => '3' },
+          gem_version: '1.0.0'
+        }
+        if suite['metadata']['protocol'] == 'api-gateway'
+          opts[:default_endpoint] = 'https://foobar.us-west-2.amazonaws.com/test'
```
mullermp: I think the service in this case would be `foo.execute-api` but totally minor

build_tools/aws-sdk-code-generator/spec/protocol_tests_spec_helper.rb
```diff
@@ -0,0 +1,291 @@
...
+      client_class.new(
+        regional_endpoint: true,
+        endpoint: suite['clientEndpoint'] || 'http://example.com',
+        region: 'us-east-1',
+        credentials: Aws::Credentials.new('akid', 'secret')
+      )
+    end
+
+    # formats response data
+    def data_to_hash(obj)
```
mullermp, line 86: Does `to_h` on `Aws::Structure` not already do this?

build_tools/aws-sdk-code-generator/spec/protocol_tests_spec_helper.rb
```diff
@@ -0,0 +1,291 @@
...
+require 'rexml/document'
+require 'rspec/expectations'
+
+PROTOCOL_TESTS_IGNORE_LIST_PATH = File.join(File.dirname(__FILE__), 'protocol-tests-ignore-list.json')
+
+# This module contains helpers relating to protocol-tests
+module ProtocolTestsHelper
+  class << self
+
+    def check_ignore_list(protocol, test_id, test_type)
```
mullermp: You can also pass `it` here and do the skipping here.

build_tools/aws-sdk-code-generator/spec/protocol_tests_spec_helper.rb
```diff
@@ -0,0 +1,291 @@
...
+          end
+        else raise "unsupported protocol: `#{protocol}`"
+        end
+        it.expect(body).to(eq(expected_body))
+      end
+    end
+
+    def match_resp_data(test_case, resp, it)
+      data = ProtocolTestsHelper.data_to_hash(resp.data)
+      expected_data = if error_case?(test_case)
```
mullermp: For formatted if/else and case, just put it on the next line.
```
expected_data =
  if ..
  else ..
```

gems/aws-sdk-core/lib/aws-sdk-core/rest/request/body.rb
```diff
@@ -45,20 +49,58 @@ def build_body(params)
...
+
+        def update_payload_content_type(req)
+          # headers could be already populated if specified on input shape
+          case @rules[:payload_member].shape
+          when BlobShape
+            req.headers['Content-Type'] ||= 'application/octet-stream'
+          when StringShape
+            req.headers['Content-Type'] ||= 'text/plain'
+          when UnionShape
+            req.headers['Content-Type'] ||= if xml_builder?
```
mullermp: Ditto on formatting, it's just easier to read the other way.

gems/aws-sdk-core/lib/aws-sdk-core/rest/response/body.rb
```diff
@@ -30,6 +31,13 @@ def apply(body, data)
 
         private
 
+        def empty_blob_body?(body)
+          return false if !@rules[:payload_member].shape.is_a?(BlobShape) ||
+                          @rules[:payload_member]['streaming']
+
+          true if body.respond_to?(:size) && body.size.zero?
```
mullermp: These operands already return a boolean, no need to return true or false accordingly, you can structure this method to just return the output of whether the shape is a Blob, body size 0, etc. Could also break this into two methods like "is_streaming?"

gems/aws-sdk-core/lib/aws-sdk-core/xml/builder.rb
mullermp: What exactly is the question? Maybe we can figure it out

gems/aws-sdk-s3/CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 Unreleased Changes
 ------------------
 
+* Issue - Ignore `Content-Type` header in the pre-signer.
+
+=======
+=======
```
mullermp: Merge issue here

gems/aws-sdk-s3/lib/aws-sdk-s3/presigner.rb
```diff
@@ -15,6 +15,7 @@ class Presigner
         'amz-sdk-request',
         'cache-control',
         'content-length', # due to a ELB bug
+        'content-type',
```
mullermp: Why are we doing this? And do presigned urls still work?

mullermp (approved):
Nice!!

Let's get that last test fixed.

Have we considered whether we need to global bump all gems? It may be a good idea to do so especially with so many fixes.

We absolutely need to make sure that this is backwards compatible (no new behaviors on returned methods especially syntax or type changes)

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/rbs/keyword_argument_builder.rb
jterapin: Ah, that is correct. The protocol-tests usage of short has been changed to Integer once there was some conversation popped up. I'll revert the commit. Thank you!

gems/aws-sdk-core/lib/aws-sdk-core/rest/request/body.rb
```diff
@@ -45,20 +49,58 @@ def build_body(params)
...
+            end
             serialize(@rules[:payload_member], params) if params
           else
             params = body_params(params)
             serialize(@rules, params) unless params.empty?
           end
         end
 
+        def update_payload_location_name
+          @rules[:payload_member].location_name =
```
alextwoods: So we're modifying the actual rules/client_api here?  It would be better to not modify these if we can.  

Instead, can we move this logic into the actual serialization logic?

gems/aws-sdk-core/lib/aws-sdk-core/xml/error_handler.rb
```diff
@@ -48,16 +47,34 @@ def error_data(context, code)
             # match modeled shape name
             error_shape_code = rule.shape['error']['code'] if rule.shape['error']
             match = (code == error_shape_code || code == rule.shape.name)
-            if match && rule.shape.members.any?
-              data = Parser.new(rule).parse(context.http_response.body_contents)
-            end
+            next unless match && rule.shape.members.any?
+
+            data = parse_error_data(rule, context.http_response.body_contents)
```
alextwoods, line 51: I know I made this change, but I'm still slightly worried about the change.  My understanding of the parsing of XML errors is that in all protocols they MUST have the `<Errors>` or `<Errors>` tags - and so this old code would NEVER have worked.  I think this ended up being fine because these services don't actually model many errors?

gems/aws-sdk-core/spec/aws/plugins/jsonvalue_converter_spec.rb
```diff
@@ -109,7 +109,8 @@ module Plugins
         }
         ruby_types.each do |_k, v|
           resp = client.foo(header_field: v)
-          expect(resp.context.params[:header_field]).to eq(v.to_json)
+          expected_value = v.is_a?(String) ? v : v.to_json
```
alextwoods: I'm not 100% certain on this behavior anymore.  This definitely passes the protocol tests as we have them written, but technically does not encode json strings correctly.

I think we may instead need to modify our protocol test runner to handle this case differently - specifically - if we see a string param for an input that is marked with the jsonvalue trait, then we do a parse on it.  I think worth asking Kevin about to confirm what the behvior here should be.

gems/aws-sdk-core/lib/aws-sdk-core/xml/builder.rb
jterapin: Discussing outside of PR

gems/aws-sdk-core/lib/aws-sdk-core/rest/request/body.rb
jterapin: I didn't realize I could do that so this is niceee

build_tools/aws-sdk-code-generator/spec/protocol_tests_spec_helper.rb
jterapin: I grandfathered this snippet from previous test runner so I'm not sure on the intention of setting that has default.  That being said -  all C2J translation of Smithy protocol tests always has these areas defined so I feel comfortable removing this fall back. 

(I will update the older "extras" tests to include this bit but there's  very few test cases)

build_tools/aws-sdk-code-generator/spec/protocol_tests_spec_helper.rb
jterapin: It is very [similiar](https://github.com/aws/aws-sdk-ruby/blob/a703da0525ac26d2ac4a68253b3718b6d86461b8/gems/aws-sdk-core/lib/aws-sdk-core/structure.rb#L31-L49) but it does not `.read` when an object is IO or StringIO. I ran the tests, using`resp.data.to_h` and there were test failures. Based on this observation, I will opt to keeping this around since we would want to keep the `Aws::Structure` alone without any modifications.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/jsonvalue_converter.rb
jterapin: I asked Kevin for clarification prior to making any changes. He said it seems about right so I'll add in a handling to resolve this test case on our end (and revert the changes I made relating to it)

gems/aws-sdk-core/spec/aws/plugins/jsonvalue_converter_spec.rb
jterapin: See my comment [here](https://github.com/aws/aws-sdk-ruby/pull/2988/files#r1550748493) on resolution

gems/aws-sdk-core/lib/aws-sdk-core/xml/error_handler.rb
jterapin: Per our offline conversation, I'm going to scan through our existing services to see if there's any existence of a modeled error response (this specific error handler is used for query, ec2 and rest-xml protocols so not many services to review) and also make some calls to see how things are being parsed if defined.

gems/aws-sdk-core/lib/aws-sdk-core/rest/request/body.rb
jterapin: Yes, agreed. I'm going to look into having a keyword arg to pass this specific information down to serialization.

build_tools/aws-sdk-code-generator/spec/protocol_tests_spec_helper.rb
jterapin: Yeah, that totally makes sense - I think this section was something that I grandfathered in from old test runner. 

Although, I would lean toward leaving as it is - to keep everything on this file consistent?

gems/aws-sdk-core/lib/aws-sdk-core/xml/builder.rb
jterapin: For transparency, I received some behavior clarification on `xmlNamespace` from Kevin. Basically - we are walking back all the changes we made here since this specific protocol-test here need to be fixed.

build_tools/aws-sdk-code-generator/spec/protocol_tests_spec_helper.rb
```diff
@@ -0,0 +1,306 @@
...
+            expected_body = expected_body.split('&').sort.join('&')
+          when 'json'
+            body = Aws::Json.load(body)
+            expected_body = Aws::Json.load(expected_body)
+          when 'rest-json'
+            if body[0] == '{'
+              body = Aws::Json.load(body)
+              expected_body =
+                case expected_body
+                  when ''
```
jterapin, line 231: Note to self, leave a comment here

gems/aws-sdk-core/lib/aws-sdk-core/json/parser.rb
```diff
@@ -69,6 +69,9 @@ def list(ref, values, target = nil)
       def map(ref, values, target = nil)
         target = {} if target.nil?
         values.each do |key, value|
+          # drops key-value pair if nil value
+          next if value.nil?
```
jterapin, line 73: Note to self - leave a comment here

build_tools/aws-sdk-code-generator/spec/protocol_tests_spec_helper.rb
mullermp: Technically it was wrong before. Regardless of engine, we'd want this file parsed the same way each time. In theory all engines should parse this the same :shrug:

build_tools/aws-sdk-code-generator/spec/protocol_tests_spec_helper.rb
mullermp: Quite possibly structure.to_h should also do that read. I think the to_s method is conscious of length. Maybe a discussion with Alex.

build_tools/aws-sdk-code-generator/spec/protocol-tests-ignore-list.json
```diff
@@ -0,0 +1,33 @@
...
+    "output" : []
+  },
+  "rest-json" : {
+    "input" : [],
+    "output" : []
+  },
+  "rest-xml" : {
+    "input" : [],
+    "output" : [
+      {"SimpleScalarPropertiesComplexEscapes": "OxEngine does not handle all escape cases but other engines supports it"}
```
jterapin, line 26: Note to self to investigate

gems/aws-sdk-s3/lib/aws-sdk-s3/presigner.rb
jterapin: The rest-xml protocol did not have content-type header in most cases... which caused failures in our protocol-tests. This PR adds this [here](https://github.com/aws/aws-sdk-ruby/blob/8df5f93263aa722e10e0fcad58bd729ef92ddf49/gems/aws-sdk-core/lib/aws-sdk-core/plugins/protocols/rest_xml.rb#L8).

We had integration tests for presigned urls that failed after the above addition. To remedy this, I updated here to blacklist the `content-type` header and the integration tests was successful. I plan to do a manual test of presigned url before PR release.

build_tools/aws-sdk-code-generator/spec/protocol_tests_spec_helper.rb
jterapin: From what I understand, introducing `.read` for IO and StringIO will be a breaking change of the behavior of `.to_h`. In addition, this will change the semantic meaning of `to_h` which should be just converting to hash.

build_tools/aws-sdk-code-generator/spec/protocol_tests_spec_helper.rb
alextwoods: Yeah, I'd lean towards consistency in the code of using the engine interface and always use AWS::Json instead of any specific library, even though this is test code vs runtime.  

Although - now that I'm looking, our specs use a mix of both approaches already :-/
* JSON.load: https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/spec/api_helper.rb#L11
* Aws::Json: https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/spec/aws/ecs_credentials_spec.rb#L305

build_tools/aws-sdk-code-generator/spec/protocol-tests-ignore-list.json
```diff
@@ -0,0 +1,35 @@
...
+    "output" : []
+  },
+  "query" : {
+    "input" : [],
+    "output" : []
+  },
+  "rest-json" : {
+    "input" : [],
+    "output" : [
+      {"RestJsonDeserializesDenseSetMapAndSkipsNull": "Seeking clarification from Smithy"}
```
jterapin, line 22: This is a new test case that was added when I sourced updated protocol-tests - I THINK this was supposed to be omitted since I previously had convo with Kevin about sparse/dense maps for C2J. Clarifying with Kevin before addressing any changes if needed

build_tools/aws-sdk-code-generator/spec/protocol_tests_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+          client = ProtocolTestsHelper.client_for(suite, test_case, "Output_#{test_id}")
+          client.handle(step: :send) do |context|
+            context.http_response.signal_headers(
+              test_case['response']['status_code'],
+              test_case['response']['headers'] ||= {}
+            )
+
+            # Base64 encoded binary body is provided for eventstream
+            body = if test_case['response']['eventstream']
```
mullermp: nit: format

build_tools/aws-sdk-code-generator/spec/protocol_tests_spec_helper.rb
```diff
@@ -0,0 +1,312 @@
...
+
+# This module contains helpers relating to protocol-tests
+module ProtocolTestsHelper
+  PROTOCOL_TESTS_IGNORE_LIST_PATH = File.join(File.dirname(__FILE__), 'protocol-tests-ignore-list.json')
+  class << self
+    def ignore_list
+      @ignore_list ||= JSON.parse(File.read(PROTOCOL_TESTS_IGNORE_LIST_PATH))
+    end
+
+    def check_ignore_list(protocol, test_id, test_type)
```
mullermp: nit: can be private now?

gems/aws-sdk-core/lib/aws-sdk-core/rest/request/endpoint.rb
```diff
@@ -30,30 +30,47 @@ def uri(base_uri, params)
...
           uri.path = path.gsub(/{.+?}/) do |placeholder|
             param_value_for_placeholder(placeholder, params)
           end
         end
 
         def param_value_for_placeholder(placeholder, params)
           name = param_name(placeholder)
-          value = params[name].to_s
+          param_shape = @rules.shape.member(name).shape
+          value = case param_shape
```
mullermp: nit: format

gems/aws-sdk-s3/lib/aws-sdk-s3/presigner.rb
mullermp: You may need to change the integration tests but keep the behavior here. Content type is necessary for PUT/POST presigned requests. It should not be present for GET presigned urls.

mullermp (approved):
Nice!

I think we need to verify presigned urls, error cases, and general backwards compatibility. The content-type should not be blocked (see comment below).

I think that after those are confirmed and fixed, any other tests that are skipped can be done as follow ups? That makes review a bit easier too.

build_tools/aws-sdk-code-generator/spec/protocol_tests_spec_helper.rb
jterapin: Yup, good catch

jterapin:
Note to self - do a final lookover to see if there's anything that need to be highlighted in the changelogs

alextwoods (approved):

gems/aws-sdk-s3/lib/aws-sdk-s3/presigner.rb
jterapin: Recent commits now resolved this

gems/aws-sdk-core/lib/aws-sdk-core/xml/error_handler.rb
jterapin: I did some digging on how error responses are coming back so posting here for historic context:

---
**Service**: S3 (rest-xml protocol)
**Operation**: `GetObject` ([model](https://github.com/aws/aws-sdk-ruby/blob/431c33756ae4c3a020a9aa26e3fce9ae0f8673d6/apis/s3/2006-03-01/api-2.json#L589-L611))
**Error Shape**: `InvalidObjectState` ([model](https://github.com/aws/aws-sdk-ruby/blob/431c33756ae4c3a020a9aa26e3fce9ae0f8673d6/apis/s3/2006-03-01/api-2.json#L4953-L4961))
* We received the following error response:
```
-> "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Error><Code>InvalidObjectState</Code><Message>The operation is not valid for the object's storage class</Message><StorageClass>DEEP_ARCHIVE</StorageClass><RequestId>fooID</RequestId><HostId>fooHostId</HostId></Error>"
```
* I was able to inspect the modeled error response like so:
```
(byebug) err.inspect
"#<Aws::S3::Errors::InvalidObjectState: The operation is not valid for the object's storage class>"
(byebug) err.class
Aws::S3::Errors::InvalidObjectState
(byebug) err.message
"The operation is not valid for the object's storage class"
(byebug) err.storage_class
"DEEP_ARCHIVE"
```

As a sanity check, I also looked at custom error code, which is very specific to query protocol: https://smithy.io/2.0/aws/protocols/aws-query-protocol.html#awsquery-error-response-code

**Service**: CloudWatch (query protocol, using the xml's error handler)
**Operation**: `GetDashboard` ([model def](https://github.com/aws/aws-sdk-ruby/blob/431c33756ae4c3a020a9aa26e3fce9ae0f8673d6/apis/monitoring/2010-08-01/api-2.json#L219-L235))
**Error Shape**: `DashboardNotFoundError` ([model def](https://github.com/aws/aws-sdk-ruby/blob/431c33756ae4c3a020a9aa26e3fce9ae0f8673d6/apis/monitoring/2010-08-01/api-2.json#L884-L895))
  *  Note that this error shape has a custom error code defined (`ResourceNotFound`)

We receive the following error response:
```
"<ErrorResponse xmlns=\"http://monitoring.amazonaws.com/doc/2010-08-01/\">\n  <Error>\n    <Type>Sender</Type>\n    <Code>ResourceNotFound</Code>\n    <Message>Dashboard DashboardName does not exist</Message>\n  </Error>\n  <RequestId>fooRequestID</RequestId>\n</ErrorResponse>\n"
```
which brings us to this raise:
```
Aws::CloudWatch::Errors::ResourceNotFound: Dashboard DashboardName does not exist
```
This is intended behavior from what I read on query protocol per Smithy Docs linked above

gems/aws-sdk-core/lib/aws-sdk-core/xml/parser/frame.rb
```diff
@@ -266,7 +262,7 @@ def result
 
       class BlobFrame < Frame
         def result
-          @text.empty? ? nil : Base64.decode64(@text.join)
+          @text.empty? ? '' : Base64.decode64(@text.join)
```
jterapin, line 265: Re-examining this .... will this be a breaking behavior for customers? 
When they expect a nil value instead of empty string?

This adjustment was made due to this test case: https://github.com/smithy-lang/smithy/blob/09a61e2165b6d970987b4f035ded20b3c1ff2962/smithy-aws-protocol-tests/model/awsQuery/xml-structs.smithy#L183-L204

mullermp: Maybe? These methods would no longer exist on it. to_h might be of concern if it were called in some nested way (have to ensure that still works). But it's probably fine? I would agree with the test that if <data> tag didn't exist, it should be nil, but it's explicit and empty. I'd say it's fine and we could patch it if it's a problem.

```
irb(main):001> nil.methods - ''.methods
=> [:to_h, :&, :to_a, :rationalize, :|, :^]
```

alextwoods: I think `to_h` in general should be okay on the containing structure (since to_h in general should be handling nils in other fields already).  I agree - I think this should be reasonable our other general behavior around nils.

gems/aws-sdk-core/CHANGELOG.md
alextwoods: While I agree with using "Issue" for all of these - I do think this change should be released as a new minor version - maybe we add a single "Feature" entry (and keep all of the issues as well just to provide the full list for posterity)?

alextwoods (approved):
Nice!! These look good to me

PR #2995, opened 2024-03-15
Ensure output unions work correctly with stub_responses

alextwoods:
When using stub_responses, union members in the stub data are treated as their Type class rather than a hash (as they are for input validation).  This causes param validators to fail because the `length` method always returns the number of members (rather than the number of members with values).

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core/param_validator.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/param_validator.rb
@@ -71,9 +71,10 @@ module Aws
         end
 
         if @validate_required && shape.union
-          if values.length > 1
+          set_values = @input ? values.length : values.to_h.length
+          if set_values > 1
             errors << "multiple values provided to union at #{context} - must contain exactly one of the supported types: #{shape.member_names.join(', ')}"
-          elsif values.length == 0
+          elsif set_values == 0
             errors << "No values provided to union at #{context} - must contain exactly one of the supported types: #{shape.member_names.join(', ')}"
           end
         end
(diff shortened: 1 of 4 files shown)


jterapin (approved):

gems/aws-sdk-core/lib/aws-sdk-core/param_validator.rb
```diff
@@ -71,9 +71,10 @@ def structure(ref, values, errors, context)
         end
 
         if @validate_required && shape.union
-          if values.length > 1
+          set_values = @input ? values.length : values.to_h.length
+          if set_values > 1
```
mullermp, line 75: What is values here? An array? This seems weird.

gems/aws-sdk-core/lib/seahorse/model/shapes.rb
```diff
@@ -87,7 +87,7 @@ def [](key)
           if @metadata.key?(key.to_s)
             @metadata[key.to_s]
           else
-            @shape[key.to_s]
+            @shape[key.to_s] if @shape
```
mullermp: elsif

gems/aws-sdk-core/lib/aws-sdk-core/param_validator.rb
alextwoods: During normal input validation this should be a hash.  However, in the case of stubbing, it will be a Union type.  the `length` method on that will just return the number of defined members rather than the number of members set.

alextwoods: The other alternative here I think is to override the `length` method in the [`union` module](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/lib/aws-sdk-core/structure.rb#L76).  But I'd be worried about that changing behavior.

PR #2998, opened 2024-03-26
Fix IMDS utilities to use port from configured endpoints

### LINKED ISSUE #2997: `AWS_EC2_METADATA_SERVICE_ENDPOINT` port is ignored
### Describe the bug

When I set a port as part of the `AWS_EC2_METADATA_SERVICE_ENDPOINT` environment variable, I expect the IMDS credential provider to attempt to get credentials using that hostname/port combo.

### Expected Behavior

`AWS_EC2_METADATA_SERVICE_ENDPOINT=http://127.0.0.1:9099`, should result in fetching credentials from `127.0.0.1:9099` (not the default port, 80)

### Current Behavior

```
Error retrieving instance profile credentials: Failed to open TCP connection to 127.0.0.1:80 (Connection refused - connect(2) for "127.0.0.1" port 80)
```

### Reproduction Steps

`Gemfile`
```ruby
source "https://rubygems.org"

gem 'base64'
gem 'nokogiri'

gem 'aws-sdk-ssm', `~> 1.166.0`
```

`repro.rb`
```ruby
#!/usr/bin/env ruby

require "aws-sdk-ssm"

client = Aws::SSM::Client.new(region: "ap-southeast-2")

client.get_parameter(name: "test-param")
```

```
$ bundle install
$ env AWS_CONFIG_FILE=/does/not/exist AWS_SHARED_CREDENTIALS_FILE=/does/not/exist AWS_EC2_METADATA_SERVICE_ENDPOINT=http://127.0.0.1:9099 bundle exec ./repro.rb

Error retrieving instance profile credentials: Failed to open TCP connection to 127.0.0.1:80 (Connection refused - connect(2) for "127.0.0.1" port 80)
...
```

### Possible Solution

Try using `uri.port` first in:

https://github.com/aws/aws-sdk-ruby/blob/64ad62e3375a02a32363441f20fbfe39f61da4cc/gems/aws-sdk-core/lib/aws-sdk-core/ec2_metadata.rb#L186

Or, default `@port` to `nil`.

_No response_

`aws-sdk-core` 3.191.4

ruby 3.3.0 (2023-12-25 revision 5124f9ac75) [arm64-darwin23]

mullermp:
Fixes #2997

Use the configured endpoint's port if provided, otherwise, use the default port configured onto the class.

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core/ec2_metadata.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/ec2_metadata.rb
@@ -183,7 +183,7 @@ module Aws
 
     def open_connection
       uri = URI.parse(@endpoint)
-      http = Net::HTTP.new(uri.hostname || @endpoint, @port || uri.port)
+      http = Net::HTTP.new(uri.hostname || @endpoint, uri.port || @port)
       http.open_timeout = @http_open_timeout
       http.read_timeout = @http_read_timeout
       http.set_debug_output(@http_debug_output) if @http_debug_output
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Fix `EC2Metadata` and `InstanceProfileCredentials` to respect the port from a configured endpoint from code, ENV, or shared config.
+
 3.191.4 (2024-03-15)
 ------------------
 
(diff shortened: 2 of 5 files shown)


gems/aws-sdk-core/lib/aws-sdk-core/ec2_metadata.rb
```diff
@@ -183,7 +183,7 @@ def http_put(connection, ttl)
 
     def open_connection
       uri = URI.parse(@endpoint)
-      http = Net::HTTP.new(uri.hostname || @endpoint, @port || uri.port)
+      http = Net::HTTP.new(uri.hostname || @endpoint, uri.port || @port)
```
alextwoods, line 186: is `@port` used anywhere else? (ie, printed in to_s or anything?) should we set it if so?

alextwoods (approved):

gems/aws-sdk-core/lib/aws-sdk-core/ec2_metadata.rb
mullermp: It doesn't appear to be. It's a direct override to the class.

PR #3000, opened 2024-03-27
Respect thread_count argument in multipart uploads

kspector:
## Observed Behavior ##
A method call in form:

`Aws::S3::Object.new(...).upload(some_file, thread_count: 20)`

only utilizes 10 threads to perform the upload.

## Reasoning ##
[MultipartFileUploader](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb#L37) accepts a `thread_count` option in its initializer, which sets instance variable `@thread_count` - however, [FileUploader](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-s3/lib/aws-sdk-s3/file_uploader.rb#L42) only passes its own `@options` to the `MultipartFileUploader` initializer, which does not include a`thread_count` key value pair.

The `thread_count` option is therefore ignored due to the preference in method [upload_in_threads](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb#L150) for the instance variable value of `@thread_count`.

## Solution ##
I moved the setting of instance variable `@thread_count` to the [upload](https://github.com/kspector/aws-sdk-ruby/blob/a2bf07e28fddaa8534e6f24afd8253651f5e09f6/gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb#L54) method, which mirrors the location of its counterpart in [FileDownloader -> download](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb#L27C9-L27C63).

I also added two tests to assert the default and specific cases, the [latter of which](https://github.com/kspector/aws-sdk-ruby/blob/a2bf07e28fddaa8534e6f24afd8253651f5e09f6/gems/aws-sdk-s3/spec/object/upload_file_spec.rb#L195-L201) fails against the current state of the codebase.

### DIFF
--- a/gems/aws-sdk-s3/CHANGELOG.md
+++ b/gems/aws-sdk-s3/CHANGELOG.md
@@ -1,7 +1,7 @@
 Unreleased Changes
 ------------------
 
-* Feature - Fix bug where thread_count option was not being respected for multipart uploads.
+* Issue - Fix bug where thread_count option was not being respected for multipart uploads.
 
 1.146.0 (2024-03-18)
 ------------------
--- a/gems/aws-sdk-s3/CHANGELOG.md
+++ b/gems/aws-sdk-s3/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Feature - Fix bug where thread_count option was not being respected for multipart uploads.
+
 1.146.0 (2024-03-18)
 ------------------
 
--- a/gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
+++ b/gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
@@ -31,8 +31,10 @@ module Aws
       )
 
       # @option options [Client] :client
+      # @option options [Integer] :thread_count (THREAD_COUNT)
       def initialize(options = {})
         @client = options[:client] || Client.new
+        @thread_count = options[:thread_count]
       end
 
       # @return [Client]
@@ -44,14 +46,11 @@ module Aws
       # @option options [Proc] :progress_callback
       #   A Proc that will be called when each chunk of the upload is sent.
       #   It will be invoked with [bytes_read], [total_sizes]
-      # @option options [Integer] :thread_count (THREAD_COUNT)
-      #   The thread count to use for multipart uploads.
       # @return [Seahorse::Client::Response] - the CompleteMultipartUploadResponse
       def upload(source, options = {})
         if File.size(source) < MIN_PART_SIZE
           raise ArgumentError, FILE_TOO_SMALL
         else
-          @thread_count = options[:thread_count] || THREAD_COUNT
           upload_id = initiate_upload(options)
           parts = upload_parts(upload_id, source, options)
           complete_upload(upload_id, parts, options)
@@ -148,7 +147,7 @@ module Aws
         if (callback = options[:progress_callback])
           progress = MultipartProgress.new(pending, callback)
         end
-        @thread_count.times do
+        (@thread_count || options[:thread_count] || THREAD_COUNT).times do
           thread = Thread.new do
             begin
               while part = pending.shift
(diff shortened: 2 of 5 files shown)


gems/aws-sdk-s3/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Feature - Fix bug where thread_count option was not being respected for multipart uploads.
```
alextwoods: I'd classify this as `Issue` instead of feature since its correcting a bug.

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
```diff
@@ -31,10 +31,8 @@ class MultipartFileUploader
       )
 
       # @option options [Client] :client
-      # @option options [Integer] :thread_count (THREAD_COUNT)
       def initialize(options = {})
         @client = options[:client] || Client.new
-        @thread_count = options[:thread_count] || THREAD_COUNT
```
alextwoods: I think we still want to set `@thread_count` here - otherwise this could be a breaking change.  However - I think we can skip setting a default ie, just `@thread_count = options[:thread_count]`.

Then in the `upload_in_threads` method we can do something like:
```ruby
      def upload_in_threads(pending, completed, options)
        # code skipped for brevity here
     
        (@thread_count || options[:thread_count] || THREAD_COUNT).times do
          thread = Thread.new do
```

alextwoods (commented):
Thanks for submitting this PR!  

Left a couple of comments.  I think we should also make the same change in the `MultipartStreamUploader` as well.

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
kspector: makes sense 👍

kspector:
Thanks for the quick review @alextwoods, I addressed your feedback. Just let me know if the commit history needs to be cleaned too.

> I think we should also make the same change in the MultipartStreamUploader as well.

It seems that the only initialization of a [`MultipartStreamUploader` in the repo](https://github.com/aws/aws-sdk-ruby/blob/d1565421f4cc4321e35f57eba1cab523cf4fe0e9/gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb#L387C20-L387C43) _does_ pass the `thread_count` option, so I held off on making the equivalent change there in favor of [an additional spec](https://github.com/aws/aws-sdk-ruby/pull/3000/commits/07ef78cfe975737f3da2b521573f644487649665).

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
mullermp: This works, but I would have suggested the opposite? The instance variable for the class is still valid, either as a passed value or a default. Any request level options should just supersede it. I would have recommended that upload_in_threads did something like `options.fetch(:thread_count, @thread_count)` and still have `@thread_count` as a user passed class option or a default to `THREAD_COUNT`. This way the hierarchy is preserved.

alextwoods: Edit: Sorry, for the initial suggestion - I do agree with @mullermp's suggestion.  We can keep the defaulting of `@thread_count` in the initializer and use an `options.fetch(:thread_count, @thread_count)` in the `upload_in_threads` method.  I agree that preserves the normal order of precedence better.

kspector: 👍 Addressed in my latest commit @mullermp @alextwoods.

For consistency, I elected to implement the same variable precedence in `MultipartStreamUploader` as well.

alextwoods (approved):
Nice - this looks good!  Thanks for submitting and making updates!

alextwoods:
I've merged, this should be released today.

PR #3003, opened 2024-04-01
Fix ordering of plugin options

### LINKED ISSUE #3002: Fix Seahorse::Client::Plugin::PluginOption shape variation warnings
### Describe the bug

Hello. :wave: I'm seeing Ruby warnings when installing the `aws-sdk-core` gem (via the `aws-sdk-s3` gem).

### Expected Behavior

I'd expect there to be no warnings when installing this gem.

### Current Behavior

This is what I see when installing the gem:

``` 
/Users/demo/.cache/frum/versions/3.3.0/lib/ruby/gems/3.3.0/gems/aws-sdk-core-3.191.5/lib/seahorse/client/plugin.rb:116: warning: The class Seahorse::Client::Plugin::PluginOption reached 8 shape variations, instance variables accesses will be slower and memory usage increased.
It is recommended to define instance variables in a consistent order, for instance by eagerly defining them all in the #initialize method.
```

### Reproduction Steps

To reproduce, run this Bundler Inline script:

``` ruby
#! /usr/bin/env ruby
# frozen_string_literal: true

# Save as `demo`, then `chmod 755 demo`, and run as `./demo`.

require "bundler/inline"

gemfile true do
  source "https://rubygems.org"

  gem "debug"
  gem "aws-sdk-s3"
  gem "nokogiri", "~> 1.13"
end
```

### Possible Solution

The problem is that `Seahorse::Client::Plugin::PluginOption` is defining instance variables in the the wrong order (or they are being initialized in random order).

_No response_

aws-sdk-core

ruby 3.3.0 (2023-12-25 revision 5124f9ac75) +YJIT [arm64-darwin23.4.0], macOS 14.4.1

jterapin:
Fixes #3002 

**Testing:**  
Ran customer-provided script and confirmed that the message is no longer showing.

---

### DIFF
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,6 @@
 Unreleased Changes
 ------------------
-* Issue - Fix ordering of plugin options for performance optimization.
+* Issue - Fix ordering of plugin options for performance optimization (#3002).
 
 3.191.5 (2024-03-26)
 ------------------
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,6 @@
 Unreleased Changes
 ------------------
-* Issue - Fix ordering of plugin options for performance optimization (#3002).
+* Issue - Performance optimization: ensure presence and order of instance variables in `PluginOptions` (#3002).
 
 3.191.5 (2024-03-26)
 ------------------
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,5 +1,6 @@
 Unreleased Changes
 ------------------
+* Issue - Fix ordering of plugin options for performance optimization.
 
 3.191.5 (2024-03-26)
 ------------------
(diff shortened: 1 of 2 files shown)


gems/aws-sdk-core/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 Unreleased Changes
 ------------------
+* Issue - Fix ordering of plugin options for performance optimization (#3002).
```
alextwoods: I'd be slightly more specific here - and say something like "Performance optimization:  ensure presence and order of instance variables in PluginOptions"

gems/aws-sdk-core/lib/seahorse/client/plugin.rb
```diff
@@ -111,7 +111,13 @@ class PluginOption
 
         def initialize(name, options = {})
           @name = name
```
alextwoods, line 113: Lets add a comment about why we're doing this.  Just something about "Prevent unstable object shapes by ensuring order and presence of instances variables"

alextwoods (approved):

gems/aws-sdk-core/lib/seahorse/client/plugin.rb
jterapin: That makes sense. Future us will thank us for this comment!

PR #3006, opened 2024-04-11
RPCv2 and cbor support

mullermp:
Implements CBOR and RPCv2

Also refactors protocols and related tests.

Encoder and decoder created by @alextwoods


build_tools/aws-sdk-code-generator/spec/protocol_tests_spec_helper.rb
```diff
@@ -314,3 +322,11 @@ def error_case?(test_case)
     end
   end
 end
+
+# @api private
+# RSpec::Matchers.define :match_cbor_float do |expected|
```
alextwoods: I think we probably want something almost exactly like the param matcher we have in hearth: https://github.com/smithy-lang/smithy-ruby/blob/main/hearth/lib/hearth/query/param_matcher.rb

We can add it to core's spec_helper (either in that file or as a separate file thats included).  And then we can use it on the cbor decode tests as well.  I think it can handle float precision and NaN equality.

gems/aws-sdk-core/lib/aws-sdk-core/rpc_v2/error_handler.rb
```diff
@@ -0,0 +1,72 @@
...
+        @handler.call(context).on(300..599) do |response|
+          response.error = error(context)
+          response.data = nil
+        end
+      end
+
+      private
+
+      def extract_error(body, context)
+        body = body.force_encoding(Encoding::BINARY)
```
alextwoods: What happens if we don't do this?

gems/aws-sdk-core/lib/aws-sdk-core/rpc_v2/error_handler.rb
```diff
@@ -0,0 +1,72 @@
...
+        json = Cbor.decode(body)
```
alextwoods: nit: not json here. this will just be a ruby hash.

gems/aws-sdk-core/lib/aws-sdk-core/cbor/engines/default_cbor_engine.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require_relative '../half'
```
alextwoods: This seems like a slightly weird file structure now. Maybe we should have a module for our engine and have these files there?

gems/aws-sdk-core/lib/aws-sdk-core/cbor/tagged.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Aws
+  module Cbor
+    Tagged = Struct.new(:tag, :value) do
```
alextwoods: the `cbor` gem also defines its own Tagged class: https://github.com/cabo/cbor-ruby/blob/master/doclib/cbor/tagged.rb#L6
Which it will return when decoding items.

I think we have two options on how to handle this:
1. In our cbor gem engine adaptor, we replace any Cbor Tagged objects with Aws::Taged objects.
2. We ensure the intrefaces are the same (ie, remove the use of Struct and just define a class with attr accessors) for our Tagged object.  Then just rely no ruby's duck typing.

build_tools/aws-sdk-code-generator/spec/protocol_tests_spec_helper.rb
mullermp: I don't think a matcher makes sense because it's not going to be reusable across specs. A simple method to iterate data members seems best so I did that.

gems/aws-sdk-core/lib/aws-sdk-core/rpc_v2/error_handler.rb
mullermp: Eventually we get a `ArgumentError: invalid byte sequence in UTF-8`

Not sure where else this goes.

We do it in parsing data too:

```
        cbor_data = context.http_response.body_contents.force_encoding(Encoding::BINARY)

```

gems/aws-sdk-core/lib/aws-sdk-core/cbor/engines/default_cbor_engine.rb
mullermp: I removed half. But I think it's fine to require relative the encoder and decoder if the engines will live in a subfolder?

gems/aws-sdk-core/lib/aws-sdk-core/cbor/decoder.rb
```diff
@@ -0,0 +1,304 @@
...
+
+      # 16 bit IEEE 754 half-precision floats
+      # Support decoding only
+      # format:
+      # sign - 1 bit
+      # exponent - 5 bits
+      # precision - 10 bits
+      def read_half
+        read_info
+        b16 = take(2).unpack1('n')
```
alextwoods, line 212: I know this is only used here, but its a lot of complexity to stick in an already hugely complex class.  I'd still potentially lean towards keeping it as a separate class

gems/aws-sdk-core/lib/aws-sdk-core/cbor/engines/cbor_engine.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+require_relative '../tagged'
+require_relative '../encoder'
```
alextwoods: This still feels like a weird structure to me.
The encoder and decoder are specific to this engine and its implementation, they feel like they should be a lower/same level and not in the top level Cbor module.

gems/aws-sdk-core/lib/aws-sdk-core/xml/parser.rb
```diff
@@ -80,7 +80,7 @@ def set_default_engine
         private
 
         def load_engine(name)
-          require "aws-sdk-core/xml/parser/engines/#{name}"
+          require "aws-sdk-core/xml/parser/#{name}_engine"
```
alextwoods, line 80: nice! I like these!

build_tools/aws-sdk-code-generator/spec/protocol_tests_spec.rb
```diff
@@ -4,68 +4,85 @@
...
-            ctx = context
-            Seahorse::Client::Response.new(context: context)
+      context "Engine: #{engine}" do
+        before(:all) do
+          engine_interface = ProtocolTestsHelper.engine_interface_for(protocol)
+          begin
+            engine_interface.engine = engine
+          rescue LoadError
+            skip "Skipping tests for missing engine: #{engine}"
           end
```
jterapin: Could we hide these into the helper?

build_tools/aws-sdk-code-generator/spec/protocol_tests_spec_helper.rb
```diff
@@ -11,9 +10,15 @@ def ignore_list
...
 
-    def skip_test_if_ignored(protocol, test_id, test_type, it)
-      if ignore_result = check_ignore_list(protocol, test_id, test_type)
-        it.skip(ignore_result[test_id])
+    #   "protocol" : {
+    #     "input" : {},
+    #     "output" : {
+    #       "some_engine": { "SomeTestId": "Description" }
+    #     }
+    #   },
```
jterapin: I'm assuming that we are leaving this here to visualize the format for this file. Could we add a comment about what this format is a reference for?

build_tools/aws-sdk-code-generator/spec/protocol_tests_spec_helper.rb
```diff
@@ -11,9 +10,15 @@ def ignore_list
...
+    def skip_test_if_ignored(protocol, test_type, engine, test_id, it)
+      if (_test_id, description = check_ignore_list(protocol, test_type, engine, test_id))
+        it.skip(description)
```
jterapin: If we have to return `_test_id`, I don't see the harm of including that in the skip message - in a case if we need to reference back to the C2J files (or Smithy version of the test)

gems/aws-sdk-core/lib/aws-sdk-core/cbor.rb
```diff
@@ -0,0 +1,89 @@
...
+
+    class OutOfBytesError < CborError
+      def initialize(n, left)
+        super("Out of bytes. Trying to read #{n} bytes but buffer contains only #{left}")
+      end
+    end
+
+    class UnknownTypeError < CborError
+      def initialize(type)
+        super("Unable to encode #{value}")
```
jterapin: I think you meant to say `type` instead of `value`

gems/aws-sdk-core/lib/aws-sdk-core/cbor/encoder.rb
```diff
@@ -0,0 +1,222 @@
...
+          when 0...MAX_INTEGER
+            [major_type + 27, value].pack('CQ>')
+          else
+            raise CborError, "Value is too large to encode: #{d}"
+          end
+      end
+
+      # streaming style, lower level interface
+      def add_integer(value)
+        major_type = if value.negative?
```
jterapin: nit: Move `if` to a newline for readability, same goes for `add_bignum` and `add_auto_integer` below

gems/aws-sdk-core/lib/aws-sdk-core/rpc_v2/handler.rb
```diff
@@ -0,0 +1,61 @@
...
+        response
+      end
+
+      private
+
+      def build_request(context)
+        context.http_request.http_method = 'POST'
+        build_url(context)
+        context.http_request.headers['smithy-protocol'] = 'rpc-v2-cbor'
+        context.http_request.headers['Accept'] = 'application/cbor' # remove?
```
jterapin: Yeah, I couldn't find any information on what to set this header on the "default" route - worth checking with Smithy?

gems/aws-sdk-core/spec/aws/cbor/encoder_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+require_relative '../../spec_helper'
+
+module Aws
+  module Cbor
+
+    describe Encoder do
+      let(:time) { Time.parse('2020-01-01 12:21:42Z') }
+
+      def cbor64_encode(value)
+        Base64.encode64(Encoder.new.add(value).bytes).strip.force_encoding("UTF-8")
```
jterapin: nit: single quotations marks - prob need to run rubocop on the file

gems/aws-sdk-core/lib/aws-sdk-core/rpc_v2/handler.rb
```diff
@@ -0,0 +1,61 @@
...
+        context.http_request.body = build_body(context)
+        apply_content_type_header(context)
+      end
+
+      def build_url(context)
+        base = context.http_request.endpoint
+        service_name = context.config.api.metadata['targetPrefix']
+        base.path += "/service/#{service_name}/operation/#{context.operation.name}"
```
jterapin, line 28: The protocol indicates that they will only use the last four segments of the URL when routing requests but still need to include the `prefix?` bit before the `/service/`

gems/aws-sdk-core/lib/aws-sdk-core/rpc_v2/handler.rb
```diff
@@ -0,0 +1,61 @@
...
+
+module Aws
+  module RpcV2
+    class Handler < Seahorse::Client::Handler
+      # @param [Seahorse::Client::RequestContext] context
+      # @return [Seahorse::Client::Response]
+      def call(context)
+        build_request(context)
+        response = @handler.call(context)
+        response.on(200..299) { |resp| resp.data = parse_body(context) }
```
jterapin, line 11: Before we parse the body, we need to check for existence of `smithy-protocol` header in the response. If this does not have the same header, it must be considered malformed. Is this handled somewhere else in the core? I may have missed it.

jterapin (commented):
I love the updates to the protocol test runner (testing with each engines, etc)! Great job. 

I don't have too much context on how eventstreams work in the SDK but that being said - I'm assuming that specific work will be included in the next round of review? I noticed that we set different headers like `Content-Type` for event streaming requests. Also - what's the remaining set of tasks to do?

build_tools/aws-sdk-code-generator/spec/protocol_tests_spec_helper.rb
```diff
@@ -82,6 +87,32 @@ def client_for(suite, test_case, id)
       )
     end
 
+    def engines_for(protocol)
```
alextwoods, line 96: Nice! I like this approach!

Gemfile
```diff
@@ -9,13 +9,13 @@ gem 'aws-crt' if ENV['CRT']
 gem 'http-2'
 gem 'jmespath'
 
-# json and xml parsers
+# protocol parsers
```
alextwoods, line 12: Can we remove the `KITCHEN_SINK` from CI now?  I don't think its actually doing anything anymore.  Esp w/ the new approach to engines in protocol tests.

mullermp: It's necessary in the matrix as a "this is not CRT" option.

gems/aws-sdk-core/lib/aws-sdk-core/rpc_v2/handler.rb
mullermp: It's in here to make the tests pass but it will be removed in tests soon.

gems/aws-sdk-core/lib/aws-sdk-core/rpc_v2/handler.rb
mullermp: Yes. It's not handled yet, there's more to do.

gems/aws-sdk-core/lib/aws-sdk-core/rpc_v2/handler.rb
mullermp: I think this is handled when resolving the endpoint prior to this.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/client_api_module.rb
```diff
@@ -27,25 +27,26 @@ class ClientApiModule < View
...
         'flattened' => true,
-        'timestampFormat' => true, # glacier api customization
+        'timestampFormat' => true,
         'xmlNamespace' => true,
-        'streaming' => true, # transfer-encoding
-        'requiresLength' => true, # transfer-encoding
-        'union' => false, # should remain false
+        'streaming' => true,
+        'requiresLength' => true,
+        'union' => false, # not ignored - should remain false
```
alextwoods: "not ignored" is slightly confusing.  Maybe "handled separately"

alextwoods (commented):
Nice - updates look good

build_tools/aws-sdk-code-generator/spec/interfaces/protocols_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+    end
+
+    it 'selects json as fallback if smithy rpc v2 is not modeled' do
+      SpecHelper.generate_service(['ProtocolsJsonQuery'], multiple_files: false)
+      client = ProtocolsJsonQuery::Client.new(stub_responses: true)
+      expect(client.handlers).to include(Aws::Json::Handler)
+      Object.send(:remove_const, :ProtocolsJsonQuery)
+    end
+
+    it 'selects query if is the only choice' do
```
jterapin: I think you meant to say, `selects query if it is the only choice` here like line 20

gems/aws-sdk-core/lib/aws-sdk-core/cbor/encoder.rb
```diff
@@ -0,0 +1,225 @@
...
+      def start_map(length)
+        head(MAJOR_TYPE_MAP, length)
+      end
+
+      def start_indefinite_map
+        head(MAJOR_TYPE_MAP + 31, 0)
+      end
+
+      def end_indefinite_collection
+        # write the stop sequenece
```
jterapin: Typo: sequence

build_tools/aws-sdk-code-generator/spec/interfaces/protocols_spec.rb
```diff
@@ -0,0 +1,90 @@
+# frozen_string_literal: true
+
+require_relative '../spec_helper'
+
+describe 'Protocols Resolution:' do
```
jterapin, line 5: Will all existing service models be onboarded to use `"protocols"` meta key? If not, maybe worth having a test case that covers a model that just have `"protocol"` to ensure that the changes that we make here works with the older model files

gems/aws-sdk-core/lib/aws-sdk-core/cbor.rb
```diff
@@ -0,0 +1,105 @@
...
+        bytes.force_encoding(Encoding::BINARY)
+        @engine.decode(bytes)
+      end
+
+      def set_default_engine
+        [:cbor].each do |name|
+          @engine ||= try_load_engine(name)
+        end
+        unless @engine
+          raise 'Unable to find a compatible cbor library. '
```
jterapin: Nit: The empty space between `.` and `'`

gems/aws-sdk-core/lib/aws-sdk-core/cbor/decoder.rb
```diff
@@ -0,0 +1,304 @@
...
+          read_end_indefinite_collection
+          value.force_encoding(Encoding::UTF_8)
+        when :tag
+          case (tag = read_tag)
+          when TAG_TYPE_EPOCH
+            type = peek_type
+            item = decode_item
+            item /= 1000.0 if type == :integer
+            Time.at(item)
+          # TODO: Consider handling of  BigDecimal, ect
```
jterapin: Will this be done for this release?

gems/aws-sdk-core/lib/aws-sdk-core/stubbing/protocols/rpc_v2.rb
```diff
@@ -0,0 +1,41 @@
...
+module Aws
+  module Stubbing
+    module Protocols
+      class RpcV2
+
+        def stub_data(api, operation, data)
+          resp = Seahorse::Client::Http::Response.new
+          resp.status_code = 200
+          resp.headers["Content-Type"] = content_type(api)
+          resp.headers["x-amzn-RequestId"] = "stubbed-request-id"
```
jterapin: rubocop: single quotes for string literals

gems/aws-sdk-core/lib/aws-sdk-core/cbor/decoder.rb
alextwoods: Currently services can't use BigDecimal.  Will fast follow with support here.

alextwoods: Edit, nvm, added support in: https://github.com/aws/aws-sdk-ruby/pull/3006/commits/3a079f6454cea6ed9b834d0ceb35065d94f86adc

build_tools/aws-sdk-code-generator/spec/interfaces/protocols_spec.rb
mullermp: Other fixtures/interfaces tests only have protocol so yes it works with older models.

gems/aws-sdk-core/lib/aws-sdk-core/cbor/decoder.rb
mullermp: This is fine.

alextwoods (approved):

build_tools/customizations.rb
```diff
@@ -85,6 +85,9 @@ def dynamodb_example_deep_transform(subsegment, keys)
         members.delete('DestinationRegion')
         members.delete('PresignedUrl')
       end
+
+      # temporary
+      api['metadata']['protocols'] = ['ec2']
```
jterapin: I'm guessing this will be removed once services using ec2 protocol are using the new `protocols` metakey?

mullermp: It can stay here forever if it needs to. It's here because what's currently in the ec2 model is wrong.

PR #3008, opened 2024-04-15
Generic Content Type plugin

mullermp:
Move rest/request plugin to generic. Will add to all services


gems/aws-sdk-core/lib/aws-sdk-core/rest/request/content_type.rb
```diff
@@ -1,51 +0,0 @@
-# frozen_string_literal: true
-
-module Aws
-  module Rest
-    # NOTE: headers could be already populated if specified on input shape
-    class ContentTypeHandler < Seahorse::Client::Handler
```
alextwoods, line 6: I don't think we can remove this.  This will break old service w/ new core

alextwoods: Edit: Actually nvm, this isn't dependend on by service clients directly, but through the protocol plugin.  So we may be okay.

gems/aws-sdk-core/lib/seahorse/client/plugins/content_type.rb
```diff
@@ -0,0 +1,58 @@
...
+          private
+
+          def non_empty_body?(body)
+            body.respond_to?(:size) && body.size.positive?
+          end
+
+          # content-type defaults as noted here:
+          # rest-json: https://smithy.io/2.0/aws/protocols/aws-restxml-protocol.html#content-type
+          # rest-xml: https://smithy.io/2.0/aws/protocols/aws-restxml-protocol.html#content-type
+          def apply_default_content_type(context)
```
alextwoods, line 40: I don't think this belongs in Seahorse - this is AWS specific logic per protocol.  I'd lean instead towards a generic base class which the protocols can extend and use.

gems/aws-sdk-core/lib/aws-sdk-core/rest/request/content_type.rb
mullermp: It hasn't released yet. This is new code too

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/plugin_list.rb
```diff
@@ -40,6 +40,7 @@ def compute_plugins(options)
     def default_plugins
       {
         'Seahorse::Client::Plugins::ContentLength' => "#{seahorse_plugins}/content_length.rb",
+        'Seahorse::Client::Plugins::ContentType' => "#{seahorse_plugins}/content_type.rb",
```
jterapin, line 43: With this being default, should we relocate the setting of `content-type` for other protocols like [json](https://github.com/aws/aws-sdk-ruby/blob/09ee4c3e0b96e807d9fd1b9331c4387746f0896f/gems/aws-sdk-core/lib/aws-sdk-core/json/handler.rb#L23) here?

PR #3016, opened 2024-04-30
Do not include sse configuration of source object in that of the destination object

matchbookmac:
When copying objects using `Aws::S3::Object.copy_to(..., multipart_copy: true)`, the `ObjectMultipartCopier` was inferring the metadata of the destination object from that of the source object. This is fine for `content_type`, `content_length`, etc.; however, the `ssekms_key_id` and `server_side_encryption` attributes were also copied. This causes the destination object to use the kms key of the source, which is not usually appropriate if the destination bucket has a different default encryption than the source. This is especially true when copying from one account to the other. This change is in line with the behavior of `Aws::S3::Client.copy_object` which does not copy the source object encryption.


alextwoods (commented):
Thanks for submitting this!  The logic looks good - can you add a test case for this?

matchbookmac:
Sure thing!

matchbookmac:
@alextwoods added

gems/aws-sdk-s3/spec/object/multipart_copy_spec.rb
```diff
@@ -257,8 +257,7 @@ module S3
             allow(client).to receive(:head_object).with({
               bucket: 'source-bucket',
               key: 'source key'
-            }).and_return(Types::HeadObjectOutput.new(
-              content_length: size, content_type: 'ContentType'))
+            }).and_return(Types::HeadObjectOutput.new(content_length: size, content_type: 'ContentType'))
```
mullermp: Yuck.. I'm not sure why this is here. Do you mind changing all of these to stub_data?

mullermp (approved):

matchbookmac:
@mullermp gladly

mullermp (approved):
Thanks for your contribution (and clean up :D)

matchbookmac:
Sure thing! Thanks for the merge.

PR #3031, opened 2024-05-28
User Agent metrics

mullermp:
Bumps user agent to 2.1 and uses new m/ tag for gathering metrics. Metrics is block based similar to how feature was.

We also remove cfg/ and ft/ usage.

We also need to re-arrange how the user agent plugin is registered - also fix CSM ordering issue discovered during this process.


gems/aws-sdk-core/lib/aws-sdk-core/plugins/user_agent.rb
```diff
@@ -4,6 +4,10 @@ module Aws
   module Plugins
     # @api private
     class UserAgent < Seahorse::Client::Plugin
+      METRICS = Aws::Json.load(
```
alextwoods: Do we need to update the gemspec to include this file?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/user_agent.rb
```diff
@@ -25,12 +29,20 @@ class UserAgent < Seahorse::Client::Plugin
 
       def self.feature(feature, &block)
         Thread.current[:aws_sdk_core_user_agent_feature] ||= []
-        Thread.current[:aws_sdk_core_user_agent_feature] << "ft/#{feature}"
+        Thread.current[:aws_sdk_core_user_agent_feature] << feature
```
alextwoods: I can't remember if we discussed this in the previous UA implementation - but does this cause a problem for things like S3 Access Grants where multiple SDK requests are made in the same thread or does the block with pop at the end fully handle that?

alextwoods (commented):
nice - looks good overall.  

Whats the reasoning for the change in handler order?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/user_agent.rb
mullermp: Yes. Good catch. I think for now I'm going to embed it as a heredoc since there are no plans to make this automated.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/user_agent.rb
mullermp: I don't believe so. I think that was the case until my last commit which made it block based. You should confirm that. But in express session auth testing, both credentials call (create session) and get object call will have these user agents:
```
user agent: aws-sdk-ruby3/3.196.1 ua/2.1 api/s3#1.151.0 os/macos#23 md/x86_64 lang/ruby#3.3.1 md/3.3.1 m/A
user agent: aws-sdk-ruby3/3.196.1 ua/2.1 api/s3#1.151.0 os/macos#23 md/x86_64 lang/ruby#3.3.1 md/3.3.1 m/A
```

mullermp: I've changed this so that create session does not have the metric. I'm getting clarification.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/client_metrics_send_plugin.rb
```diff
@@ -62,17 +64,27 @@ def call(context)
...
           end
           call_attempt.http_status_code = context.http_response.status_code
-          if e = resp.error
+          context.metadata[:current_call_attempt] = call_attempt
+          request_metrics.add_call_attempt(call_attempt)
+          resp
+        end
+      end
+
+      class ErrorHandler < Seahorse::Client::Handler
```
alextwoods, line 73: Nice! This is a good way of fixing this.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/user_agent.rb
```diff
@@ -23,12 +39,15 @@ class UserAgent < Seahorse::Client::Plugin
         app_id
       end
 
-      def self.feature(feature, &block)
-        Thread.current[:aws_sdk_core_user_agent_feature] ||= []
-        Thread.current[:aws_sdk_core_user_agent_feature] << "ft/#{feature}"
+      def self.metric(metric, &block)
+        Thread.current[:aws_sdk_core_user_agent_metric] ||= []
```
alextwoods, line 49: Is there an old service gem (relying on `:aws_sdk_core_user_agent_feature`) + new core risk here?

gems/aws-sdk-core/spec/aws/plugins/client_metrics_spec.rb
```diff
@@ -29,15 +29,14 @@ module Plugins
         client.handlers.add(
           ClientMetricsSendPlugin::AttemptHandler,
           step: :sign,
+          priority: 39
```
alextwoods, line 32: I'm worried about future disconnects in this test (as we had before...).

Stub resposnes removes the CSM handler - which is why we need to add them back in here, right?

Could we call the `add_handlers` function in `ClientMetricsSendPlugin`? like:

```
ClientMetricsSendPlugin.new.add_handlers(client.handlers, client.config)
```

Either that or add constants on the class for the step and priority?

gems/aws-sdk-securityhub/lib/aws-sdk-securityhub/client.rb
```diff
@@ -5158,7 +5158,7 @@ def get_configuration_policy(params = {}, options = {})
     #   resp.to_h outputs the following:
     #   {
     #     association_status: "FAILED", 
-    #     association_status_message: "Configuration Policy a1b2c3d4-5678-90ab-cdef-EXAMPLE11111 couldn\u2019t be applied to account 111122223333 in us-east-1 Region. Retry your request.", 
+    #     association_status_message: "Configuration Policy a1b2c3d4-5678-90ab-cdef-EXAMPLE11111 couldn’t be applied to account 111122223333 in us-east-1 Region. Retry your request.",
```
alextwoods: Whats this change from? This doesn't seem like this file should be getting changed here.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/user_agent.rb
```diff
@@ -4,6 +4,22 @@ module Aws
...
+          "WAITER": "B",
+          "PAGINATOR": "C",
+          "RETRY_MODE_LEGACY": "D",
+          "RETRY_MODE_STANDARD": "E",
+          "RETRY_MODE_ADAPTIVE": "F",
+          "S3_TRANSFER": "G",
+          "S3_CRYPTO_V1N": "H",
+          "S3_CRYPTO_V2": "I",
+          "S3_EXPRESS_BUCKET": "J",
+          "S3_ACCESS_GRANTS": "K"
```
jterapin: Missing `"GZIP_REQUEST_COMPRESSION": "L"`

gems/aws-sdk-s3/lib/aws-sdk-s3/file_uploader.rb
```diff
@@ -37,7 +37,7 @@ def initialize(options = {})
       #   objects smaller than the multipart threshold.
       # @return [void]
       def upload(source, options = {})
-        Aws::Plugins::UserAgent.feature('s3-transfer') do
+        Aws::Plugins::UserAgent.metric('S3_TRANSFER') do
```
jterapin, line 40: I recall hearing that there was some differences with our implementation of "S3 Transfer Manager" vs the most recently defined specifications. Even so, we are still collecting the metrics under this ID, right?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/client_metrics_send_plugin.rb
jterapin: What was the context of the bug relating to the CSM? Was it not capturing errors?

mullermp: I think it wasn't capturing errors for 6 years and nobody complained (I guess since this feature is effectively not used)

gems/aws-sdk-core/lib/aws-sdk-core/plugins/user_agent.rb
mullermp: Oh.. yeah I think so. I think feature() will have to persist but as a no-op.

gems/aws-sdk-core/spec/aws/plugins/client_metrics_spec.rb
mullermp: I'm not too worried about this. Yeah this bug existed for 6 years but nobody noticed or cared. I just wanted to get the tests to pass "correctly"

gems/aws-sdk-securityhub/lib/aws-sdk-securityhub/client.rb
mullermp: I ran rake build. I think it's a difference in kramdown or something? I reverted it manually, but I think internal versions are dated and should be updated.

gems/aws-sdk-s3/lib/aws-sdk-s3/file_uploader.rb
mullermp: Yeah. This is effectively the same feature and we should track it.

alextwoods (approved):

alextwoods (commented):
Do we need to bump the min core version in services?

mullermp:
We don't need a minimum core version technically but good call out that it may be a good idea. I've added it.

mullermp:
Yes, we do need a minimum core version. Good call out. New service could have old core and not have the metric() method.

PR #3037, opened 2024-06-04
Fix UFC offset when deserializing `iso8601` timestamp values

jterapin:
This fix is to ensure that there are no UFC offset when deserializing. 

Previously, there were some adjustments made to deserializing timestamps so that it handles fractional seconds (to pass smithy protocol-tests).  However, when `Time.at(...)` is used, the value always reflect local system time by default. 

Also, see [Smithy Docs](https://smithy.io/2.0/spec/protocol-traits.html#timestampformat-trait) on timestamp formats:
> Date time as defined by the date-time production in [RFC 3339#section-5.6](https://datatracker.ietf.org/doc/html/rfc3339.html#section-5.6) with optional millisecond precision but no UTC offset (for example, 1985-04-12T23:20:50.520Z). Values that are more granular than millisecond precision SHOULD be truncated to fit millisecond precision. Deserializers SHOULD parse date-time values that contain offsets gracefully by normalizing them to UTC.

---

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core/util.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/util.rb
@@ -90,7 +90,7 @@ module Aws
         end
       end
 
-      # @param [String, Integer] value
+      # @param [String] value
       # @return [Time]
       def deserialize_time(value)
         case value
--- a/gems/aws-sdk-core/lib/aws-sdk-core/util.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/util.rb
@@ -95,11 +95,11 @@ module Aws
       def deserialize_time(value)
         case value
         when nil then nil
-        when /^\d+$/ then Time.at(value.to_i)
+        when /^\d+$/ then Time.at(value.to_i).utc
         else
           begin
-            fractional_time = Time.parse(value).utc.to_f
-            Time.at(fractional_time)
+            fractional_time = Time.parse(value).to_f
+            Time.at(fractional_time).utc
           rescue ArgumentError
             raise "unhandled timestamp format `#{value}'"
           end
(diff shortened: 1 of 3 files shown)


gems/aws-sdk-core/lib/aws-sdk-core/util.rb
```diff
@@ -95,11 +95,11 @@ def deserialize_number(str)
       def deserialize_time(value)
         case value
         when nil then nil
-        when /^\d+$/ then Time.at(value.to_i)
+        when /^\d+$/ then Time.at(value.to_i).utc
```
alextwoods: Do we have any specs for this?  I can't seem to find any.  I guess it is covered by protocol tests, but I think probably worth adding a specific spec for this method.

gems/aws-sdk-core/spec/aws/util_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+      end
+
+      it 'correctly parses when given value is a numeric string' do
+        expect(Util.deserialize_time('946702800').to_s)
+          .to eq('2000-01-01 05:00:00 UTC')
+      end
+
+      it 'correctly parses when given value is a string' do
+        expect(Util.deserialize_time('2000-01-02T20:34:56.123Z').to_s)
+          .to eq('2000-01-02 20:34:56 UTC')
```
alextwoods: This should include the fractional seconds right?

Maybe instead of comparing with to_s we can check that its utc and that the time matches instead?

alextwoods (approved):

gems/aws-sdk-core/spec/aws/util_spec.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+require_relative '../spec_helper'
+
+module Aws
+  describe Util do
```
mullermp, line 6: We should probably add tests for the other stuff while we're at it..

jterapin: Yeah, I think I plan to update this in a separate PR since we want this fix out with today's release.

PR #3038, opened 2024-06-05
Handle S3 200 errors for all operations

mullermp:
Handles s3 200 errors for all operations by checking for a root Error and if it's not a streaming output.


mullermp:
TODO: fix broken tests. Stubbing has empty body.

alextwoods (approved):

PR #3046, opened 2024-06-13
Handle S3 200 errors for all operations #3 (#3040)

mullermp:
Attempt #3 of https://github.com/aws/aws-sdk-ruby/pull/3040

First attempt failed for event stream.
Second attempt failed for JSON payloads in an XML format.

This attempt looks for string payload first and skips checking members of the body.

### DIFF
--- a/gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/http_200_errors.rb
+++ b/gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/http_200_errors.rb
@@ -39,45 +39,43 @@ module Aws
             end
           end
 
-          # Payloads that are strings will not have serialized XML unless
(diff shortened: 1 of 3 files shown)


gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/http_200_errors.rb
```diff
@@ -15,22 +15,67 @@ class Handler < Seahorse::Client::Handler
...
               end
+
+            if structure_shape?(shape)
+              shape.members.any? { |_, k| k.location.nil? }
+            else
+              false
             end
           end
 
+          def structure_shape?(shape)
```
alextwoods, line 62: This covers union shape as well right?

alextwoods (approved):

PR #3048, opened 2024-06-18
Support array arguments for process credentials

mullermp:
Use `system` instead of backticks for executing processes in process credentials, allowing for more secure argument passing. Adds a warning for existing usage of a single string and general refactors in this class.

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core/process_credentials.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/process_credentials.rb
@@ -2,9 +2,15 @@
 
 module Aws
   # A credential provider that executes a given process and attempts
-  # to read its stdout to recieve a JSON payload containing the credentials.
+  # to read its stdout to receive a JSON payload containing the credentials.
   #
-  #     credentials = Aws::ProcessCredentials.new('/usr/bin/credential_proc')
+  #     credentials = Aws::ProcessCredentials.new(['/usr/bin/credential_proc'])
+  #     ec2 = Aws::EC2::Client.new(credentials: credentials)
+  #
+  # Arguments should be provided as strings in the array, for example:
+  #
+  #     process = ['/usr/bin/credential_proc', 'arg1', 'arg2']
+  #     credentials = Aws::ProcessCredentials.new(process)
   #     ec2 = Aws::EC2::Client.new(credentials: credentials)
   #
   # Automatically handles refreshing credentials if an Expiration time is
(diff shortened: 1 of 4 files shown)


gems/aws-sdk-core/lib/aws-sdk-core/process_credentials.rb
```diff
@@ -19,40 +25,50 @@ class ProcessCredentials
     # Creates a new ProcessCredentials object, which allows an
     # external process to be used as a credential provider.
     #
-    # @param [String] process Invocation string for process
-    # credentials provider.
+    # @param [Array<String>, String] process An array of strings including
+    #  the process name and its arguments to execute, or a single string to be
+    #  executed by the shell (deprecated and insecure).
     def initialize(process)
+      if process.is_a?(String)
```
mullermp, line 32: Ideally we would rather splat arguments like `Aws::ProcessCredentials.new('some', 'arg')` however we could not correctly warn in these cases because we already support the string case.

gems/aws-sdk-core/lib/aws-sdk-core/process_credentials.rb
```diff
@@ -19,40 +25,50 @@ class ProcessCredentials
...
 
       super
     end
 
     private
-    def credentials_from_process(proc_invocation)
-      begin
-        raw_out = `#{proc_invocation}`
-        process_status = $?
-      rescue Errno::ENOENT
```
mullermp, line 37: `Errno::ENOENT` is no longer raised when `system` is used.

gems/aws-sdk-core/spec/aws/process_credentials_spec.rb
```diff
@@ -5,47 +5,85 @@
...
-    it 'will throw an error when the process credentials payload is malformed' do 
+    it 'will throw an error when the process credentials payload is malformed' do
+      process = %w[echo {"Version":1}]
       expect {
-        creds = ProcessCredentials.new('echo \'{"Version":1}\'').credentials
+        ProcessCredentials.new(process).credentials
       }.to raise_error(Errors::InvalidProcessCredentialsPayload)
     end
 
-    it 'will throw an error and expose the stderr output when the credential process has a nonzero exit status' do
```
mullermp, line 38: stderr case no longer applies. `fake_proc` will trigger a nonzero exit status.

alextwoods (approved):

alextwoods (approved):

PR #3049, opened 2024-06-18
Service specific protocol tests

mullermp:
Handwritten service tests from https://github.com/smithy-lang/smithy/blob/main/smithy-aws-protocol-tests/model/

Also a few related fixes.


alextwoods (approved):

jterapin (approved):
Looks good! I like the comment reference to the smithy version of these test cases.

build_tools/services.rb
```diff
@@ -10,10 +10,10 @@ class ServiceEnumerator
     MANIFEST_PATH = File.expand_path('../../services.json', __FILE__)
 
     # Minimum `aws-sdk-core` version for new gem builds
-    MINIMUM_CORE_VERSION = "3.197.0"
+    MINIMUM_CORE_VERSION = "3.197.2"
```
alextwoods: I actually don't think we need to bump this.  This is just a minor fix in core.  All service gems will continue to work (ie, this does not change service gems in any way that requires the new core version to work).

mullermp: I am worried about old service, new core, and new service, old core. I've reverted this.

PR #3051, opened 2024-06-19
Instance plugins

mullermp:
Support instance plugin configuration passed to clients or using global config.

WIP - needs tests and documentation.


gems/aws-sdk-core/lib/seahorse/client/base.rb
```diff
@@ -60,12 +59,19 @@ def operation_names
...
+        config.add_option(:plugins)
+        instance_plugins = build_plugins(options.fetch(:plugins, []))
+        (plugins + instance_plugins).each do |plugin|
           plugin.add_options(config) if plugin.respond_to?(:add_options)
         end
         config.build!(options.merge(api: self.class.api))
       end
 
+      # Builds a list of plugins from a configured list of plugins
+      def build_plugins(plugins)
```
alextwoods: This looks duplicated, could we just call the class method above?

mullermp: I think there was an issue doing this but I can try again.

gems/aws-sdk-core/lib/seahorse/client/base.rb
```diff
@@ -209,21 +208,62 @@ def define_operation_methods
...
+          super
           subclass.instance_variable_set('@plugins', PluginList.new(@plugins))
         end
 
       end
     end
   end
 end
+
+class TestPlugin < Seahorse::Client::Plugin
```
alextwoods: Probably can remove this from here

alextwoods (approved):

jterapin (approved):
🚢

PR #3062, opened 2024-07-03
Do not warn using process credentials in shared config

### LINKED ISSUE #3061: "Passing a single string to Aws::ProcessCredentials.new is insecure" now appearing when using credential_process option in an aws config
### Describe the bug

After upgrading to the most recent version of `aws-sdk-core`, I started receiving the following warning: 
`Passing a single string to Aws::ProcessCredentials.new is insecure, please use use an array of system arguments instead`

This appears to happen because I have the "credential_process" option set in my aws config file like so:
```
[profile my-account]
role_arn = << role_arn >>
source_profile = << profile >>
region= << region >>

[profile my-account-creds]
credential_process = aws-vault export --format=json my-account
region=us-east-2
```

Then in ruby I try to use the profile:
```
require 'aws-sdk-core'
require 'aws-sdk-s3'
Aws.config.update(profile: 'my-account-creds')
Aws::S3::Client.new
Passing a single string to Aws::ProcessCredentials.new is insecure, please use use an array of system arguments instead
=> #<Aws::S3::Client>
```

It seems like this warning was adding recently in https://github.com/aws/aws-sdk-ruby/pull/3048

Is there a proper way to set `credential_process` in the aws config file to avoid this warning? I couldn't find any example of it taking an array.

### Expected Behavior

Not receive a warning.

### Current Behavior

A warning appears: `Passing a single string to Aws::ProcessCredentials.new is insecure, please use use an array of system arguments instead`

### Reproduction Steps

See the bug description.

### Possible Solution

_No response_

_No response_

aws-sdk-core 3.201.0

Ruby 3.3.1 OSX

mullermp:
Closes #3061 

Shared config values are strings only. Strings will always warn for Process Credentials. So any time Process Credentials come from shared config, wrap it in an array.

### DIFF
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Fix `Aws::ProcessCredentials` warning in cases where shared config is used.
+
 3.201.0 (2024-07-02)
 ------------------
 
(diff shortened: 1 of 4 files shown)


jterapin (approved):
LGTM

PR #3073, opened 2024-07-25
Handle expected boolean values for waiter error matcher

jterapin:
All services currently define waiter error acceptors that have expected string values that represent an AWS Error code like below:
```
{
  "expected":"ResourceNotFoundFault",
  "matcher":"error",
  "state":"success"
}
```

A waiter with this acceptor will continue to wait until an error the the code `ResourceNotFoundFault` is received or the max amount of attempts is reached.

However, it's possible for service teams to define a waiter error acceptor like below which has an expected boolean (true or false) value. The expected behavior is defined below:

**true** - Match an error code with any value.
**false** - Match if no errors were encountered

Example:
```
{
  "matcher" : "error",
  "expected" : false,
  "state" : "success"
}
```
A waiter with this acceptor will continue to wait until a response with no error is received or the max amount of attempts is reached.

This PR adds the ability for waiters to handle the acceptors described above.

----


jterapin:
I plan do some rubocop adjustments once the PR has been approved. Keeping as it is to help out with readability.

gems/aws-sdk-core/lib/aws-sdk-core/waiters/poller.rb
```diff
@@ -96,8 +96,13 @@ def matches_status?(acceptor, response)
       end
 
       def matches_error?(acceptor, response)
-        Aws::Errors::ServiceError === response.error &&
-        response.error.code == acceptor['expected'].delete('.')
+        case acceptor['expected']
+        when 'false' then response.error.nil?
+        when 'true' then !!response.error
```
mullermp: I think you can just check not nil here instead of coerce.

gems/aws-sdk-core/spec/aws/waiters_spec.rb
```diff
@@ -190,14 +189,60 @@ module Waiters
...
+          it 'fails when an expected status is not encountered' do
             expect {
               client.wait_until(:table_not_exists, table_name:'table') do |w|
                 w.delay = 0
               end
             }.to raise_error(Errors::TooManyAttemptsError)
           end
+        end
+
+        describe 'error matcher' do
```
mullermp, line 207: Do we have test cases for all of the possible options? This fixture seems pretty lightweight. Let's talk offline but I think we should have a test suite of all the possible options (and remove the ddb fixture).

jterapin:
Moving to Draft mode since I need to do some audits and add/update existing waiter specs

jterapin:
Self-reminder: Revert [1a61acd](https://github.com/aws/aws-sdk-ruby/pull/3073/commits/1a61acd59ed92095fa0e40e9391426f484a0c28f) after approvals

gems/aws-sdk-core/spec/aws/temp_waiters_spec.rb
```diff
@@ -0,0 +1,368 @@
...
+    describe 'Waiters' do
+      dir = File.expand_path('../../fixtures', __FILE__)
+      WaiterTest = ApiHelper.sample_service(
+        api: JSON.load_file("#{dir}/waiters/service.json"),
+        waiters: JSON.load_file("#{dir}/waiters/waiters.json")
+      )
+
+      let(:client) { WaiterTest::Client.new(stub_responses: true) }
+
+      let(:setup_stub_resp) do
```
mullermp: These don't seem to get reused, I would just remove them.

mullermp (approved):

alextwoods (approved):
Nice - great additional tests.

PR #3074, opened 2024-07-26
Benchmark refactor

mullermp:
Refactor benchmarking to use a shared tooling repository


alextwoods (approved):
Nice, looks good!

.github/workflows/benchmark.yml
```diff
@@ -17,13 +17,14 @@ jobs:
...
+      GH_EVENT: ${{github.event_name}}
       GH_REF: ${{github.head_ref}}
       GH_REPO: ${{github.repository}}
-      GH_EVENT: ${{github.event_name}}
+      EXECUTION_ENV: github-action
     strategy:
       fail-fast: false
       matrix:
-        ruby: [jruby-9.2, 2.6, 2.7, 3.2, 3.3]
+        ruby: [jruby-9.4, 3.3]
```
alextwoods: Probably a bigger discussion on what versions we want to track against.  I see a point in not necessarily tracking this data for every ruby version, esp on PRs.  But for continuity over time, we likely want to keep a complete list for releases (ie, maybe we need separate tasks that run on version-3 branch vs PRs?).  But we can address that in the future.

tasks/gems/cloudwatchlogs.rb
```diff
@@ -0,0 +1,77 @@
...
+        'Aws::CloudWatchLogs'
+      end
+
+      def operation_benchmarks
+        {
+          put_log_events_small: {
+            setup: proc do |_client|
+              { log_group_name: 'log_group', log_stream_name: 'log_stream', log_events: (0...5).map do |i|
+                                                                                          { timestamp: Time.now.to_i, message: "test log event #{i}" }
+                                                                                        end }
```
jterapin: Could cut down the line length like follows? For this and others alike in this PR
```
def operation_benchmarks
  {
    put_log_events_small: {
      setup: proc do |_client|
        {
          log_group_name: 'log_group',
          log_stream_name: 'log_stream',
          log_events: (0...5).map do |i|
            { timestamp: Time.now.to_i, message: "test log event #{i}" }
          end
        }
      end
      ...
```

.github/workflows/benchmark.yml
```diff
@@ -32,29 +33,31 @@ jobs:
...
-      - name: configure aws credentials
+      - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@v4
         with:
-          role-to-assume: arn:aws:iam::469596866844:role/aws-sdk-ruby-performance-reporter
-          role-session-name: benchmark-report
-          aws-region: us-west-2
+          role-to-assume: arn:aws:iam::373952703873:role/BenchmarkReporter
+          role-session-name: benchmark-reporter
+          aws-region: us-east-1
```
jterapin, line 58: Curious to know why we are switching regions here?

jterapin (approved):

.github/workflows/benchmark.yml
mullermp: We can do latest versions or all supported versions.

.github/workflows/benchmark.yml
mullermp: No particular reason other than our other team resources are in us-east-1 here.

tasks/gems/cloudwatchlogs.rb
mullermp: Fixed up all gem files.

PR #3078, opened 2024-07-31
Refreshing credentials in SharedCredentials

zendesk-duncanboynton:
This pull request introduces a new feature to the `SharedCredentials` class, enabling opt-in credential refreshing. This functionality allows credentials to be automatically refreshed when they are near expiration, based on user-defined settings.

This functionality is crucial for long-running applications that would otherwise need to use unconventional or impractical methods to refresh credentials (e.g., using `cat` to access PATH variables during refresh). With this feature, `SharedCredentials` can handle refreshing directly, simplifying the process.

[PR #1619](https://github.com/aws/aws-sdk-ruby/pull/1619) proposed this feature but was closed due to being an opt-out implementation, which posed a potential breaking change. This PR addresses those concerns by making it opt-in and maintaining backward compatibility.

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
@@ -89,6 +89,10 @@ module Aws
 
     private
 
+    def refresh_if_necessary
+      refresh! if @enable_refresh && near_expiration?
+    end
+
     def sync_expiration_length
       @refresh_interval || self.class::SYNC_EXPIRATION_LENGTH
     end
--- a/gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
@@ -70,7 +70,6 @@ module Aws
     end
 
     # Refreshes credentials from the shared credentials file.
-    # Refreshes credentials from the shared credentials file.
     def refresh
       shared_config = Aws.shared_config
 
@@ -88,13 +87,6 @@ module Aws
 
       # Set expiration time if credentials are present
       @expiration = @credentials ? (Time.now + @refresh_interval) : nil
-    end
-
-    # For testing purposes to check the refresh logic
-    # This method triggers the internal refresh logic as if the credentials
-    # were near expiration.
-    def force_refresh_check
-      refresh_if_near_expiration! if @enable_refresh
     end
 
     private
--- a/gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
@@ -45,9 +45,7 @@ module Aws
       @path = options[:path] || Aws.shared_config.credentials_path
       @profile_name = options[:profile_name] || ENV['AWS_PROFILE'] || Aws.shared_config.profile_name
 
-      super(options) # This will call RefreshingCredentials#initialize
-
-      refresh # Initially load credentials
+      super(options)
     end
 
     # @return [String] The path to the credentials file
@@ -90,10 +88,6 @@ module Aws
     end
 
     private
-
-    def refresh_if_necessary
-      refresh! if @enable_refresh && near_expiration?
-    end
 
     def sync_expiration_length
       @refresh_interval || self.class::SYNC_EXPIRATION_LENGTH
--- a/gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
@@ -1,17 +1,18 @@
 # frozen_string_literal: true
 
 require_relative 'ini_parser'
+require_relative 'refreshing_credentials'
 
 module Aws
   class SharedCredentials
-
     include CredentialProvider
+    include RefreshingCredentials
 
     # @api private
     KEY_MAP = {
       'aws_access_key_id' => 'access_key_id',
       'aws_secret_access_key' => 'secret_access_key',
-      'aws_session_token' => 'session_token',
+      'aws_session_token' => 'session_token'
     }
 
     # Constructs a new SharedCredentials object. This will load static
@@ -25,37 +26,37 @@ module Aws
     # You may access the resolved credentials through
     # `client.config.credentials`.
     #
-    # @option [String] :path Path to the shared file.  Defaults
+    # @option [String] :path Path to the shared file. Defaults
     #   to "#{Dir.home}/.aws/credentials".
     #
     # @option [String] :profile_name Defaults to 'default' or
     #   `ENV['AWS_PROFILE']`.
     #
+    # @option [Integer] :refresh_interval The duration (in seconds) before the
+    #   credentials are considered near expiration and should be refreshed.
+    #   Defaults to 10 minutes (600 seconds).
+    #
+    # @option [Boolean] :enable_refresh If true, the credentials will be automatically
+    #   refreshed when they are near expiration. Defaults to false.
+    #
     def initialize(options = {})
-      shared_config = Aws.shared_config
-      @path = options[:path]
-      @path ||= shared_config.credentials_path
-      @profile_name = options[:profile_name]
-      @profile_name ||= ENV['AWS_PROFILE']
-      @profile_name ||= shared_config.profile_name
-      if @path && @path == shared_config.credentials_path
-        @credentials = shared_config.credentials(profile: @profile_name)
-      else
-        config = SharedConfig.new(
-          credentials_path: @path,
-          profile_name: @profile_name
-        )
-        @credentials = config.credentials(profile: @profile_name)
-      end
+      @enable_refresh = options[:enable_refresh] || false # Refresh disabled by default
+      @refresh_interval = options[:refresh_interval] || 10 * 60 # Default refresh interval: 10 minutes
+      @path = options[:path] || Aws.shared_config.credentials_path
+      @profile_name = options[:profile_name] || ENV['AWS_PROFILE'] || Aws.shared_config.profile_name
+
+      super(options) # This will call RefreshingCredentials#initialize
+
+      refresh # Initially load credentials
     end
 
-    # @return [String]
+    # @return [String] The path to the credentials file
     attr_reader :path
 
-    # @return [String]
+    # @return [String] The name of the profile being used
     attr_reader :profile_name
 
-    # @return [Credentials]
+    # @return [Credentials] The loaded credentials
     attr_reader :credentials
 
     # @api private
@@ -63,19 +64,58 @@ module Aws
       parts = [
         self.class.name,
         "profile_name=#{profile_name.inspect}",
-        "path=#{path.inspect}",
+        "path=#{path.inspect}"
       ]
       "#<#{parts.join(' ')}>"
     end
 
-    # @deprecated This method is no longer used.
-    # @return [Boolean] Returns `true` if a credential file
-    #   exists and has appropriate read permissions at {#path}.
-    # @note This method does not indicate if the file found at {#path}
-    #   will be parsable, only if it can be read.
-    def loadable?
-      !path.nil? && File.exist?(path) && File.readable?(path)
+    # Refreshes credentials from the shared credentials file.
+    def refresh
+      shared_config = Aws.shared_config
+
+      if @path && @path == shared_config.credentials_path
+        # Use credentials from shared config if paths match
+        @credentials = shared_config.credentials(profile: @profile_name)
+      else
+        # Load credentials from specified path and profile
+        config = SharedConfig.new(
+          credentials_path: @path,
+          profile_name: @profile_name
+        )
+        @credentials = config.credentials(profile: @profile_name)
+      end
+
+      if @credentials && @credentials.set?
+        # Credentials successfully loaded and set
+        @access_key_id = @credentials.access_key_id
+        @secret_access_key = @credentials.secret_access_key
+        @session_token = @credentials.session_token
+        @expiration = Time.now + @refresh_interval
+      else
+        # Incomplete or missing credentials
+        @credentials = nil
+        @access_key_id = nil
+        @secret_access_key = nil
+        @session_token = nil
+        @expiration = nil
+      end
     end
 
+    # For testing purposes to check the refresh logic
+    # This method triggers the internal refresh logic as if the credentials
+    # were near expiration.
+    def force_refresh_check
+      refresh_if_near_expiration! if @enable_refresh
+    end
+
+    private
+
+    def sync_expiration_length
+      @refresh_interval || self.class::SYNC_EXPIRATION_LENGTH
+    end
+
+    def async_expiration_length
+      @refresh_interval || self.class::ASYNC_EXPIRATION_LENGTH
+    end
   end
 end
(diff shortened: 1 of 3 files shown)


gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
```diff
@@ -1,17 +1,18 @@
...
+require_relative 'refreshing_credentials'
 
 module Aws
   class SharedCredentials
-
     include CredentialProvider
+    include RefreshingCredentials
 
     # @api private
     KEY_MAP = {
```
mullermp: I'm pretty sure this thing can be deleted now.

gems/aws-sdk-core/spec/aws/shared_credentials_spec.rb
```diff
@@ -121,20 +121,59 @@ module Aws
...
+    it 'refreshes credentials when near expiration' do
+      refresh_interval = 2 # Short interval for testing
+      creds = SharedCredentials.new(path: mock_credential_file, refresh_interval: refresh_interval,
+                                    enable_refresh: true)
+
+      initial_access_key = creds.credentials.access_key_id
+      allow(creds).to receive(:refresh).and_call_original
+
+      # Simulate time passing to near expiration
+      sleep(refresh_interval + 1)
```
mullermp: Instead of sleeping in tests, you should mock `expiration` - there are examples of this in assume_role_credentials_spec and other refreshing credential provider specs.

gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
```diff
@@ -25,57 +26,96 @@ class SharedCredentials
...
-        @credentials = shared_config.credentials(profile: @profile_name)
-      else
-        config = SharedConfig.new(
-          credentials_path: @path,
-          profile_name: @profile_name
-        )
-        @credentials = config.credentials(profile: @profile_name)
-      end
+      @enable_refresh = options[:enable_refresh] || false # Refresh disabled by default
+      @refresh_interval = options[:refresh_interval] || 10 * 60 # Default refresh interval: 10 minutes
```
mullermp, line 31: Is interval a necessary config? Aren't the defaults of 10 minute async and 5 min sync sufficient? If you need a refresh immediately, refresh can be called on the instance.

gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
```diff
@@ -25,57 +26,96 @@ class SharedCredentials
...
+      @path = options[:path] || Aws.shared_config.credentials_path
+      @profile_name = options[:profile_name] || ENV['AWS_PROFILE'] || Aws.shared_config.profile_name
+
+      super(options) # This will call RefreshingCredentials#initialize
```
mullermp: The comments here and above are probably overkill.. the code describes itself here.

gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
```diff
@@ -25,57 +26,96 @@ class SharedCredentials
...
-    # @deprecated This method is no longer used.
-    # @return [Boolean] Returns `true` if a credential file
-    #   exists and has appropriate read permissions at {#path}.
-    # @note This method does not indicate if the file found at {#path}
-    #   will be parsable, only if it can be read.
-    def loadable?
-      !path.nil? && File.exist?(path) && File.readable?(path)
+    # Refreshes credentials from the shared credentials file.
+    def refresh
+      shared_config = Aws.shared_config
```
mullermp: You can remove this one too like you did in initialize.

gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
```diff
@@ -25,57 +26,96 @@ class SharedCredentials
...
       ]
       "#<#{parts.join(' ')}>"
     end
 
-    # @deprecated This method is no longer used.
-    # @return [Boolean] Returns `true` if a credential file
-    #   exists and has appropriate read permissions at {#path}.
-    # @note This method does not indicate if the file found at {#path}
-    #   will be parsable, only if it can be read.
-    def loadable?
```
mullermp, line 76: I'm in favor of removing this but it may break someone using this class directly. We can take that risk.

gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
```diff
@@ -25,57 +26,96 @@ class SharedCredentials
...
+        @session_token = @credentials.session_token
+        @expiration = Time.now + @refresh_interval
+      else
+        # Incomplete or missing credentials
+        @credentials = nil
+        @access_key_id = nil
+        @secret_access_key = nil
+        @session_token = nil
+        @expiration = nil
+      end
```
mullermp: Hmm, there are no accessors for access key id, etc, I think only credentials and expiration are necessary for refreshing credential provider? I don't think you also need to check set? because something else that uses refreshing credential provider will already check this. I would move expiration up into the branched code above as well.

gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
```diff
@@ -25,57 +26,96 @@ class SharedCredentials
...
+    end
+
+    # For testing purposes to check the refresh logic
+    # This method triggers the internal refresh logic as if the credentials
+    # were near expiration.
+    def force_refresh_check
```
mullermp: We should not have a public interface just for testing. Were you looking for the `refresh!` method here? You can mock expiration.

mullermp (commented):
This is an interesting idea. I will discuss with the team.

gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
alextwoods: I think the refresh_interval is necessary here because `expiration` isn't available from the credentials file, so the code below in `refresh` is using it to trigger when a refresh will happen.

gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
```diff
@@ -25,57 +26,96 @@ class SharedCredentials
...
-        @credentials = config.credentials(profile: @profile_name)
-      end
+      @enable_refresh = options[:enable_refresh] || false # Refresh disabled by default
+      @refresh_interval = options[:refresh_interval] || 10 * 60 # Default refresh interval: 10 minutes
+      @path = options[:path] || Aws.shared_config.credentials_path
+      @profile_name = options[:profile_name] || ENV['AWS_PROFILE'] || Aws.shared_config.profile_name
+
+      super(options) # This will call RefreshingCredentials#initialize
+
+      refresh # Initially load credentials
```
alextwoods: I believe the `RefreshingCredentials#initialize` will call this already.

gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
```diff
@@ -25,57 +26,96 @@ class SharedCredentials
...
-      if @path && @path == shared_config.credentials_path
-        @credentials = shared_config.credentials(profile: @profile_name)
-      else
-        config = SharedConfig.new(
-          credentials_path: @path,
-          profile_name: @profile_name
-        )
-        @credentials = config.credentials(profile: @profile_name)
-      end
+      @enable_refresh = options[:enable_refresh] || false # Refresh disabled by default
```
alextwoods, line 30: Its not clear to me how `enable_refresh` works - when set to false, how does this prevent refreshing?  As far as I can tell, its only used in the `force_refresh_check` method, which is only used in testing unless I'm missing something.

gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
zendesk-duncanboynton: I had issues taking it out and still refreshing the credentials, I am relatively new to Ruby

gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
zendesk-duncanboynton: I had issues taking this line out here

gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
zendesk-duncanboynton: Should be fixed!

zendesk-duncanboynton:
@alextwoods @mullermp I'd greatly appreciate another look- I did my best to refine my changes and incorporate your comments

gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
```diff
@@ -25,57 +26,82 @@ class SharedCredentials
...
-    #   will be parsable, only if it can be read.
-    def loadable?
-      !path.nil? && File.exist?(path) && File.readable?(path)
+    # Refreshes credentials from the shared credentials file.
+    def refresh
+      shared_config = Aws.shared_config
+
+      if @path && @path == shared_config.credentials_path
+        # Use credentials from shared config if paths match
+        @credentials = shared_config.credentials(profile: @profile_name)
```
alextwoods: Does this actually refresh credentials? `Aws.shared_config` I don't think will re-load/parse the config files, so I'm not sure you would actually get new credentials from this.

gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
```diff
@@ -25,57 +26,82 @@ class SharedCredentials
...
+        # Load credentials from specified path and profile
+        config = SharedConfig.new(
+          credentials_path: @path,
+          profile_name: @profile_name
+        )
+        @credentials = config.credentials(profile: @profile_name)
+      end
+
+      # Set expiration time if credentials are present
+      @expiration = @credentials ? (Time.now + @refresh_interval) : nil
```
alextwoods: We may not actually want to set this as it has implications for *using* the credentials (Eg: the credential's expiration is used when creating a presigned url).  Instead we may need an internal variable to track this.

gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
```diff
@@ -25,57 +26,82 @@ class SharedCredentials
...
+    end
+
+    private
+
+    def refresh_if_necessary
+      refresh! if @enable_refresh && near_expiration?(@refresh_interval)
+    end
+
+    def sync_expiration_length
```
alextwoods: I don't think we should need to define these here

gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
mullermp: Hmm. I don't see KEY_MAP used anywhere in the SDK anymore. What issue did you see deleting it?

gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
mullermp: My original comment may have been misleading. I think this is right.

gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
mullermp: This would have to be SharedConfig.new, always, right?

gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
mullermp: I think we could probably just overwrite `near_expiration?(expiration_length)`, and remove all of these private methods, and remove credentials method overwrite.

mullermp:
Can I ask more about your use case? Is it not possible to write your own custom CredentialProvider that sources your credentials from a dynamic source? Why are your credentials on disk (meant to be static) changing?

gems/aws-sdk-core/lib/aws-sdk-core/shared_credentials.rb
zendesk-duncanboynton: turns out I think this is needed to get creds from shared cfg

zendesk-duncanboynton:
> Can I ask more about your use case? Is it not possible to write your own custom CredentialProvider that sources your credentials from a dynamic source? Why are your credentials on disk (meant to be static) changing?

@mullermp Our use case involves long-running applications that require dynamic, temporary credentials for an IAM Role. These credentials need periodic refreshing to maintain security and access.

mullermp:
Have you looked at ProcessCredentials? You can have some process that is responsible for returning credentials, and that process is responsible for vending new and rotated credentials. Shared config is for static credentials typically.

zendesk-duncanboynton:
@mullermp that works but if the credentials are in a json file we have to do something like `Aws::ProcessCredentials.new('cat $SHARED_CREDENTIALS_JSON')` which is not a good experience

mullermp:
Hmm. I see these as functionally equivalent. If you're writing out new rotated credentials on disk, you can write them to a json and echo them in process credentials. That's functionally equivalent to re-reading configuration in Ruby with parsing (which may be more expensive). Why do you feel it's a bad experience?

alextwoods:
A slightly different question - how is the credentials file getting updated with the new temporary IAM credentials?  I assume there is some process running which manages that.

Also - RE: using the ProcessCredentials as suggested by @mullermp  - I believe that would only refresh if the credentials have an expiration time on them (in which case that expiration will be used to manage when the refresh happens).

derricklam-zendesk:
> A slightly different question - how is the credentials file getting updated with the new temporary IAM credentials? 

Hi @alextwoods and @mullermp, the credentials file is refreshed by a secret sidecar. The issue we are attempting to solve is the ability for CredentialProvider to consume a credentials filepath. ex:

```
Aws::ProcessCredentials.new("/path/to/creds/file")
```  

It looks like [SharedCredentials](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/SharedCredentials.html) supports a `path` parameter today, but [ProcessCredentials](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/ProcessCredentials.html) does not. Would it be possible for ProcessCredentials to accept a path as well?

mullermp:
The process that is executed can be literally whatever you want at whatever path you want. For your use case, are you not able to do simply `echo path/to/creds.json`?

derricklam-zendesk:
Right, `Aws::ProcessCredentials.new('cat $SHARED_CREDENTIALS_JSON')` is our existing approach today

However, we are looking to make this small improvement so that we don't have to run process `cat` or `echo`

alextwoods:
I think what I'm trying to understand a little bit is how your process updates the shared credentials file.  

It sounds like you are using short term/temporary IAM credentials which should come with an expiration.  A limitation of the shared credentials file (eg `.aws/credentials`) used by the `SharedCredentials` is that is does NOT support expiration.  The refreshing of the credentials used by the SDK should be based on that expiration, which the `ProcessCredentials` do support automatically (as long as the expiration is included in the returned json).  This seems like a better solution than relying on a configured refresh interval (which is easy to misconfigure and may not get updated correctly if you update the expiration times for your temporary credentials).

mullermp:
I agree with Alex. I'm also wondering what you mean by "small improvement" - process credentials calling echo or cat is much faster than parsing an entire shared config file in Ruby. I actually think this would be a performance hit. We also discussed this with SDK teams internally and decided this is not something we should support in SDKs.

mullermp:
I'm going to close this after further discussions with SDK engineers. We believe ProcessCredentials is the correct way to support this use case both from a practical and performance standpoint.

PR #3079, opened 2024-08-01
Support non x-amz headers in s3 presigned request

mullermp:
With this change, add all headers (regardless if they are x-amz or not) to presigned request.

### DIFF
--- a/gems/aws-sdk-s3/lib/aws-sdk-s3/presigner.rb
+++ b/gems/aws-sdk-s3/lib/aws-sdk-s3/presigner.rb
@@ -192,7 +192,7 @@ module Aws
       def sign_but_dont_send(
         req, expires_in, secure, time, unsigned_headers, hoist = true
       )
-        x_amz_headers = {}
+        headers = {}
 
         http_req = req.context.http_request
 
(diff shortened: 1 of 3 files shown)


gems/aws-sdk-s3/lib/aws-sdk-s3/presigner.rb
```diff
@@ -211,16 +211,18 @@ def sign_but_dont_send(
 
           query = http_req.endpoint.query ? http_req.endpoint.query.split('&') : []
           http_req.headers.each do |key, value|
-            next unless key =~ /^x-amz/i
+            next if key == 'user-agent'
 
             if hoist
+              next unless key =~ /^x-amz/i
```
alextwoods, line 217: is there a case where we would want to hoist range?

alextwoods (approved):

gems/aws-sdk-s3/lib/aws-sdk-s3/presigner.rb
mullermp: I'm not convinced we should merge this PR yet, but I made it anyway. I don't think we want to hoist non x-amz headers. I actually couldn't get this to work with a live test (maybe you can give it a try with get_object range and put_object with content-md5).

PR #3083, opened 2024-08-07
Reduce initial memory usage by auto-loading bundled gems 

alextwoods:
Reduce initial memory usage by auto-loading bundled gems (STS, SSO, SSOOIDC).

These gems are used only by certain credential providers in core and are usually not all needed in any given application.

Memory used when requiring aws-sdk-core:
* BEFORE Total allocated: 22028954 bytes (183674 objects)
* AFTER   Total allocated: 19133824 bytes (161203 objects)


gems/aws-sdk-core/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Feature - Reduce initial memory usage by auto-loading bundled gems (STS, SSO, SSOOIDC).
```
mullermp: I think this should be an issue bump.

gems/aws-sdk-core/lib/aws-sdk-core.rb
```diff
@@ -205,3 +197,14 @@ def eager_autoload!(*args)
 
   end
 end
+
+# Autoload bundled service gems used in credential providers
+# need to ensure that the files are the local files from aw-sdk-core
```
mullermp: Typo

gems/aws-sdk-resources/bin/aws-v3.rb
```diff
@@ -107,10 +107,11 @@ def env_bool key, default
 require 'aws-sdk-resources'
 
 # Finally load all of the gems. Core is loaded a second time because of STS.
```
mullermp: Is this comment still relevant? Also I think this gem should be bumped too and it probably needs a minimum core version bump because of that.

tasks/update-aws-sdk-dependencies.rake
```diff
@@ -19,11 +19,12 @@ task 'update-aws-sdk-dependencies' do
   )
 
   # update the module autoloads
+  autoload_services = BuildTools::Services.select { |svc| svc.gem_name != 'aws-sdk-core' }
   BuildTools.replace_lines(
     filename: "#{$GEMS_DIR}/aws-sdk-resources/lib/aws-sdk-resources.rb",
     start: /# service gems/,
     stop: /# end service gems/,
-    new_lines: BuildTools::Services.map { |service|
+    new_lines: autoload_services.map { |service|
```
mullermp, line 28: Why is core ignored here?

gems/aws-sdk-resources/bin/aws-v3.rb
alextwoods: I'm not sure it ever was actually correct. I've updated.

aws-sdk-resources will get a version bump because of the codegen changes (removing the autoload for sts,sso,ssoidc).  
I don't believe it needs a new minimum core version - with old (current) core versions SSO,STS,SSOOIDC are all explicitly required so the autoload thats being removed would not have done anything.

tasks/update-aws-sdk-dependencies.rake
alextwoods: This is a little confusing - this is actually removing the bundled service gems from the autoload.  They have a gem name of aws-sdk-core (since they are bundled in it). I've added comments.

mullermp (approved):

PR #3090, opened 2024-08-20
Default checksum calculation for httpChecksum to CRC32

mullermp:
Defaults CRC32 for httpChecksum implementations and adds CRC64 as a new algorithm.


gems/aws-sdk-core/CHANGELOG.md
```diff
@@ -1,10 +1,15 @@
 Unreleased Changes
 ------------------
 
+* Feature - Always calculate checksums for operations that support or require it. Supported config options are `WHEN_SUPPORTED` and `WHEN_REQUIRED`. The default value is `WHEN_SUPPORTED`. This option is configured in code with `:request_checksum_calculation`, in the shared config file as `request_checksum_calculation`, and in the ENV as `ENV['AWS_REQUEST_CHECKSUM_CALCULATION']`.
```
alextwoods: Nit, mention request.  eg: "Always calculate request checksums..."

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
```diff
@@ -55,79 +150,59 @@ def base64digest
...
-          if (ChecksumAlgorithm.request_validation_mode(context))
-            # Compute an ordered list as the union between priority supported and the
-            # operation's modeled response algorithms.
-            validation_list = CHECKSUM_ALGORITHM_PRIORITIES &
-              ChecksumAlgorithm.operation_response_algorithms(context)
-            context[:http_checksum][:validation_list] = validation_list
+          # Default checksum member to CRC32 if not set
+          default_request_algorithm_member(context)
+          # Not modeled with httpChecksum
+          if context.operation_name == :create_multipart_upload
```
alextwoods: This potentially could cause an issue with a non-s3 service w/ a create_multipart_upload operation...

It might be better to move this to a handler in S3 that sets a value on context that we check here? (like we did with the skip on prefix)

gems/aws-sdk-core/lib/aws-sdk-core/plugins/http_checksum.rb
```diff
@@ -12,7 +12,7 @@ class Handler < Seahorse::Client::Handler
 
         def call(context)
           if checksum_required?(context) &&
-             !context[:checksum_algorithms] && # skip in favor of flexible checksum
+             !context[:http_checksum][:request_algorithm] && # skip in favor of flexible checksum
```
alextwoods, line 15: Do we need to leave the old check in place for old service gem + new core?

Edit: NVM, that shouldn't be an issue since the context values get set by handlers in core, not the service.

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
```diff
@@ -159,14 +159,16 @@ def upload_in_threads(pending, completed, options)
...
-                  completed_part[k] = resp[k]
+                completed_part = {
+                  etag: resp.etag,
+                  part_number: part[:part_number]
+                }.tap do |h|
+                  # get checksum algorithm from request params (default or part)
+                  algorithm = resp.context.params[:checksum_algorithm]
+                  k = "checksum_#{algorithm.downcase}".to_sym
+                  # get the requested checksum from the response
+                  h[k] = resp.send(k)
```
alextwoods: Is there a reason we do `send` instead of `[k]` here?

alextwoods (commented):
Nice - looks good overall.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
mullermp: Good catch. I'll refactor that plugin to be generic and add two handlers.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/http_checksum.rb
mullermp: I think so yeah, it's been "promoted" to core in this case.

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
mullermp: Using the data accessor instead as intended. It shouldn't conflict with any ruby names..

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
```diff
@@ -13,34 +13,129 @@ class ChecksumAlgorithm < Seahorse::Client::Plugin
...
+             docstring: <<~DOCS) do |cfg|
+               Determines when a checksum will be calculated for request payloads. Values are:
+
+               * `when_supported` - (default) When set, a checksum will be
+                 calculated for all request payloads of operations modeled with the
+                 `httpChecksum` trait where `requestChecksumRequired` is `true` and/or a
+                 `requestAlgorithmMember` is modeled.
+               * `when_required` - When set, a checksum will only be calculated for
+                 request payloads of operations modeled with the  `httpChecksum` trait where
+                 `requestChecksumRequired` is `true` or where a requestAlgorithmMember
```
jterapin: missing backticks around `requestAlgorithmMember` to match the previous value option

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
```diff
@@ -13,34 +13,129 @@ class ChecksumAlgorithm < Seahorse::Client::Plugin
...
+            raise ArgumentError,
+                  'expected :request_checksum_calculation or' \
+                  " ENV['AWS_REQUEST_CHECKSUM_CALCULATION'] to be " \
+                  '`when_supported` or `when_required`.'
+          end
+          mode
+        end
+
+        def resolve_response_checksum_validation(cfg)
+          mode = ENV['AWS_response_checksum_validation'] ||
```
jterapin: Shouldn't the ENV variable be all in caps?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
```diff
@@ -49,122 +144,207 @@ def update(chunk)
...
+          add_request_config_metric(config, metrics)
+          add_response_config_metric(config, metrics)
+          add_request_checksum_metrics(algorithm, metrics)
+          Aws::Plugins::UserAgent.metric(*metrics, &block)
+        end
+
+        def add_request_config_metric(config, metrics)
+          case config.request_checksum_calculation
+          when 'when_supported'
+            metrics << 'FLEXIBLE_CHECKSUMS_REQ_when_supported'
```
jterapin: `'...when_supported'` and `'....when required'` should be all caps based on features chart

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
```diff
@@ -49,122 +144,207 @@ def update(chunk)
...
+          !checksum_provided_as_header?(context.http_request.headers) &&
+            request_algorithm_selection(context) &&
+            (checksum_required?(context) || checksum_optional?(context))
+        end
+
+        def choose_request_algorithm!(context)
+          algorithm = request_algorithm_selection(context).upcase
+          return algorithm if CLIENT_ALGORITHMS.include?(algorithm)
+
+          if %w[CRC32C CRC64NVME].include?(algorithm)
```
jterapin: During review, I was wondering which checksum algos needed the CRT gem - might worth turning this into a constant?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
mullermp: Yeah, bad search and replace. Good catch

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
mullermp: They should be lower case options.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
```diff
@@ -49,122 +144,207 @@ def update(chunk)
...
-        def should_verify_response_checksum?(context)
-          context[:http_checksum][:validation_list] && !context[:http_checksum][:validation_list].empty?
+        def checksum_provided_as_header?(headers)
+          headers.any? { |k, _| k.start_with?('x-amz-checksum-') }
+        end
+
+        def should_calculate_request_checksum?(context)
+          !checksum_provided_as_header?(context.http_request.headers) &&
+            request_algorithm_selection(context) &&
+            (checksum_required?(context) || checksum_optional?(context))
```
alextwoods, line 292: I believe there is an issue here with a user configures a client with an invalid `request_checksum_calculation`.  We don't do any validation of that when set on the client and then these two are mutually exclusive (and check only the known values).  

We could solve that in a few ways, but I'd vote for just a validation in client after init personally.  Alternatively, checksum_optional? could be changed to check `request_checksum_calculation != 'when_required'`

gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/md5s.rb
```diff
@@ -6,81 +6,26 @@ module Aws
...
+               default: true,
+               doc_type: 'Boolean',
+               docstring: <<~DOCS)
+                 This option is deprecated. Please use `:request_checksum_calculation`
+                 instead. When `true`, `request_checksum_calculation` is set to `when_supported`,
+                 and if `false` it is set to `when_required`.
+               DOCS
+
+        def after_initialize(client)
+          client.config.request_checksum_calculation =
```
alextwoods: I see why we need to do this for the legacy option.... but I am pretty concerned about the customer behavior from this.  Doing this effectively means that `request_checksum_calculation` does not work at all for S3 - that is, its value is ignored no matter where it is set (client, ENV, shared config).  I think thats pretty confusing.

mullermp: Yeah, agreed. I think `client.config.request_checksum_calculation = 'when_required' unless client.config.compute_checksums` satisfies the concern?

User doesn’t set compute checksums:
* This value defaults to true
* User can freely set new config to required or optional and it will be honored.

User set compute checksums to false:
* Setting new config to when_required basically does the same thing
* Setting to when_optional would maybe be confusing because it would be overwritten, but they have explicitly already chosen a conflicting option of compute checksums false.

I can update the documentation but the "false" case and "when_optional" is probably a less likely use case. However computing checksums in general would make sense to override the when optional anyway.

gems/aws-sdk-core/CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 Unreleased Changes
 ------------------
 
+* Feature - Always calculate request checksums for operations that support or require it. Supported config options are `when_supported` and `when_required`. The default value is `when_supported`. This option is configured in code with `:request_checksum_calculation`, in the shared config file as `request_checksum_calculation`, and in the ENV as `ENV['AWS_REQUEST_CHECKSUM_CALCULATION']`.
+
+* Feature - Always validate response checksums for operations that support or require it. Supported config options are `when_supported` and `when_required`. The default value is `when_supported`. This option is configured in code with `:response_checksum_validation`, in the shared config file as `response_checksum_validation`, and in the ENV as `ENV['AWS_response_checksum_validation']`.
```
jterapin: Fix the reference ENV here to all caps as mentioned previously

PR #3092, opened 2024-08-20
Observability Support

jterapin:
Following changes has been made to implement Observability:

**Build Tools**
* Add telemetry plugin to plugin list
* Update `async_client_class` and `client_class` files to include `tracer` when creating `RequestContext`

**Core Changes**
* Add Telemetry plugin, `telemetry_provider` config and Telemetry handler
* Add the following telemetry provider support: no-op and otel
* Add span wrappers for the following handlers: StubResponses, NetHttp and H2
* Update RequestContext to include a `tracer` attribute
* Add specs
* Add RBS support

**Other Changes**
* Add `opentelemetry-sdk` to the Gemfile (for testing purposes)

---


gems/aws-sdk-core/lib/seahorse/client/request_context.rb
```diff
@@ -9,11 +9,14 @@ class RequestContext
       # @option options [required,Symbol] :operation_name (nil)
       # @option options [required,Model::Operation] :operation (nil)
       # @option options [Model::Authorizer] :authorizer (nil)
+      # @option options [Client] :client (nil)
       # @option options [Hash] :params ({})
       # @option options [Configuration] :config (nil)
       # @option options [Http::Request] :http_request (Http::Request.new)
       # @option options [Http::Response] :http_response (Http::Response.new)
-      #   and #rewind.
```
jterapin, line 16: This looked really odd to me so I removed it.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/plugin_list.rb
```diff
@@ -63,7 +63,8 @@ def default_plugins
         'Aws::Plugins::ChecksumAlgorithm' => "#{core_plugins}/checksum_algorithm.rb",
         'Aws::Plugins::RequestCompression' => "#{core_plugins}/request_compression.rb",
         'Aws::Plugins::DefaultsMode' => "#{core_plugins}/defaults_mode.rb",
-        'Aws::Plugins::RecursionDetection' => "#{core_plugins}/recursion_detection.rb"
+        'Aws::Plugins::RecursionDetection' => "#{core_plugins}/recursion_detection.rb",
+        'Aws::Plugins::Telemetry' => "#{core_plugins}/telemetry.rb"
```
alextwoods, line 67: I think we'll need to update the min core version for all gems (otherwise this may fail for users with old versions of core).

gems/aws-sdk-core/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Feature - Add support for Observability.
```
alextwoods: May want to mention the config option + otel support as part of the feature here.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/telemetry.rb
```diff
@@ -0,0 +1,75 @@
...
+
+module Aws
+  module Plugins
+    # @api private
+    class Telemetry < Seahorse::Client::Plugin
+      option(
+        :telemetry_provider,
+        default: Aws::Telemetry::NoOpTelemetryProvider,
+        doc_type: Aws::Telemetry::TelemetryProviderBase,
+        rbs_type: 'untyped',
```
alextwoods: we do have rbs types for this right?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/telemetry.rb
```diff
@@ -0,0 +1,75 @@
...
+        docstring: <<~DOCS) do |_cfg|
+Allows you to provide a telemetry provider, which is used
+to emit telemetry data. By default, will use `NoOpTelemetryProvider`
```
alextwoods: May be worth mentioning the otel provider here for discoverability.

build_tools/aws-sdk-code-generator/templates/client_class.mustache
```diff
@@ -105,7 +105,11 @@ module {{module_name}}
         authorizer: authorizer,{{/authorizer?}}
         client: self,
         params: params,
-        config: config)
+        config: config,
+        tracer: config.telemetry_provider.tracer_provider.tracer(
```
alextwoods: Do we need to account for any cases where either:
1. config does not have telemetry provider
2. the telemtry_provider is nil

gems/aws-sdk-core/lib/aws-sdk-core/plugins/telemetry.rb
```diff
@@ -0,0 +1,75 @@
...
+which will not record or emit any telemetry data.
+        DOCS
+        resolve_provider
```
alextwoods: Since this is just a default rather than a resolution (search ENV/shared_config), I'd lean towards getting rid of the `resolve_provider` method here and just returning the NoOp impl.

gems/aws-sdk-core/lib/aws-sdk-core/telemetry.rb
```diff
@@ -0,0 +1,92 @@
...
+  # defined:
+  # * {TelemetryProviderBase}
+  # * {ContextManagerBase}
+  # * {TracerProviderBase}
+  # * {TracerBase}
+  # * {SpanBase}
+  module Telemetry
+    class << self
+      # @api private
+      def otel_loaded?
```
alextwoods: Is there a reason we need this in the base `Telemetry` module instead of just defining it in the Otel implementation (eg on the `OTelProvider` class)?

alextwoods (commented):
Nice - Looks good, some minor comments!

gems/aws-sdk-core/lib/aws-sdk-core/plugins/telemetry.rb
jterapin: Ahh, right. Thanks for the reminder!

gems/aws-sdk-core/lib/aws-sdk-core/telemetry.rb
jterapin: No particular reason, but I agree that make more sense under `OTelProvider` so I'll move it.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/plugin_list.rb
jterapin: Good call, I'll update both `MINIMUM_CORE_VERSION` and `MINIMUM_S3_VERSION`

build_tools/aws-sdk-code-generator/templates/client_class.mustache
jterapin: 1. When new gems are rebuilt, using this mustache template will always have a some kind of configuration for `telemetry_provider`, whether it is an no-op telemetry provider or otel-based telemetry provider. Upon a client creation, should no provider be given as a config, we will opt to using the no-op telemetry provider as intended for default.
2. I think this is covered by the validation that I have in the telemetry plugin. During client creation, if the `telemetry_provider` given is `nil`, it is caught at that moment.
```ruby
[5] pry(Aws)> Aws::S3::Client.new(telemetry_provider: nil)
ArgumentError: Must provide a telemetry provider for the `telemetry_provider` configuration option.
```

jterapin:
Hold off reviewing, I need to do another run-through to catch anything that I missed in the AM

build_tools/aws-sdk-code-generator/templates/client_class.mustache
mullermp: I think nil and no-op should probably be the same. Maybe when configured with nil, it can default to no-op? If it defaults as no-op that's fine too.

build_tools/aws-sdk-code-generator/templates/async_client_class.mustache
```diff
@@ -89,7 +89,11 @@ module {{module_name}}
         client: self,
         params: params,
         http_response: Seahorse::Client::Http::AsyncResponse.new,
-        config: config)
+        config: config,
+        tracer: config.telemetry_provider.tracer_provider.tracer(
```
mullermp: Config is already piped through. Couldn't request context have a method for tracer that just returned this thing?

gems/aws-sdk-core/lib/aws-sdk-core.rb
```diff
@@ -45,24 +43,23 @@
...
 
 # logging
-
 require_relative 'aws-sdk-core/log/formatter'
 require_relative 'aws-sdk-core/log/param_filter'
 require_relative 'aws-sdk-core/log/param_formatter'
 
-# stubbing
+# telemetry
+require_relative 'aws-sdk-core/telemetry'
```
mullermp: I'd classify this as "client metrics" below.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/stub_responses.rb
```diff
@@ -49,17 +49,21 @@ def after_initialize(client)
...
+            if Hash === stub && stub[:mutex]
+              stub[:mutex].synchronize { apply_stub(stub, resp, async_mode) }
+            else
+              apply_stub(stub, resp, async_mode)
+            end
+            async_mode ? Seahorse::Client::AsyncResponse.new(
+              context: context,
+              stream: context[:input_event_stream_handler].event_emitter.stream,
+              sync_queue: Queue.new
+            ) : resp
```
mullermp: If we're wrapping this in a block, probably best to refactor this into a method or methods now.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/telemetry.rb
```diff
@@ -0,0 +1,70 @@
...
+which will not record or emit any telemetry data. The SDK currently supports
+OpenTelemetry (OTel) as a telemetry provider. To use the OTel provider, 
+require the`opentelemetry-sdk` gem and then, pass in an instance of a 
+`Aws::Telemetry::OTelProvider` for telemetry provider.
+        DOCS
+        Aws::Telemetry::NoOpTelemetryProvider.new
+      end
+
+      def after_initialize(client)
+        validate_telemetry_provider(client.config)
```
mullermp, line 26: If it's set to nil, I think just make it a no-op here.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/telemetry.rb
```diff
@@ -0,0 +1,70 @@
...
+            'rpc.service' => service_id,
+            'rpc.method' => context.operation.name,
+            'code.function' => context.operation_name.to_s,
+            'code.namespace' => 'Aws::Plugins::Telemetry'
+          }
+          context.tracer.in_span(
+            parent_span_name(context, service_id),
+            attributes: attributes,
+            kind: Aws::Telemetry::SpanKind::CLIENT
+          ) do
```
mullermp: Maybe hide this behind a method like you did previously?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/telemetry.rb
```diff
@@ -0,0 +1,70 @@
...
+            @handler.call(context)
+          end
+        end
+
+        private
+
+        def service_id(context)
+          context.config.api.metadata['serviceId'] ||
```
mullermp, line 62: I think serviceId exists on all models now, you should check.

gems/aws-sdk-core/lib/aws-sdk-core/telemetry.rb
```diff
@@ -0,0 +1,78 @@
...
+      end
+
+      # @api private
+      def http_request_attrs(context)
+        {
+          'http.method' => context.http_request.http_method,
+          'net.protocol.name' => 'http'
+        }.tap do |h|
+          h['net.protocol.version'] =
+            if context.client.is_a? Seahorse::Client::AsyncBase
```
mullermp, line 42: Are you not able to get the version from our underlying handler/dependency?

gems/aws-sdk-core/lib/seahorse/client/h2/handler.rb
```diff
@@ -27,61 +27,62 @@ module H2
...
-              input_emitter.stream = stream
-              # request sigv4 serves as the initial #prior_signature
-              input_emitter.encoder.prior_signature =
-                context.http_request.headers['authorization'].split('Signature=').last
-              input_emitter.validate_event = context.config.validate_params
-            else
-              _send_initial_headers(context.http_request, stream)
-              _send_initial_data(context.http_request, stream)
+          span_wrapper(context) do
+            stream = nil
```
mullermp: Just like before, probably can refactor this. Easiest way to do this would probably make a `_call` method.

gems/aws-sdk-core/lib/aws-sdk-core.rb
jterapin: I will group them into a single section but update the comment to `client metrics/telemetry` since "client metrics" by itself does not encapsulate what telemetry does

gems/aws-sdk-core/lib/aws-sdk-core/plugins/telemetry.rb
jterapin: Yeah, I could do that for readability.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/telemetry.rb
jterapin: I took a quick look and not all models have `serviceId` - e.g. [example](https://github.com/aws/aws-sdk-ruby/blob/version-3/apis/importexport/2010-06-01/api-2.json#L3-L11).

gems/aws-sdk-core/lib/aws-sdk-core/telemetry.rb
jterapin: From what I looked (within core and http/2 dependency), I couldn't see anything within our realm or theirs to retrieve the HTTP version - I'm thinking that it may due to that HTTP/2 - there's only one? (unlike HTTP/1.0 and HTTP/1.1)

build_tools/aws-sdk-code-generator/templates/client_class.mustache
jterapin: Defaults to no-op so we are good here.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/telemetry.rb
jterapin: See [here](https://github.com/aws/aws-sdk-ruby/pull/3092/files#r1725679161).

build_tools/aws-sdk-code-generator/templates/async_client_class.mustache
jterapin: We could but I think this is simpler than creating a method and memorizing it.

mullermp: What do you mean? I'm suggesting a "tracer" method in RequestContext that looks at config for the tracer.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/telemetry.rb
mullermp: Importexport is deprecated and uses sigv2. We could add a customization to add service id to it just to make this work. All active services should have service id. I think we could actually delete this from the repo.

build_tools/aws-sdk-code-generator/templates/async_client_class.mustache
alextwoods: The tracer isn't on the config though right?  We just use the tracer provider to create a new tracer for this request and Its scoped to a request and should only be created once per request I think?

mullermp: But a request context is made each request already right?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/telemetry.rb
alextwoods: we have similar code that falls back to service abbreviation elsewhere right? I think its fine to just keep this or rather than add a customization (just a slight preference not a strong feeling).  Its true that all services going forward must have serviceId, so maybe it does make sense to just customize + remvove these special cases.
Or maybe we can put this fallback logic once, in our build tools somewhere so that it ensures the api in ruby runtime always has serviceId?

mullermp: All services should have serviceId backfilled. Other than importexport, which ones don't? They should arguably be removed.

mullermp: The logic we have I believe in release tools and I think maybe in gem name in build tools. I would be in favor of adding to customizations.rb to add service id for the few services missing it and then get them removed.

build_tools/aws-sdk-code-generator/templates/async_client_class.mustache
jterapin: Per our discussion, we are keeping this as it is.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/telemetry.rb
jterapin: Per our discussion, we are keeping the fall back here to cover cases with old services, new core scenario. That being said, added customization in the build_tools to resolve the missing serviceId for those three services.

build_tools/customizations.rb
```diff
@@ -67,9 +67,7 @@ def dynamodb_example_deep_transform(subsegment, keys)
     api('CloudFront') do |api|
 
       api['shapes'].each do |_, shape|
-        if shape['members'] && shape['members']['MaxItems']
-          shape['members']['MaxItems']['shape'] = 'integer'
-        end
+        shape['members']['MaxItems']['shape'] = 'integer' if shape['members'] && shape['members']['MaxItems']
```
alextwoods: >80 characters?

alextwoods (approved):
LGTM!

build_tools/customizations.rb
jterapin: Heck, I thought this was fixed - dang IDE. I'm on it!

gems/aws-sdk-core/lib/aws-sdk-core/plugins/stub_responses.rb
```diff
@@ -57,9 +65,11 @@ def call(context)
           else
             apply_stub(stub, resp, async_mode)
           end
-
           async_mode ? Seahorse::Client::AsyncResponse.new(
```
mullermp: Multiline ternaries are awful.. I'd convert this to regular if-else.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/telemetry.rb
```diff
@@ -0,0 +1,73 @@
...
+
+module Aws
+  module Plugins
+    # @api private
+    class Telemetry < Seahorse::Client::Plugin
+      option(
+        :telemetry_provider,
+        default: Aws::Telemetry::NoOpTelemetryProvider,
+        doc_type: Aws::Telemetry::TelemetryProviderBase,
+        rbs_type: Aws::Telemetry::TelemetryProviderBase,
```
mullermp, line 11: Does rbs_type default to doc_type?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/telemetry.rb
```diff
@@ -0,0 +1,73 @@
...
+        docstring: <<~DOCS) do |_cfg|
+Allows you to provide a telemetry provider, which is used
+to emit telemetry data. By default, uses `NoOpTelemetryProvider`
+which will not record or emit any telemetry data. The SDK currently supports
+OpenTelemetry (OTel) as a telemetry provider. To use the OTel provider,
```
mullermp: I'd convert this to a list of .. one thing .. that can be added to later.

gems/aws-sdk-core/spec/spec_helper.rb
```diff
@@ -70,6 +70,16 @@ def client_with_plugin(options = {}, &block)
       client_class.new(options)
     end
 
+    # clears opentelemetry-sdk configuration state between specs
+    # https://github.com/open-telemetry/opentelemetry-ruby/blob/main/test_helpers/lib/opentelemetry/test_helpers.rb#L18
+    def reset_opentelemetry_sdk
```
mullermp, line 75: You can possibly just move this to an after {} in shared_spec_helper and have it run all the time.

mullermp (approved):

gems/aws-sdk-core/lib/aws-sdk-core/plugins/telemetry.rb
jterapin: Far as I looked, I don't think so for V3 - I didn't see anything special done in [PluginOption](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/lib/seahorse/client/plugin.rb#L110) class when it comes to `rbs_type` so I will be leaving this as it is.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/telemetry.rb
jterapin: I will do it now to set up for future :)

gems/aws-sdk-core/spec/spec_helper.rb
jterapin: I think I'll opt to having this method here to use when I need it.

alextwoods (approved):

PR #3099, opened 2024-09-06
Additional metrics collection

mullermp:


gems/aws-sdk-core/spec/aws/plugins/request_compression_spec.rb
```diff
@@ -121,13 +121,6 @@ def expect_uncompressed_body(resp, body)
       end
 
       context 'request compression operation' do
-        it 'sets user-agent metric for the operation' do
```
jterapin, line 124: Are we opting not to write unit tests for user-agent metrics in general? Or will be there one spec that covers all current + newly added metric values?

build_tools/aws-sdk-code-generator/templates/spec/endpoint_provider_spec_class.mustache
```diff
@@ -11,7 +11,7 @@ module {{module_name}}
     subject { {{module_name}}::EndpointProvider.new }
 
     {{#endpoint_tests}}
-    context '{{documentation}}' do
+    context "{{{documentation}}}" do
```
jterapin, line 14: Do we expect special symbols in the context description?

jterapin (approved):
LGTM ✨

gems/aws-sdk-core/spec/aws/plugins/request_compression_spec.rb
mullermp: I don't think it makes sense to test these. The feature is generally tested already in the user agent spec IIRC.

build_tools/aws-sdk-code-generator/templates/spec/endpoint_provider_spec_class.mustache
mullermp: Yeah, this fixes an issue currently.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/endpoints_module.rb
```diff
@@ -42,10 +42,6 @@ def initialize(options)
 
         # @return [Array<EndpointParameter>]
         attr_reader :parameters
-
-        def has_endpoint_built_in?
```
alextwoods, line 46: Why is this being removed?

build_tools/aws-sdk-code-generator/templates/endpoints_plugin.mustache
```diff
@@ -44,11 +44,27 @@ module {{module_name}}
           context[:auth_scheme] =
             Aws::Endpoints.resolve_auth_scheme(context, endpoint)
 
-          @handler.call(context)
+          with_endpoint_metric(context.config) do
```
alextwoods: I find these a little clunky.  Two thoughts on making it a little easier - 

1. For many metric cases (like these), its scoped within a single request.  We can add these metrics to the context instead of to a thread local.  That way, they do not need to a call a block and pop the metric off.   This both lets do conditional metrics more easily and avoids having to wrap the handler call, which isn't really necessary for these.  (Note: I'm suggesting a second way to set metrics - scoped only to the request.  Not replacing the thread local, which I know we need for other cases).

2.   If we do not want to support multiple places/scopes metrics can be added, then we can have `with_metric` take an array of metrics, and allow that array to be empty.  That will let us easily support conditional metrics and multiple ones.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/endpoints_module.rb
mullermp: I have this in the account id change that was reviewed by us .. but I decided to carry it forward here - it's really just a syntax change. Is it incorrect? I think Endpoint is ALWAYS present in rules is it not? In the switch case below, it is switching on built in, and on SDK::Endpoint, we do the same check that was previously in the mustache template - check regional endpoint and either nil or endpoint.

build_tools/aws-sdk-code-generator/templates/endpoints_plugin.mustache
mullermp: Yeah, it's pretty clunky especially as these add. I think we want to keep a single mechanism (thread local) but probably this can just take an array and we build the array and block call with it.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/endpoints_module.rb
alextwoods: I don't think it MUST be present in the rules, but the code we have for constructing the SDK::Endpoint parameter is already conditional, so I agree this is fine, I just wanted to check.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/user_agent.rb
```diff
@@ -48,15 +48,13 @@ def self.feature(_feature, &block)
         block.call
       end
 
-      def self.metric(metric, &block)
+      def self.metric(*metrics, &block)
         Thread.current[:aws_sdk_core_user_agent_metric] ||= []
-        Thread.current[:aws_sdk_core_user_agent_metric] << METRICS[metric]
+        metrics = metrics.map { |metric| METRICS[metric] }
```
alextwoods: I think we may want to do a `compact` here, this lets us avoid creating an array and passing it and we can do something like:
```ruby
Aws::Plugins::UserAgent.metric((METRIC_1 if condition1), (METRIC_2 if condition2)) do

end
```

alextwoods (approved):

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/endpoints_module.rb
mullermp: Even if Endpoint isn't guaranteed to exist, I think it's still functionally equivalent, since the switch case looks for Endpoint anyway.

PR #3100, opened 2024-09-09
Add support for Http client cert and key to support mTLS

alextwoods:

### DIFF
--- a/gems/aws-sdk-core/lib/seahorse/client/net_http/connection_pool.rb
+++ b/gems/aws-sdk-core/lib/seahorse/client/net_http/connection_pool.rb
@@ -34,7 +34,9 @@ module Seahorse
           ssl_ca_bundle: nil,
           ssl_ca_directory: nil,
           ssl_ca_store: nil,
-          ssl_timeout: nil
+          ssl_timeout: nil,
+          http_client_cert: nil,
+          http_client_key: nil
         }
 
         # @api private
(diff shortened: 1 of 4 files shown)


gems/aws-sdk-core/lib/seahorse/client/net_http/connection_pool.rb
```diff
@@ -34,7 +34,9 @@ class ConnectionPool
           ssl_ca_bundle: nil,
           ssl_ca_directory: nil,
           ssl_ca_store: nil,
-          ssl_timeout: nil
+          ssl_timeout: nil,
+          http_client_cert: nil,
+          http_client_key: nil
```
jterapin: Is it necessary to have `client` included in the option name? Could we just do `http_cert` / `http_key`? 

I also feel like the `OPTIONS` are in alphabetical order so could re-order

jterapin (approved):
Might as well update [`net_http_spec.rb`](https://github.com/aws/aws-sdk-ruby/blob/c11769cb38fc091807c95b0872f099e67ef23193/gems/aws-sdk-core/spec/seahorse/client/plugins/net_http_spec.rb) with the new config options.

gems/aws-sdk-core/lib/seahorse/client/net_http/connection_pool.rb
alextwoods: yeah we could remove the `client` - I was debating it - I was struggling to name these.

The net http options don't have "client" in their names.  And these are really tls options and not http.  Like the net http option for ca_store/ca_path which we prefix with ssl (eg ssl_ca_store).    

After thinking a bit more, I've updated to `ssl_cert` and `ssl_key` which matches our other parameter naming.

mullermp (approved):

PR #3107, opened 2024-09-16
Remove CRT SigV4a

mullermp:
Removes CRT as a requirement for Sigv4a and with endpoint auth.

### DIFF
--- a/build_tools/services.rb
+++ b/build_tools/services.rb
@@ -10,10 +10,10 @@ module BuildTools
     MANIFEST_PATH = File.expand_path('../../services.json', __FILE__)
 
     # Minimum `aws-sdk-core` version for new gem builds
-    MINIMUM_CORE_VERSION = "3.206.0"
+    MINIMUM_CORE_VERSION = "3.205.0"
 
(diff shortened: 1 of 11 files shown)


build_tools/services.rb
```diff
@@ -10,10 +10,10 @@ class ServiceEnumerator
     MANIFEST_PATH = File.expand_path('../../services.json', __FILE__)
 
     # Minimum `aws-sdk-core` version for new gem builds
-    MINIMUM_CORE_VERSION = "3.205.0"
+    MINIMUM_CORE_VERSION = "3.206.0"
```
alextwoods: I don't think we need to bump min core for this? If you rely on an old version of core, everything will continue to work as is.

gems/aws-sigv4/lib/aws-sigv4/signer.rb
```diff
@@ -801,131 +778,7 @@ def presigned_url_expiration(options, expiration, datetime)
         end
       end
 
-      ### CRT Code
```
alextwoods, line 804: nice removal.

alextwoods (approved):

PR #3116, opened 2024-09-25
Add all core plugins to autoloads

alextwoods:


gems/aws-sdk-core/lib/aws-sdk-core/plugins.rb
```diff
@@ -4,8 +4,35 @@ module Aws
   # setup autoloading for Plugins
   # Most plugins are required explicitly from service clients
```
mullermp, line 5: Maybe update this comment seeing as we should always be doing this?

gems/aws-sdk-core/lib/aws-sdk-core/assume_role_credentials.rb
```diff
@@ -66,7 +66,7 @@ def refresh
       creds = c.credentials
       account_id =
         begin
-          ARNParser.parse(c.assumed_role_user.arn).account_id
+          ARNParser.parse(c.assumed_role_user&.arn).account_id
```
mullermp: Don't you need this on assume role web identity credentials? It would be nice if you can make a private method for parsing both of those things to make the code nicer.

gems/aws-sdk-core/lib/aws-sdk-core.rb
```diff
@@ -75,6 +75,8 @@ module Aws
   # dynamic endpoints
   autoload :Endpoints, 'aws-sdk-core/endpoints'
 
+  autoload :Sigv4, 'aws-sigv4'
```
mullermp: Why is this not in the sigv4 gem? Or why even autoload this at all?

mullermp (commented):
Can we make the minor buildtools changes I commented on earlier?

gems/aws-sdk-core/lib/aws-sdk-core/assume_role_credentials.rb
alextwoods: yeah, good catch. Cleaned up both.

gems/aws-sdk-core/lib/aws-sdk-core.rb
alextwoods: We were previously requiring the aws-sigv4 gem from core, so this keeps that functionality.

mullermp (approved):

PR #3122, opened 2024-10-07
Reduce memory usage from aws-partitions

alextwoods:
The endpoints (partitions.json) loaded by aws-partitions account for a large amount of memory required when creating/using service clients and post Endpoints2.0, are required only for legacy features.

**Benchmarking**
code being benchmarked:
```ruby
require 'aws-sdk-lambda'
client = Aws::Lambda::Client.new
client.list_functions; nil
```

* Before: memory retained: 8010 kb (69488 objects)
* After: memory retained: 3732 kb (30878 objects)

A decrease of more than 50%.

This PR does the following to reduce memory usage:
* Adds a new partitions metadata module that loads only the partition metadata.
* Change the aws_partitions Endpoints matcher to use the new, metadata only methods.
* Make `client.config.endpoint` set using an after_initailize and using the configured endpoint resolver

Additional endpoints changes:
Added a `create` class method to EndpointParameters which takes values from config.  Operation endpoint parameter classes now are only generated when there is operation specific values and use the `create` method to handle setting values from config.  Example generated code for S3:

```ruby
# EndpointParameters#create
    def self.create(config, options={})
      new({
        region: config.region,
        use_fips: config.use_fips_endpoint,
        endpoint: config.regional_endpoint ? nil : config.endpoint.to_s,
        force_path_style: config.force_path_style,
        use_global_endpoint: config.s3_us_east_1_regional_endpoint == 'legacy',
        disable_multi_region_access_points: config.s3_disable_multiregion_access_points,
        use_arn_region: config.s3_use_arn_region,
        disable_s3_express_session_auth: config.disable_s3_express_session_auth,
      }.merge(options))
    end

# Endpoint module
    class ListObjectsV2
      def self.build(context)
        Aws::S3::EndpointParameters.create(
          context.config,
          bucket: context.params[:bucket],
          use_dual_stack: context[:use_dualstack_endpoint],
          accelerate: context[:use_accelerate_endpoint],
          prefix: context.params[:prefix],
        )
      end
    end

```

**Alternatives**
Instead of using `define_singleton_method` to override the `endpoint` method on the config struct, we could instead:
1. resolve with the new endpoints resolver from config (fallback to legacy if we can’t infer parameters).  This is fairly complex code and requires us to make some assumptions about how parameters are bound (but reasonable within all existing AWS services).  See [demo implementation](https://github.com/aws/aws-sdk-ruby/pull/3122/commits/13f2c57f5a89290d534d7386ecfa5e7128a7a047) - this demo does NOT work for S3, but works for most other services.
2. (Slight breaking change) Use a simple templated/pattern resolver instead of the legacy resolver.  Requires writing the basic template code, but is relatively simple.  It will produce "incorrect" answers in some cases, but these are NOT used when making requests (but may be used by customers).
3. (Breaking change) Set endpoint to nil.  Customers that were relying on `client.config.endpoint` may be broken. 


gems/aws-sdk-core/lib/aws-sdk-core/plugins/regional_endpoint.rb
```diff
@@ -150,7 +150,18 @@ def resolve_endpoint(cfg)
           # that a custom endpoint has NOT been configured by the user
           cfg.override_config(:regional_endpoint, true)
 
-          resolve_legacy_endpoint(cfg)
+          # preserve legacy (pre EP2) client.config.endpoint still resolves a value
+          # when needed.
+          struct = cfg.instance_variable_get(:@struct)
```
mullermp: I don't understand this - and it's very implementation specific. When would this thing not exist?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/regional_endpoint.rb
```diff
@@ -150,7 +150,18 @@ def resolve_endpoint(cfg)
...
+          if struct
+            # need a separate proc to get namespace/private access to resolve_legacy_endpoint
```
mullermp: Should the method just be public to the plugin then?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/regional_endpoint.rb
```diff
@@ -150,7 +150,18 @@ def resolve_endpoint(cfg)
           # that a custom endpoint has NOT been configured by the user
           cfg.override_config(:regional_endpoint, true)
```
mullermp, line 187: What about using cfg.override_config for endpoint to be the value of resolve legacy endpoint?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/regional_endpoint.rb
```diff
@@ -213,7 +224,7 @@ def resolve_legacy_endpoint(cfg)
             sts_regional = cfg.sts_regional_endpoints
           end
 
-          Aws::Partitions::EndpointProvider.resolve(
+          endpoint = Aws::Partitions::EndpointProvider.resolve(
```
mullermp: I thought the seahorse plugin will make this a URI already?

gems/aws-sdk-core/lib/seahorse/client/plugins/endpoint.rb
```diff
@@ -42,7 +40,11 @@ def after_initialize(client)
         class Handler < Client::Handler
 
           def call(context)
-            context.http_request.endpoint = URI.parse(context.config.endpoint.to_s)
+            if !context.config.respond_to?(:regional_endpoint) || # generic
```
mullermp: When would regional endpoint not exist? I think this plugin should continue to blindly set endpoint and logic like this belongs somewhere else.. it should not know or care about endpoints 2.0 logic.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/regional_endpoint.rb
alextwoods: Was being defensive here.  If I add an accessor for struct to ConfigResolver, then I can skip checking probably

gems/aws-sdk-core/lib/aws-sdk-core/plugins/regional_endpoint.rb
alextwoods: what do you mean by public to the plugin?  I can add a public accessor for struct to ConfigResolver

gems/aws-sdk-core/lib/aws-sdk-core/plugins/regional_endpoint.rb
alextwoods: That won't work - override_config sets a value at the time its called.  Calling resolve_legacy_endpoint loads the entire endpoints.json which I am trying to avoid.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/regional_endpoint.rb
alextwoods: Yeah, the seahorse plugin did make it a URI - but because we don't want to call the endpoint method unless its required, I moved this here (see the changes in the searhorse endpoints plugin).

gems/aws-sdk-core/lib/seahorse/client/plugins/endpoint.rb
alextwoods: regional endpoint doesn't exist in old services.  I think this logic must exist here to fully handle old service cases. 

I guess one option is to actually remove the Searhose Endpoints plugin when endpoints2.0 is used?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/regional_endpoint.rb
alextwoods: Check the most recent commit - I've added an accessor for struct and cleaned this code up.  Let me know if thats what you were thinking

gems/aws-sdk-core/lib/seahorse/client/plugins/endpoint.rb
alextwoods: Check the most recent commit - I've updated our codegen to remove the seahorse endpoints plugin when ep2 is enabled for a service (slightly tricky since its a searhose base default plugin). 

Let me know which of these approaches you prefer.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/regional_endpoint.rb
```diff
@@ -93,6 +96,66 @@ def after_initialize(client)
...
+
+      def build_parameters(params, config)
+        # generic parameters set for most (but not all) services
+        params.region = config.region if params.respond_to?(:region=)
+        params.use_dual_stack = config.use_dualstack_endpoint if params.respond_to?(:use_dual_stack=)
+        params.use_fips = config.use_fips_endpoint if params.respond_to?(:use_fips=)
+
+        # service specific
+        if config.respond_to?(:sts_regional_endpoints) && params.respond_to?(:use_global_endpoint=)
+          params.use_global_endpoint = (config.sts_regional_endpoints == 'legacy')
```
mullermp: Do these not have defaults in the params class? Maybe the params class can take config to construct defaults.

alextwoods: These do have defaults. But this is required to get the "correct" values based on customers config.

alextwoods: I'll look into taking config when constructing params. That would be a codegen change that would regenerate every service and I'm not sure want to couple params to config since we have a separate way to construct it (per operation)

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/endpoint_parameter.rb
```diff
@@ -0,0 +1,200 @@
...
+    end
+
+    def default?
+      !@default.nil?
+    end
+
+    def validate_required?
+      required && !default?
+    end
+    def key
```
mullermp, line 63: Newline

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/endpoint_parameter.rb
```diff
@@ -0,0 +1,200 @@
...
+              'config.use_accelerate_endpoint'
+            end
+          when 'AWS::S3::ForcePathStyle'
+            'config.force_path_style'
+          when 'AWS::S3::UseArnRegion', 'AWS::S3Control::UseArnRegion'
+            'config.s3_use_arn_region'
+          when 'AWS::S3::DisableMultiRegionAccessPoints'
+            'config.s3_disable_multiregion_access_points'
+          when 'SDK::Endpoint'
+            'config.regional_endpoint ? nil : config.endpoint.to_s'
```
mullermp: This can be simplified to "(config.endpoint.to_s if config.regional_endpoint)"

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/endpoint_parameters_class.rb
```diff
@@ -9,14 +9,20 @@ def initialize(options)
         @service = options.fetch(:service)
         if (parameters = @service.endpoint_rules&.fetch('parameters'))
           @parameters = parameters.map do |k,p|
-            EndpointParameter.new(k, p)
+            value, source = EndpointParameter.endpoint_parameter_value(@service,  k, p)
+            EndpointParameter.new(k, p, value, source)
```
mullermp: This is weird, using a class method to construct an instance of that class. Can new() just take the service, k, and p?

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/endpoints_module.rb
```diff
@@ -34,157 +34,35 @@ def module_name
...
-
-        return nil unless binding
-
-        "JMESPath.search(\"#{Underscore.underscore_jmespath(binding['path'])}\", context.params)"
-      end
-
-      def static_context_param(operation, param_name)
-        operation.fetch('staticContextParams', {})
-                 .fetch(param_name, {}).fetch('value', nil)
+        end.select { |p| p.source == 'operation' }
```
mullermp: This could probably be moved to the loop for readability and efficiency

build_tools/aws-sdk-code-generator/spec/interfaces/plugins/endpoints_spec.rb
```diff
@@ -124,6 +125,10 @@
         EndpointsPrecedence::Client.new(stub_responses: true, region: 'config')
       end
 
+      before(:each) do
```
mullermp, line 128: Is this still necessary? If so how can we avoid doing this?

build_tools/aws-sdk-code-generator/templates/endpoints_module.mustache
```diff
@@ -11,19 +11,21 @@ module {{module_name}}
...
           {{key}}: {{{value}}},
-          {{/static_string?}}
           {{/parameters}}
         )
       end
     end
 
     {{/endpoint_classes}}
+
+    OPERATION_PARAM_MAP = {
```
mullermp: Why this over using the method with a case statement?

build_tools/aws-sdk-code-generator/templates/endpoints_plugin.mustache
```diff
@@ -86,12 +86,8 @@ The endpoint provider used to resolve endpoints. Any object that responds to
...
 
         def parameters_for_operation(context)
-          case context.operation_name
-          {{#endpoint_classes}}
-          when :{{operation_name}}
-            {{module_name}}::Endpoints::{{class_name}}.build(context)
-          {{/endpoint_classes}}
-          end
+          operation_class = {{module_name}}::Endpoints::OPERATION_PARAM_MAP[context.operation_name]
+          operation_class ? operation_class.build(context) : {{module_name}}::EndpointParameters.create(context.config)
```
mullermp: Parameters for operation should probably fail if the operation doesn't exist, maybe an argument or runtime error is appropriate. If the default was somehow used, that could be difficult to track down.

Regardless, if we must keep this, please don't make it a ternary.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/regional_endpoint.rb
```diff
@@ -93,6 +96,49 @@ def after_initialize(client)
         client.config.sigv4a_signing_region_set = region_set
       end
 
+      private
+
+      def initialize_default_endpoint(client)
+        client_module = Object.const_get(client.class.name.rpartition('::').first)
+        if client.config.respond_to?(:endpoint_provider) &&
```
mullermp: There's a lot of nesting here. Can we break this up a bit with either methods and/or guard clauses?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/regional_endpoint.rb
```diff
@@ -93,6 +96,49 @@ def after_initialize(client)
...
+          client_module.const_defined?(:EndpointParameters) &&
+          (param_class = client_module.const_get(:EndpointParameters)) &&
+          param_class.respond_to?(:create) &&
```
mullermp: We could potentially just call create and add NoMethodError in our rescue. We could do the same with fetching endpoint_provider off config (would or would not exist) and potentially also with const_get (does this error or return nil?)

gems/aws-sdk-core/lib/aws-sdk-core/plugins/regional_endpoint.rb
```diff
@@ -93,6 +96,49 @@ def after_initialize(client)
...
+          begin
+            endpoint = endpoint_provider.resolve_endpoint(params)
+            client.config.endpoint = endpoint.url
+          rescue ArgumentError
+            # fallback to legacy
+            client.config.endpoint = resolve_legacy_endpoint(client.config)
+          end
+        else
+          # fallback to legacy
+          client.config.endpoint = resolve_legacy_endpoint(client.config)
```
mullermp: Ditto for guard clauses and simplify

mullermp (commented):
Nice. I like this approach better. Just needs some clean up I think.

build_tools/aws-sdk-code-generator/templates/endpoints_module.mustache
alextwoods: We know have classes only for certain operations (ones which need operation specific values).  I wanted to keep that knowledge isolated here rather than duplicating it in the endpoints module.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/endpoint_parameter.rb
alextwoods: not sure its really a simplification, but changed.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/endpoint_parameters_class.rb
alextwoods: Yeah, that makes sense.  Orig was trying to match old usage from both classes that used different endpointParameter classes. 

Updated.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/endpoints_module.rb
alextwoods: how? I guess I could return a nil and then compact instead of select? But that doesn't seem any better.

mullermp: It would have to go back to each with object instead of map, and then next item if source is not operation. If it's operation then add the parameter.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/endpoints_module.rb
```diff
@@ -9,13 +9,13 @@ def initialize(options)
...
-          |(name, op), array|
-          array << EndpointClass.new(
+        @endpoint_classes = @service.api['operations'].map do
+          |name, op|
+          EndpointClass.new(
             name: name,
             parameters: endpoint_parameters_for_operation(op)
           )
-        end
+        end.reject { |ec| ec.parameters.empty? }
```
mullermp: Ditto on my other comment. Could this just stay each with object and when you find empty params, skip item

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/endpoints_module.rb
alextwoods: Hmm, that seems a lot more complicated than using a simple map/select, but happy to change it back if you have strong feelings.

build_tools/aws-sdk-code-generator/templates/endpoints_module.mustache
mullermp: That's fine

build_tools/aws-sdk-code-generator/spec/interfaces/plugins/endpoints_spec.rb
alextwoods: Added a comment.  This is neccessary because config after_initialize will call the endpoint resolver once.  This lets us achieve that before the actual tests with their assertions about how the resolver/parameters are called.

build_tools/aws-sdk-code-generator/templates/endpoints_plugin.mustache
alextwoods: Parameter classes now only exist when the operation needs specific configuration (which reduces size a ton since most operations dont) so I think we still need this.

Removed the ternary.

mullermp: Oh I see. So this isn't every operation, just ones that don't have config based ones?

mullermp: Maybe the map name could be different to indicate that it's not all the operations.

alextwoods: yeah exactly. Changed to OPERATION_CUSTOM_PARAMS_MAP

gems/aws-sdk-core/lib/aws-sdk-core/plugins/regional_endpoint.rb
alextwoods: yeah, just rescuing exceptions cleans this up a ton.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/endpoints_module.rb
alextwoods: Changed to each.with_object.

build_tools/aws-sdk-code-generator/templates/endpoints_plugin.mustache
alextwoods: Moved method to endpoints module and use case statement.

mullermp (approved):

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/endpoint_parameter.rb
```diff
@@ -0,0 +1,190 @@
...
+    end
+
+    def operation_context_param_value(operation, param_name)
+      return nil unless operation['input']
+
+      binding = operation.fetch('operationContextParams', {})[param_name]
+
+      return nil unless binding
+
+      "JMESPath.search(\"#{Underscore.underscore_jmespath(binding['path'])}\", context.params)"
```
HeyNonsterTemp, line 177: We're wondering if `context.config.simple_json` needs to be considered here to conditionally generate `camel_case` or `snake_case`. cc: @bquorning

HeyNonster, line 177: We're wondering if `context.config.simple_json` needs to be considered here to conditionally generate `camel_case` or `snake_case`. cc: @bquorning

bquorning: I tried adding this patch:

```diff
diff --git i/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/endpoint_parameter.rb w/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/endpoint_parameter.rb
index 36d22c08d..aadeca202 100644
--- i/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/endpoint_parameter.rb
+++ w/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/endpoint_parameter.rb
@@ -174,7 +174,9 @@ module AwsSdkCodeGenerator
 
       return nil unless binding
 
-      "JMESPath.search(\"#{Underscore.underscore_jmespath(binding['path'])}\", context.params)"
+      camel_cased = binding['path']
+      snake_cased = Underscore.underscore_jmespath(binding['path'])
+      "JMESPath.search(context.config.simple_json ? \"#{camel_cased}\" : \"#{snake_cased}\", context.params)"
     end
 
     def static_context_param(operation, param_name)
```

Which produces the following diff (from c29196866947b3fe2fb3a049a1d966c9884953a1) for the generated code:

```diff
diff --git i/gems/aws-sdk-dynamodb/lib/aws-sdk-dynamodb/endpoints.rb w/gems/aws-sdk-dynamodb/lib/aws-sdk-dynamodb/endpoints.rb
index 0f9a6b7bb..41359a745 100644
--- i/gems/aws-sdk-dynamodb/lib/aws-sdk-dynamodb/endpoints.rb
+++ w/gems/aws-sdk-dynamodb/lib/aws-sdk-dynamodb/endpoints.rb
@@ -16,7 +16,7 @@ module Aws::DynamoDB
       def self.build(context)
         Aws::DynamoDB::EndpointParameters.create(
           context.config,
-          resource_arn_list: JMESPath.search("keys(request_items)", context.params),
+          resource_arn_list: JMESPath.search(context.config.simple_json ? "keys(RequestItems)" : "keys(request_items)", context.params),
         )
       end
     end
@@ -25,7 +25,7 @@ module Aws::DynamoDB
       def self.build(context)
         Aws::DynamoDB::EndpointParameters.create(
           context.config,
-          resource_arn_list: JMESPath.search("keys(request_items)", context.params),
+          resource_arn_list: JMESPath.search(context.config.simple_json ? "keys(RequestItems)" : "keys(request_items)", context.params),
         )
       end
     end
@@ -241,7 +241,7 @@ module Aws::DynamoDB
       def self.build(context)
         Aws::DynamoDB::EndpointParameters.create(
           context.config,
-          resource_arn: JMESPath.search("table_creation_parameters.table_name", context.params),
+          resource_arn: JMESPath.search(context.config.simple_json ? "TableCreationParameters.TableName" : "table_creation_parameters.table_name", context.params),
         )
       end
     end
@@ -358,7 +358,7 @@ module Aws::DynamoDB
       def self.build(context)
         Aws::DynamoDB::EndpointParameters.create(
           context.config,
-          resource_arn_list: JMESPath.search("transact_items[*].get.table_name", context.params),
+          resource_arn_list: JMESPath.search(context.config.simple_json ? "TransactItems[*].Get.TableName" : "transact_items[*].get.table_name", context.params),
         )
       end
     end
@@ -367,7 +367,7 @@ module Aws::DynamoDB
       def self.build(context)
         Aws::DynamoDB::EndpointParameters.create(
           context.config,
-          resource_arn_list: JMESPath.search("transact_items[*].[condition_check.table_name, put.table_name, delete.table_name, update.table_name][]", context.params),
+          resource_arn_list: JMESPath.search(context.config.simple_json ? "TransactItems[*].[ConditionCheck.TableName, Put.TableName, Delete.TableName, Update.TableName][]" : "transact_items[*].[condition_check.table_name, put.table_name, delete.table_name, update.table_name][]", context.params),
         )
       end
     end
```

This fixes a problem we’re seeing in an internal library (which uses `simple_json: true` and fails on the `keys(request_items)` expressions), but I am not sure if the other changes might break things for others.

bquorning: I should add that our problems started on v1.138.0 of the gem, after commit ce5d80077de14fa4c9f84b022a726d25f8eacc2b.

mullermp: I'm just wondering what your use case is for simple json. I was not planning on supporting this for a new major version. If possible, I would move away from it. Can you also be specific about what bug you encountered and can you make a new issue?

PR #3128, opened 2024-10-14
Improve awsQueryCompatible support for JSON and CBOR

mullermp:
Also makes some misc improvements to JSON around query compatible, adds stubbing and error handling tests to rpc v2, and improves general stubbing tests.

Fixes to existing cbor/rpcv2 tests after merging


gems/aws-sdk-core/spec/api_helper.rb
```diff
@@ -45,6 +45,14 @@ def sample_ec2 # ec2 has its own protocol
       end
     end
 
+    def sample_rpcv2_cbor # cloudwatch logs changed to cbor
```
alextwoods, line 48: this comment is a little confusing.  is the fixture itself already set to use cbor?  If so, maybe rename to logs_cbor or something?

gems/aws-sdk-core/spec/aws/stubbing/protocols/rpc_v2_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+                  inherited_properties: [ 'ACCOUNT_DATA_PROTECTION' ],
+                  log_group_class: 'STANDARD',
+                  log_group_arn: 'arn:aws:logs:us-west-2:012345678910:log-group:my-log-group:*'
+                }
+              ]
+            }
+
+            resp = RpcV2.new.stub_data(api, operation, data)
+            actual = normalize(Cbor.decode(resp.body.string))
+            expect(actual).to eq(normalize(JSON.parse(<<-JSON)))
```
alextwoods, line 57: It seems a little weird here to use json to compare the body (I know we're not actually comparing/expecting json).... but I guess, why not compare the returned response to the expected stub values?

alextwoods (approved):
Looks good, just minor comments.

gems/aws-sdk-core/spec/api_helper.rb
mullermp: It's a sample rpcv2 cbor service. The comment is just indicating that logs isn't actually cbor (I did a manual change since no service exists yet)

gems/aws-sdk-core/spec/aws/stubbing/protocols/rpc_v2_spec.rb
mullermp: Existing tests did this - I think it makes sense to keep. We want to check that we are generating a response that can be parsed (with a string body)

alextwoods (approved):

PR #3132, opened 2024-10-29
Docs: do not escape curly braces in example

hannahramadan:
In the example regex pattern, the curly braces are escaped, but stream patterns don't appear to typically expect `{` or `}`. 

The Java [example](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-kinesis/src/main/java/com/amazonaws/services/kinesis/AmazonKinesisClient.java#L625C1-L626C1) seems to be what is intended. This PR makes that update.


mullermp:
This is making a change to a code generated file. The upstream docs would need to be fixed or translation logic in our build_tools folder.

PR #3139, opened 2024-11-02
Remove extra backslashes from code snippets

mullermp:
Closes #3132 

We were addressing Yard warnings by escaping backslashes, but this also applied to code snippets. I tried a solution where we only escaped when not in a code section but this approach failed. This is because we have a multi-tier doc system with html -> markdown (rdiscount) -> yard, each handling escaping differently. This approach fixes the docs visibly but we will have warning outputs when curly braces are used outside of code blocks. This doesn't happen often, a majority of docs use code blocks.


mullermp:
See generated changes

gems/aws-sigv4/lib/aws-sigv4/asymmetric_credentials.rb
```diff
@@ -11,8 +11,6 @@ module AsymmetricCredentials
 
       N_MINUS_2 = 0xFFFFFFFF00000000FFFFFFFFFFFFFFFFBCE6FAADA7179E84F3B9CAC2FC632551 - 2
 
-      # @param [String] :access_key_id
```
alextwoods, line 14: Why remove these?  They are still the correct paramters here?

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/client_operation_documentation.rb
```diff
@@ -203,15 +202,11 @@ def shared_examples(examples, operation, api)
         end
         example_block.join("\n")
       rescue
-        puts "Invalid example for operation: #{@name}"
+        puts "Invalid example for operation: #{@module_name} - #{@name}"
         nil
       end
     end
 
-    def generated_examples(operation, api)
```
alextwoods, line 211: Was this just not used?

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/docstring.rb
```diff
@@ -118,10 +118,6 @@ def html_to_markdown(html, options = {})
         if html
           html = "<p>#{html}</p>" unless html.match(/<\w+>/)
 
-          # unescaped curly braces cause YARD errors, they are interpreted
-          # as code links.
-          html = html.gsub('{', "\\{").gsub('}', "\\}")
```
alextwoods, line 123: could we take an `escape_curly_braces` option and pass it only when this is called for code docs?

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/client_operation_documentation.rb
mullermp: I don't think so...

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/docstring.rb
mullermp: That doesn't fix inline code snippets in the documentation. This applies on the whole block of documentation text.

gems/aws-sigv4/lib/aws-sigv4/asymmetric_credentials.rb
mullermp: The module is API private and these were wrong. These are positional parameters but they were symbolized name, so it produced a warning. I just deleted them because it's clear what they are (and API private).

alextwoods (approved):

PR #3140, opened 2024-11-06
Declare dependency on json 2.4.0+

byroot:
If the gems use `JSON.load_file` and are compatible with Ruby >= 2.5 they must declare that they require `json 2.4.0` otherwise they may get a much older version.

cc @jterapin, I don't know this project enough to know if `aws-sdk-core` is the right gemspec to add the dependency to, but that should solve your issues.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,8 @@
 
 source 'https://rubygems.org'
 
+gemspec path: "gems/aws-sdk-core/"
+
 gem 'rake', require: false
 # SDK feature dependencies
 gem 'aws-crt' if ENV['CRT']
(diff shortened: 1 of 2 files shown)


byroot:
Ah, I see. The gemspec isn't loaded by the `Gemfile`, so that dependency doesn't surface, and need to be repeated there too.

Gemfile
```diff
@@ -2,6 +2,8 @@
 
 source 'https://rubygems.org'
 
+gemspec path: "gems/aws-sdk-core/"
```
byroot, line 5: This should hopefully do the trick.

byroot:
🍏

jterapin:
Hi! Due to the structure of how we run specs that involves `load_file` and the Ruby SDK team discussion, we decided to pin the dependency. Thank you for all your help! ✨

PR #3150, opened 2024-11-27
Add an RBS signature for `Aws::Query::Param`

k-in0ue:
Here is my first PR to the aws-sdk-core repository.
I kindly ask for your understanding if there are any mistakes.

I have added an RBS type definition file, starting with just one file as a trial.

If this PR is successfully merged, I would like to continue contributing by adding more RBS type definitions in the future.

Thank you for taking the time to review this. I look forward to your feedback.

### DIFF
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Feature - Add RBS signature files to support static type checking
+
 3.214.0 (2024-11-25)
 ------------------
 
--- a/gems/aws-sdk-core/sig/aws-sdk-core/query/param.rbs
+++ b/gems/aws-sdk-core/sig/aws-sdk-core/query/param.rbs
@@ -1,0 +1,12 @@
+module Aws
+  module Query
+    class Param
+      def initialize: (String | Symbol, ?(String | nil)) -> void
+
+      attr_reader name: String
+      attr_reader value: String?
+
+      def to_s: -> String
+    end
+  end
+end


mullermp:
Thanks for opening a pull request. We are happy to take RBS contributions. I'm wondering, why did you choose `Aws::Query::Param` specifically? I believe this a private API not intended to be used by customers. I think there is more value to RBS for public classes. That said, I'm not against taking this, but I'd like to see more than just this class added before we merge and bump the SDK version.

k-in0ue:
Thank you for your feedback, and I apologize for my misunderstanding.

I initially chose Aws::Query::Param because I assumed that any class without `# @api private` was part of the public API. I now realize this was incorrect. Moving forward, I’ll focus on contributing RBS for public classes.

Are there any specific indicators within the project, such as directory structures or naming conventions, to distinguish public APIs from internal ones? Or is comparing with the official documentation the primary way to make this determination?

Thank you for your patience and guidance.

mullermp:
No worries :) You are correct that the API private documentation indicates that. Sometimes we've forgotten to add that comment, or it ends up not being necessary because it's [already omitted from docs](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/).

As a trial run, I think a good place to start might be to add RBS for the [sigv4 gem](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/Sigv4.html) since it is not too large and it's self contained. You could add RBS for the public errors, the request, the signature, and the signer and signer methods.

k-in0ue:
Thank you for the helpful clarification and guidance!

I’ll take your suggestion and work on adding RBS for the sigv4 gem. It seems like a great starting point, and I’m excited to dive in!

To keep things clean, I’ll close this PR and submit a new one focused on the sigv4 gem. I’ll do my best to contribute, even though I still have a lot to learn!

Thanks again for your support!

PR #3152, opened 2024-12-04
Add an RBS signature for sigv4

k-in0ue:
I received some advice suggesting, "Why not start with aws-sigv4?" in this PR: https://github.com/aws/aws-sdk-ruby/pull/3150.

Following that advice, I have added type definition files to `aws-sigv4`.


mullermp:
Thank you, we will review this. I think you will also need: a changelog entry, add sig/ to gemspec files, and ensure rbs test passes (see tasks/rbs.rake).

mullermp:
@ksss may also be interested in reviewing if you'd like.

gems/aws-sigv4/sig/error.rbs
```diff
@@ -0,0 +1,13 @@
+module Aws
+  module Sigv4
+    module Errors
+      class MissingCredentialsError < ArgumentError
+        def initialize: (?String? msg) -> void
```
alextwoods: I think this is technically correct (since you could explicitly pass nil here), but I think we may want to just type it as `?String`

gems/aws-sigv4/sig/error.rbs
```diff
@@ -0,0 +1,13 @@
...
+      end
+
+      class MissingRegionError < ArgumentError
+        def initialize: (*untyped args) -> void
```
alextwoods: The args are actually not used at all... we should potentially change the code to be just `def initialize(*)`

gems/aws-sigv4/sig/request.rbs
```diff
@@ -0,0 +1,24 @@
+module AWS
+  module Sigv4
+    class Request
+      def initialize: (
+          http_method: String,
+          endpoint: URI::HTTP | URI::HTTPS | String,
+          headers?: Hash[String, String],
```
alextwoods: Should be these `?headers` and `?body` instead since they're optional arguments?

gems/aws-sigv4/sig/signer.rbs
```diff
@@ -0,0 +1,70 @@
+module Aws
+  module Sigv4
+    class Signer
+      def initialize: (
+        service: String,
+        region: String,
+        access_key_id: String,
+        secret_access_key: String,
+        session_token?: String?,
```
alextwoods: Should these be `?session_token` for optional rather than nullable?

gems/aws-sigv4/sig/signer.rbs
```diff
@@ -0,0 +1,70 @@
...
+        unsigned_headers?: Array[String]?,
+        uri_escape_path?: bool,
+        apply_checksum_header?: bool,
+        signing_algorithm?: Symbol,
```
alextwoods: Since this is a limited set of valid symbols (sigv4, sigv4a, sigv4-s3express), we could limit that here.  Not sure if we should or not though. Thoughts?

gems/aws-sigv4/sig/signer.rbs
```diff
@@ -0,0 +1,70 @@
...
+        url: String,
+        headers: Hash[String, String],
+        body?: String,
+      ) -> Signature
+
+      def sign_event: (
+        prior_signature: String,
+        payload: String,
+        encoder: Object
+      ) -> Object
```
alextwoods: I think this actually returns a tuple of [Hash, String]

alextwoods (commented):
Thanks for the contribution, this is looking good!

gems/aws-sigv4/sig/credentials.rbs
```diff
@@ -0,0 +1,30 @@
+module Aws
+  module Sigv4
+    class Credentials
```
ksss: Looks like a private API.
https://github.com/aws/aws-sdk-ruby/blob/8c6f65feb0500775b50153e0c24a56381509c873/gems/aws-sigv4/lib/aws-sigv4/credentials.rb#L5-L8

gems/aws-sigv4/sig/credentials.rbs
```diff
@@ -0,0 +1,30 @@
...
+      ) -> void
+
+      attr_reader access_key_id: String
+      attr_reader secret_access_key: String
+      attr_reader session_token: String?
+
+      def set?: () -> bool
+    end
+
+    class StaticCredentialsProvider
```
ksss: ditto.
https://github.com/aws/aws-sdk-ruby/blob/8c6f65feb0500775b50153e0c24a56381509c873/gems/aws-sigv4/lib/aws-sigv4/credentials.rb#L43-L46

gems/aws-sigv4/sig/request.rbs
ksss: I agree with @alextwoods .

The syntax for optional keyword arguments is like this.

```suggestion
          ?headers: Hash[String, String],
```

https://github.com/aws/aws-sdk-ruby/blob/8c6f65feb0500775b50153e0c24a56381509c873/gems/aws-sigv4/lib/aws-sigv4/request.rb#L11

gems/aws-sigv4/sig/request.rbs
```diff
@@ -0,0 +1,24 @@
+module AWS
+  module Sigv4
+    class Request
+      def initialize: (
+          http_method: String,
```
ksss, line 5: This rbs syntax cannot support the following ruby syntaxes in steep.

```rb
# With Hash argument
hash = {http_method: 'get'}
Aws::Sigv4::Request.new(hash)

# With Hash directly
Aws::Sigv4::Request.new({http_method: 'get'})
```

I solve this problem this way.

```rb
def initialize: (?foo: String, ?bar: Integer, ?baz: bool) -> void # Support keyword arguments
              | (?Hash[Symbol, untyped]) -> void                  # Support others
```

This is true for almost all methods in aws-sdk.

gems/aws-sigv4/sig/request.rbs
```diff
@@ -0,0 +1,24 @@
+module AWS
```
ksss:
```suggestion
module Aws
```

gems/aws-sigv4/sig/signer.rbs
```diff
@@ -0,0 +1,70 @@
...
+        http_method: String,
+        url: String,
+        headers: Hash[String, String],
+        body?: String,
+      ) -> Signature
+
+      def sign_event: (
+        prior_signature: String,
+        payload: String,
+        encoder: Object
```
ksss: Use `untyped` when you do not know what the type is or when you want to put off a decision.

gems/aws-sigv4/sig/signer.rbs
```diff
@@ -0,0 +1,70 @@
...
+        service: String,
+        region: String,
+        credentials_provider: Object,
+        unsigned_headers?: Array[String]?,
+        uri_escape_path?: bool,
+        apply_checksum_header?: bool,
+        signing_algorithm?: Symbol,
+        omit_session_token?: bool,
+        normalize_path?: bool,
+      ) -> void
```
ksss, line 18: I think you are writing the overload to match the YARD notation, but there is a tradeoff with maintenance costs.
Either way, we can't make a strict type determination when dealing with hash arguments.

ksss (commented):
- Private API types should be removed.
- The intent and syntax are different.
- Hash arguments should be supported as well as keyword arguments.
- What about RBS testing?

gems/aws-sigv4/sig/error.rbs
k-in0ue: Fixed!
5c6600f4cf845af9eedd8ebee14c1c8175dcd544

gems/aws-sigv4/sig/error.rbs
k-in0ue: Fixed!
5c6600f4cf845af9eedd8ebee14c1c8175dcd544

gems/aws-sigv4/sig/request.rbs
k-in0ue: Fixed!
a522cc1b8a06496539730bbfedd6ab3110050792

gems/aws-sigv4/sig/request.rbs
k-in0ue: Fixed!
a522cc1b8a06496539730bbfedd6ab3110050792

gems/aws-sigv4/sig/request.rbs
k-in0ue: Fixed!
a522cc1b8a06496539730bbfedd6ab3110050792

gems/aws-sigv4/sig/credentials.rbs
k-in0ue: Sorry, I deleted it.
a72b3bcfb6b8ab60ae2fd5cb968b861c7f02fe84

gems/aws-sigv4/sig/credentials.rbs
k-in0ue: Sorry, I deleted it.
a72b3bcfb6b8ab60ae2fd5cb968b861c7f02fe84

gems/aws-sigv4/sig/signer.rbs
k-in0ue: Thank you, I fixed it.
2a780e6cc1c843a179acdebeb5382297ca390aba

gems/aws-sigv4/sig/signer.rbs
k-in0ue: I've tried combining all the overloads of the initialize method into a single definition. Would you mind sharing your thoughts on this approach?
2a780e6cc1c843a179acdebeb5382297ca390aba

gems/aws-sigv4/sig/signer.rbs
k-in0ue: Thank you, I fixed it.
2a780e6cc1c843a179acdebeb5382297ca390aba

gems/aws-sigv4/sig/signer.rbs
k-in0ue: I thought it would be better to restrict it to the valid symbols only, so I’ve updated the code accordingly.
2a780e6cc1c843a179acdebeb5382297ca390aba

k-in0ue:
I replaced `URI::HTTP | URI::HTTPS` with `untyped` because using `URI::HTTP | URI::HTTPS` caused an RBS::NoTypeFoundError.

mullermp:
Let me know when this is ready for re-review and I can allocate some time.

gems/aws-sigv4/sig/signer.rbs
```diff
@@ -0,0 +1,50 @@
+module Aws
+  module Sigv4
+    class Signer
+      def initialize: (
+        service: String,
+        region: String,
+        access_key_id?: String,
```
ksss: Same as https://github.com/aws/aws-sdk-ruby/pull/3152#discussion_r1870483168

`foo?: String` is not optional keyword syntax.
It support to call method like `m(foo?: 'bar')`. This is a very rare case.

k-in0ue: Sorry about that, I've updated the code.

k-in0ue:
I've made changes based on your feedback. Please review the updated code at your convenience.

gems/aws-sigv4/sig/signer.rbs
ksss: Hash arguments should be supported.

ksss (commented):
We should modify tasks/rbs.rake to run the RBS test.

mullermp:
Is this ready for re-review from our side?

gems/aws-sigv4/sig/signer.rbs
```diff
@@ -0,0 +1,50 @@
...
+  module Sigv4
+    class Signer
+      def initialize: (
+        service: String,
+        region: String,
+        ?access_key_id: String,
+        ?secret_access_key: String,
+        ?session_token: String,
+        ?credentials: untyped,
+        ?credentials_provider: untyped,
```
mullermp: I think there are classes in this gem you could use for these, or possibly use some interface for "responds to"

mullermp (approved):
Looks good and definitely an improvement!

gems/aws-sigv4/sig/signer.rbs
k-in0ue: Let me confirm: Do you mean I should revise it as follows?

```ruby
def initialize: (
  options: {
    service: String,
    region: String,
    access_key_id?: String,
    secret_access_key?: String,
    session_token?: String,
    credentials?: untyped,
    credentials_provider?: untyped,
    unsigned_headers?: Array[String],
    uri_escape_path?: bool,
    apply_checksum_header?: bool,
    signing_algorithm?: :sigv4 | :sigv4a | :'sigv4-s3express',
    omit_session_token?: bool,
    normalize_path?: bool,
  }
) -> void
```

ksss: No. "Hash arguments" is same as https://github.com/aws/aws-sdk-ruby/pull/3152#discussion_r1870488007

How are you testing the behavior?
I don't think it will pass in steep with the proposed signature.

gems/aws-sigv4/sig/signer.rbs
k-in0ue: I noticed that `aws-sigv4/credentials.rb` is marked `@api private`, so I removed its RBS definitions. Do you think it’s better to restore them or just define an interface (requiring `#credentials` returning access keys, etc.) directly in `signer.rbs`? I’m unsure which approach is more suitable, given the private status. Any thoughts?

gems/aws-sigv4/sig/signer.rbs
k-in0ue: Sorry about that, I've updated the code.
3c34687562bbc42d3dbbf4372b96410fd5d32a83

gems/aws-sigv4/sig/signer.rbs
mullermp: Yes, this is true. I think `sigv4/credentials.rb` serves as an intermediary. Technically these would be `Aws::Credentials` from aws-sdk-core. We can leave these untyped for now, and if you're interested in adding rbs for credentials in aws-sdk-core, we can fill these in. I'm not sure how rbs works with across packages, since aws-sigv4 does not depend on aws-sdk-core.

k-in0ue: So you’re suggesting that if we create RBS files corresponding to `gems/aws-sdk-core/lib/aws-sdk-core/credentials.rb` and `gems/aws-sdk-core/lib/aws-sdk-core/credential_provider.rb`, we could then make use of those type definitions, right? For now, I'll keep it `untyped`, but afterward I'm thinking of adding RBS definitions for `aws-sdk-core` credentials. How does that sound?

alextwoods: While in most use cases, `Aws::Credentials` from `aws-sdk-core` would be used, this library is intended as a stand alone and doesn't depend on aws-sdk-core - I think from a typing perspective maybe we should define a credentials interface (ie: any object that responds to access_key_id, secret_access_key and session_token) since thats actually what is required.  And probably an interface for credentials provider as well.

mullermp:
I added the sigv4 gem to testing but there's a couple failures. So I think we're really close. I'm not sure how to add Tempfile to rbs - the signature doesn't exist, and then we just need credentials/credential provider interfaces.

gems/aws-sigv4/sig/signer.rbs
k-in0ue: I'm sorry I wasn't able to address this. I was busy during the year-end and New Year holidays and couldn't work on it. Thank you for adding the interface on my behalf.🙇‍♂️

ksss:
To add stdlib (e.g. tempfile) dependencies that are not in the Gemfile.lock, you need to add the manifest.yaml file.

https://dev.to/pocke/rbs-introduced-manifestyaml-2idc

rbs_collection.yaml
```diff
@@ -0,0 +1,28 @@
...
+  - type: git
+    name: ruby/gem_rbs_collection
+    remote: https://github.com/ruby/gem_rbs_collection.git
+    revision: main
+    repo_dir: gems
+
+path: .gem_rbs_collection
+
+gems:
+  - name: addressable
```
alextwoods, line 17: I'm not sure if theres preferred ways to organize this.  It might be easier to read if all of the included stdlib (tempfile/stringio) are listed first, then all of the ignored ones?

alextwoods (approved):
Nice!

rbs_collection.yaml
jterapin: Yeah I think that might be better - all the ignored ones will be alphabetical

mullermp:
Thank you @kazuyainoue0124

PR #3158, opened 2024-12-27
docs: Fix documentation that references a non-existent method

nekketsuuu:
YARD comments for Aws::ARN says there is Aws::ARNParser#parse_resource method, which is not. It seems that this is just a documentation bug. So I've removed the lines.

Follow-up: fdb74d66458dfb9e37bdf28f8240359fbd6fdba1

### DIFF
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Fix documentation that references a non-existent method.
+
 3.214.0 (2024-11-25)
 ------------------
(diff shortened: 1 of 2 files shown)


mullermp:
Thanks for opening a pull request. It probably makes sense to update this note instead of removing it. There is a parse method on this class for parsing into an ARN. https://github.com/aws/aws-sdk-ruby/blob/9541f3d3f487cabe692f9bfb78c703f546aebbb8/gems/aws-sdk-core/lib/aws-sdk-core/arn_parser.rb#L13

mullermp (approved):

nekketsuuu:
Thanks!

PR #3173, opened 2025-01-24
[DO NOT MERGE] CBOR Performance Testing

richardwang1124:
These changes are intended to stay in the cbor-perf-testing branch and **NOT** to be merged to version-3. This PR was opened to facilitate review and discussion.

Run tests with `zsh perf-testing/run-perf-tests.sh <iteration_number>`. The actual test with 500 iterations can be executed with `zsh perf-testing/run-perf-tests.sh 500`.

The test output for each suite (Local only, Secrets Manager, CloudWatch) can be found in their respective directories under perf-testing/test-output. The data.txt file contains the properly formatted results and the raw.txt files contains the measured data points for each test case. The combined outputs are under perf-testing/test-output/combined.


.gitignore
```diff
@@ -15,3 +15,12 @@ benchmark_report.json
 .idea
 
 .ruby-version
+
+perf-testing/test-output/local-only/*.txt
```
alextwoods: Why not just `perf-test/test-output` - That should ignore everything in the output directory.

gems/aws-sdk-core/lib/aws-sdk-core/json/handler.rb
```diff
@@ -9,10 +9,25 @@ class Handler < Seahorse::Client::Handler
...
         response = @handler.call(context)
-        response.on(200..299) { |resp| parse_response(resp) }
+        response.on(200..299) do |resp|
+          success = true
+          t_deser_start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+          parse_response(resp)
+          t_deser_end = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+        end
         response.on(200..599) { |_resp| apply_request_id(context) }
+        if thread[:warm] && success
```
alextwoods: Do we care if the request succeeded or not? in general for this testing (where we aren't expecting/handeling errors) I think we can skip the check to simplify the code.

gems/aws-sdk-core/lib/aws-sdk-core/json/handler.rb
```diff
@@ -9,10 +9,25 @@ class Handler < Seahorse::Client::Handler
...
+          thread[:json_serde_data] << [format('%.3f', (t_ser_end - t_ser_start) * 1000.0).to_f,
```
alextwoods: Is there a reason to combine the serialization and deseralization into the same thread local?  It might simplify to store them separately.

gems/aws-sdk-core/lib/aws-sdk-core/json/handler.rb
```diff
@@ -9,10 +9,25 @@ class Handler < Seahorse::Client::Handler
...
+                                       format('%.3f', (t_deser_end - t_deser_start) * 1000.0).to_f]
```
alextwoods: Whats the motivation for using the %.3f here? I assume you're truncating to milliseconds?  Does the performance testing doc require us to do that?  In general our clock does give higher resolution that milliseconds, and some of these values may be low enough that we care about that extra precision.

alextwoods: Note: If we *must* use milliseconds, consider using the `Aws::Util.monotonic_milliseconds` method (or passing `:milliseconds` as the second argument to clock_gettime.

perf-testing/stats.rb
```diff
@@ -0,0 +1,9 @@
+module Stats
+  def p50(data)
```
alextwoods: These only work on sorted data right?  Maybe document that here (or change the parameter name, like `def p50(sorted_data)`

perf-testing/test_cloudwatch.rb
```diff
@@ -0,0 +1,246 @@
...
+  raw.puts(test_case)
+  output_raw(thread, raw)
+
+  separated_request_size = separate_byte_data(thread[:request_size_data])
+  # pp separated_request_size
+  separated_response_size = separate_byte_data(thread[:response_size_data])
+  # pp separated_response_size
+
+  separated_query_serde = separate_serde_data(thread[:query_serde_data])
+  write_test_output(test_case, 'Query', metrics, 'Serialization time (ms)', separated_query_serde[0], data)
```
alextwoods: Looking at Go's output the final file they produce is a json file with a list of these outupts.  Eg:
```
[{
        "service": "SecretsManager",
        "test_case": "Describe Secret",
        "protocol": "awsJson1_1",
        "dimension_value": 4096,
        "metric": "Total request time (ms)",
        "p50": 7.254758499999999,
        "p90": 8.358251000000001,
        "max": 2088.567724
    },
    {
        "service": "SecretsManager",
        "test_case": "Describe Secret",
        "protocol": "awsJson1_1",
        "dimension_value": 4096,
        "metric": "Deserialization time (ms)",
        "p50": 0.16587000000000002,
        "p90": 0.27500600000000003,
        "max": 10.494192
    },
....
]
```

I think the code could be simplified a lot by building a single array of Ruby Hash's and then writing out once with JSON.dump

perf-testing/test_local_only.rb
```diff
@@ -0,0 +1,267 @@
...
+    current[measurements[1]] = separated[1]
+    current[measurements[2]] = cbor_full_request_data[i]
+    current[measurements[3]] = cbor_full_response_data[i]
+    cbor_full_test_cases[test_cases[i]] = current
+  end
+  analyze(test_cases, measurements, cbor_full_test_cases, 'CBOR', data)
+end
+
+thread = Thread.current
+thread[:json_serde_data] = []
```
alextwoods: could use clear_thread_data for this?

perf-testing/test_local_only.rb
```diff
@@ -0,0 +1,267 @@
...
+thread[:cbor_serde_data] = []
+thread[:request_size_data] = []
+thread[:response_size_data] = []
+thread[:warm] = false
```
alextwoods: Rather than having a warm value on Thread and then needing to check it in each handler where we're recording values, you could call `clear_thread_data` after the warm up calls are done.

perf-testing/test_local_only.rb
```diff
@@ -0,0 +1,267 @@
...
+
+Aws::Echo::Client.remove_plugin(Aws::Plugins::Protocols::JsonRpc)
+json_echo = Aws::Echo::Client.new(plugins: [Aws::Plugins::Protocols::JsonRpc], stub_responses: true)
+cbor_echo = Aws::Echo::Client.new(plugins: [Aws::Plugins::Protocols::RpcV2], stub_responses: true)
+cbor_echo.config.api = cbor_echo.config.api.dup
+cbor_echo.config.api.metadata = cbor_echo.config.api.metadata.dup
+cbor_echo.config.api.metadata['protocol'] = 'smithy-rpc-v2-cbor'
+
+data = File.open('perf-testing/test-output/local-only/data.txt', 'w')
```
alextwoods: Should this be a json file?

perf-testing/test_local_only.rb
```diff
@@ -0,0 +1,267 @@
...
+$LOAD_PATH.unshift(File.expand_path('../gems/aws-sdk-echo/lib', __dir__))
+
+require 'aws-sdk-core'
+require 'aws-sdk-echo'
+require 'aws-sdk-core/plugins/protocols/json_rpc'
+require 'aws-sdk-core/plugins/protocols/rpc_v2'
+require_relative 'Stats'
+include Stats
+
+ITERATIONS = ARGV.first.to_i
```
alextwoods: While this value is technically not changing within this file, I think its a little misleading to use ALL_CAPS for it.

perf-testing/test_local_only.rb
```diff
@@ -0,0 +1,267 @@
...
+ASCII = "!\"#$%&\'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~".chars
+INT_MIN = -2_147_483_648
+INT_MAX = 2_147_483_647
+LONG_MIN = -9_223_372_036_854_775_808
+LONG_MAX = 9_223_372_036_854_775_807
+FLOAT_MIN = -2_147_483_648.0
+FLOAT_MAX = 2_147_483_647.0
+DOUBLE_MIN = -9_223_372_036_854_775_808.0
+DOUBLE_MAX = 9_223_372_036_854_775_807.0
+RUN_START_TIMESTAMP = Time.now
```
alextwoods: Ditto here - this value changes each run, so I'd potentially not case it as a constant.

perf-testing/test_secretsmanager.rb
```diff
@@ -0,0 +1,367 @@
...
+      separated[1] << i
+    end
+  end
+  separated
+end
+
+def write_test_output(operation, protocol, dimension, metric, input, outfile)
+  input.sort!
+  result = {
+    "service": 'SecretsManager',
```
alextwoods: Ruby has weird syntax for Hashes.  What you're doing here is actually creating a symbol key and doesn't require you to use quotes. eg:
```
result = {
  service: 'SecretsManager'
...
}
```

perf-testing/test_cloudwatch.rb
```diff
@@ -0,0 +1,246 @@
...
+    t_query_total_start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+    query_cloudwatch.put_metric_data(request)
+    t_query_total_end = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+    thread[:query_total_data] << format('%.3f', (t_query_total_end - t_query_total_start) * 1000.0).to_f
+
+    t_cbor_total_start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+    cbor_cloudwatch.put_metric_data(request)
+    t_cbor_total_end = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+    thread[:cbor_total_data] << format('%.3f', (t_cbor_total_end - t_cbor_total_start) * 1000.0).to_f
+    sleep(2) if (i % 50).zero?
```
alextwoods, line 203: Was this specified as part of the testing doc?

Edit: NVM - I see that it is - maybe add a comment for it.

perf-testing/test_cloudwatch.rb
```diff
@@ -0,0 +1,246 @@
...
+end
+thread[:warm] = true
+
+METRIC_COUNTS.each do |metrics|
+  (0...ITERATIONS).each do |i|
+    request = generate_put_metric_data_request(metrics, BASE_TIME, SUITE_ID)
+    t_query_total_start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+    query_cloudwatch.put_metric_data(request)
+    t_query_total_end = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+    thread[:query_total_data] << format('%.3f', (t_query_total_end - t_query_total_start) * 1000.0).to_f
```
alextwoods: We're doing this sort of behavior a lot - as a cleanup suggestion, consider a some sort of `benchmark` utility function that takes a block and would let you do:

```ruby
thread[:query_total_data] << Aws::Util.benchark do
  query_cloudwatch.put_metric_data(request)
end
```

alextwoods (commented):
Nice!  Its looking good overall and I think the general approach to measuring the data looks correct.

.gitignore
richardwang1124: I think my main issue was that the subdirectories (local-only, cloudwatch, secretsmanager) need to exist in order to create the data files and those files were what I wanted to hide. However, I realize now that I could gitignore everything in the test-output file and create the subdirectories as needed in the code. Would that be better?

gems/aws-sdk-core/lib/aws-sdk-core/json/handler.rb
richardwang1124: I think it's true that we aren't expecting errors and Kevin's doc specifies that any test runs that encounter errors must be considered invalid. I will remove it.

gems/aws-sdk-core/lib/aws-sdk-core/json/handler.rb
richardwang1124: Yeah initially I thought it might be better to group serialization and deserialization together but since we're sorting each group separately anyways storing them in separate arrays will simplify the data handling later. I'll update this.

gems/aws-sdk-core/lib/aws-sdk-core/json/handler.rb
richardwang1124: The time is stored as milliseconds already but I was truncating the value to three places past the decimal. I wasn't sure how high of a resolution the clock had so I chose an arbitrary value just to keep the number of decimal places consistent across measurements. Should I remove this entirely or choose a higher value to provide better accuracy and keep the number of decimal places consistent?

perf-testing/stats.rb
richardwang1124: Sounds good, will add.

perf-testing/test_local_only.rb
richardwang1124: I used ARGV just for creating the tests so I can use a smaller iteration number to check my work before jumping to 500. Should we continue to use ARGV or hardcode this to 500 for the actual test runs?

perf-testing/test_cloudwatch.rb
richardwang1124: Sure, I'll add a comment.

perf-testing/test_cloudwatch.rb
richardwang1124: My main concern with `benchmark` was that it seemed to print out the result to the terminal every time but from [here](https://stackoverflow.com/questions/36969333/run-benchmark-but-dont-print-the-result) I see I might be able to save the result instead of printing it. I'll try that out.

perf-testing/test_local_only.rb
alextwoods: Taking the argument is good, its nice to be able to control the iterations - my comment was just that the all caps casing of `ITERATIONS` to me implies this is a constant rather than an argument.

.gitignore
alextwoods: Yeah, I think that would be preferable.  You can use [mkdir_p](https://apidock.com/ruby/FileUtils/mkdir_p) which handles directories already existing (and ensuring all parents exist as well).

perf-testing/test_cloudwatch.rb
alextwoods: oh sorry - to clarify, I didn't mean use the `benchmark` library, I just meant write your own utility function to do this. (you can stick it in the Aws::Util namespace to make it easily accessible in the handlers and in the scripts).

richardwang1124: Ah I see. Okay I'll try that out.

gems/aws-sdk-core/lib/aws-sdk-core/json/handler.rb
richardwang1124: Possibly because the resolution is not high enough, when I use `Aws::Util.monotonic_milliseconds` or `:milliseconds` the results all display `0.0`...

alextwoods: I think its reasonable to do no truncation (so I'd vote for remove this entirely) - its fine if the number of decimal places differs across measurements.

perf-testing/stats.rb
```diff
@@ -1,9 +1,16 @@
 module Stats
-  def p50(data)
-    data[(data.length - 1) / 2]
+  MEASUREMENTS = ['Total request time (ms)', 'Serialization time (ms)', 'Deserialization time (ms)',
+                  'Request payload size (bytes)', 'Response payload size (bytes)'].freeze
+
+  def get_measurements
```
alextwoods: Ruby style - its generally considered reasonable to access constants directly rather than defining a method that returns them.  Also ruby method naming usually prefers shorter names and drops the `get`.

perf-testing/test_cloudwatch.rb
```diff
@@ -94,32 +85,36 @@ def separate_byte_data(data)
...
+    protocol: protocol,
+    dimension_value: dimension,
+    metric: metric,
+    p50: p50(input),
+    p90: p90(input),
+    max: input.last,
+    n: iterations
   }
-  outfile.puts(JSON.pretty_generate(result))
+  output << result
```
alextwoods: Nice, I like this over writing each time to the output file.

Generally however I'd lean towards not modifying the arguments: return the result and append it outside the call.

alextwoods (commented):
nice - updates look good.

alextwoods:
> You have made a change to core without a corresponding change to the CHANGELOG.md. This change will not result in a new version and will not published unless an entry is added to CHANGELOG.md.

You could probably remove the CI task that checks for the changelog entry in this branch since we never intend to merge it.

Or event invert the logic of it - comment if we DO have a changelog (since not having a changelog for these changes protects us from accidently releasing them if we did accidentally merge this branch).

PR #3175, opened 2025-01-29
Allow other retry configs to influence DynamoDB extended retries

### LINKED ISSUE #3172: Retries on DynamoDB do not match documentation
### Describe the bug

When reading the documentation for the retry configuration on DynamoDB, we can check the `initialize` documentation on the Client object:

```
:retry_backoff (Proc) — A proc or lambda used for backoff. Defaults to 2**retries * retry_base_delay. This option is only used in the legacy retry mode.
:retry_base_delay (Float) — default: 0.3 — The base delay in seconds used by the default backoff function. This option is only used in the legacy retry mode.
```

This shows that retries should have a base of 300ms, with a exponential backoff strategy that doubles the number on every retry. 
In practice, the default configuration of `retry_base_delay` is not the one shared in this documentation.

In practice, the default backoff proc is the one loaded from one of the plugins:

Here is an example dump from a ruby console (doing `.config` on the client) with a client with the default configuration:
```
retry_backoff=#<Proc:0x00007f2e35e53f70 /usr/local/bundle/ruby/3.2.0/gems/aws-sdk-dynamodb-1.106.0/lib/aws-sdk-dynamodb/plugins/extended_retries.rb:20 (lambda)>,
 ```

After opening this plugin, we saw that the retry method has a [hardcoded](https://github.com/aws/aws-sdk-ruby/blob/1ddf07f18fb78d3996f7265e1434404c6d04f356/gems/aws-sdk-dynamodb/lib/aws-sdk-dynamodb/plugins/extended_retries.rb#L22C13-L22C69) param of 50ms:

```ruby
Kernel.sleep(50 * (2 ** (context.retries - 1)) / 1000.0)
```

This differs completely with the default client configuration from `aws-sdk-core`:

https://github.com/aws/aws-sdk-ruby/blob/1ddf07f18fb78d3996f7265e1434404c6d04f356/gems/aws-sdk-core/lib/aws-sdk-core/plugins/retry_errors.rb#L29

The behaviour aligns with the traces we see on our side, where the retries match the 50ms base delay.

Also, from this default proc, we don't see how the `retry_base_delay` would be applied if configured.

### Regression Issue

### Expected Behavior

The retry base delay should be 300ms as per documentation.

### Current Behavior

The retry has a default base of 50ms instead of the 300ms explained on the documentation.

### Reproduction Steps

Create a default DynamoDB client and get the configuration

```
client = Aws::DynamoDB::Client.new
client.config
```

Here you can verify that the default configuration for the retry proc is obtained from the plugin.

### Possible Solution

I wanted to create a PR to fix this issue, but I wanted to know what would be the right approach on this.
There are two possible solutions that I can come up with:

- Remove the plugin. I am not sure why there is a plugin with this kind of configuration, but if removed the client would use the default proc from `aws-sdk-core`.
- Update the plugin. I can update the plugin to use the right configuration of retries.

The concern with these solution is also that it would change the default configuration. I wouldn't change the current default for dynamo (50ms) to the overall default from `aws-sdk-core` (300ms) because it might impact a lot of services. 

A change in the documentation would be needed to cover for this, to clarify that the default for dynamo is 50ms, independent from which fix is chosen.

_No response_

aws-sdk-dynamodb

Ruby 3.2

mullermp:
Fixes #3172

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,6 +5,7 @@ source 'https://rubygems.org'
 gem 'rake', require: false
 # SDK feature dependencies
 gem 'aws-crt' if ENV['CRT']
+gem 'base64'
 gem 'http-2'
 gem 'jmespath'
 if defined?(JRUBY_VERSION)
--- a/gems/aws-sdk-dynamodb/lib/aws-sdk-dynamodb/plugins/extended_retries.rb
+++ b/gems/aws-sdk-dynamodb/lib/aws-sdk-dynamodb/plugins/extended_retries.rb
@@ -4,6 +4,20 @@ module Aws
   module DynamoDB
     module Plugins
       class ExtendedRetries < Seahorse::Client::Plugin
+        DEFAULT_BACKOFF = lambda do |c|
+          return if c.retries < 1
+
+          delay = 2**(c.retries - 1) * c.config.retry_base_delay
+          if (c.config.retry_max_delay || 0) > 0
+            delay = [delay, c.config.retry_max_delay].min
+          end
+          jitter = c.config.retry_jitter
+          jitter = JITTERS[jitter] if jitter.is_a?(Symbol)
+          delay = jitter.call(delay) if jitter
+          Kernel.sleep(delay)
+        end
+
+
         option(
           :retry_limit,
           default: 10,
@@ -24,6 +38,15 @@ This option is only used in the `legacy` retry mode.
           docstring: <<-DOCS)
 The base delay in seconds used by the default backoff function. This option
 is only used in the `legacy` retry mode.
+        DOCS
+
+        option(
+          :retry_backoff,
+          default: DEFAULT_BACKOFF,
+          doc_type: Proc,
+          docstring: <<-DOCS)
+A proc or lambda used for backoff. Defaults to 2**retries * retry_base_delay.
+This option is only used in the `legacy` retry mode.
         DOCS
       end
     end
--- a/gems/aws-sdk-dynamodb/lib/aws-sdk-dynamodb/plugins/extended_retries.rb
+++ b/gems/aws-sdk-dynamodb/lib/aws-sdk-dynamodb/plugins/extended_retries.rb
@@ -5,7 +5,7 @@ module Aws
     module Plugins
       class ExtendedRetries < Seahorse::Client::Plugin
         DEFAULT_BACKOFF = lambda do |c|
-          return if c.retries < 1
+          return unless c.retries > 1
 
           delay = 2**(c.retries - 1) * c.config.retry_base_delay
           if (c.config.retry_max_delay || 0) > 0
--- a/gems/aws-sdk-dynamodb/lib/aws-sdk-dynamodb/plugins/extended_retries.rb
+++ b/gems/aws-sdk-dynamodb/lib/aws-sdk-dynamodb/plugins/extended_retries.rb
@@ -12,11 +12,10 @@ module Aws
             delay = [delay, c.config.retry_max_delay].min
           end
           jitter = c.config.retry_jitter
-          jitter = JITTERS[jitter] if jitter.is_a?(Symbol)
+          jitter = Aws::Plugins::RetryErrors::JITTERS[jitter] if jitter.is_a?(Symbol)
           delay = jitter.call(delay) if jitter
           Kernel.sleep(delay)
         end
-
 
         option(
           :retry_limit,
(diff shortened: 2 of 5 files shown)


mullermp:
@dankochime

gems/aws-sdk-dynamodb/lib/aws-sdk-dynamodb/plugins/extended_retries.rb
```diff
@@ -4,25 +4,27 @@ module Aws
...
           doc_type: Integer,
           docstring: <<-DOCS)
 The maximum number of times to retry failed requests.  Only
 ~ 500 level server errors and certain ~ 400 level client errors
 are retried.  Generally, these are throttling errors, data
-checksum errors, networking errors, timeout errors and auth
-errors from expired credentials.
-          DOCS
-
-        option(:retry_backoff, default: lambda { |context|
```
alextwoods, line 20: I'm concerned about this change in behavior:
```
retries: 0       old: 0         new: 0.05
retries: 1       old: 0         new: 0.1
retries: 2       old: 0.1       new: 0.2
retries: 3       old: 0.2       new: 0.4
retries: 4       old: 0.4       new: 0.8
retries: 5       old: 0.8       new: 1.6
retries: 6       old: 1.6       new: 3.2
retries: 7       old: 3.2       new: 6.4
retries: 8       old: 6.4       new: 12.8
retries: 9       old: 12.8      new: 25.6
retries: 10      old: 25.6      new: 51.2
```
This is a pretty big change, in particular for the first and second retry which are probably the most important and additionally ends up with about double the time on most retries. 

I understand and agree with trying to make sure that the other legacy retry settings are respected, but I think we need to be careful changing default retry backoffs here.

mullermp: The intent was to be the same. I can re-define retry backoff again.

alextwoods: I think we may want to still define the retry_backoff in extended retries, but we can add those other configs to it, but default (and document) them in a way that the behavior doesn't change by default

mullermp: That's the current experience. I think we want base delay, jitter, and max delay to still apply.

gems/aws-sdk-dynamodb/lib/aws-sdk-dynamodb/plugins/extended_retries.rb
```diff
@@ -4,25 +4,50 @@ module Aws
   module DynamoDB
     module Plugins
       class ExtendedRetries < Seahorse::Client::Plugin
+        DEFAULT_BACKOFF = lambda do |c|
+          return if c.retries < 1
```
alextwoods: Needs to be `<=` I think.

gems/aws-sdk-dynamodb/lib/aws-sdk-dynamodb/plugins/extended_retries.rb
```diff
@@ -4,25 +4,50 @@ module Aws
...
 
-        option(:retry_limit,
+          delay = 2**(c.retries - 1) * c.config.retry_base_delay
+          if (c.config.retry_max_delay || 0) > 0
+            delay = [delay, c.config.retry_max_delay].min
+          end
+          jitter = c.config.retry_jitter
+          jitter = JITTERS[jitter] if jitter.is_a?(Symbol)
```
alextwoods: Will this work? I don't think `JITTERS` is defined here

mullermp: Ah good catch. It looks like this is not really tested either.

Gemfile
```diff
@@ -5,6 +5,7 @@ source 'https://rubygems.org'
 gem 'rake', require: false
 # SDK feature dependencies
 gem 'aws-crt' if ENV['CRT']
+gem 'base64'
```
alextwoods, line 8: Unrelated to this change, but should we add BigDecimal as well here?

mullermp: I'll do a separate PR to add ruby 3.4 to CI and also that.

alextwoods (approved):

PR #3184, opened 2025-02-06
Improve code generation to avoid unused variable warnings

byroot:
The Rails test suite had these warnings for ages, which I'd like to fix:

```
aws-sdk-s3/endpoint_provider.rb:22: warning: assigned but unused variable - key
aws-sdk-s3/endpoint_provider.rb:24: warning: assigned but unused variable - copy_source
aws-sdk-s3/endpoint_provider.rb:23: warning: assigned but unused variable - prefix
```

Ultimately I don't think pre-assigning to variables is really worth it, as the `parameters` argument is a T_STRUCT so access is really fast.

But I suspect this was done to make the code generation simpler. In the end this adds a little bit of complexity, but not that much.

NB: I only regenerated the aws-sdk-s3 gem to show what the change looks like. But if this change is accepted I can regenerate all gems, and add a CHANGELOG entry.

### DIFF
--- a/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/endpoint_provider_class.rb
+++ b/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/endpoint_provider_class.rb
@@ -7,6 +7,7 @@ module AwsSdkCodeGenerator
       # @option options [required, Service] :service
       # @option options [required, Hash] :endpoint_rules
       def initialize(options)
+        @assigned_variables = []
         @service = options.fetch(:service)
         @endpoint_rules = @service.endpoint_rules
         # Used to collect metrics in the generated endpoint provider
@@ -33,10 +34,6 @@ module AwsSdkCodeGenerator
 
       def endpoint_rules_code
         res = StringIO.new
-        # map parameters first
-        @endpoint_rules["parameters"].each do |k,_v|
-          res << indent("#{underscore(k)} = parameters.#{underscore(k)}\n", 3)
-        end
 
         # map rules
         @endpoint_rules["rules"].each do |rule|
@@ -164,6 +161,7 @@ module AwsSdkCodeGenerator
 
       def condition(condition)
         if condition['assign']
+          @assigned_variables << condition['assign']
           "(#{underscore(condition['assign'])} = #{fn(condition)})"
         else
           fn(condition)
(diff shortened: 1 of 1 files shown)


byroot:
An alternative strategy would be to prefix all local variables with `_` to tell the parser not to emit a warning. I assumed it wasn't a preferable solution, but it's an option.

mullermp:
Thanks for opening a pull request. Either approach is fine. This seems ok at a glance, I will review today. We should not include generated changes in the commit because our release system will handle that, and a changelog would be applied for all gems automatically (though it's a generic one about updating generating code - I think that's ok).

byroot:
> We should not include generated changes in the commit

Right, I remembered that from previous contributions, but I figured you'd maybe want to see the result?

I can revert that part whenever you like.

mullermp:
Yeah, perfectly fine for review! Your post said you would generate all gems unless I misunderstood.

mullermp:
I enabled CI for you. It looks like there are failures. `rake test:spec:aws-sdk-s3` is a local way to run these.

byroot:
Thank you. I think I fixed the problem, and I removed the generated files.

gems/aws-sdk/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Fix - Improve code generation to avoid triggering Ruby unused variable warnings.
```
mullermp: We should not need any changelog entry. Our release system will add one to all gems when they are built. This changelog is also for the aws-sdk gem which pulls in all gems. It is a gem we don't want to update or maintain but exists for compatibility.

mullermp (approved):

gems/aws-sdk/CHANGELOG.md
byroot: Ah my bad.

byroot: Removed.

alextwoods:
Adding some historical context here - Smithy validates endpoint rulesets to ensure that all parameters are used.  However, S3 needed an exception to support using endpoint rules for a few complex auth resolution cases (eg access grants), so those unused parameters violate our normal validations.  We would expect to see these warnings only in S3.

alextwoods (approved):

PR #3186, opened 2025-02-07
add base64 as explicit dependency

### LINKED ISSUE #2984: Gem displays deprecation warning in ruby 3.3

HoneyryderChuck:
closes #2984

was made into a bundled gem in ruby 3.4

### DIFF
--- a/gems/aws-sdk-core/aws-sdk-core.gemspec
+++ b/gems/aws-sdk-core/aws-sdk-core.gemspec
@@ -1,9 +1,8 @@
 # frozen_string_literal: true
 
 Gem::Specification.new do |spec|
-
   spec.name          = 'aws-sdk-core'
-  spec.version       = File.read(File.expand_path('../VERSION', __FILE__)).strip
+  spec.version       = File.read(File.expand_path('VERSION', __dir__)).strip
   spec.summary       = 'AWS SDK for Ruby - Core'
   spec.description   = 'Provides API clients for AWS. This gem is part of the official AWS SDK for Ruby.'
   spec.author        = 'Amazon Web Services'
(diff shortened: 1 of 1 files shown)


mullermp:
I was looking at doing this soon. I would like to add it to all gems where it is used.

HoneyryderChuck:
@mullermp feel free to close it if you are already working on this. my change was mostly "drive by" to fix my local issue, but you're right that all aws-sdk-* gems should declare new bundled gems individually.

mullermp:
I haven't started. We can use this PR. I'll look today.

mullermp:
I just took a look. I see there are some individual gems that need base64 but I think it's entirely reasonable to get those through cores. What I was looking at was non aws-sdk gems such as aws-sigv4, aws-eventstreams, etc, and it appears base64 is used during tests, which does not need to be included. So, I think as a whole, this change is just fine.

alextwoods (approved):

malcolm-pro:
This change is causing problems with builds where the gem is pulled into a cache in a Ruby compatible with 3.1.x. Could this dependency be flagged so it only appears in builds where Ruby >= 3.4? Ie in the gemspec:
```
if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('3.4.0')
  spec.add_dependency('base64')
end
```

mullermp:
I don't think that approach works unfortunately. Are you able to get that working locally? I checked out ruby 3.1, added that to the core gemspec, and bundle install brought in base64.

Can you be more specific about your cache problem? That seems like a very odd case. The default gem is supposed to work backwards compatibly here.

malcolm-pro:
I dug a little deeper and found that version 0.1.1 was popping up somewhere in my build and causing trouble; installing over that fixed the problem, sorry for the trouble!

PR #3188, opened 2025-02-12
Remove event stream max payload size and headers length checks

richardwang1124:
Removes the event stream max payload size and headers length checks to align with new size limits.

---


gems/aws-eventstream/spec/encoder_spec.rb
```diff
@@ -29,18 +29,18 @@ module EventStream
...
+        it 'does not raise an error when payload exceeds' do
           payload = double('payload', :length => 16777217)
           message = Aws::EventStream::Message.new(
             headers: {},
             payload: payload
           )
           expect {
             Encoder.new.encode(message)
-          }.to raise_error(Aws::EventStream::Errors::EventPayloadLengthExceedError)
+          }.not_to raise_error(Aws::EventStream::Errors::EventPayloadLengthExceedError)
```
jterapin: Do we need to keep this error classes around? Might not need it anymore since we don't raise on our end.

jterapin (commented):
Looks great so far.

gems/aws-eventstream/spec/encoder_spec.rb
richardwang1124: We could remove these error classes and replace it with just `raise_error` but that leads to errors related to  `#<RSpec::Mocks::MockExpectationError: #<Double "payload"> received unexpected message :read with (no args)>` and `#<NoMethodError: undefined method 'bytesize' for nil>`. Something may be wrong with the test's payload/message?

mullermp: You should be able to remove the classes since we don't catch them outside of this gem. Though someone might, it is unlikely - this gem is not intended to be used directly by customers. The error are you seeing is because the test was only made to trip the error case. I honestly don't think you need the test case at all.

alextwoods: I do think its reasonable to remove these errors, though it is technically a breaking change (if a customer had code that referenced these classes.  I agree these aren't intended for direct users). 

I do think we should have tests the cover the case, but rather than framing them as "does not raise" - I think we can change them to something like: `'encodes large payloads'` and it can just have `Encoder.new.encode(message)` with no expect around it.

jterapin: To give context on why this is happening - it is related to how `payload` is defined here.  Our mock responds to only `:length` method so when lines 84-86 were removed - the code does the `payload.read` but hey! that method is not defined. If you update the payload like below, it would pass.

```
payload = double('payload', :read => '') # syntax prob needs to be updated here
``` 


`:read` probably should return some big payload object.

richardwang1124: Removed the error classes and updated the test. Thanks all!

alextwoods (approved):

gems/aws-eventstream/spec/encoder_spec.rb
```diff
@@ -25,33 +25,25 @@ module EventStream
             expect(test_io.string.freeze).to eq(expectation)
           end
         end
-      end
-
-      describe '#encode error' do
 
-        it 'raises an error when payload exceeds' do
-          payload = double('payload', :length => 16777217)
+        it 'encodes large payloads' do
```
mullermp, line 29: This is what I wanted, but I thought Alex wanted to keep tests around that assert no error raised?

mullermp (approved):

gems/aws-eventstream/spec/encoder_spec.rb
alextwoods: Ah - I think maybe we had a miscommunication on it.  I thought you didn't want to test these cases at all rather than just removing the expect not_to raise.

PR #3195, opened 2025-02-18
Fix broken smoke tests with customizations

mullermp:
These smoke tests are broken for various reasons, such as the service is discontinued, or bad modeling. This adds some customization to fix the tests so that failures are documented.


gems/aws-sdk-observabilityadmin/features/smoke.feature
```diff
@@ -17,7 +17,7 @@ Feature: Smoke tests for ObservabilityAdmin
       """
 {}
       """
-    Then I expect a 'Aws::ObservabilityAdmin::Errors::AccessDeniedException' was raised
+    Then I expect an error was not raised
```
jterapin, line 20: Getting the following error when running this services test:
```
RSpec::Expectations::ExpectationNotMetError: expected: nil
     got: #<Aws::ObservabilityAdmin::Errors::ValidationException: Caller is not a member of an organization.>
```

jterapin (approved):
LGTM

PR #3197, opened 2025-02-19
Do not use connection-pooling in single-threaded process.

tubsandcans:
Connection pooling in a single-threaded process does not provide any benefit, and in certain cases can cause processes to become non-reusable.


mullermp:
I am going to close this. Feel free to reopen with new information - but seeing as this is a singular report, it is unlikely an issue with the SDK.

PR #3204, opened 2025-03-03
Stub config per request

mullermp:
An approach that sticks stubs, api requests, and their mutexes on undocumented config, rather than instance variables.

This also implements the per request stub approach originally from: https://github.com/aws/aws-sdk-ruby/pull/2679/commits/5d26627f10a03bbffaae086a1594ccb4b715722f

This pushes stub resolution at request time rather than up front on client initialize or stub_responses call.


build_tools/aws-sdk-code-generator/spec/interfaces/client/bearer_authorization_spec.rb
```diff
@@ -25,6 +25,7 @@
       it 'sets the Authorization header' do
         client.stub_responses(:bearer_auth, -> (context) {
           expect(context.http_request.headers['Authorization']).to eq('Bearer token')
+          {}
```
mullermp, line 28: We were not doing these correctly - dynamic stubs should return a stub.

gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb
```diff
@@ -228,54 +212,42 @@ def api_requests(options = {})
...
+        stubs = @config.stubs[operation_name] || []
         case stubs.length
-        when 0 then default_stub(operation_name)
+        when 0 then stub_data(operation_name)
         when 1 then stubs.first
         else stubs.shift
         end
       end
-      Proc === stub ? convert_stub(operation_name, stub.call(context)) : stub
+      convert_stub(operation_name, stub, context)
```
mullermp: On next stub call, convert the stub recursively, even if it's a nested proc (which was not supported before)

gems/aws-sdk-core/lib/aws-sdk-core/plugins/stub_responses.rb
```diff
@@ -46,8 +54,20 @@ def after_initialize(client)
         end
       end
 
-      class Handler < Seahorse::Client::Handler
+      class ApiRequestsHandler < Seahorse::Client::Handler
```
mullermp, line 57: Logs the api request (the anonymous handle in setup_stubbing) but prior to sending so presigned urls are still tracked.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/stub_responses.rb
```diff
@@ -57,14 +77,10 @@ def call(context)
...
-          stub = context.client.next_stub(context)
           resp = Seahorse::Client::Response.new(context: context)
           async_mode = context.client.is_a? Seahorse::Client::AsyncBase
-          if Hash === stub && stub[:mutex]
-            stub[:mutex].synchronize { apply_stub(stub, resp, async_mode) }
-          else
-            apply_stub(stub, resp, async_mode)
-          end
+          stub = context.client.next_stub(context)
+          apply_stub(stub, resp, async_mode)
```
mullermp: We should not need a per-stub only mutex anymore? I will try to confirm.

gems/aws-sdk-dynamodb/spec/client_spec.rb
```diff
@@ -161,14 +161,10 @@ module DynamoDB
 
         it 'observes the :simple_attributes configuration option' do
```
mullermp, line 162: Validation gets pushed to call time instead of stub_responses time. I'm not sure if this is better or worse.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/stub_responses.rb
alextwoods: I'm not 100% sure, but I *think* we might still need the mutex.  

I think this fixes the issue for requests that generate a stub (ie data stubs), but not for raw http stubs.  

In the case where there is only one HTTP response stub (so the same stub would get returned multiple times), we would do `hash_to_http_resp` and then still call `signal_http(stub[:http], http_resp, async_mode)`.  If an operation was invoked multiple times concurrently, we could end up with the same race condition where two threads are signaling from the same body.

gems/aws-sdk-dynamodb/spec/client_spec.rb
alextwoods: I think stub_responses time validation is still probably ideal - but we could call the param validator up front but not actually do the conversion until call time.

gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb
```diff
@@ -15,27 +15,11 @@ module ClientStubs
...
     def setup_stubbing
-      @stubs = {}
-      @stub_mutex = Mutex.new
       if Hash === @config.stub_responses
         @config.stub_responses.each do |operation_name, stubs|
           apply_stubs(operation_name, Array === stubs ? stubs : [stubs])
         end
       end
-
-      # When a client is stubbed allow the user to access the requests made
```
alextwoods, line 26: Nice - I like this in the plugin's handler much better.

gems/aws-sdk-dynamodb/spec/client_spec.rb
mullermp: I'm not certain that works in all cases, such as if the stub was a proc, you'd have to resolve it then.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/stub_responses.rb
mullermp: I think you're right. I can add back mutex per stub. I think upfront converting will simplify the code (removes branching about whether it exists or not).

Gemfile
```diff
@@ -6,6 +6,7 @@ gem 'rake', require: false
 # SDK feature dependencies
 gem 'aws-crt' if ENV['CRT']
 gem 'base64'
+gem 'byebug', platforms: :ruby
```
alextwoods: Should probalby be moved to development section

alextwoods (approved):

Gemfile
mullermp: I didn't plan to keep this change but I could put it there yeah.

jterapin: Ooo I actually would like that

jterapin (approved):

rafal-brize:
Hey folks, I observe a different breaking change with this, than the one above.
Is this intentional?

```
stubbed_aws_client =    Aws::CloudFront::Client.new(stub_responses: true)
stubbed_data = stubbed_aws_client.stub_data(:get_distribution, distribution: {distribution_config: {viewer_certificate: {acm_certificate_arn: "limey"}}}) # this is a partial data
stubbed_aws_client.stub_responses(:get_distribution, stubbed_data)

stubbed_aws_client.get_distribution(id: "limes") # does not cause errors in  3.220.0 and raises in 3.220.1
```

```
Error: the evaluation of `req.send_request(options)` failed with the exception 'parameter validator found 11 errors:   - missing required parameter params[:distribution][:id]   - missing required parameter params[:distribution][:arn]   - missing required parameter params[:distribution][:status]   - missing required parameter params[:distribution][:last_modified_time]   - missing required parameter params[:distribution][:in_progress_invalidation_batches]   - missing required parameter params[:distribution][:domain_name]   - missing required parameter params[:distribution][:distribution_config][:caller_reference]   - missing required parameter params[:distribution][:distribution_config][:origins]   - missing required parameter params[:distribution][:distribution_config][:default_cache_behavior]   - missing required parameter params[:distribution][:distribution_config][:comment]   - missing required parameter params[:distribution][:distribution_config][:enabled]'
```

mullermp:
I'm sorry this broke you. Yes, this now causes validation of the stub. I can see how this might be a sharp edge for a service such as cloudfront that has a response that requires so much output. I can look for solutions of this - but the easiest way to unblock you is to set the required parameters in your partial stub. You can also pass the hash directly to stub_responses instead of getting an intermediary stub_data call.

mullermp:
A couple options:
* (easy) add validate_required: false here: https://github.com/aws/aws-sdk-ruby/blob/568213726f240ba67f15f675f7429a0d77ea8958/gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb#L278. The biggest con is that there is a tradeoff of usability vs correctness. Ideally your testing should work with as close to as a real service response as possible - which means you should see an id, an arn, etc, requiring you to stub that.
* (harder?) stub_data should fill in required members. Arguably [stub_data](https://github.com/aws/aws-sdk-ruby/blob/568213726f240ba67f15f675f7429a0d77ea8958/gems/aws-sdk-core/lib/aws-sdk-core/stubbing/stub_data.rb#L55) should never have used validate_required: false for the same reason as above, but it does. It should have stopped you from creating a partial stub. We can consider filling required members into the [empty stub](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/lib/aws-sdk-core/stubbing/empty_stub.rb#L5) but it would not handle the case where you call stub_responses directly with a hash, so the experience is still inconsistent.
* No action - accept this as a partial breaking change - Probably a bit unpopular in your case, but customers are already required to add required input to the operation call and required output (in the case of hashes to stub_responses) to their stubs to simulate a real response. This was not the case with stub_data, which is a rarer case, but now all is consistent.

I can talk with the team on this, and also if you'd like to provide your thoughts too.

rafal-brize:
Thanks for a quick reply @mullermp 

I liked using stub data so far because it was essentially separating me from the http layer.
Same time i understand the need to unify things.

Would be great if we could have stub data that is valid in itself and allows customisation replacing certain attributes, i also understand that would require a lot of work to prepare those.

With current approach things like Cloudfront `list_distributions` requires me to pass 16 additional complex attributes per single item in the list. It starts resembling testing i would do with the use of a http mocking library like webmock.

> You can also pass the hash directly to stub_responses instead of getting an intermediary stub_data call.

In order to do that my tests need to start using http lingo:

https://github.com/aws/aws-sdk-ruby/blob/568213726f240ba67f15f675f7429a0d77ea8958/gems/aws-sdk-core/lib/aws-sdk-core/client_stubs.rb#L260-L263

which i tried to avoid in the first place.

Thank you for all your work.

mullermp:
You shouldn't need to make an http response or a hash with http response parts, you can take a hash of data that you otherwise would have called stub_data with and pass it directly to stub_responses.

```
stubbed_aws_client.stub_responses(:get_distribution, distribution: { ... })
```

Essentially this is the common pattern and this approach prior to my change was already validating required members were set.

mullermp:
@rafal-brize One thing you can do for convenience:

```
stubbed_aws_client =    Aws::CloudFront::Client.new(stub_responses: true)
stubbed_data = stubbed_aws_client.stub_data(:get_distribution)
stubbed_data.distribution.distribution_config[:viewer_certificate][:acm_certificate_arn] = 'limey'
```

agrare:
Hey @mullermp we ran into the same issue as @rafal-brize above https://github.com/aws/aws-sdk-ruby/pull/3204#issuecomment-2709856255.  This appears to have been introduced in 3.220.1 as 3.220.0 is not impacted.  Given that this gem follow semantic versioning would you consider this a breaking change?  I realize it is only in specs so just curious

mullermp:
Yes, this was not entirely intended, I'm sorry for the breakage there. The `stub_responses(:operation, hash)` usage will validate these things already, but it wasn't with stub data. I don't think the issue is widespread to warrant any rollback. You should be able to use the stub_data approach above as a quick fix.

PR #3208, opened 2025-03-06
Add back validation for the eventstream payload and header

mullermp:
Partial revert of https://github.com/aws/aws-sdk-ruby/pull/3188/files but increases the limit.

### DIFF
--- a/gems/aws-eventstream/lib/aws-eventstream/errors.rb
+++ b/gems/aws-eventstream/lib/aws-eventstream/errors.rb
@@ -32,6 +32,18 @@ module Aws
         end
       end
 
+      class EventPayloadLengthExceedError < RuntimeError
+        def initialize(*args)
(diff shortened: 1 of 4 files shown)


richardwang1124 (approved):

gems/aws-eventstream/lib/aws-eventstream/errors.rb
```diff
@@ -32,6 +32,18 @@ def initialize(*args)
         end
       end
 
+      class EventPayloadLengthExceedError < RuntimeError
+        def initialize(*args)
+          super("Payload length of a message should be under 16mb.")
```
richardwang1124: This should be updated to 24 right?

mullermp: Ah yeah oops.

PR #3214, opened 2025-03-21
Fix h2 connections to use config values

mullermp:
Actually set the default this time.

### DIFF
--- a/gems/aws-sdk-core/lib/seahorse/client/h2/connection.rb
+++ b/gems/aws-sdk-core/lib/seahorse/client/h2/connection.rb
@@ -10,55 +10,48 @@ module Seahorse
   module Client
     # @api private
     module H2
-
       # H2 Connection build on top of `http/2` gem
-      # (requires Ruby >= 2.1)
-      # with TLS layer plus ALPN, requires:
-      # Ruby >= 2.3 and OpenSSL >= 1.0.2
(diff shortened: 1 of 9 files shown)


alextwoods (approved):
We should potentially add a comment to the h2 plugin that calls out that the defaults aren't used and that we need to also update them in the h2 connection class.

gems/aws-sdk-core/lib/seahorse/client/h2/connection.rb
```diff
@@ -10,56 +10,49 @@ module Seahorse
...
-          connection_read_timeout: 60,
-          http_wire_trace: false,
-          logger: nil,
-          ssl_verify_peer: true,
-          ssl_ca_bundle: nil,
-          ssl_ca_directory: nil,
-          ssl_ca_store: nil,
-          enable_alpn: false
-        }
+        OPTIONS = [
```
alextwoods: Are there backwards/forwards risks in removing the defaults from here? 
IE - old service gem + new core? I think because the changes to async base client are also in core, we should be good, but just want to check.

mullermp: I don't think so. It's all contained in the connection/base.

alextwoods (approved):

PR #3221, opened 2025-03-25
[aws-sdk-core] Fix typo in method's name in Seahorse::Client::H2::Connection

D-system:


mullermp:
Thanks. This happened from some churn/revert. I've removed the changelog entry and it will get published next time (since it's benign).

mullermp (approved):

PR #3223, opened 2025-03-26
Ensure that client plugin options doc are generated

jterapin:
Any Doc changes made to default base client plugins were not being recognized by the codegenerator. This PR addresses that and as well - adds a new method to the `BuildTools::ServiceEnumerator` to reset services rather than re-assigning the const to overwrite the service updates.

---


build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/plugin_list.rb
```diff
@@ -8,8 +8,12 @@ class PluginList
     def initialize(options)
       @aws_sdk_core_lib_path = options.fetch(:aws_sdk_core_lib_path)
       @plugins = compute_plugins(options)
+      @client_plugins = compute_client_plugins(options)
```
jterapin, line 11: Few things I want to mention here:
* I decided to move the "default plugins" logic like [this](https://github.com/aws/aws-sdk-ruby/blob/version-3/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/async_client_class.rb#L27-L29) into here since I could take an advantage of existing functionalities and the aws-sdk-core path!
*  I didn't think "default plugins" was sufficient to describe what makes it about them as default so I changed these to "client plugins" - that might be subjective so open to changing it back.
* These plugins shouldn't be combined with `@plugins` since `client_plugins` are only intended to use for codegenerating documentation.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/plugin_list.rb
```diff
@@ -27,44 +31,74 @@ def compute_plugins(options)
...
+        'Seahorse::Client::Plugins::NetHttp' => "#{seahorse_plugins_path}/h2.rb",
+        'Seahorse::Client::Plugins::RaiseResponseErrors' => "#{seahorse_plugins_path}/raise_response_errors.rb",
+        'Seahorse::Client::Plugins::ResponseTarget' => "#{seahorse_plugins_path}/response_target.rb",
+        'Seahorse::Client::Plugins::RequestCallback' => "#{seahorse_plugins_path}/request_callback.rb",
+      }
+    end
+
     def default_plugins
       {
-        'Seahorse::Client::Plugins::ContentLength' => "#{seahorse_plugins}/content_length.rb",
```
jterapin, line 42: I looked into listing this as a base client plugins but decided not to make any changes. There might be some edge cases around `ContentLength` plugin so I left it alone.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/plugin_list.rb
```diff
@@ -116,36 +150,27 @@ def signature_plugins(options)
...
+    def core_plugins_path
       File.join(@aws_sdk_core_lib_path, 'aws-sdk-core/plugins')
     end
 
-    def seahorse_plugins
+    def seahorse_plugins_path
       File.join(@aws_sdk_core_lib_path, 'seahorse/client/plugins')
     end
 
-    def core_lib
```
jterapin, line 138: I couldn't find any usage for this method so removed.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/plugin_list.rb
```diff
@@ -116,36 +150,27 @@ def signature_plugins(options)
...
-      #        than have the hard-coded here as a relative path
-      File.expand_path('../../../../../gems/aws-sdk-core/lib', __FILE__)
-    end
-
     def const_get(class_name)
-      const_names = class_name.split('::')
-      const_names.inject(Kernel) do |const, const_name|
-        const.const_get(const_name)
-      end
+      Object.const_get(class_name)
```
jterapin, line 173: This is super old code from 9 years ago to get the const. I think the replacing line is fine? If so, I will get rid of the entire method entirely and just use `Object.const_get(...)` in the above code.

jterapin:
Demo commit: https://github.com/aws/aws-sdk-ruby/pull/3223/commits/cf024170e38dcfe77fa4ccfd23d0cd1238abb0d3

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/plugin_list.rb
mullermp: The old code is fetching constant iteratively. I'm not sure if it's necessary anymore so it can probably be deleted. As long as it still works.

```
[1] pry(Aws)> Object.const_get('Seahorse::Client::Response')
=> Seahorse::Client::Response
```

mullermp (approved):
Seems fine. Ensure rake build does not produce anything wildly different.

jterapin:
Confirming here that `rake build` is working as intended

PR #3237, opened 2025-04-28
Support disable host prefixing in env and shared config

mullermp:
Allow disabling of host prefixing using ENV and shared config.


gems/aws-sdk-core/lib/aws-sdk-core/plugins/endpoint_pattern.rb
```diff
@@ -4,62 +4,70 @@ module Aws
...
+      option(
+        :disable_host_prefix_injection,
         default: false,
         doc_type: 'Boolean',
-        docstring: <<-DOCS
-Set to true to disable SDK automatically adding host prefix
-to default service endpoint when available.
-        DOCS
-      )
+        docstring: 'When true, the SDK will not prepend the modeled host prefix to the endpoint.'
```
jterapin:
```suggestion
        docstring: 'When `true`, the SDK will not prepend the modeled host prefix to the endpoint.'
```

gems/aws-sdk-core/lib/aws-sdk-core/plugins/endpoint_pattern.rb
```diff
@@ -4,62 +4,70 @@ module Aws
...
-      class Handler < Seahorse::Client::Handler
+      class << self
+        private
+
+        def resolve_disable_host_prefix_injection(cfg)
+          value = ENV['AWS_DISABLE_HOST_PREFIX_INJECTION'] ||
+                  Aws.shared_config.disable_host_prefix_injection(profile: cfg.profile) ||
+                  'false'
+          # Raise if provided value is not true or false
+          if value != 'true' && value != 'false'
```
jterapin: Nit: Could use our handy `Aws::Util.str_2_bool` but this is fine too

gems/aws-sdk-core/lib/aws-sdk-core/plugins/endpoint_pattern.rb
```diff
@@ -4,62 +4,70 @@ module Aws
...
+            raise ArgumentError,
+                  'Must provide either `true` or `false` for '\
+                    'adaptive_retry_wait_to_fill profile option or for '\
+                    'ENV[\'AWS_ADAPTIVE_RETRY_WAIT_TO_FILL\']'
```
jterapin: I think this is what you meant:
```suggestion
                  'Must provide either `true` or `false` for '\
                    'disable_host_prefix_injection profile option or for '\
                    'ENV[\'AWS_DISABLE_HOST_PREFIX_INJECTION\']'
```

gems/aws-sdk-core/lib/aws-sdk-core/plugins/endpoint_pattern.rb
```diff
@@ -4,62 +4,70 @@ module Aws
...
+      # @api private
+      class Handler < Seahorse::Client::Handler
         def call(context)
-          if !context.config.disable_host_prefix_injection
+          unless context.config.disable_host_prefix_injection
             endpoint_trait = context.operation.endpoint_pattern
-            if endpoint_trait && !endpoint_trait.empty?
-              _apply_endpoint_trait(context, endpoint_trait)
-            end
+            _apply_endpoint_trait(context, endpoint_trait) if endpoint_trait && !endpoint_trait.empty?
```
jterapin: Nit: I'm guessing we are using `_` to indicate private methods (noticed that these code are pretty old) - could update it. Same goes for `_replace_label_value` method

jterapin (approved):
Just minor feedbacks.

PR #3247, opened 2025-05-13
Only load what is required from `cgi` for Ruby 3.5

Earlopain:
In Ruby 3.5, most of the `cgi` gem is removed. Only methods relating to escaping/unescaping are retained.

On versions before Ruby 3.5, `cgi/util` must be required for correct functionality.

https://bugs.ruby-lang.org/issues/21258

-------


gems/aws-sdk-s3/spec/object/presigned_url_spec.rb
```diff
@@ -159,7 +159,7 @@ module S3
           allow(Time).to receive(:now).and_return(now)
           url, headers = obj.presigned_request(
             :get, expires_in: 86_400, request_payer: 'peccy')
-          expect(CGI.parse(url)).to eq(CGI.parse(
+          expect(URI.decode_www_form(url)).to eq(URI.decode_www_form(
```
mullermp, line 162: Are these actually equivalent? Is there any supporting documentation for these implementations?

mullermp (commented):
Do we need cgi as a dependency as well? Is it being removed as a bundled gem entirely, or are you saying that parts are being removed but cgi is still bundled?

mullermp:
@Earlopain I've performed a review. Do you know for sure this change is being made? It seems a bit premature since Ruby 3.5 is not released and the cgi work is not complete.

gems/aws-sdk-s3/spec/object/presigned_url_spec.rb
Earlopain: `CGI.parse`: https://docs.ruby-lang.org/en/3.4/CGI.html#method-c-parse
`URI.decode_www_form`: https://docs.ruby-lang.org/en/3.4/URI.html#method-c-decode_www_form

It treats it as `x-www-form-urlencoded`, which uses the query string format of `key=value&key2=value`.

They return data in a slightly different format. `CGI.parse` returns a hash, `URI.decode_www_form` an array of 2-element arrays. For tests to just check against this is fine, if it were exposed to users it would need a different solution.

Earlopain:
Thanks for taking a look.

> Do we need cgi as a dependency as well? Is it being removed as a bundled gem entirely, or are you saying that parts are being removed but cgi is still bundled?

Things like `CGI::Cookie` are being removed. All the escape/unescape methods will continue to work _without_ `cgi` as a dependency. So there is no need to declare it. It is being done a bit differently than the other bundled gems in the past (and rightly so, since every other gem uses these escape/unescape methods).

> Do you know for sure this change is being made? It seems a bit premature since Ruby 3.5 is not released and the cgi work is not complete.

Ruby already did the work to handle this change in their own code (`erb`, `bundler`, `rdoc`, others I can't think about). Of course I can't say it definitly, but I am very sure. But we can wait until 3.5-preview2 (preview1 was released in april) or something along the lines, it's no rush from me.

mullermp:
Thanks. I think it would be prudent to wait until Ruby 3.5 preview confirms the changes. I've prepared this release with changelogs.

mullermp (approved):

mullermp:
It looks like the upstream work was completed. We can possibly merge this before.

richardwang1124 (approved):

Earlopain:
Thanks!

PR #3248, opened 2025-05-13
Abort multi part download if the object is modified during download

richardwang1124:
Adds an etag [if match statement](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html#API_GetObject_RequestSyntax) to the get by parts/ranges functionality during multipart download of the transfer manager. In the event a file is changed mid operation, multipart download will abort and no file will be downloaded.

---

### DIFF
--- a/gems/aws-sdk-s3/CHANGELOG.md
+++ b/gems/aws-sdk-s3/CHANGELOG.md
@@ -1,5 +1,6 @@
 Unreleased Changes
 ------------------
+* Issue - Add if_match to multipart download.
 
 1.186.0 (2025-05-12)
 ------------------
--- a/gems/aws-sdk-s3/CHANGELOG.md
+++ b/gems/aws-sdk-s3/CHANGELOG.md
@@ -1,6 +1,6 @@
 Unreleased Changes
 ------------------
-* Issue - Add if_match to multipart download.
+* Issue - Abort multipart download if object is modified during download.
 
 1.186.0 (2025-05-12)
 ------------------
--- a/gems/aws-sdk-s3/spec/object/download_file_spec.rb
+++ b/gems/aws-sdk-s3/spec/object/download_file_spec.rb
@@ -237,7 +237,7 @@ module Aws
 
           client.stub_responses(:get_object, -> (ctx) {
             expect(ctx.params[:if_match]).to eq('test-etag')
-            Aws::S3::Errors::PreconditionFailed.new(nil, nil)
+            'PreconditionFailed'
           })
 
           thread = double(value: nil)
@@ -262,7 +262,7 @@ module Aws
 
           client.stub_responses(:get_object, -> (ctx) {
             expect(ctx.params[:if_match]).to eq('test-etag')
-            Aws::S3::Errors::PreconditionFailed.new(nil, nil)
+            'PreconditionFailed'
           })
 
           thread = double(value: nil)
--- a/gems/aws-sdk-s3/spec/object/download_file_spec.rb
+++ b/gems/aws-sdk-s3/spec/object/download_file_spec.rb
@@ -222,14 +222,23 @@ module Aws
         end
 
         it 'does not download object when ETAG does not match during multipart get by ranges' do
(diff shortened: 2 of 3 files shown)


gems/aws-sdk-s3/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 Unreleased Changes
 ------------------
+* Issue - Add if_match to multipart download.
```
jterapin: I would be more descriptive on what has been changed. The CHANGELOG entry should be easy to understand without knowing any context for the customer's POV

gems/aws-sdk-s3/spec/object/download_file_spec.rb
```diff
@@ -221,6 +221,42 @@ module S3
...
+            if_match: 'ETag'
+          }).and_raise(Aws::S3::Errors::PreconditionFailed.new(nil, nil))
+
+          expect(client).to_not receive(:write)
+
+          thread = double(value: nil)
+          expect(Thread).to receive(:new).and_yield.and_return(thread)
+
+          expect do
+            single_obj.download_file(path)
```
jterapin, line 247: Is there a reason why these tests uses different object sizes? Just wondering.

jterapin (approved):
LGTM

gems/aws-sdk-s3/spec/object/download_file_spec.rb
richardwang1124: Talked offline, different objects needed to test get by parts vs get by ranges.

gems/aws-sdk-s3/spec/object/download_file_spec.rb
```diff
@@ -221,6 +221,42 @@ module S3
           end.to raise_error(Aws::Errors::ChecksumError)
         end
 
+        it 'does not download object when ETAG does not match during multipart get by ranges' do
+          expect(client).to receive(:get_object).with({
```
mullermp: Use stub responses feature. You can use a proc stub that asserts etag is present, and have it raise the error without you needing to make it here. Where is the Etag value coming from? You could double assert this by stubbing the response of head object and ensuring you are passing it through to if_match if that makes sense.

gems/aws-sdk-s3/spec/object/download_file_spec.rb
```diff
@@ -221,6 +221,42 @@ module S3
...
+            bucket: 'bucket',
+            key: 'single',
+            range: "bytes=0-5242879",
+            if_match: 'ETag'
+          }).and_raise(Aws::S3::Errors::PreconditionFailed.new(nil, nil))
+
+          expect(client).to_not receive(:write)
```
mullermp: What is this? This method doesn't even exist does it?

gems/aws-sdk-s3/spec/object/download_file_spec.rb
```diff
@@ -221,6 +221,58 @@ module S3
...
+              :head_object,
+              content_length: 15 * one_meg,
+              parts_count: nil,
+              etag: 'test-etag'
+            )
+          )
+
+          client.stub_responses(:get_object, -> (ctx) {
+            expect(ctx.params[:if_match]).to eq('test-etag')
+            Aws::S3::Errors::PreconditionFailed.new(nil, nil)
```
mullermp: You don't need to initialize the errors. Just use the class.

gems/aws-sdk-s3/spec/object/download_file_spec.rb
```diff
@@ -221,6 +221,58 @@ module S3
           end.to raise_error(Aws::Errors::ChecksumError)
         end
 
+        it 'does not download object when ETAG does not match during multipart get by ranges' do
+          allow(client).to receive(:head_object).with({
```
mullermp, line 225: I guess this is fine seeing as other tests in this file are doing it.. but the tests here should be refactored (transfer manager makes these kind of obsolete anyway).

mullermp (approved):

PR #3249, opened 2025-05-22
Dropping Ruby 2.5 and 2.6 Runtime support

jterapin:
Context: [AWS SDK for Ruby Deprecating Ruby 2.5 and 2.6 runtime support](https://aws.amazon.com/blogs/developer/aws-sdk-for-ruby-deprecating-ruby-2-5-2-6-runtime-supports-and-future-compatibility/)

--- 


build_tools/services.rb
```diff
@@ -9,10 +9,10 @@ class ServiceEnumerator
     MANIFEST_PATH = File.expand_path('../../services.json', __FILE__)
 
     # Minimum `aws-sdk-core` version for new gem builds
-    MINIMUM_CORE_VERSION = "3.216.0"
+    MINIMUM_CORE_VERSION = "3.225.0"
 
     # Minimum `aws-sdk-core` version for new S3 gem builds
-    MINIMUM_CORE_VERSION_S3 = "3.216.0"
+    MINIMUM_CORE_VERSION_S3 = "3.225.0"
```
jterapin: Self-Reminder to double check before merging

Gemfile
```diff
@@ -17,7 +17,7 @@ if defined?(JRUBY_VERSION)
 end
 
 # protocol parsers
-gem 'json', '>= 2.4.0' # due to load_file support
+gem 'json', '>= 2.4.0' # due to load_file support, see: https://github.com/ruby/json/issues/696
 gem 'nokogiri', '>= 1.6.8.1'
```
jterapin: Curious if anyone know of any context for this version min....

bquorning: The specific version (`= 1.6.8.1`) was added in 6a0dd35acab3ff8b1e4e971ce7c3d5d452ad8195 as the last version being compatible with Ruby 1.x. The `>=` was added in e62c1d553cc9c16b17b79866aad7f9fbb9ceeaf8.

Since the original argument for listing that specific version number no longer applies, I would say it’s ok to just remove it.

mullermp: Yep, I think you can remove it too. I recall some issue with nokogiri years ago where the supported ruby version requirement of the latest version did not work with the syntax of the shipped gem (newer syntax used in a gem that said it supported older version).

Gemfile
```diff
@@ -44,7 +44,7 @@ end
 
 group :development do
   gem 'byebug', platforms: :ruby
-  gem 'rubocop', '1.28.0'
+  gem 'rubocop', '1.51.0'
```
mullermp: https://docs.rubocop.org/rubocop/compatibility.html#support-matrix

This can be any version currently. You can make this open ended, and we can add a constraint later.

build_tools/services.rb
mullermp: Keep in mind that STS, SSO, etc will also cause this to increase, so you cannot select a patch version, only a minor version.

mullermp (approved):

PR #3256, opened 2025-06-06
Prioritize JSON over CBOR during protocol selection

richardwang1124:
Currently, RPCv2 CBOR is prioritized over JSON during protocol selection. However, due to JSON's better performance and robust implementation compared to CBOR, protocol selection will now prioritize JSON over CBOR for all services.

---


build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/service.rb
```diff
@@ -4,11 +4,12 @@ module AwsSdkCodeGenerator
   class Service
     # Ordered priority list of supported protocols
     # api-gateway is a special case and is always first.
+    # Prioritize JSON over CBOR due to better performance.
     SUPPORTED_PROTOCOLS = %w[
       api-gateway
-      smithy-rpc-v2-cbor
       json_1.0
       json_1.1
+      smithy-rpc-v2-cbor
```
jterapin, line 12: Maybe we should put this lower than `rest-json` since it uses json builder/parser?

build_tools/aws-sdk-code-generator/spec/interfaces/protocols_spec.rb
```diff
@@ -24,14 +25,14 @@
       Object.send(:remove_const, :ProtocolsRpc)
     end
 
-    it 'selects smithy rpc v2 out of a list' do
-      SpecHelper.generate_service(['ProtocolsRpcJsonQuery'], multiple_files: false)
-      client = ProtocolsRpcJsonQuery::Client.new(stub_responses: true)
+    it 'selects smithy rpc v2 out of a list without json' do
```
jterapin: The service model only supports rpc-v2 or query so I think the description should be more like
> "selects smithy rpc v2 over query...."

jterapin (commented):
Looks good but let's have Matt review.

RE: `protocols_spec.rb`
Might be worth to take an overview look to see if any of the previous tests does not make sense to keep.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/service.rb
mullermp: I don't think that matters. JSON rpc and cbor are the same type of protocol - they do not use http bindings. I think it's fine here, above query.

build_tools/aws-sdk-code-generator/spec/fixtures/interfaces/protocols_rpc_query/api.json
```diff
@@ -0,0 +1,18 @@
+{
```
mullermp: This makes sense - are there any test models that are now obsolete that you can delete?

richardwang1124: Removed models for json+query and rpc+json since rpc+json+query covers all cases

jterapin (approved):
LGTM!

build_tools/aws-sdk-code-generator/spec/interfaces/protocols_spec.rb
```diff
@@ -9,12 +9,13 @@
...
 
+  # Prioritize JSON over CBOR due to better performance.
   context 'with smithy rpc protocol support' do
-    it 'selects smithy rpc v2 with priority' do
-      SpecHelper.generate_service(['ProtocolsRpcJson'], multiple_files: false)
-      client = ProtocolsRpcJson::Client.new(stub_responses: true)
-      expect(client.handlers).to include(Aws::RpcV2::Handler)
-      Object.send(:remove_const, :ProtocolsRpcJson)
+    it 'selects json with priority' do
+      SpecHelper.generate_service(['ProtocolsRpcJsonQuery'], multiple_files: false)
```
mullermp: The file name has the priority list. This implies rpc, json, query as the list. Please change to ProtocolsJsonRpcQuery.

build_tools/aws-sdk-code-generator/spec/interfaces/protocols_spec.rb
```diff
@@ -53,13 +47,6 @@
       stub_const('AwsSdkCodeGenerator::Service::SUPPORTED_PROTOCOLS', protocols)
     end
 
-    it 'selects json as a fallback' do
```
mullermp, line 56: I think we want to test query as a fallback.

build_tools/aws-sdk-code-generator/spec/interfaces/protocols_spec.rb
```diff
@@ -73,13 +60,6 @@
       Object.send(:remove_const, :ProtocolsRpcJsonQuery)
     end
 
-    it 'selects json out of a list' do
```
mullermp, line 76: Why delete this? I believe this should still pass - and these tests follow the SEP.

build_tools/aws-sdk-code-generator/spec/interfaces/protocols_spec.rb
```diff
@@ -24,18 +25,11 @@
...
-      SpecHelper.generate_service(['ProtocolsRpcJsonQuery'], multiple_files: false)
-      client = ProtocolsRpcJsonQuery::Client.new(stub_responses: true)
+    it 'selects smithy rpc v2 with priority over query' do
+      SpecHelper.generate_service(['ProtocolsRpcQuery'], multiple_files: false)
+      client = ProtocolsRpcQuery::Client.new(stub_responses: true)
       expect(client.handlers).to include(Aws::RpcV2::Handler)
-      Object.send(:remove_const, :ProtocolsRpcJsonQuery)
-    end
-
-    it 'selects json as fallback if smithy rpc v2 is not modeled' do
```
mullermp, line 34: I think we want to test query as a fallback instead of json here.

build_tools/aws-sdk-code-generator/spec/interfaces/protocols_spec.rb
richardwang1124: The first test case tests that JSON is selected with priority, so I thought that case would cover this case. If it follows the SEP I'll add this test back.

build_tools/aws-sdk-code-generator/spec/interfaces/protocols_spec.rb
richardwang1124: I added a new test case in the `without smithy rpc protocol support` context to test fallback to query.

build_tools/aws-sdk-code-generator/spec/interfaces/protocols_spec.rb
mullermp: If it's duplicated, it's fine to remove. We do want cases where rpc is supported or not however.

PR #3257, opened 2025-06-09
Ensure checksum is not recalculated on retries

mullermp:
This is already done because of how handlers are set up, but add a test to ensure.

### DIFF
--- a/gems/aws-sdk-core/spec/aws/plugins/checksum_algorithm_spec.rb
+++ b/gems/aws-sdk-core/spec/aws/plugins/checksum_algorithm_spec.rb
@@ -193,6 +193,27 @@ module Aws
           end.to raise_error(ArgumentError)
         end
 
+        it 'does not recompute checksums on a retry' do
+          original_checksum = nil
+          t = Tempfile.new
+          t.write('original')
+          t.rewind
+          client.stub_responses(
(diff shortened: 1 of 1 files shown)


gems/aws-sdk-core/spec/aws/plugins/checksum_algorithm_spec.rb
```diff
@@ -193,6 +193,27 @@ module Plugins
           end.to raise_error(ArgumentError)
         end
 
+        it 'does not recompute checksums on a retry' do
+          original_checksum = nil
+          t = Tempfile.new
+          t.write('original')
+          t.rewind
```
richardwang1124, line 200: Is this `rewind` necessary if you're already rewinding in the proc below?

richardwang1124 (approved):
LGTM. Left a question.

gems/aws-sdk-core/spec/aws/plugins/checksum_algorithm_spec.rb
mullermp: Sort of. It's necessary because if the body is re-read, we would want to ensure that it is the new value. However the body is not re-read (by design). It otherwise would make a checksum of an empty string which is still different but not really the ideal test.

PR #3263, opened 2025-06-30
IMDS/EC2Metadata Refactor

jterapin:
See #3255 

Contains IMDS/EC2Metadata refactoring w/o IMDS 2.1 specifications.

--- 

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core/ec2_metadata.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/ec2_metadata.rb
@@ -8,7 +8,7 @@ module Aws
   class EC2Metadata
     # Path for PUT request for token
     # @api private
-    METADATA_TOKEN_PATH = '/latest/api/token'.freeze
+    METADATA_TOKEN_PATH = '/latest/api/token'
 
     # Raised when the PUT request is not valid. This would be thrown if `token_ttl` is not an integer.
     # @api private
@@ -32,12 +32,10 @@ module Aws
     # @see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html#instance-metadata-transition-to-version-2
     #
     # @param [Hash] options
-    # @option options [Boolean] :disable_imds_v1 (false) Deprecated. The legacy EC2 Metadata Service v1
-    #   has been retired. Only IMDSv2 is supported.
     # @option options [Integer] :token_ttl (21600) The session token's TTL, defaulting to 6 hours.
-    # @option options [Integer] :retries (0) The number of retries for failed requests.
+    # @option options [Integer] :retries (3) The number of retries for failed requests on EC2 Metadata.
     # @option options [String] :endpoint ('http://169.254.169.254') The IMDS endpoint.
-    #   This option has precedence over the :endpoint_mode.
+    #   This option has precedence over the `:endpoint_mode`.
     # @option options [String] :endpoint_mode ('IPv4') The endpoint mode for  the instance metadata service.
     #   This is either 'IPv4'  ('http://169.254.169.254') or 'IPv6' ('http://[fd00:ec2::254]').
     # @option options [Integer] :port (80) The IMDS endpoint port.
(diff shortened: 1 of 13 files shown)


gems/aws-sdk-core/spec/aws/credential_resolution_chain_spec.rb
```diff
@@ -5,21 +5,11 @@
 module Aws
   describe 'Credential Resolution Chain' do
     let(:mock_credential_file) do
```
jterapin, line 7: Freed up bunch of lines since my rubocop is now configured correctly (yay)

gems/aws-sdk-core/lib/aws-sdk-core/ec2_metadata.rb
```diff
@@ -27,110 +26,130 @@ class MetadataNotFoundError < RuntimeError; end
...
+      @backoff = resolve_backoff(options[:backoff])
+      @endpoint = resolve_endpoint(options)
       @port = options[:port] || 80
-
       @http_open_timeout = options[:http_open_timeout] || 1
       @http_read_timeout = options[:http_read_timeout] || 1
       @http_debug_output = options[:http_debug_output]
 
+      @disable_imds_v1 = options[:disable_imds_v1].nil? || options[:disable_imds_v1]
+      @imds_v1_fallback = false # flag for if v2 flow fails, skip future attempts
```
jterapin, line 58: I'm re-introducing the v1 support here but I'm leaving them out from EC2 Metadata documentation- since these really matter when instance profile creds init.  

Set `true` as default disable_imds_v1 to prevent imds fallback for ec2 metadata init

gems/aws-sdk-core/lib/aws-sdk-core/ec2_metadata.rb
```diff
@@ -176,8 +192,12 @@ def http_put(connection, ttl)
         ]
       when 400
         raise TokenRetrievalError
-      when 403
+      when 403, 405
+        @imds_v1_fallback = true unless @disable_imds_v1
         raise RequestForbiddenError
+      when 404
+        @imds_v1_fallback = true unless @disable_imds_v1
+        raise MetadataNotFoundError
```
jterapin, line 200: Currently - the way we handle non-200 response from fetch token is slightly different depending on the context:
* Instance Profile Credentials: https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb#L236-L240
* EC2 Metadata: https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/lib/aws-sdk-core/ec2_metadata.rb#L177-L181

I went back to the SEP to see what the expected behavior around v1 fallback and the above error codes is where we care to set the fallback

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
```diff
@@ -153,204 +94,70 @@ def refresh
...
-      raise Non200Response
+    def empty_credentials?(creds)
+      creds.nil? || !creds.set?
     end
 
-    def token_set?
-      @token && !@token.expired?
-    end
+    def fetch_credentials
+      return '{}' if ec2_metadata_disabled?
```
jterapin, line 123: In the previous PR, we removed this but after taking a look deeper - I _think_ we need to keep it since customers could create this directly and expect a certain type of behavior.

When going through credential provider chain, we don't create instance profile at all when this is set

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
```diff
@@ -153,204 +94,70 @@ def refresh
...
-
-              # disable insecure flow if we couldn't get token
-              # and imds v1 is disabled
-              raise TokenRetrivalError if token.nil? && @disable_imds_v1
-
-              _get_credentials(conn, token)
-            end
-          end
-        rescue => e
+          retry_json_errors { Aws::Json.load(fetch_credentials) }
```
jterapin, line 99: Currently... atop of json error retries, there's two separate retries happening:
* fetch token: https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb#L227
* fetch creds: https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb#L205


Current state of PR does not combine json error retries with ec2 metadata retries - which means that there is more than 3 retries. 

Alternatively - we could take a step further to combine json retries with ec2 metadata retries: https://github.com/aws/aws-sdk-ruby/pull/3263/commits/e4153b314c28122627b5c610e5eae33440d9df83
* Key here is to set ec2 metadata retries, backoff to `0` - allowing any errors to bubble up to instance profile retries. 
* From InstanceProfile Credentials init, Customer will not be explicitly alter ec2 metadata retries or backoff unless they create a ec2 client and pass it in themselves.

PR #3266, opened 2025-07-02
Support bearer tokens through environment for Bedrock

mullermp:
Supports bearer tokens through ENV only for Bedrock. Scoped to two clients for now (others are planned). Precedence of the bearer token will also enforce this auth scheme preference (not yet implemented). A metric is added when used.

Alternate approach is to put logic in TokenProviderChain to read from ENV (perhaps a future generalized approach can do this).


gems/aws-sdk-bedrock/lib/aws-sdk-bedrock/plugins/bearer_authorization.rb
```diff
@@ -0,0 +1,34 @@
...
+        def call(context)
+          # This also sets the preferred auth scheme even if the code token has precedence.
+          context[:auth_scheme] = { 'name' => 'bearer' } if ENV['AWS_BEARER_TOKEN_BEDROCK']
+          with_metric { @handler.call(context) }
+        end
+
+        private
+
+        def with_metric(&block)
+          if ENV['AWS_BEARER_TOKEN_BEDROCK']
```
alextwoods: Possible nit - this should also check that the token wasn't overridden in code

gems/aws-sdk-bedrock/spec/client_spec.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+require_relative 'spec_helper'
+
+module Aws
+  module Bedrock
+    describe Client do
```
alextwoods, line 7: Could use test case for setting the business metric

gems/aws-sdk-bedrock/spec/client_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      end
+
+      it 'uses explicit config over the environment token' do
+        ENV['AWS_BEARER_TOKEN_BEDROCK'] = 'bedrock-token'
+        client = Client.new(
+          stub_responses: true,
+          token_provider: Aws::StaticTokenProvider.new('explicit-code-token')
+        )
+        resp = client.list_imported_models
+        expect(resp.context.http_request.headers['Authorization']).to eq('Bearer explicit-code-token')
```
alextwoods, line 44: maybe also assert business metric not set. here

alextwoods (approved):
Looks good overall - minor comments.

gems/aws-sdk-bedrock/lib/aws-sdk-bedrock/plugins/bearer_authorization.rb
mullermp: I'm clarifying this with Aaron.

gems/aws-sdk-bedrock/spec/client_spec.rb
mullermp: Related to above. Not sure if that's intended.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/sign.rb
```diff
@@ -13,8 +13,7 @@ class Sign < Seahorse::Client::Plugin
       option(:sigv4_region)
       option(:unsigned_operations, default: [])
 
-      supported_auth_types = %w[sigv4 bearer sigv4-s3express sigv4a none]
-      SUPPORTED_AUTH_TYPES = supported_auth_types.freeze
```
mullermp, line 17: Prior to this, I believe we added sigv4a if CRT was loaded, but now we just support everything.

gems/aws-sdk-core/lib/aws-sdk-core/token_provider.rb
```diff
@@ -6,6 +6,10 @@ module TokenProvider
     # @return [Token]
     attr_reader :token
 
+    # @api private
+    # Returns UserAgent metrics for tokens.
+    attr_accessor :metrics
```
mullermp, line 11: Copies the same Credentials metrics approach.

gems/aws-sdk-core/lib/aws-sdk-sts/presigner.rb
```diff
@@ -53,13 +53,9 @@ def get_caller_identity_presigned_url(options = {})
...
         )
-        endpoint = context.config.endpoint_provider
-                          .resolve_endpoint(endpoint_params)
+        endpoint = context.config.endpoint_provider.resolve_endpoint(endpoint_params)
         auth_scheme = Aws::Endpoints.resolve_auth_scheme(context, endpoint)
-
-        signer = Aws::Plugins::Sign.signer_for(
-          auth_scheme, context.config
-        )
+        signer = Aws::Plugins::Sign.signer_for(auth_scheme, context)
```
mullermp: While this is in a service gem, STS is part of core, so this should be a safe refactor.

gems/aws-sdk-core/spec/aws/endpoints_spec.rb
```diff
@@ -114,13 +114,6 @@ module Aws
           end
 
           context 'sigv4a defaults' do
-            before do
```
mullermp, line 117: This was when we needed to mock CRT loaded.

gems/aws-sdk-core/spec/sigv4_helper.rb
```diff
@@ -1,29 +0,0 @@
-module Sigv4Helper
```
mullermp, line 1: Made this more generic, but most of this code was mocking CRT signing (for sigv4a) and stopped working when I made the refactor.

gems/aws-sigv4/lib/aws-sigv4/signer.rb
```diff
@@ -111,41 +111,42 @@ class Signer
...
       def initialize(options = {})
         @service = extract_service(options)
         @region = extract_region(options)
         @credentials_provider = extract_credentials_provider(options)
-        @unsigned_headers = Set.new((options.fetch(:unsigned_headers, [])).map(&:downcase))
+        @unsigned_headers = Set.new(options.fetch(:unsigned_headers, []).map(&:downcase))
         @unsigned_headers << 'authorization'
         @unsigned_headers << 'x-amzn-trace-id'
         @unsigned_headers << 'expect'
+        @signing_algorithm = options.fetch(:signing_algorithm, 'sigv4').to_s
```
mullermp: This used to be a symbol but I made it a string, and call to_s for compatibility. s3 express symbolized naming made this weird.

gems/aws-sigv4/lib/aws-sigv4/signer.rb
```diff
@@ -165,12 +166,22 @@ def initialize(options = {})
...
 
-      # @return [Set<String>] Returns a set of header names that should not be signed.
-      #   All header names have been downcased.
+      # @return [Set<String>]
       attr_reader :unsigned_headers
 
-      # @return [Boolean] When `true` the `x-amz-content-sha256` header will be signed and
-      #   returned in the signature headers.
+      # @return [String]
+      attr_reader :signing_algorithm
```
mullermp: Added these accessors for our test helper but they should just exist anyway.

alextwoods (approved):
Nice - LGTM

richardwang1124 (approved):

PR #3267, opened 2025-07-03
Refactor IMDS credentials provider

jterapin:

### DIFF
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Skip `Aws::InstanceProfileCredentials` instantiation when `ENV['AWS_EC2_METADATA_DISABLED']` is set to `true` in the credential resolution chain.
+
 * Issue - Refactor `InstanceProfileCredentials` to improve code clarity and documentation. 
 
 3.226.2 (2025-07-01)
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Document retry behaviors in `InstanceProfileCredentials`.
+
 3.226.2 (2025-07-01)
 ------------------
 
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,7 +1,7 @@
 Unreleased Changes
 ------------------
 
-* Issue - Refactor `InstanceProfileCredentials` to be simpler and increase network retries to be more resilient.
+* Issue - Document retry behaviors in `InstanceProfileCredentials`.
 
 3.226.2 (2025-07-01)
 ------------------
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,7 +1,7 @@
 Unreleased Changes
 ------------------
 
-* Issue - Document retry behaviors in `InstanceProfileCredentials`.
+* Issue - Refactor `InstanceProfileCredentials` to be simpler and increase network retries to be more resilient.
 
 3.226.2 (2025-07-01)
 ------------------
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,7 +1,7 @@
 Unreleased Changes
 ------------------
 
-* Issue - Document retry behaviors in `InstanceProfileCredentials`.
+* Issue - Refactor `InstanceProfileCredentials` to improve code clarity and documentation. 
 
 3.226.2 (2025-07-01)
 ------------------
--- a/gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
@@ -10,15 +10,9 @@ module Aws
   #     ec2 = Aws::EC2::Client.new(credentials: instance_credentials)
   #
   # ## Retries
-  # When initialized from the default credential chain, this provider defaults to `0` retries.
-  # Breakdown of retries is as follows:
   #
-  #  * **Configurable retries** (defaults to `1`): these retries handle errors when communicating
-  #     with the IMDS endpoint. There are two separate retry mechanisms within the provider:
-  #       * Entire token fetch and credential retrieval process
-  #       * Token fetching
-  #  * **JSON parsing retries**: Fixed at 3 attempts to handle cases when IMDS returns malformed JSON
-  #     responses. These retries are separate from configurable retries.
+  # When initialized from the default credential chain, this provider defaults to `0` retries.
+  # Otherwise, it defaults to 3 retries and will retry network and parsing errors.
   #
   # @see https://docs.aws.amazon.com/sdkref/latest/guide/feature-imds-credentials.html IMDS Credential Provider
   class InstanceProfileCredentials
@@ -28,24 +22,15 @@ module Aws
     # @api private
     class Non200Response < RuntimeError; end
 
+    # @deprecated Unfortunate spelling name.
     # @api private
     class TokenRetrivalError < RuntimeError; end
 
     # @api private
-    class TokenExpiredError < RuntimeError; end
+    class TokenRetrievalError < TokenRetrivalError; end
 
-    # These are the errors we trap when attempting to talk to the instance metadata service.
-    # Any of these imply the service is not present, no responding or some other non-recoverable error.
     # @api private
-    NETWORK_ERRORS = [
-      Errno::EHOSTUNREACH,
-      Errno::ECONNREFUSED,
-      Errno::EHOSTDOWN,
-      Errno::ENETUNREACH,
-      SocketError,
-      Timeout::Error,
-      Non200Response
-    ].freeze
+    class TokenExpiredError < RuntimeError; end
 
     # Path base for GET request for profile and credentials
     # @api private
@@ -56,9 +41,10 @@ module Aws
     METADATA_TOKEN_PATH = '/latest/api/token'.freeze
 
     # @param [Hash] options
-    # @option options [Integer] :retries (1) Number of times to retry when retrieving credentials.
+    # @option options [Integer] :retries (3) Number of times to retry when retrieving credentials. Defaults to 0 when
+    #   resolving from the default credential chain.
     # @option options [String] :endpoint ('http://169.254.169.254') The IMDS endpoint. This option has precedence
-    #    over the `:endpoint_mode`.
+    #   over the `:endpoint_mode`.
     # @option options [String] :endpoint_mode ('IPv4') The endpoint mode for the instance metadata service. This is
     #   either 'IPv4' (`169.254.169.254`) or IPv6' (`[fd00:ec2::254]`).
     # @option options [Boolean] :disable_imds_v1 (false) Disable the use of the legacy EC2 Metadata Service v1.
@@ -84,20 +70,19 @@ module Aws
       @http_read_timeout = options[:http_read_timeout] || 1
       @http_debug_output = options[:http_debug_output]
       @port = options[:port] || 80
-      @retries = options[:retries] || 1
+      @retries = options[:retries] || 3
       @token_ttl = options[:token_ttl] || 21_600
 
-      @async_refresh = false
-      # Flag for if v2 flow fails, skip future attempts
       @imds_v1_fallback = false
-      @no_refresh_until = nil
       @token = nil
+      @no_refresh_until = nil
+
+      @async_refresh = false
       @metrics = ['CREDENTIALS_IMDS']
       super
     end
 
-    # @return [Integer] Number of times to retry when retrieving credentials from the instance metadata service.
-    #   Defaults to 0 when resolving from the default credential chain.
+    # @return [Integer]
     attr_reader :retries
 
     private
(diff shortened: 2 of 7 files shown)


gems/aws-sdk-core/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Document retry behaviors in `InstanceProfileCredentials`.
```
mullermp: It does more than this doesn't it? It still refactors credential fetching.

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
```diff
@@ -4,11 +4,23 @@
...
-  # An auto-refreshing credential provider that loads credentials from
-  # EC2 instances.
+  # An auto-refreshing credential provider that loads credentials from EC2 instances.
   #
   #     instance_credentials = Aws::InstanceProfileCredentials.new
   #     ec2 = Aws::EC2::Client.new(credentials: instance_credentials)
+  #
+  # ## Retries
+  # When initialized from the default credential chain, this provider defaults to `0` retries.
+  # Breakdown of retries is as follows:
```
mullermp, line 14: Much of this information doesn't matter to customers. I think you just want to say that it defaults to 0 from the chain or if you initialize an instance it has a default of 1 (or 3). Customer doesn't care about malformed json or network conditions.

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
```diff
@@ -46,97 +56,88 @@ class TokenExpiredError < RuntimeError; end
...
-    # @option options [String] :endpoint ('http://169.254.169.254') The IMDS
-    #   endpoint. This option has precedence over the :endpoint_mode.
-    # @option options [String] :endpoint_mode ('IPv4') The endpoint mode for
-    #   the instance metadata service. This is either 'IPv4' ('169.254.169.254')
-    #   or 'IPv6' ('[fd00:ec2::254]').
-    # @option options [Boolean] :disable_imds_v1 (false) Disable the use of the
-    #  legacy EC2 Metadata Service v1.
-    # @option options [String] :ip_address ('169.254.169.254') Deprecated. Use
-    #   :endpoint instead. The IP address for the endpoint.
+    # @option options [Integer] :retries (1) Number of times to retry when retrieving credentials.
```
mullermp, line 59: We should consider defaulting this to 3. What are the implications?

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
```diff
@@ -46,97 +56,88 @@ class TokenExpiredError < RuntimeError; end
...
       @http_debug_output = options[:http_debug_output]
-      @backoff = backoff(options[:backoff])
+      @port = options[:port] || 80
+      @retries = options[:retries] || 1
       @token_ttl = options[:token_ttl] || 21_600
-      @token = nil
-      @no_refresh_until = nil
+
       @async_refresh = false
+      # Flag for if v2 flow fails, skip future attempts
```
mullermp: Comment is kind of useless - instance variable name is pretty clear.

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
```diff
@@ -46,97 +56,88 @@ class TokenExpiredError < RuntimeError; end
...
+      @no_refresh_until = nil
+      @token = nil
       @metrics = ['CREDENTIALS_IMDS']
       super
     end
 
-    # @return [Integer] Number of times to retry when retrieving credentials
-    #   from the instance metadata service. Defaults to 0 when resolving from
-    #   the default credential chain ({Aws::CredentialProviderChain}).
+    # @return [Integer] Number of times to retry when retrieving credentials from the instance metadata service.
```
mullermp: I think you can just have the `@return [Integer]` as you've already documented what this value is and what it can return. It's also pretty clear what retries means.

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
```diff
@@ -153,98 +154,80 @@ def refresh
...
+        warn_expired_credentials
+      else
+        # credentials are empty or successfully retrieved, update them
+        update_credentials(new_creds)
       end
     end
 
-    def get_credentials
+    def retrieve_credentials
+      return '{}' if ec2_metadata_disabled?
```
mullermp: I don't think this belongs here? We still want the credentials provider chain change I think - where we check disabled in the chain and only make an instance of this class when not disabled.

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
```diff
@@ -153,98 +154,80 @@ def refresh
...
-                @no_refresh_until = Time.now + refresh_offset
-                warn_expired_credentials
-              end
-            end
+      new_creds =
+        begin
+          # Retry loading credentials up to 3 times is the instance metadata
+          # service is responding but is returning invalid JSON documents
+          # in response to the GET profile credentials call.
+          retry_errors([Aws::Json::ParseError], max_retries: 3) do
```
mullermp, line 184: Now that we aren't using the EC2 Metadata client, I think we should consider consolidating this. You can use `retry_errors(NETWORK_ERRORS, Aws::Json::ParseError, max_retries: retries)` to lump all of the error cases into a retriable block, making this simpler code.

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
```diff
@@ -153,98 +154,80 @@ def refresh
...
-          warn("Error retrieving instance profile credentials: #{e}")
-          '{}'
         end
+      rescue StandardError => e
+        warn("Error retrieving instance profile credentials: #{e}")
+        '{}'
       end
     end
 
+    def skip_token?
```
mullermp: Not good naming. I think treat imds_v1_fallback and token expired as two different cases. It's easier to refactor that way too.

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
```diff
@@ -153,98 +154,80 @@ def refresh
...
-            end
+      begin
+        retry_errors(NETWORK_ERRORS, max_retries: @retries) do
+          open_connection do |conn|
+            # attempt to fetch token to start secure flow first
+            # and rescue to failover
+            fetch_token(conn) unless skip_token?
+
+            # disable insecure flow if we couldn't get token and imds v1 is disabled
+            raise TokenRetrivalError if @token.nil? && @disable_imds_v1
```
mullermp, line 213: This is unfortunate error naming. You can consider changing this - technically breaking but I don't know if customers are relying on it. A safe way could be to create a `TokenRetrievalError` that inherits from `TokenRetrivalError` then raise the new error. Old usage error catching would still work. Or, decide to do nothing.

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
```diff
@@ -153,98 +154,80 @@ def refresh
...
-          @token = Token.new(token_value, ttl, created_time) if token_value && ttl
-        end
-      end
+      created_time = Time.now
+      token_value, ttl = http_put(conn)
+      @token = Token.new(token_value, ttl, created_time) if token_value && ttl
     rescue *NETWORK_ERRORS
       # token attempt failed, reset token
       # fallback to non-token mode
-      @token = nil
```
mullermp, line 239: I think you still need this right? If token refresh failed (and there was previously a token) then we need to clear it.

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
```diff
@@ -153,98 +154,80 @@ def refresh
...
+          retry_errors([Aws::Json::ParseError], max_retries: 3) do
+            Aws::Json.load(retrieve_credentials.to_s)
           end
+        rescue Aws::Json::ParseError
+          raise Aws::Errors::MetadataParserError
         end
-      rescue Aws::Json::ParseError
-        raise Aws::Errors::MetadataParserError
+
+      if !empty_credentials?(@credentials) && (!new_creds['AccessKeyId'] || new_creds['AccessKeyId'].empty?)
```
mullermp: Couldn't you also check empty_credentials? with new_creds? set? will check if access key id is not nil or empty (and also secret access key).

gems/aws-sdk-core/spec/aws/instance_profile_credentials_spec.rb
```diff
@@ -641,21 +578,22 @@ module Aws
...
       before(:each) do
-        stub_request(:put, "http://169.254.169.254#{token_path}")
+        stub_request(:put, ipv4_endpoint + token_path)
           .to_return(
             status: 200,
             body: "my-token\n",
             headers: { 'x-aws-ec2-metadata-token-ttl-seconds' => '21600' }
           )
-        stub_request(:get, "http://169.254.169.254#{path}")
+        stub_request(:get, ipv4_endpoint + path)
```
mullermp, line 522: Rubocop typically complains about using interpolation here like you did below with ` "#{ipv4_endpoint}#{path}profile-name"`. You've got many offenders above. You also have a lot of usage of ipv4 endpoint + path or token_path so you can consider just making a variable for those, like `ipv4_endpoint_token_path` or `ipv4_endpoint_credentials_path`.

gems/aws-sdk-core/spec/aws/instance_profile_credentials_spec.rb
```diff
@@ -171,31 +138,21 @@ module Aws
...
-          .and_return('false')
+        allow_any_instance_of(Aws::SharedConfig).to receive(:ec2_metadata_v1_disabled).and_return('false')
         ENV['AWS_EC2_METADATA_V1_DISABLED'] = 'false'
-        subject = InstanceProfileCredentials.new(
-          disable_imds_v1: disable_imds_v1
-        )
-        expect(subject.instance_variable_get(:@disable_imds_v1))
-          .to eq disable_imds_v1
+        subject = InstanceProfileCredentials.new(disable_imds_v1: disable_imds_v1)
+        expect(subject.instance_variable_get(:@disable_imds_v1)).to eq disable_imds_v1
```
mullermp: We should surface all of these accessors. Checking instance variable state is not great.

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
jterapin: This was carried over but yeah, I can delete it.

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
jterapin: After few considerations, I decided against making any changes to the existing retries behavior. We could streamline retries but I personally feel sketchy about making any changes and I rather tread on caution without worrying about any unintended side effects. However in new MV, we should streamline the retries and remove IMDS v1 support completely.

gems/aws-sdk-core/CHANGELOG.md
jterapin: I felt that the most of the refactors was just re-arranging the existing code into a easier format to digest but not necessarily changing any behaviors so I felt that I didn't need to highlight anything significant here.

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
jterapin: When I think of retries, I expect retries that I passed into to explicitly happen and NOT expect any other set of retries to occur. Therefore, I felt that this was worth highlighting.

Yeah, we could refactor to make it more streamline but I would like to avoid any changes to current retrying behavior.

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
jterapin: For customers who create this provider directly, they would be affected by any changes here. I rather leave it as it is. We can update it in next MV

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
jterapin: This was carried over but yes, I can delete it since we already expanded on it above.

gems/aws-sdk-core/spec/aws/instance_profile_credentials_spec.rb
jterapin: I could pick out the ones that is worth resurfacing, like endpoint and disable_imds_v1

gems/aws-sdk-core/spec/aws/instance_profile_credentials_spec.rb
jterapin: I didn't get any complains from rubocop when I ran it. But yeah, I could do the latter suggestions.

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
jterapin: `fetch_token` will only happen when either token is not set or expired (aside from imds v1 disable) so I didn't feel the need to set it `nil`  - but I can clear the state anyway

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
jterapin: Probably not worth having this method honestly

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
jterapin: Yeah, I meant to mention this but I think I'll leave it as it is.

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
jterapin: I could bring that over from previouis PR but I think we should keep this around for customers who uses instance profile creds directly.

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
jterapin: If I'm following you correctly - the returned new creds are in a parsed hash format - not an instance of Credentials.

gems/aws-sdk-core/CHANGELOG.md
mullermp: Refactoring is still a change. If we break something, this is an untrackable changelog.

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
mullermp: If you think so, but I'd rather not bog down documentation with technical details. From the customer POV, they just want to know that retries happen and how many. We don't really care to surface the exact cases.

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
mullermp: If you want to leave it that's fine, but I think it's perfectly reasonable to change within the current major version.

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
mullermp: I think this would greatly simplify the logic and makes more sense than nested retry cases. I think the retry count of 1 + json parse error of 3 results in 6 attempts total, where instead we can just have 3 retries (4 attempts) total - I can make the change if you want.

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
mullermp: Ah yes. Still weird - maybe empty_credentials? should check hash version and you can simply use `@credentials&.set?`

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
mullermp: I think we discussed this before - I don't think this makes sense if you are creating this instance directly. That environment variable is intended for the chain and not when explicitly creating providers through code.

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
mullermp: Yes.

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
mullermp: As long as you ensure fetch_token is not called in imds v1 cases or if token is not expired then it's fine.

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
jterapin: Right but I would argue that there's certainly more than default retries are happening here.

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
```diff
@@ -257,7 +258,7 @@ def token_set?
       @token && !@token.expired?
     end
 
-    def _metadata_disabled?
+    def ec2_metadata_disabled?
```
mullermp: If we're committing to the chain approach, we can remove this here. Otherwise, revert the chain changes and I can include it in the follow up. Or, you can leave it, and have it do double work.

mullermp (approved):

gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
jterapin: I'll just get rid of it here and add a changelog entry.

PR #3268, opened 2025-07-07
Support auth scheme preference config

mullermp:
Support auth scheme preference config using `ENV['AWS_AUTH_SCHEME_PREFERENCE']` or `auth_scheme_preference` shared config. Configuration values are an array in code and a comma separated string in env/shared config and do not use name spaces. Invalid values are ignored. If an auth scheme is resolved and preferred, the credentials/token is expected. This differs from original design because v3 does not resolve auth per operation but rather on client initialize.


gems/aws-sdk-core/lib/aws-sdk-core/endpoints.rb
```diff
@@ -42,6 +51,16 @@ def resolve_auth_scheme(context, endpoint)
 
       private
 
+      def endpoint_auth_scheme_preference(auth_schemes, preferred_auth)
+        ordered_auth = preferred_auth.each_with_object([]) do |pref, list|
+          next unless ENDPOINT_AUTH_PREFERENCE_MAP.key?(pref)
+
+          ENDPOINT_AUTH_PREFERENCE_MAP[pref].each { |name| list << { 'name' => name } }
```
richardwang1124, line 58: Why is the list adding hashes of `name` to auth scheme names rather than just being a list of names? Is `auth_schemes` a list of `name`, `auth_scheme_name` key pairs?

richardwang1124 (commented):
Overall approach looks good to me! Left a question.

gems/aws-sdk-core/lib/aws-sdk-core/endpoints.rb
mullermp: The endpoint auth schemes model has a list of maps in the same way, so I transformed the preference list to be the same, then added them together. This overall made the logic and change simpler.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/credentials_configuration.rb
```diff
@@ -17,61 +17,61 @@ class CredentialsConfiguration < Seahorse::Client::Plugin
       option(:profile,
         doc_default: 'default',
         doc_type: String,
-        docstring: <<-DOCS)
-Used when loading credentials from the shared credentials file
-at HOME/.aws/credentials.  When not specified, 'default' is used.
+        docstring: <<~DOCS)
+          Used when loading credentials from the shared credentials file at HOME/.aws/credentials.
```
jterapin:
```suggestion
          Used when loading credentials from the shared credentials file at `HOME/.aws/credentials`.
```

gems/aws-sdk-core/lib/aws-sdk-core/plugins/credentials_configuration.rb
```diff
@@ -17,61 +17,61 @@ class CredentialsConfiguration < Seahorse::Client::Plugin
...
+          * ENV['AWS_ACCESS_KEY_ID'], ENV['AWS_SECRET_ACCESS_KEY'],
+            ENV['AWS_SESSION_TOKEN'], and ENV['AWS_ACCOUNT_ID']
+          * `~/.aws/credentials`
+          * `~/.aws/config`
+          * EC2/ECS IMDS instance profile - When used by default, the timeouts
+            are very aggressive. Construct and pass an instance of
+            `Aws::InstanceProfileCredentials` or `Aws::ECSCredentials` to
+            enable retries and extended timeouts. Instance profile credential
+            fetching can be disabled by setting ENV['AWS_EC2_METADATA_DISABLED']
+            to true.
```
jterapin: The backticks are not rendering correctly so we need to add a line in-between to bullet pts

```suggestion
          * `Aws.config[:credentials]`
          
          * The `:access_key_id`, `:secret_access_key`, `:session_token`, and
            `:account_id` options.
            
          * `ENV['AWS_ACCESS_KEY_ID']`, `ENV['AWS_SECRET_ACCESS_KEY']`,
            `ENV['AWS_SESSION_TOKEN']`, and `ENV['AWS_ACCOUNT_ID']`.
            
          * `~/.aws/credentials`
          
          * `~/.aws/config`
          
          * EC2/ECS IMDS instance profile - When used by default, the timeouts
            are very aggressive. Construct and pass an instance of
            `Aws::InstanceProfileCredentials` or `Aws::ECSCredentials` to
            enable retries and extended timeouts. Instance profile credential
            fetching can be disabled by setting `ENV['AWS_EC2_METADATA_DISABLED']`
            to `true`.
```

jterapin (approved):
🚢✨

PR #3273, opened 2025-07-21
Add bigdecimal to core gemspec

### LINKED ISSUE #3271: cannot load such file -- bigdecimal in Ruby 3.4+
### Describe the bug

The current version of aws-sdk-dynamodb causes our build to fail in Ruby 3.4 due to a missing dependency of bigdecimal. There was a previous attempt to address this issue but it appears to have caused an issue for a major sdk user and was subsequently reverted. 

Opening this bug request to create a central place to discuss possible solutions.  

### Regression Issue

### Expected Behavior

All required gems are included in the gemspec file.

### Current Behavior

LoadError: cannot load such file -- bigdecimal when attempting to use aws-sdk-dynamodb in a ruby application targeting Ruby 3.4.

### Reproduction Steps

Create a an application in Ruby 3.4+. Include the aws-sdk-dynamodb gem. Build the application. 

### Possible Solution

I don't have the full context of why this failed for the existing customer. I believe adding the dependency will cause bundle install to pull the latest version even if they are on an older version of Ruby where it is still included as a default gem.

For versions of Ruby < 3.4 it appears that you can prevent bundler from attempting to install bigdecimal and using the default gem bundled with Ruby by doing the following:

- add the bigdecimal default gem version packaged with Ruby (3.1.3 for Ruby 3.2) in the Gemfile.lock
- use `bundle install --prefer-local` to prevent it from fetching remote ruby gems that exist locally

Those two steps allowed me to specify the bigdecimal dependency to be pulled in for Ruby 3.4, or to avoid pulling in any external dependency in Ruby 3.2.

Added missing dependencies: https://github.com/aws/aws-sdk-ruby/pull/2986
Reverted and removed dependencies: https://github.com/aws/aws-sdk-ruby/pull/2987
Related issue opened with Ruby: https://github.com/ruby/bigdecimal/issues/289
Bigdecimal history as a default and bundled gem: https://stdgems.org/bigdecimal/

aws-sdk-dynamodb

Ruby 3.4

mullermp:
Closes #3271

Adds back `bigdecimal` as a required dependency of `aws-sdk-core` (for cbor) and also dynamodb transitively. We can consider pushing this dependency only to dynamodb as cbor is not used currently, limiting blast radius.

Originally, we pushed this change but reverted it, because of customer issues installing the bigdecimal dependency on systems without native extension gems. Bundled `bigdecimal` in ruby can be used with `bundle install --prefer-local` but it would require a change for those customers in their workflow, so there is risk to breaking. Customers can also pin their gem versions until the flow is fixed.


Gemfile
```diff
@@ -3,25 +3,22 @@
...
 gem 'http-2'
+gem 'irb'
 gem 'jmespath'
-if defined?(JRUBY_VERSION)
-  # get the latest jruby-openssl to support sigv4a
-  # see: https://github.com/jruby/jruby-openssl/issues/30
-  gem 'jruby-openssl'
-end
+gem 'jruby-openssl' if defined?(JRUBY_VERSION)
+gem 'rdoc'
```
jterapin, line 15: for context purposes: where do we use rdoc gem?

jterapin (approved):
Looks good.

Gemfile
mullermp: I'm not sure - we got the warning. Maybe from yard.

evman182:
@jterapin  Can this be pinned to a specific major bigdecimal version? The fact that it wasn't means that when v4 of the bigdecimal gem was released, the ruby lambda runtime builds were based off of v4, breaking existing Lambdas that were dependent on v3 (as there was a runtime error due to use of both versions within the Lambda)

jterapin:
@evman182 Hi! Thanks for raising this and sorry for the trouble. To figure out the best path forward, could I get more details?

* Is the error surfacing from SDK Code or application code that has direct dependency on `bigdecimal` V3? If from SDK code, could you share a snippet of the stack trace?
* Which Lambda runtime version and deployment type (managed zip vs. container image) are you using?

Adding a strict major-version pin to the SDK gem has broader implications since it can conflict with customer Gemfiles and make updates harder over time. That being said, we generally only add constraints when we know the SDK itself is incompatible with a given dependency version.

evman182:
1) Here's the error/stack:
```
{
    "errorMessage": "You have already activated bigdecimal 4.0.1, but your Gemfile requires bigdecimal 3.1.8. Prepending `bundle exec` to your command may solve this.",
    "errorType": "Init<Gem::LoadError>",
    "stackTrace": [
        "/var/lang/lib/ruby/site_ruby/3.3.0/bundler/runtime.rb:328:in `check_for_activated_spec!'",
        "/var/lang/lib/ruby/site_ruby/3.3.0/bundler/runtime.rb:25:in `block in setup'",
        "/var/lang/lib/ruby/site_ruby/3.3.0/bundler/spec_set.rb:224:in `each'",
        "/var/lang/lib/ruby/site_ruby/3.3.0/bundler/spec_set.rb:224:in `each'",
        "/var/lang/lib/ruby/site_ruby/3.3.0/bundler/runtime.rb:24:in `map'",
        "/var/lang/lib/ruby/site_ruby/3.3.0/bundler/runtime.rb:24:in `setup'",
        "/var/lang/lib/ruby/site_ruby/3.3.0/bundler.rb:165:in `setup'",
        "/var/lang/lib/ruby/site_ruby/3.3.0/bundler/setup.rb:32:in `block in <top (required)>'",
        "/var/lang/lib/ruby/site_ruby/3.3.0/bundler/ui/shell.rb:177:in `with_level'",
        "/var/lang/lib/ruby/site_ruby/3.3.0/bundler/ui/shell.rb:122:in `silence'",
        "/var/lang/lib/ruby/site_ruby/3.3.0/bundler/setup.rb:32:in `<top (required)>'",
        "<internal:/var/lang/lib/ruby/site_ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:139:in `require'",
        "<internal:/var/lang/lib/ruby/site_ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:139:in `require'",
        "/var/task/designer_loader.rb:4:in `<top (required)>'",
        "<internal:/var/lang/lib/ruby/site_ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:139:in `require'",
        "<internal:/var/lang/lib/ruby/site_ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:139:in `require'",
        "/var/runtime/gems/aws_lambda_ric-3.1.3/lib/aws_lambda_ric.rb:46:in `run'",
        "/var/runtime/gems/aws_lambda_ric-3.1.3/lib/aws_lambda_ric.rb:219:in `bootstrap_handler'",
        "/var/runtime/gems/aws_lambda_ric-3.1.3/lib/aws_lambda_ric.rb:201:in `start'",
        "/var/runtime/index.rb:4:in `<main>'"
    ]
}
```
2. Managed zip. Here's the INIT line from the logs:

INIT_START Runtime Version: ruby:3.3.mainline.v94	Runtime Version ARN: arn:aws:lambda:us-east-1::runtime:58713c1f627a6d04dfda3ea4e11fae1cbca3ac46a4becad26f739c38911423d9

jterapin:
@evman182  

Thanks for the details and stack trace. I've opened an issue with the Lambda runtime team to get their guidance on this: https://github.com/aws/aws-lambda-ruby-runtime-interface-client/issues/65

PR #3275, opened 2025-07-24
S3 Multipart upload/download update

jterapin:
Objectives:
* Ensure that temp file is deleted when multipart download fails (and do not overrwrite if a file exists)
* Added various validation checks
---


gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -3,10 +3,13 @@
 require 'pathname'
 require 'thread'
 require 'set'
-require 'tmpdir'
 
 module Aws
   module S3
+    # TODO - move to another location
+    # Error raised when file download operations fail
```
mullermp: You can probably make this similar to MultipartUploadError and make it a MultipartDownloadError and define it at the same scope. This is the only time this is raised, right? What errors are raised if the file cannot be written to, or it's a single request download?

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -26,56 +29,56 @@ def download(destination, options = {})
         @mode = options[:mode] || 'auto'
         @thread_count = options[:thread_count] || THREAD_COUNT
         @chunk_size = options[:chunk_size]
-        @params = {
-          bucket: options[:bucket],
-          key: options[:key]
-        }
-        @params[:version_id] = options[:version_id] if options[:version_id]
+        @params = set_params(options)
```
mullermp: I think we weren't doing this correctly. The download_file method says:
```
      # @param [Hash] options
      #   Additional options for {Client#get_object} and #{Client#head_object}
      #   may be provided.
```

however it looks like we only cherrypick these 3. I think instead, we probably want to delete the other options like mode, thread count, chunk size, etc, then the remaining options are all API options.

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -151,50 +128,61 @@ def multithreaded_get_by_parts(n_parts, total_size, etag)
...
                       progress.call(part.part_number, bytes, total)
                     end
                 end
                 resp = @client.get_object(part.params)
+
+                if part.params[:range]
+                  range = resp.content_range.split(' ').last.split('/').first
+                  expected_range = part.params[:range].split('=').last
+                  raise FileDownloadError, 'file download integrity checked failed' unless expected_range == range
+                end
```
mullermp: What is this being parsed and what is compared? Perhaps pulling this out into a method can better document what is happening.

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -151,50 +128,61 @@ def multithreaded_get_by_parts(n_parts, total_size, etag)
...
+
                 write(resp)
                 if @on_checksum_validated && resp.checksum_validated
                   @on_checksum_validated.call(resp.checksum_validated, resp)
                 end
+                mutex.synchronize { total_requests += 1 }
```
mullermp: I think the mutex may not be necessary here, because at no point are you retrieving total_requests until after all the threads have completed.

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
```diff
@@ -67,19 +57,23 @@ def initiate_upload(options)
...
       end
 
-      def complete_upload(upload_id, parts, options)
+      def complete_upload(upload_id, parts, source, options)
         @client.complete_multipart_upload(
           **complete_opts(options).merge(
             upload_id: upload_id,
-            multipart_upload: { parts: parts }
+            multipart_upload: { parts: parts },
+            mpu_object_size: File.size(source)
```
mullermp, line 65: For my understanding, why is this needed?

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
```diff
@@ -136,28 +122,23 @@ def has_checksum_key?(keys)
...
       def upload_part_opts(options)
-        UPLOAD_PART_OPTIONS.inject({}) do |hash, key|
-          if options.key?(key)
-            # don't pass through checksum calculations
-            hash[key] = options[key] unless checksum_key?(key)
-          end
-          hash
+        UPLOAD_PART_OPTIONS.each_with_object({}) do |key, hash|
+          # don't pass through checksum calculations
+          hash[key] = options[key] if options.key?(key) && !checksum_key?(key)
```
mullermp, line 140: I remember adding this check, but is it still necessary?

gems/aws-sdk-s3/spec/object/download_file_spec.rb
```diff
@@ -184,8 +184,13 @@ module S3
             part_number: 1
           }).exactly(1).times
 
-          client.stub_responses(:get_object, -> (_ctx) {
-            { body: 'body', content_range: 'bytes 0-4/4' }
+          client.stub_responses(:get_object, -> (context) {
```
mullermp: You are missing validation cases - I'm assuming you will add these.

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
jterapin: That's a good idea - I will extract this into a method. The intention here is to compare requested range (if it applies) against the range given in the response.

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
jterapin: Ahh good callout - I'll remove..

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
jterapin: Another validation check - from [AWS Documentation](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html):
> The expected total object size of the multipart upload request. If there’s a mismatch between the specified object size value and the actual object size value, it results in an HTTP 400 InvalidRequest error.

Now that I think about this - I need to set up abort upload if this call fails

gems/aws-sdk-s3/spec/object/download_file_spec.rb
jterapin: Yeah I intend to add test cases this week.

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
jterapin: Cool, I really liked the `MultipartDownloadError` naming to match so I'll go with that. I'm thinking that any errors encountered during using this download utility should return `MultipartDownloadError` - runtime error when file cannot be written to and around single request download errors.

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
jterapin: I believe so - I think we want our checksum plugin to handle the checksums on our behalf.

gems/aws-sdk-s3/spec/object/download_file_spec.rb
```diff
@@ -312,69 +129,163 @@ module S3
...
+          it 'raises error when range validation fails' do
+            client.stub_responses(:get_object, { body: 'body', content_range: 'bytes 0-4/4' })
+            expect { large_obj.download_file(path, mode: 'get_range', chunk_size: one_meg) }
+              .to raise_error(Aws::S3::MultipartDownloadError)
+          end
+
+          it 'deletes partially downloaded file when download fails' do
+            client.stub_responses(:get_object, [{ body: 'body', content_range: 'bytes 0-5242879/15728640' }])
+            expect { single_obj.download_file(path) }.to raise_error(Aws::S3::MultipartDownloadError)
+            expect(File.exist?(path)).to be(false)
```
jterapin: Here are the new additions around validations

gems/aws-sdk-s3/spec/object/upload_file_spec.rb
```diff
@@ -266,15 +239,35 @@ module S3
...
+
+          it 'raises an error when expected parts count differs from completed requests' do
+            completed = double('PartList')
+            allow(completed).to receive(:count).and_return(1)
+            allow(MultipartFileUploader::PartList).to receive(:new).and_return(completed)
+
+            pending = double('PartList')
+            allow(pending).to receive(:count).and_return(2)
+            allow(pending).to receive(:shift).and_return(nil)
+            allow(MultipartFileUploader::PartList).to receive(:new).with(anything).and_return(pending)
```
jterapin: If there's a better way to test these "expected parts" - let me know.

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -151,50 +124,63 @@ def multithreaded_get_by_parts(n_parts, total_size, etag)
...
+        msg = "multipart download failed: expected #{max_requests} parts but received #{total_requests}"
+        raise MultipartDownloadError, msg
+      end
+
+      def validate_range(resp, part)
+        range = resp.content_range.split(' ').last.split('/').first
+        expected_range = part.params[:range].split('=').last
+        return if expected_range == range
+
+        raise MultipartDownloadError, 'multipart download failed: file integrity checked failed'
```
jterapin: I could be more descriptive - like "failed to validate content range of a part" or something.

gems/aws-sdk-s3/lib/aws-sdk-s3/customizations.rb
```diff
@@ -10,6 +10,7 @@ module S3
     autoload :FileUploader, 'aws-sdk-s3/file_uploader'
     autoload :FileDownloader, 'aws-sdk-s3/file_downloader'
     autoload :LegacySigner, 'aws-sdk-s3/legacy_signer'
+    autoload :MultipartDownloadError, 'aws-sdk-s3/file_downloader'
```
mullermp: I think define this in its own file OR consider moving multipart upload error out of its own file.

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -24,58 +24,57 @@ def initialize(options = {})
...
-            raise ArgumentError, msg
+            raise ArgumentError, "Invalid mode #{@mode} provided, mode should be :single_request, :get_range or :auto"
           end
         end
+      rescue StandardError => e
+        File.delete(@path) if File.exist?(@path)
+        raise e
       end
 
       private
```
mullermp, line 51: I think it's much easier to reason about and more efficient if in download() you just call `options.delete(:mode) || <default>` on all the options you are extracting, rather than iterating keys after reading through it.

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -151,50 +124,63 @@ def multithreaded_get_by_parts(n_parts, total_size, etag)
...
+              pending.clear! # keep other threads from downloading other parts
+              raise e
             end
           end
           threads << thread
         end
         threads.map(&:value).compact
+        return if max_requests == total_requests
+
+        msg = "multipart download failed: expected #{max_requests} parts but received #{total_requests}"
```
mullermp: `but made #{total_requests} requests"`

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
mullermp: Yes, and I would include the range information of expected and got.

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -24,58 +24,57 @@ def initialize(options = {})
       def download(destination, options = {})
         @path = destination
         @mode = options[:mode] || 'auto'
```
mullermp: Check the documentation, but the error message below seems to indicate that these should be symbols instead of strings. We should fix that by either changing the error message and documentation to be string config or converting everything to symbols (for backwards compatibility) and documenting symbols.

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -151,50 +124,63 @@ def multithreaded_get_by_parts(n_parts, total_size, etag)
...
         end
         threads.map(&:value).compact
+        return if max_requests == total_requests
+
+        msg = "multipart download failed: expected #{max_requests} parts but received #{total_requests}"
+        raise MultipartDownloadError, msg
+      end
+
+      def validate_range(resp, part)
+        range = resp.content_range.split(' ').last.split('/').first
```
mullermp: It's hard to tell what this is extracting or parsing. I am assuming you are trying to parse out a number from a string like `bytes 0-3/4`.I would also consider a regex because it would be faster and also be more clear about the format of the header. Does this header always exist?

For parsing out `bytes 0-3/4` you can do this regex: `actual = resp.content_range.match(/bytes (?<range>(\d+)-(\d+))\/(\d+)/)[:range]`

For expected range, similarly `expected = part.params[:range].match(/bytes=(?<range>(\d+)-(\d+))/)[:range]`

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -151,50 +124,63 @@ def multithreaded_get_by_parts(n_parts, total_size, etag)
...
-              while part = pending.shift
+              while (part = pending.shift)
                 if progress
                   part.params[:on_chunk_received] =
                     proc do |_chunk, bytes, total|
                       progress.call(part.part_number, bytes, total)
                     end
                 end
                 resp = @client.get_object(part.params)
+                validate_range(resp, part) if part.params[:range]
```
mullermp: nit - pass in resp.content_range and part.params[:range]. You can also do resp.content_range in write() below.

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -151,50 +124,63 @@ def multithreaded_get_by_parts(n_parts, total_size, etag)
...
+        expected_range = part.params[:range].split('=').last
+        return if expected_range == range
+
+        raise MultipartDownloadError, 'multipart download failed: file integrity checked failed'
       end
 
       def write(resp)
-        range, _ = resp.content_range.split(' ').last.split('/')
-        head, _ = range.split('-').map {|s| s.to_i}
+        range = resp.content_range.split(' ').last.split('/').first
```
mullermp: I would also use the regex here for extracting the range. You can consider a small method for this extraction. You could either call it twice in write() and validate() or once in download() and pass the value down to the two methods.

The range split('-') is weird - why are we mapping all values to integers just to get the first? You can do this with `range.split('-').first.to_i`

gems/aws-sdk-s3/spec/object/download_file_spec.rb
```diff
@@ -45,241 +24,83 @@ module S3
...
-            key: 'single',
-            part_number: 1
-          }).and_return(
-            client.stub_data(
-              :head_object,
-              content_length: 15 * one_meg,
-              parts_count: nil
-            )
-          )
+          allow(client)
```
mullermp: Should we be using `stub_responses` for this rather than rspec mocks? I suppose this is enforcing params but we could do that with dynamic stubs too.

gems/aws-sdk-s3/spec/object/download_file_spec.rb
```diff
@@ -45,241 +24,83 @@ module S3
...
-              n_calls += 1
-            end
-          end
+          expect(client)
+            .to receive(:head_object)
+            .with({ bucket: 'bucket', key: 'large', part_number: 1 })
+            .exactly(1).times
 
-          large_obj.download_file(path, progress_callback: callback)
+          client.stub_responses(:get_object, ->(_ctx) { { body: 'body', content_range: 'bytes 0-4/4' } })
```
mullermp: I'm pretty sure it's `0-3/4`. You should check with a live test, but 0-4 would indicate 5 bytes not 4, and you should ensure everything works as expected - I would expect this test to fail validation of some kind. There are many instances of this throughout the file but they should be fixed too.

gems/aws-sdk-s3/spec/object/download_file_spec.rb
```diff
@@ -45,241 +24,83 @@ module S3
...
-          }).exactly(1).times
+          expect(client)
+            .to receive(:get_object)
+            .with({ bucket: 'bucket', key: 'small', response_target: path })
+            .exactly(1).times
 
           small_obj.download_file(path)
         end
 
-        it 'reports progress for single part objects' do
```
mullermp, line 118: Are these deleted or refactored below? It's hard to tell what's new and what's a refactor.

gems/aws-sdk-s3/spec/object/download_file_spec.rb
```diff
@@ -45,241 +24,83 @@ module S3
...
-            { body: 'body', content_range: 'bytes 0-4/4' }
+          expect(client)
+            .to receive(:head_object)
+            .with({ bucket: 'bucket', key: 'single', part_number: 1 })
+            .exactly(1).times
+
+          client.stub_responses(:get_object, lambda { |context|
+            responses = {
+              'bytes=0-5242879' => { body: 'body', content_range: 'bytes 0-5242879/15728640' },
+              'bytes=10485760-15728639' => { body: 'body', content_range: 'bytes 10485760-15728639/15728640' },
```
mullermp: Can you order this hash by bytes so it's more clear that you are mocking this correctly?

gems/aws-sdk-s3/spec/object/download_file_spec.rb
```diff
@@ -312,69 +129,163 @@ module S3
...
+          client.stub_responses(:get_object, lambda { |context|
             ranges = context.params[:range].split('=').last.split('-')
             expect(ranges[1].to_i - ranges[0].to_i + 1).to eq(one_meg)
             { content_range: "bytes #{ranges[0]}-#{ranges[1]}/#{20 * one_meg}" }
           })
 
           large_obj.download_file(path, chunk_size: one_meg)
         end
+
+        context 'multipart progress' do
```
mullermp, line 117: Is there anything new here?

gems/aws-sdk-s3/spec/object/download_file_spec.rb
```diff
@@ -312,69 +129,163 @@ module S3
...
+              expect(bytes).to eq([small_file_size])
+              expect(part_sizes).to eq([small_file_size])
+              expect(total).to eq(small_file_size)
+              n_calls += 1
+            end
+
+            small_obj.download_file(path, progress_callback: callback)
+            expect(n_calls).to eq(1)
+          end
+          it 'reports progress for files downloaded in parts' do
```
mullermp, line 138: spacing

gems/aws-sdk-s3/spec/object/download_file_spec.rb
```diff
@@ -312,69 +129,163 @@ module S3
...
+              .to raise_error(ArgumentError, ":chunk_size shouldn't exceed total file size.")
+          end
+
+          it 'raises an error if :on_checksum_validated is not callable' do
+            expect { large_obj.download_file(path, on_checksum_validated: 'string') }
+              .to raise_error(ArgumentError, 'on_checksum_validated must be callable')
+          end
+
+          it 'raises an error when expected parts count differs from completed requests' do
+            # Mock PartList to simulate missing parts
```
mullermp: It would be better to have this occur through behavior. This is very implementation specific. How can a part be missing from a part list? One part/call can throw an error.

gems/aws-sdk-s3/spec/object/download_file_spec.rb
mullermp: Does this test actually fail when you remove the new code you added?

gems/aws-sdk-s3/spec/object/upload_file_spec.rb
```diff
@@ -266,15 +239,35 @@ module S3
...
-            )
+            client.stub_responses(:abort_multipart_upload, [RuntimeError.new('network-error')])
             expect { object.upload_file(one_hundred_seventeen_meg_file) }.to raise_error(
               S3::MultipartUploadError,
               /failed to abort multipart upload: network-error. Multipart upload failed: part failed/
             )
           end
+
+          it 'raises an error when expected parts count differs from completed requests' do
+            completed = double('PartList')
```
mullermp: ditto about forcing this - consider driving this with behavior.

gems/aws-sdk-s3/spec/object/upload_file_spec.rb
mullermp: After forcing this with behavior, you can expect the multipart upload error to have a message `/parts validation failed/` or whatever. That message should probably be more descriptive too. The test below this as well.

gems/aws-sdk-s3/lib/aws-sdk-s3/customizations.rb
jterapin: I was about to move the other one but realized that it's used in few different files so I ended up making this own file to keep it consistent

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
jterapin: Far as I read and examined - this header should be always included in a response. Good idea, I'll update here with the regex format.

gems/aws-sdk-s3/spec/object/download_file_spec.rb
jterapin: Yeah we could make those improvements.

gems/aws-sdk-s3/spec/object/download_file_spec.rb
jterapin: nope, i made a comment below about what's new

gems/aws-sdk-s3/spec/object/download_file_spec.rb
jterapin: re-arranged below.

gems/aws-sdk-s3/spec/object/download_file_spec.rb
jterapin: I mentioned this offline but I'm unsure if we need this validation since we have a pretty firm error handling around the requests. Other errors will surface prior to hitting this.

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
jterapin: Yeah, the error message are inconsistent based on documentation. Given that documentation is already in strings, I will resort to updating the error message.

gems/aws-sdk-s3/spec/object/download_file_spec.rb
jterapin: I updated here and other parts that references the same content range. For validations, this test did not fail any validations since it is strictly downloaded by parts with no `get_range` involved (hence no range validations)

gems/aws-sdk-s3/spec/object/download_file_spec.rb
jterapin: Yup, failed.

gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
```diff
@@ -421,8 +421,8 @@ def upload_stream(options = {}, &block)
       # You can provide a callback to monitor progress of the upload:
       #
       #     # bytes and totals are each an array with 1 entry per part
-      #     progress = Proc.new do |bytes, totals|
-      #       puts bytes.map.with_index { |b, i| "Part #{i+1}: #{b} / #{totals[i]}"}.join(' ') + "Total: #{100.0 * bytes.sum / totals.sum }%" }
+      #     progress = proc do |bytes, totals|
```
jterapin, line 423: Mostly doc rendering fixes here

gems/aws-sdk-s3/spec/object/download_file_spec.rb
```diff
@@ -11,369 +11,221 @@ module S3
...
+            client.stub_responses(:get_object, { body: 'body', content_range: 'bytes 0-3/4' })
+            expect { large_obj.download_file(path, mode: 'get_range', chunk_size: one_meg) }
+              .to raise_error(Aws::S3::MultipartDownloadError)
+          end
+
+          it 'deletes partially downloaded file when download fails' do
+            client.stub_responses(:get_object, { body: 'body', content_range: 'bytes 0-5242879/15728640' })
+            expect { single_obj.download_file(path) }.to raise_error(Aws::S3::MultipartDownloadError)
+            expect(File.exist?(path)).to be(false)
+          end
```
jterapin: newly added test cases

gems/aws-sdk-s3/spec/object/upload_file_spec.rb
jterapin: I actually think these whole "total requests" validations are not useful for us so I ended up removing it. Our error handling will happen first - should a request get timed out or "goes missing"

gems/aws-sdk-s3/spec/object/download_file_spec.rb
jterapin: Update - I ended up removing the validation per other comments

gems/aws-sdk-s3/spec/object/upload_file_spec.rb
jterapin: removed it entirely per above

gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
```diff
@@ -498,64 +484,57 @@ def upload_file(source, options = {})
...
+      #   Additional options for {Client#get_object} and #{Client#head_object} may be provided.
+      #
+      # @option options [String] :mode ("auto") `"auto"`, `"single_request"` or `"get_range"`
       #
-      # @option options [String] mode `auto`, `single_request`, `get_range`
-      #  `single_request` mode forces only 1 GET request is made in download,
-      #  `get_range` mode allows `chunk_size` parameter to configured in
-      #  customizing each range size in multipart_download,
-      #  By default, `auto` mode is enabled, which performs multipart_download
+      #  * `"single_request`" mode forces only 1 GET request is made in download
```
mullermp, line 497: Wrong ordering of "` - check for other cases.

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -23,59 +20,53 @@ def initialize(options = {})
...
+        @mode = options.delete(:mode) || 'auto'
+        @thread_count = options.delete(:thread_count) || 10
+        @chunk_size = options.delete(:chunk_size)
+        @on_checksum_validated = options.delete(:on_checksum_validated)
+        @progress_callback = options.delete(:progress_callback)
+        @params = options
         validate!
 
         Aws::Plugins::UserAgent.metric('S3_TRANSFER') do
           case @mode
```
mullermp, line 33: I'm fine with string config because that's what it is now, but we should consider symbol in the new version, or do symbol now and just call to_s here. Either is fine. Symbol is more idiomatic though.

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_download_error.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module Aws
+  module S3
+    class MultipartDownloadError < StandardError; end
```
mullermp, line 6: Is this a public error? Some light weight documentation should be here.

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_upload_error.rb
```diff
@@ -4,7 +4,7 @@ module Aws
   module S3
     class MultipartUploadError < StandardError
```
mullermp, line 6: Similarly, some documentation would be good here too.

mullermp (approved):
Seems fine in general. Be sure to run integration tests and also do manual tests.

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
jterapin: Agreed, I prefer symbol-format for these. We can adjust it for V4 👍

gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
```diff
@@ -435,49 +435,35 @@ def upload_stream(options = {}, &block)
       #   will be empty.
       #
       # @param [Hash] options
-      #   Additional options for {Client#put_object}
-      #   when file sizes below the multipart threshold. For files larger than
-      #   the multipart threshold, options for {Client#create_multipart_upload},
-      #   {Client#complete_multipart_upload},
-      #   and {Client#upload_part} can be provided.
+      #   Additional options for {Client#put_object}  when file sizes below the multipart threshold.
```
mullermp: extra space

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -152,49 +117,58 @@ def multithreaded_get_by_parts(n_parts, total_size, etag)
...
-              raise error
+            rescue StandardError => e
+              pending.clear! # keep other threads from downloading other parts
+              raise e
             end
           end
           threads << thread
         end
         threads.map(&:value).compact
+        File.rename(@temp_path, @path)
```
mullermp: I think push this up to the download() method as the last thing that happens, next to where we do the ensure for clean up. This method is for downloading parts.

mullermp (approved):
Looks good, be sure to run integ tests and some manual testing. Even better if you want to add a cucumber test.

PR #3281, opened 2025-08-12
S3 Transfer Manager

jterapin:
S3 Transfer Manager implementation that uses existing internals for multipart upload/download. This utility intends to replace the resource-model based utility.

---


gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
```diff
@@ -8,6 +8,15 @@ class Object
...
       alias_method :copy_from, :copy_from
 
+      # @api private
+      def self.deprecation_msg(method)
+        "#################### DEPRECATION WARNING ####################\n"\
+          "Called deprecated method `#{method}` of #{self}.\n"\
+          "Use method `#{method}` from Aws::S3::TransferManager instead.\n"\
+          "#{self} support will be removed in next major version.\n"\
+          '#############################################################'
+      end
```
jterapin: As we introduce more functionalities into S3 transfer manager, this will be used to mark deprecated methods. I could create a top level deprecation warning for object/buckets init as an addition.

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -29,18 +32,16 @@ def download(destination, options = {})
         @params = options
         validate!
 
-        Aws::Plugins::UserAgent.metric('S3_TRANSFER') do
```
jterapin: With the introduction of S3 TM utility, I felt that the feature tracking should happen at TM-level.  But removing this means losing data on resource model multipart usage. Any thoughts? Ideally - I would like to track the actual S3 TM usage.

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_stream_uploader.rb
```diff
@@ -9,48 +9,40 @@ module Aws
   module S3
     # @api private
     class MultipartStreamUploader
```
jterapin, line 11: Cleaned up this file (just moving things around, renaming, etc) + calls abort MPU when complete MPU fails (per SEP)

gems/aws-sdk-s3/spec/file_downloader_spec.rb
```diff
@@ -0,0 +1,254 @@
+# frozen_string_literal: true
+
+require_relative 'spec_helper'
+require 'tempfile'
+
+module Aws
+  module S3
+    describe FileDownloader do
```
jterapin, line 8: I made some heavy refactoring for specs. 

Currently - our multipart upload/download internals are tightly coupled with object specs (instead of having `file_downloader_spec`, there was `download_file_spec` that was encapsulated with resource-model interface). I would like to easily carry these specs over to V4 so I separate them out per functionality classes.

gems/aws-sdk-s3/spec/multipart_file_uploader_spec.rb
```diff
@@ -0,0 +1,169 @@
...
+        it 'constructs a default s3 client when not given' do
+          client = double('client')
+          expect(S3::Client).to receive(:new).and_return(client)
+
+          uploader = MultipartFileUploader.new
+          expect(uploader.client).to be(client)
+        end
+
+        it 'sets a default thread count when not given' do
+          expect(subject.instance_variable_get(:@thread_count)).to be(MultipartFileUploader::DEFAULT_THREAD_COUNT)
```
jterapin: This is not ideal but I want to ensure that these are set accordingly. Alternatively, I can update all init variables  to have reader access (but I didn't think it was that important to have access). Thoughts?

gems/aws-sdk-s3/spec/object/download_file_spec.rb
```diff
@@ -7,236 +7,34 @@ module Aws
   module S3
     describe Object do
       let(:client) { S3::Client.new(stub_responses: true) }
-      let(:tmpdir) { Dir.tmpdir }
+      let(:subject) { S3::Object.new(bucket_name: 'bucket', key: 'key', client: client) }
```
jterapin, line 10: all the specs under `object/` had been slimmed down to testing the behavior only, since our components are throughly tested.

gems/aws-sdk-s3/spec/transfer_manager_spec.rb
```diff
@@ -0,0 +1,119 @@
+# frozen_string_literal: true
+
+require_relative 'spec_helper'
+require 'tempfile'
+
+module Aws
+  module S3
+    describe TransferManager do
```
jterapin, line 8: Should we have cucumber tests? Not sure if this is worth having and possibly cause ops pain (flaky tests).

jterapin (commented):
Passes all unit tests and existing integration tests. 
I will be doing some manual tests on the side.

gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
mullermp: Are the options that already exist in the `Aws::Deprecations#deprecated` method not sufficient?

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
mullermp: Good question. We could go either way. I would think we want to track high level usage whether it's at the transfer manager class or the resource model class. You'd probably want to duplicate the metric() block both here and in the resource model. In the resource model, it also has "resource" metric, so we can distinguish both usages of them that way.

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
```diff
@@ -8,29 +8,36 @@ module S3
     # @api private
     class MultipartFileUploader
 
+      # @api private
       MIN_PART_SIZE = 5 * 1024 * 1024 # 5MB
```
mullermp, line 11: If the class is already `@api private` then all of these are redundant (in other classes too).

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_stream_uploader.rb
```diff
@@ -155,48 +139,56 @@ def upload_in_threads(read_pipe, completed, options, thread_errors)
         part_number = 0
         options.fetch(:thread_count, @thread_count).times.map do
           thread = Thread.new do
-            begin
```
mullermp, line 158: Is there anything new here or only refactor?

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_stream_uploader.rb
```diff
@@ -155,48 +139,56 @@ def upload_in_threads(read_pipe, completed, options, thread_errors)
...
+        return completed_part unless part[:checksum_algorithm]
+
+        # get the requested checksum from the response
+        k = "checksum_#{part[:checksum_algorithm].downcase}".to_sym
+        completed_part[k] = resp[k]
+        completed_part
+      end
+
+      def ordered_parts(parts)
+        parts.size.times.map { parts.pop }.sort_by { |part| part[:part_number] }
```
mullermp: You should benchmark this against the other, but I think the other would be faster and is probably a bit more readable. Both are wonky. This might be a better case for bsearch_index (binary search)? https://ruby-doc.org/core-2.4.0/Array.html#method-i-bsearch_index Basically you'd allocate Array with size parts.size, and while parts is not empty, pop one off, find the insertion index, then insert.

gems/aws-sdk-s3/lib/aws-sdk-s3/transfer_manager.rb
```diff
@@ -0,0 +1,247 @@
...
+      #   The S3 client to use for {TransferManager} operations. If not provided, a new default client
+      #   will be created automatically.
+      def initialize(options = {})
+        @client = options.delete(:client) || Client.new
+      end
+
+      # @return [S3::Client]
+      attr_reader :client
+
+      # Downloads a file in S3 to a path on disk.
```
mullermp, line 26: I didn't review this in depth but I assume it's mostly copy pasted.

gems/aws-sdk-s3/spec/file_downloader_spec.rb
mullermp: Yep, makes sense to test these individual components. Transfer Manager spec would be light weight.

gems/aws-sdk-s3/spec/transfer_manager_spec.rb
mullermp: It's up to you - I don't think we need them particularly.

mullermp (commented):
Nice. I didn't do a deep dive into test review yet.

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_stream_uploader.rb
jterapin: Just refactor. I did separate some logic to its own method for readability.

gems/aws-sdk-s3/lib/aws-sdk-s3/transfer_manager.rb
jterapin: Mostly copy - I think the most significant parts are examples since I have to change up the syntax.

gems/aws-sdk-s3/spec/transfer_manager_spec.rb
jterapin: Yeah, I heard that Java SDK have TM integration tests and it has been a pain point for them so I think I will omit for now. We still have resource object integ tests - which tests the core functionalities.

gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
jterapin: We could - looking at it now, I guess below format warning doesn't look off. I'll reuse.
```
#################### DEPRECATION WARNING ####################
Called deprecated method `upload_file` of Aws::S3::Object. 
Use `Aws::S3::TransferManager.new.upload_file` instead.
Method `upload_file` will be removed in next major version.
#############################################################

```

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
jterapin: Good to know, l misread the documentation when I was looking this up. Let me clean it up since I'm at it.

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_stream_uploader.rb
jterapin: Yeah, good callout. I ran few benchmarking sessions (based on max number of parts) and it seems that bsearch is most optimal .

gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
```diff
@@ -398,6 +398,7 @@ def upload_stream(options = {}, &block)
         end
         true
       end
+      deprecated(:upload_stream, use: 'Aws::S3::TransferManager.new.upload_stream', version: 'next major version')
```
mullermp: These are fine, but you could also use `Aws::S3::TransferManager#<method>`

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_stream_uploader.rb
```diff
@@ -155,48 +133,62 @@ def upload_in_threads(read_pipe, completed, options, thread_errors)
...
+        completed_part = { etag: resp.etag, part_number: part[:part_number] }
+        return completed_part unless part[:checksum_algorithm]
+
+        # get the requested checksum from the response
+        k = "checksum_#{part[:checksum_algorithm].downcase}".to_sym
+        completed_part[k] = resp[k]
+        completed_part
+      end
+
+      def ordered_parts(parts)
```
mullermp, line 174: Nice, curious to see if you benchmarked this vs the other.

gems/aws-sdk-s3/spec/multipart_file_uploader_spec.rb
mullermp: I don't think these two tests really provide value. If you are already testing thread upload behavior then it's covered, and you could run the same kind of test with a different thread count option and assert at most that many threads are made. There's no guarantee that setting this instance variable ever does the thing you want.

gems/aws-sdk-s3/spec/multipart_stream_uploader_spec.rb
```diff
@@ -0,0 +1,274 @@
...
+      describe '#initialize' do
+        it 'constructs a default s3 client when none provided' do
+          client = double('client')
+          expect(S3::Client).to receive(:new).and_return(client)
+
+          uploader = MultipartStreamUploader.new
+          expect(uploader.client).to be(client)
+        end
+
+        it 'sets default configuration values when none provided' do
```
mullermp: Ditto for my previous comment - does not really test anything. If you want to test tempfile option does something, then assert one is made and used in an upload test. Assert part sizes and thread counts are what you expect in an upload test.

mullermp (approved):

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_stream_uploader.rb
jterapin: Yeah, I did a three-way comparison between:
```
                           user     system      total        real
revised method         0.003576   0.000133   0.003709 (  0.003722)
original method        0.000009   0.000012   0.000021 (  0.000020)
bsearch method         0.000005   0.000000   0.000005 (  0.000005)

```

gems/aws-sdk-s3/spec/multipart_file_uploader_spec.rb
jterapin: Yeah, I'll remove actually since we covered the custom thread upload behavior in another test.

gems/aws-sdk-s3/spec/multipart_stream_uploader_spec.rb
jterapin: Removing since we have test cases that covers the temp file option

fschwahn:
@jterapin just curious if the deprecation of `Aws::S3::Object#upload_file` is really necessary? I assume there's a lot of code relying on this, and rewriting all of that will take a lot of time. Couldn't `Aws::S3::Object#upload_file` use the new API under the hood?

jterapin:
Hello, @fschwahn. Thank you for reaching out! 

Unfortunately, AWS services have discontinued their maintenance over resource models (such as S3 and EC2). The Ruby SDK uses these models to generate classes like `Aws::S3::Object` and `Aws::S3::Bucket`. As a result, we will be phasing out support for these resource classes in the upcoming MV of the Ruby SDK. Deprecation warnings are introduced to prepare for this transition ahead of time.

PR #3283, opened 2025-08-21
FEATURE: Add support for credential_source=Environment in SharedConfig

ducks:
### Problem

The [AWS CLI documentation](https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-files.html) clearly states that `credential_source` supports three values:

- `Environment` – Specifies that the AWS CLI is to retrieve source credentials from environment variables
- `Ec2InstanceMetadata` – Specifies that the AWS CLI is to use the IAM role attached to the EC2 instance profile  
- `EcsContainer` – Specifies that the AWS CLI is to use the IAM role attached to the ECS container

However, the Ruby SDK only implemented the latter two, causing `InvalidCredentialSourceError` when users try to use the documented `Environment` option.

### Solution

This adds the missing `Environment` case in `SharedConfig#credentials_from_source` that retrieves credentials from:
- `AWS_ACCESS_KEY_ID`
- `AWS_SECRET_ACCESS_KEY` 
- `AWS_SESSION_TOKEN`

This matches the documented AWS CLI behavior.

### Testing

- Added comprehensive test coverage for the new `Environment` credential source
- All existing gems tests pass (19620 examples, 0 failures)
- Verified no regressions across AWS SDK ecosystem

### Impact

Fixes a gap where documented AWS credential functionality throws `InvalidCredentialSourceError` in the Ruby SDK.

### DIFF
--- a/build_tools/services.rb
+++ b/build_tools/services.rb
@@ -9,10 +9,10 @@ module BuildTools
     MANIFEST_PATH = File.expand_path('../../services.json', __FILE__)
 
     # Minimum `aws-sdk-core` version for new gem builds
-    MINIMUM_CORE_VERSION = "3.228.0"
+    MINIMUM_CORE_VERSION = "3.231.0"
 
     # Minimum `aws-sdk-core` version for new S3 gem builds
-    MINIMUM_CORE_VERSION_S3 = "3.228.0"
+    MINIMUM_CORE_VERSION_S3 = "3.231.0"
 
     EVENTSTREAM_PLUGIN = "Aws::Plugins::EventStreamConfiguration"
 
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Feature - Add support for ENV as credential source for `AssumeRoleCredentials`.
+
 3.230.0 (2025-08-21)
 ------------------
 
--- a/gems/aws-sdk-core/lib/aws-sdk-core/shared_config.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/shared_config.rb
@@ -370,11 +370,14 @@ module Aws
       when 'EcsContainer'
         ECSCredentials.new
       when 'Environment'
-        Credentials.new(
+        creds = Credentials.new(
           ENV['AWS_ACCESS_KEY_ID'],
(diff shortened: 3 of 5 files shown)


jterapin:
Hi! Thank you for your contribution. I will be taking a look.

richardwang1124:
Hey, thanks for opening a PR. I agree that we should support this and I may make some changes to this PR as well.

ducks:
cool cool. thanks. I'm open to any feedback/help. seemed like too simple a fix but those are the best kinds I guess.

build_tools/services.rb
```diff
@@ -9,10 +9,10 @@ class ServiceEnumerator
     MANIFEST_PATH = File.expand_path('../../services.json', __FILE__)
 
     # Minimum `aws-sdk-core` version for new gem builds
-    MINIMUM_CORE_VERSION = "3.228.0"
+    MINIMUM_CORE_VERSION = "3.231.0"
 
     # Minimum `aws-sdk-core` version for new S3 gem builds
-    MINIMUM_CORE_VERSION_S3 = "3.228.0"
+    MINIMUM_CORE_VERSION_S3 = "3.231.0"
```
jterapin, line 15: I actually think we don't need to bump s3 min core since this `credential_source` usage only applies within Amazon EC2 instances or Amazon Elastic Container Service containers

jterapin (approved):
Nice! Thanks y'all.

build_tools/services.rb
richardwang1124: Historically, numerous S3 customizations led to many min core version bumps for S3, and to prevent all other gems from needing to bump their min core versions every time as well, two separate min core versions were created. However, in this case we're bumping all gems anyway, so we can also bump S3.

ducks:
wow, thanks to y'all as well for helping me get this merged in so fast.

One question I had was on this comment:

> I actually think we don't need to bump s3 min core since this credential_source usage only applies within Amazon EC2 instances or Amazon Elastic Container Service containers

I am trying to use this on a self-hosted, non ec2 metal server and was seeing the same error when trying to use `credential_source = Environment`.

I see the other 2 options for the case are for ec2 instances but it seems like this should work elsewhere fwict.

richardwang1124:
You're correct, `credential_source = Environment` is not specific to EC2/ECS, it should work everywhere. Thanks again for your contribution! The changes should be going out as part of today's release.

PR #3288, opened 2025-08-29
[DO NOT MERGE] TransferManager: DirectoryUploader & DirectoryDownloader

jterapin:
Experimental prototype

---


gems/aws-sdk-s3/lib/aws-sdk-s3/default_executor.rb
```diff
@@ -0,0 +1,55 @@
...
+
+module Aws
+  module S3
+    # @api private
+    class DefaultExecutor
+      def initialize(options = {})
+        @queue = Queue.new
+        @max_threads = options[:max_threads] || 10
+        @pool = []
+        @running = true
```
mullermp: Does running state default to true in the initializer for other implementations of this in ruby-concurrency?

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
```diff
@@ -0,0 +1,155 @@
+# frozen_string_literal: true
+
+module Aws
+  module S3
+    # Raised when DirectoryDownloader fails to download objects from S3 bucket
+    class DirectoryDownloadError < StandardError
```
mullermp, line 6: By convention we were putting these in separate files right? If you want to promote the other two (multipart errors) to the files where they are used that's fine too, but let's stay consistent.

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
```diff
@@ -25,6 +24,8 @@ class MultipartFileUploader
       def initialize(options = {})
         @client = options[:client] || Client.new
         @thread_count = options[:thread_count] || DEFAULT_THREAD_COUNT
```
mullermp: The thread count wouldn't matter in this class anymore right?

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
```diff
@@ -38,11 +39,14 @@ def initialize(options = {})
...
+        file_size = File.size(source)
+        raise ArgumentError, 'unable to multipart upload files smaller than 5MB' if file_size < MIN_PART_SIZE
 
         upload_id = initiate_upload(options)
-        parts = upload_parts(upload_id, source, options)
-        complete_upload(upload_id, parts, source, options)
+        parts = upload_parts(upload_id, source, file_size, options)
+        complete_upload(upload_id, parts, file_size, options)
+      ensure
+        @executor.shutdown if @executor.running? && @options[:executor].nil?
```
mullermp: This seems odd. I would think the executor always exists because we provide a default - if nil we still provide a default. Also you would check if it were nil before you attempt to call a method on it?

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
```diff
@@ -0,0 +1,155 @@
...
+      # TODO: need to add progress tracker
+      def download(destination, bucket:, **options)
+        if File.exist?(destination)
+          raise ArgumentError 'invalid destination, expected a directory' unless File.directory?(destination)
+        else
+          FileUtils.mkdir_p(destination)
+        end
+
+        download_opts = options.dup
+        @bucket = bucket
```
mullermp: It's odd to set instance state like bucket in a method. Wouldn't you pass bucket, errors, configuration, etc, down to relevant methods and return back errors?

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
```diff
@@ -0,0 +1,155 @@
...
+        @errors = []
+
+        downloader = FileDownloader.new(client: client, executor: @executor)
+        producer = ObjectProducer.new(destination, build_producer_opts(download_opts))
+        producer.run
+
+        downloads = process_download_queue(producer, downloader, download_opts)
+        build_result(downloads)
+      ensure
+        @executor.shutdown unless @options[:executor]
```
mullermp: We should always assume an executor I think.

mullermp (commented):
I think this is on the right path but we should break this down into chunks. I think you should first refactor file uploader/downloader to use the executor, and we should release that, then add directory upload/download after that.

gems/aws-sdk-s3/lib/aws-sdk-s3/default_executor.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      def post(*args, &block)
+        raise 'Executor is not running' unless @running
+
+        @queue << [args, block]
+        ensure_worker_available
+      end
+
+      def shutdown
+        @running = false
```
alextwoods: Should we track state of running -> shutting down -> shutdown?  There is a gap in between calling shutdown and being fully shutdown (Where tasks cannot be submitted but tasks are still completing).

Additionally - we should probably offer a "kill" method that will immediately kill all threads.  And finally - shutdown methods usually offer the ability to specify a timeout after which remaining tasks are killed (I think you can use the limit value in `join` for that).

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
alextwoods: Since this is an internal/private api - I think I'd agree that its reasonable to always require an executor to be provided, and then we don't ever shut it down

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
alextwoods: Yeah, I'd agree.  I'm not against this being a "one-shot" class - IE, for each directory download, you must create this object again - in which case bucket/destination would be set on initialization instead.

Either way though - setting it as a member variable here is a little weird.

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
```diff
@@ -0,0 +1,155 @@
...
+          result = { completed_downloads: downloads, failed_downloads: @errors.count }
+          result[:errors] = @errors if @errors.any?
+          result
+        end
+      end
+
+      def process_download_queue(producer, downloader, opts)
+        download_attempts = 0
+        completion_queue = Queue.new
+        queue_executor = DefaultExecutor.new
```
alextwoods, line 82: Do we need an executor here? I know this needs to be done async/in a thread, but would it work to just spawn a single thread here?

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
```diff
@@ -0,0 +1,155 @@
...
+        def initialize(destination_dir, options = {})
+          @destination_dir = destination_dir
+          @client = options[:client]
+          @bucket = options[:bucket]
+          @s3_prefix = options[:s3_prefix]
+          @ignore_failure = options[:ignore_failure]
+          @filter_callback = options[:filter_callback]
+          @errors = options[:errors]
+          @directory_downloader = options[:directory_downloader]
+          @object_queue = SizedQueue.new(100)
```
alextwoods: make this a constant?

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
```diff
@@ -0,0 +1,155 @@
...
+
+            downloader.download(o[:path], opts.merge(bucket: @bucket, key: o[:key]))
+          rescue StandardError => e
+            @errors << e
+            @abort_download = true unless @ignore_failure
+          ensure
+            completion_queue << :done
+          end
+        end
+        download_attempts.times { completion_queue.pop }
```
alextwoods: I think I'm probably missing something here - what is download_attemps - what is it tracking and how is it used?

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
alextwoods: Edit: NVM, I understand why we're doing this to limit the max concurrent downloads at one time.  Maybe add comments or have a parameter/constant for that?

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
alextwoods: If we're using this to wait for all tasks to complete - could we use the shutdown method of the executor instead?

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
```diff
@@ -0,0 +1,155 @@
...
+        end
+
+        download_opts = options.dup
+        @bucket = bucket
+        @ignore_failure = download_opts.delete(:ignore_failure) || false
+        @errors = []
+
+        downloader = FileDownloader.new(client: client, executor: @executor)
+        producer = ObjectProducer.new(destination, build_producer_opts(download_opts))
+        producer.run
```
alextwoods: It seems a little weird to need to call `run` on the producer here - see my comments below on the ObjectProducer's interface for a suggestion

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
```diff
@@ -0,0 +1,155 @@
...
+          result[:errors] = @errors if @errors.any?
+          result
+        end
+      end
+
+      def process_download_queue(producer, downloader, opts)
+        download_attempts = 0
+        completion_queue = Queue.new
+        queue_executor = DefaultExecutor.new
+        while (object = producer.object_queue.shift) != :done
```
alextwoods: Suggestion on the Producer interface - I think the object_queue should be an internal detail here.  I'd lean towards having this implement enumerable, so here you would just do:

```
producer.each do |object| 
  break if @abord_download
  # rest of code
end
```

calling `each` on the producer would start it (so no need to call `run`) and would handle yielding objects from the internal object_queue.

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_uploader.rb
```diff
@@ -0,0 +1,215 @@
...
+          result[:errors] = @errors if @errors.any?
+          result
+        end
+      end
+
+      def process_upload_queue(producer, uploader, opts)
+        upload_attempts = 0
+        completion_queue = Queue.new
+        queue_executor = DefaultExecutor.new
+        while (file = producer.file_queue.shift) != :done
```
alextwoods: Ditto here - I think this producer could follow the same pattern.

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_uploader.rb
```diff
@@ -0,0 +1,215 @@
...
+          break if @abort_upload
+
+          upload_attempts += 1
+          queue_executor.post(file) do |f|
+            uploader.upload(f[:path], opts.merge(key: f[:key]))
+          rescue StandardError => e
+            @errors << e
+            @abort_upload = true unless @ignore_failure
```
alextwoods: naming nit - what about `upload_aborted` instead to make it clear that this is state rather than configuration

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_uploader.rb
```diff
@@ -0,0 +1,215 @@
...
+      # @return [Boolean]
+      attr_accessor :abort_upload
+
+      # TODO: Need to add progress tracker
+      def upload(source, bucket:, **options)
+        raise ArgumentError, 'Invalid directory' unless Dir.exist?(source)
+
+        upload_opts = options.dup
+        @ignore_failure = upload_opts.delete(:ignore_failure) || false
+        @errors = []
```
alextwoods: Same comment as on the DirectoryDownloader - I think we need to decide if these classes are "one shot" or re-usable.  If they are re-usable, then we cannot set state on the class. 

Because the File uploader/downloader are NOT one shot, I guess I'd lean towards consistency here and having a `DirectoryUploader` that is re-useable and then I think the easiest way to manage this is to create a private class `DirectoryUpload` which *is* one-shot (ie, represents the state of a single directory upload. 

As part of this - I think we would move management of the queue executor to the top level object- that is, the same queue executor would be shared across any concurrent DirectoryUploads (so that effectively a configured `max_concurrent_file_uploads` setting would apply across all concurrent uploads started from the same DirectoryUploader).

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -8,151 +8,199 @@ module Aws
...
+            resp = @client.head_object(opts[:params])
+            set_temp_path(opts)
+            multithreaded_get_by_ranges(resp.content_length, resp.etag, opts)
           end
         end
-        File.rename(@temp_path, @destination) if @temp_path
+        File.rename(opts[:temp_path], destination) if opts[:temp_path]
       ensure
-        File.delete(@temp_path) if @temp_path && File.exist?(@temp_path)
+        @executor.shutdown if @executor.running? && @options[:executor].nil?
```
alextwoods: I think there is an issue with this - the `FileDowloader` can be used to download multiple files right (ie, without creating a new one)?  so here, if we shutdown the executor on the first download call, we'll never be able to call download again.

I think we'd probably want to call shutdown only when the downloader is GC'ed (and maybe at that point, we would just call kill on it instead?)

gems/aws-sdk-s3/lib/aws-sdk-s3/default_executor.rb
jterapin: Eventually yes - concurrent-ruby's implementation goes a bit deeper but I simplified our implementation. Links below (most common executors eventually inherits `RubyExecutorService`: 

* AbstractExecutorService: https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent-ruby/concurrent/executor/abstract_executor_service.rb#L52-L54
* RubyExecutorService: https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent-ruby/concurrent/executor/ruby_executor_service.rb#L70-L72
* SimpleExecutorService: https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent-ruby/concurrent/executor/simple_executor_service.rb#L98

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
jterapin: Yup, I'm planning to separate them out.

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
jterapin: After thinking a bit through - given that executor is always passed in at init - i think we can omit this/shutdown outside of the FileDownloader (at Object/TM level).

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
jterapin: Initially, my thought process is that `completion_queue` acts as a synchronization barrier that guarantees all posted work completes before proceeding. Given the recent updates I made to `DefaultExecutor` and their shutdown method. That really will simplify things here. Thanks!

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
jterapin: Yeah, I was still trying to figure out executor but I have streamlined the executor flow a bit.

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
jterapin: I streamlined a bit here so there's less of floating instance states

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
jterapin: Thanks for the feedback! I made some changes on the interface and I really like it! It flows pretty well :)

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_uploader.rb
jterapin: I made some heavy changes around this. I'm not sure if I want to abstract out one-shot DirectoryUpload yet. 🤔  open to any thoughts after you take a look at the revisions!

PR #3292, opened 2025-09-12
Add ECDSA support to CloudFront signer

jterapin:
#3291 

---


bensie:
Thanks for working up an implementation so quickly! I pulled in these changes to an application for initial testing and it works great. Much faster signature generation and way shorter URLs!

gems/aws-sdk-cloudfront/spec/signer_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      let(:rsa_key) { OpenSSL::PKey::RSA.new(1024).to_pem }
+
+      describe '#initialize' do
+        it 'accepts RSA private key' do
+          options[:private_key] = rsa_key
+          expect { klass.new(options) }.to_not raise_error
+        end
+
+        it 'accepts ECDSA private key' do
+          options[:private_key_path] = "#{File.dirname(__FILE__)}/ecdsa_dummy_key"
```
jterapin: This test fails in JRuby environment since JRuby have stricter PEM parsing (in regards to ECDSA) so I'm avoiding dynamically generating the ECDSA key here.

gems/aws-sdk-cloudfront/spec/url_signer_spec.rb
```diff
@@ -6,32 +6,15 @@ module Aws
...
-          :private_key_path =>"#{File.dirname(__FILE__)}/unit_test_dummy_key"
+          key_pair_id: 'CF_KEYPAIR_ID',
+          private_key_path: "#{File.dirname(__FILE__)}/rsa_dummy_key"
         }
-      }
+      end
       let(:signer) { Aws::CloudFront::UrlSigner.new(options) }
       let(:expires) { 1357034400 } # January 1, 2013 10:00 am UTC (Unix timestamp)
 
-      describe '#initialize' do
```
mullermp, line 18: Why did you remove these? We still want to verify these are required arguments.

gems/aws-sdk-cloudfront/spec/url_signer_spec.rb
```diff
@@ -6,32 +6,15 @@ module Aws
   module CloudFront
     describe UrlSigner do
```
mullermp, line 7: You can probably run rubocop -A on this file instead.

gems/aws-sdk-cloudfront/spec/signer_spec.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+require_relative 'spec_helper'
```
mullermp: The public interfaces are cookie signer and url signer, not the signer module. Signer is an included module to share implementation. I don't think this file should exist, and I think cookie signer and url signer need to test both key types (RSA / ECDSA) and argument presence (private key, private key path, key pair id). This pattern is better because url signer and cookie signer (or other signer) implementations can be refactored without modifying tests.

gems/aws-sdk-cloudfront/spec/cookie_signer_spec.rb
```diff
@@ -6,13 +6,12 @@ module Aws
...
     describe CookieSigner do
 
-      let(:options) {
+      let(:options) do
         {
-          :key_pair_id => "CF_KEYPAIR_ID",
-          # private key is a .pem file
-          :private_key_path =>"#{File.dirname(__FILE__)}/unit_test_dummy_key"
+          key_pair_id: 'CF_KEYPAIR_ID',
+          private_key_path: "#{File.dirname(__FILE__)}/rsa_dummy_key"
```
mullermp, line 11: Should you have also tested ECDSA too?

jterapin: See: https://github.com/aws/aws-sdk-ruby/pull/3292/files#r2349019750

gems/aws-sdk-cloudfront/spec/url_signer_spec.rb
jterapin: Yeah I decided not to right now since it makes it harder to read what the actual changes are. 

I will be running rubocop on old files that I have touched after the approval

gems/aws-sdk-cloudfront/spec/url_signer_spec.rb
jterapin: I was consolidating to having them in `signer_spec.rb` but after you feedback on `signer_spec` - I can bring them back + add rsa/ecdsa tests

gems/aws-sdk-cloudfront/spec/signer_spec.rb
jterapin: When I was reviewing the existing specs, `CookieSigner` did not contain any tests relating to validating RSA key/arguments presence and only `UrlSigner` had them so I wanted to consolidate the shared behavior within `signer_spec` - that being said, I'm fine with just adding these into both existing spec files

mullermp (approved):

PR #3294, opened 2025-09-23
Update documentation for `S3::FileDownloader` behavior

jterapin:


gems/aws-sdk-s3/lib/aws-sdk-s3/transfer_manager.rb
```diff
@@ -48,7 +48,12 @@ def initialize(options = {})
       # @param [String, Pathname, File, Tempfile] destination
       #   Where to download the file to. This can either be a String or Pathname to the file, an open File object,
       #   or an open Tempfile object. If you pass an open File or Tempfile object, then you are responsible for
-      #   closing it after the download completes.
+      #   closing it after the download completes. Download behavior varies by destination type:
+      #
+      #   * **String/Pathname paths**: Downloads to a temporary file first, then atomically moves to the final
+      #    destination. This ensures the original file is not corrupted if the download fails.
+      #   * **File/Tempfile objects**: Downloads directly to the file object without using temporary files.
```
alextwoods: May want to also mention you are responsible for closing here as well

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -14,6 +14,7 @@ class FileDownloader
 
       def initialize(options = {})
         @client = options[:client] || Client.new
+        @legacy = options[:legacy] || false
```
alextwoods: I would tend to prefer a flag name that is longer, but more descriptive of what the legacy behavior is - naming is hard... but maybe something like `disable_tempfile_copy_on_path` or  `legacy_direct_path_write`

I dont know, its hard to describe succinctly, but even though its an internal flag, I think we'll quickly forget the context of what "legacy" means.

jterapin: Good callout. I definitely have experienced that when looking over old code. 
I'll rename to `@enable_direct_writes` since that should be clear. Thanks!

mullermp (approved):

Issue #3193, opened 2025-02-17
S3 access results in improperly terminated TCP sockets.


tubsandcans:
### Describe the bug

I notice that when reading (or writing) objects out of an S3 bucket using the aws-sdk-s3 (1.176.0) gem, a TCP connection is left in the CLOSE_WAIT state. This only goes away when the process running the Ruby program that initiated the connection dies. 

This isn't really an issue locally, but in the production server which is a multi-process web-server (Passenger+Nginx), it appears these persistent connections slowly consume system resources to the point of complete exhaustion. I'm genuinely curious if this decision to leave these connections in CLOSE_WAIT was intentional and just a really poor fit for our production environment, or if it was done by mistake. Regardless, it seems at the very least imprudent to keep these connections in CLOSE_WAIT for the lifecycle of the initiating process.

I noticed this after enabling active-storage for our application, and it would slowly consume all RAM until the application crashes. Passenger is unable to re-use these web processes due to them having a TCP socket in CLOSE_WAIT, so it keeps spawning more until all system resources are maxed out.

Ruby version: 3.3.6

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

I would not expect any evidence of a TCP connection to the S3 service once transmission is completed.

### Current Behavior

After reading from an S3 bucket, a TCP socket is left in CLOSE_WAIT state.

### Reproduction Steps

1. Run `lsof -i -P | grep CLOSE_WAIT` and notice there are no TCP connections to an AWS resource in CLOSE_WAIT. 
2. Read from an S3 bucket in a Ruby console
3. Run `lsof -i -P | grep CLOSE_WAIT` and notice a TCP socket for AWS in CLOSE_WAIT.
4. Terminate the Ruby console 
5. Run `lsof -i -P | grep CLOSE_WAIT` and notice this CLOSE_WAIT socket goes away.

### Possible Solution

Close the TCP socket upon completion of transmission.

### Gem name ('aws-sdk', 'aws-sdk-resources' or service gems like 'aws-sdk-s3') and its version

aws-sdk-s3 (1.176.0)

### Environment details (Version of Ruby, OS environment)

Ruby 3.3.6, macOS 14.6.1

mullermp:
[...]

PR #3296, opened 2025-09-30
fix: honor closed connection to avoid timeouts on all AWS API calls

victorboissiere:
We have several microservices (all using this library) that fail to open a TCP connection (execution expired), mainly in STS and S3.

When investigating and going to a node that hosts the pod with the issue, we see many connections left in CLOSE_WAIT. When checking the issue https://github.com/aws/aws-sdk-ruby/issues/3193#issuecomment-2667052220, we seem not to be alone in having this issue.

Since the only way connections are closed and not put in the queue it's either if:
- keep alive is reached
- exception is raised

But if AWS server send `Connection: close` the app will never close it.

This MR attempts to fix that behaviour to avoid any impact to apps using this lib.

## More context

```
ruby '3.3.6'

gem 'rails', '~> 7.2'
    aws-sdk-core (3.232.0)
      aws-eventstream (~> 1, >= 1.3.0)
      aws-partitions (~> 1, >= 1.992.0)
      aws-sigv4 (~> 1.9)
      base64
      bigdecimal
      jmespath (~> 1, >= 1.6.1)
      logger
    aws-sdk-kms (1.96.0)
      aws-sdk-core (~> 3, >= 3.210.0)
      aws-sigv4 (~> 1.5)
    aws-sdk-lambda (1.148.0)
      aws-sdk-core (~> 3, >= 3.216.0)
      aws-sigv4 (~> 1.5)
    aws-sdk-s3 (1.199.0)
      aws-sdk-core (~> 3, >= 3.231.0)
      aws-sdk-kms (~> 1)
      aws-sigv4 (~> 1.5)
    aws-sdk-ssm (1.191.0)
      aws-sdk-core (~> 3, >= 3.216.0)
      aws-sigv4 (~> 1.5)
    aws-sigv4 (1.10.1)
      aws-eventstream (~> 1, >= 1.0.2)
```

Example logs track traces

```
Seahorse::Client::NetworkingError (Failed to open TCP connection to XXXX.s3.eu-central-1.amazonaws.com:443 (execution expired)):
  
app/contexts/attachments/commands/create_attachment.rb:40:in `create_attachment'
app/contexts/attachments/commands/create_attachment.rb:22:in `find_or_create_attachment'
app/contexts/attachments/commands/create_attachment.rb:10:in `call'
app/contexts/attachments/services/create_attachment.rb:21:in `create_attachment'
app/contexts/attachments/services/create_attachment.rb:12:in `call'
app/service_controllers/attachment_service_controller.rb:4:in `create'
app/controllers/v3/attachments_controller.rb:44:in `create'
lib/middlewares/prometheus/team_tagger.rb:14:in `call'
```

When we go to the node where the issue happens, we have this
```
tcp      6 3380 CLOSE_WAIT src=10.XX.XX.XX dst=3.5.136.54 sport=40820 dport=443 src=3.5.136.54 dst=10.XX.XX.XX sport=443 dport=10434 [ASSURED] mark=128 secctx=system_u:object_r:unlabeled_t:s0 use=1
```

It's been nearly 8 MIN since the TCP received the FIN from S3 (meaning, S3 said to the app => please close this, the app waits for things to finish).
It's consistent with what AWS said; the app is somehow not closing its connection (we were in a live call with support).

All Ruby apps using this SDK are impacted by the issue (CLOSE_WAIT), and other calls to external services that don't rely on this SDK work fine.

When checking other issues we can see that we're not the only one impacted by these connection expired.

### DIFF
--- a/gems/aws-sdk-core/lib/seahorse/client/net_http/connection_pool.rb
+++ b/gems/aws-sdk-core/lib/seahorse/client/net_http/connection_pool.rb
@@ -76,8 +76,14 @@ module Seahorse
         #
         # @return (see #session_for)
         def request(endpoint, request, &block)
-          session_for(endpoint) do |http|
-            yield(http.request(request))
+          response = nil
+          should_close = { value: false }
+
+          session_for(endpoint, close_connection: should_close) do |http|
+            response = http.request(request)
+            # Check if server wants to close the connection
+            should_close[:value] = response['connection']&.downcase == 'close'
+            yield(response)
           end
         end
 
@@ -87,7 +93,7 @@ module Seahorse
         # @yieldparam [Net::HTTPSession] session
         #
         # @return [nil]
-        def session_for(endpoint, &block)
+        def session_for(endpoint, close_connection: { value: false }, &block)
           endpoint = remove_path_and_query(endpoint)
           session = nil
 
@@ -109,10 +115,15 @@ module Seahorse
             session.finish if session
             raise
           else
-            # No error raised? Good, check the session into the pool.
-            @pool_mutex.synchronize do
-              @pool[endpoint] = [] unless @pool.key?(endpoint)
-              @pool[endpoint] << session
+            # Only pool if server didn't request close
+            if close_connection[:value]
+              session.finish
+            else
+              # No error raised? Good, check the session into the pool.
+              @pool_mutex.synchronize do
+                @pool[endpoint] = [] unless @pool.key?(endpoint)
+                @pool[endpoint] << session
+              end
             end
           end
           nil


mullermp:
Thanks for opening a different pull request. I'm taking a look - however I'm still skeptical of this change. Can you be more specific about your environment? Are you using a single threaded, multi process setup?

victorboissiere:
> Thanks for opening a different pull request. I'm taking a look - however I'm still skeptical of this change. Can you be more specific about your environment? Are you using a single threaded, multi process setup?

multi process setup

mullermp:
But how many threads per process? I think when I last looked at this type of issue, the most plausible thing was that your environment (and Ruby) is running out of file descriptors, so it can't create a new connection. Normally those close_wait connections are just simply reused because it was previously a good connection.

victorboissiere:
> But how many threads per process? I think when I last looked at this type of issue, the most plausible thing was that your environment (and Ruby) is running out of file descriptors, so it can't create a new connection. Normally those close_wait connections are just simply reused because it was previously a good connection.

```
        MAX_THREADS: '5'
        SIDEKIQ_CONCURRENCY: '30'
```

In term of FD we're using very little (this is our usage)

image

This is the min of our capacity
image

mullermp:
Have you tried this code in production via a fork or monkey patch?

What's confusing to me is that stale sessions should be reaped by the `_clean` function prior to creating a new session. I'm not able to reproduce locally and can confirm this. After making a request to S3, and waiting a few seconds, I can see the close_wait connection as expected. Then when I make a new request, I see the open session in the pool, though it's stale, and it gets removed before it's fetched out of the pool because it has idled too long.

What configuration are you using for http_idle_timeout? If you are able to reproduce locally, that would be helpful.

mullermp:
The TAM told me that you were starting to see issues on August 18. What changed on that date?

What happens with all of the close_wait connections - is your process stalling?

The one thing in common between you and the other customer was ActiveStorage usage. Can you confirm if symptoms still happen outside of ActiveStorage? You mentioned STS - what is STS being used for in this flow?

One other thing to check - can you inspect the `@pools` state in the connection pool class via monkey patch? Endpoint is used as the connection key - are you seeing large and abnormal amounts of these?

jterapin:
Hello! We heard from TAM that the issue has been resolved. We will be closing this PR for now until additional
information or related issues come up. If you have any questions or run into similar problems, please don't
hesitate to reach out.

PR #3302, opened 2025-10-07
S3 Executor support

jterapin:
This PR introduces a lightweight `DefaultExecutor` to replace the legacy `:thread_count` approach, laying the foundation for directory upload/download feature. Additionally, we support the following:
* Custom executor support - Customers are able to pass in their own executor implementation (must implement `DefaultExecutor` interface)
* `:thread_count` configuration is still respected - Existing `:thread_count` option continues to work by
automatically creating a DefaultExecutor with the specified thread count

---


gems/aws-sdk-s3/lib/aws-sdk-s3/default_executor.rb
```diff
@@ -0,0 +1,101 @@
+# frozen_string_literal: true
+
+module Aws
+  module S3
+    # @api private
```
mullermp, line 5: Is this private or public?

gems/aws-sdk-s3/spec/multipart_file_uploader_spec.rb
```diff
@@ -7,7 +7,7 @@ module Aws
   module S3
     describe MultipartFileUploader do
       let(:client) { S3::Client.new(stub_responses: true) }
-      let(:subject) { MultipartFileUploader.new(client: client) }
+      let(:subject) { MultipartFileUploader.new(client: client, executor: DefaultExecutor.new) }
```
mullermp, line 10: executor is already defaulted right? Why is this supplied here and on subsequent options?

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -8,184 +8,243 @@ module Aws
...
+          mode: opts.delete(:mode) || 'auto',
+          chunk_size: opts.delete(:chunk_size),
+          on_checksum_validated: opts.delete(:on_checksum_validated),
+          progress_callback: opts.delete(:progress_callback),
+          params: opts,
+          temp_path: nil
+        }
+      end
+
+      def cleanup_temp_file!(opts)
```
mullermp: This method does not raise or mutate, so it should not be a bang method.

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
```diff
@@ -108,24 +109,20 @@ def checksum_key?(key)
         CHECKSUM_KEYS.include?(key)
       end
 
-      def has_checksum_key?(keys)
+      def has_checksum_keys?(keys)
```
mullermp: I think it's only ever one key.

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
```diff
@@ -135,43 +132,43 @@ def upload_part_opts(options)
...
-      def upload_in_threads(pending, completed, options)
-        threads = []
+      def upload_with_executor(pending, completed, options)
+        upload_attempts = 0
+        completion_queue = Queue.new
+        abort_upload = false
+        errors = []
+
         if (callback = options[:progress_callback])
           progress = MultipartProgress.new(pending, callback)
```
mullermp: You can simplify the usages of these I think by always making a multipart progress object and short circuiting its callback if it doesn't exist.

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
```diff
@@ -135,43 +132,43 @@ def upload_part_opts(options)
...
+            abort_upload = true
+            errors << e
+          ensure
+            completion_queue << :done
           end
-          threads << thread
         end
-        threads.map(&:value).compact
+
+        upload_attempts.times { completion_queue.pop }
```
mullermp, line 163: Why is this (and other usage) necessary?

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_stream_uploader.rb
```diff
@@ -54,28 +52,30 @@ def complete_upload(upload_id, parts, options)
...
-              completed,
-              upload_part_opts(options).merge(upload_id: upload_id),
-              thread_errors
-            )
-            begin
-              block.call(write_pipe)
-            ensure
-              # Ensure the pipe is closed to avoid https://github.com/jruby/jruby/issues/6111
-              write_pipe.close
+            upload_thread = Thread.new do
```
mullermp, line 60: Why is another thread made to wrap the executor that uses threads?

gems/aws-sdk-s3/lib/aws-sdk-s3/transfer_manager.rb
```diff
@@ -12,17 +12,47 @@ module S3
     # * track transfer progress by using progress listener
     #
     class TransferManager
+      # @example Using default executor (automatic creation and shutdown)
```
mullermp: We can consider using at_exit to handle shutdown automatically for all cases, rather than for only the default and having a customer managed executor. We can just require it to implement a shutdown method and we always call it. https://docs.ruby-lang.org/en/3.4/Kernel.html#method-i-at_exit

gems/aws-sdk-s3/lib/aws-sdk-s3/transfer_manager.rb
```diff
@@ -74,10 +104,10 @@ def initialize(options = {})
       # @option options [Integer] :chunk_size required in `"get_range"` mode.
       #
       # @option options [Integer] :thread_count (10) Customize threads used in the multipart download.
+      #   Only used when no custom executor is provided (creates {DefaultExecutor} with given thread count).
       #
-      # @option options [String] :version_id The object version id used to retrieve the object.
-      #
-      #     @see https://docs.aws.amazon.com/AmazonS3/latest/dev/ObjectVersioning.html ObjectVersioning
+      # @option options [String] :version_id The object version id used to retrieve the object. See
```
mullermp: Is this a param of get object? I think you can link the documentation for the operation instead of this. I think also checksum_mode as well.

gems/aws-sdk-s3/lib/aws-sdk-s3/transfer_manager.rb
```diff
@@ -102,8 +132,11 @@ def initialize(options = {})
       # @see Client#get_object
       # @see Client#head_object
       def download_file(destination, bucket:, key:, **options)
-        downloader = FileDownloader.new(client: @client)
+        download_opts = options.dup
+        executor = @executor || DefaultExecutor.new(max_threads: download_opts.delete(:thread_count))
+        downloader = FileDownloader.new(client: @client, executor: executor)
         downloader.download(destination, options.merge(bucket: bucket, key: key))
+        executor.shutdown unless @options[:executor]
```
mullermp: You could just inspect `@executor` here and not store all of the options.

gems/aws-sdk-s3/spec/default_executor_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    describe DefaultExecutor do
+      let(:subject) { DefaultExecutor.new }
+
+      describe '#post' do
+        it 'executes a block with arguments' do
+          block = double('block')
+          expect(block).to receive(:call).with('hello')
+
+          subject.post('hello') { |arg| block.call(arg) }
+          sleep 0.1
```
mullermp: I'd prefer tests without actual sleeping somehow - these make tests slower.

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -8,184 +8,243 @@ module Aws
...
+          download_attempts += 1
+          @executor.post(part) do |p|
+            update_progress(progress, p) if progress
+
+            resp = @client.get_object(p.params)
+            range = extract_range(resp.content_range)
+            validate_range(range, p.params[:range]) if p.params[:range]
+            write(resp.body, range, opts)
+
+            if opts[:on_checksum_validated] && resp.checksum_validated
```
mullermp: You can similarly wrap this like progress callback to make it cleaner.

mullermp (commented):
Nice so far!

gems/aws-sdk-s3/lib/aws-sdk-s3/default_executor.rb
jterapin: You know - I was super torn whether I should make this private or not. It is definitely an inner detail but at the same time, it's an interface that customer will have to abide when implementing their own executor. Thoughts?

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
jterapin: Agreed!

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
jterapin: We typically would have `threads.map(&:value).compact` or `thread.join` to wait for all the work to be done. I needed a way for the main thread to wait until all the background threads are done processing.

gems/aws-sdk-s3/lib/aws-sdk-s3/transfer_manager.rb
jterapin: That's a pretty neat. I'm still a bit worried about custom executors though, especially if customers are reusing that executor elsewhere in their app and have their own `at_exit` hooks (not sure if this is a valid case). We might run into exit ordering issues that we can't control.

gems/aws-sdk-s3/lib/aws-sdk-s3/transfer_manager.rb
jterapin: I can just get rid of it since we already have a [reference](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-s3/lib/aws-sdk-s3/transfer_manager.rb#L65-L66) and [here](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-s3/lib/aws-sdk-s3/transfer_manager.rb#L102-L103) but I'll keep `checksum_mode` around since its API documentation is rather lacking. I'll update Object documentation as well.

gems/aws-sdk-s3/lib/aws-sdk-s3/transfer_manager.rb
jterapin: This is kind of how I know whether if `@executor` is custom or not - so that I don't shut down custom executors.

gems/aws-sdk-s3/spec/multipart_file_uploader_spec.rb
jterapin: Executor should be passed down from the Object/TM level. My thinking is that these private classes should always expect an executor to be provided - this streamlines passing the executor from top to bottom. Open to thoughts on this approach though.

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_stream_uploader.rb
jterapin: I need an additional thread to submit tasks to the executor, since main thread writes data to the pipe.  Alternatively, I could use `@executor.post` to do this task submission work but that means we're technically not uploading concurrently at 10 threads.

Otherwise, I could leave `MultipartStreamUploader` (continue using `:thread_count`) as we had before to keep the current behavior but in the next MV, I could update to follow this pattern. Thoughts?

gems/aws-sdk-s3/spec/default_executor_spec.rb
jterapin: Fair. I shorten the timing so maybe this will help. 
Alternatively, I was thinking of using `Thread.pass` but I'm bit worried about the global effect of it:
```
describe '#post' do
  it 'executes a block with arguments' do
    result = nil
    subject.post('hello') { |a| result = a }

    Thread.pass until result
    expect(result).to eq('hello')
  end
```

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
jterapin: Let me know if the revision makes it better

gems/aws-sdk-s3/lib/aws-sdk-s3/default_executor.rb
mullermp: I guess it's private since thread count option is passed through. However if we haven't already, make clear how to provide a custom one.

gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
```diff
@@ -539,10 +543,13 @@ def upload_file(source, options = {})
       # @see Client#get_object
       # @see Client#head_object
       def download_file(destination, options = {})
-        downloader = FileDownloader.new(client: client)
+        download_options = options.dup
```
mullermp: As an optimization, rather than dup, you can just merge bucket and key which creates new hash then you can delete off keys.

gems/aws-sdk-s3/lib/aws-sdk-s3/file_downloader.rb
```diff
@@ -8,184 +8,245 @@ module Aws
...
-        @destination = destination
-        @mode = options.delete(:mode) || 'auto'
-        @thread_count = options.delete(:thread_count) || 10
-        @chunk_size = options.delete(:chunk_size)
-        @on_checksum_validated = options.delete(:on_checksum_validated)
-        @progress_callback = options.delete(:progress_callback)
-        @params = options
-        validate!
+        validate_destination!(destination)
+        opts = build_download_opts(destination, options.dup)
```
mullermp: I don't think we care about dup here either because we should be given a new hash already at this point.

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_stream_uploader.rb
mullermp: Well if I set thread count to 10 but 11 threads are used, that's maybe undesirable. Using post maybe makes sense so that it's at most 10 for the entire operation but that's up to you.

gems/aws-sdk-s3/lib/aws-sdk-s3/transfer_manager.rb
mullermp: I'm not sure I follow. If they are using a custom executor then we would only shut it down when the program exits. If they shut it down before us in the program or at their own exit hook, our shutdown call happens but becomes no op. I do think you should investigate this option more.

gems/aws-sdk-s3/lib/aws-sdk-s3/transfer_manager.rb
```diff
@@ -102,8 +129,11 @@ def initialize(options = {})
       # @see Client#get_object
       # @see Client#head_object
       def download_file(destination, bucket:, key:, **options)
-        downloader = FileDownloader.new(client: @client)
+        download_opts = options.dup
```
mullermp: Ditto about dup comments (applies to other places)

gems/aws-sdk-s3/spec/default_executor_spec.rb
mullermp: I suspect these kinds of tests will be flakey. I think you should find a way to do it with expectations rather than blocking with sleep.

gems/aws-sdk-s3/spec/default_executor_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      describe '#shutdown' do
+        it 'waits for running tasks to be complete' do
+          result = nil
+          subject.post { result = 'done' }
+          expect(subject.shutdown).to be(true)
+          expect(result).to eq('done')
+        end
+
+        it 'kills threads after timeout' do
+          result = nil
```
mullermp: Should you have asserted result is done at some point?

mullermp (approved):

gems/aws-sdk-s3/spec/default_executor_spec.rb
jterapin: I updated the specs to reduce flakiness by replacing (most) sleep with blocking coordination using queues.

gems/aws-sdk-s3/spec/default_executor_spec.rb
jterapin: I re-did the specs to verify the thread would have completed if not killed.

gems/aws-sdk-s3/lib/aws-sdk-s3/transfer_manager.rb
jterapin: Discussed offline

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_stream_uploader.rb
jterapin, line 60: I went back to using separate thread than using an executor's pool. This isolates the executor to do upload work only.

mullermp (approved):
Looks good!

gems/aws-sdk-s3/lib/aws-sdk-s3/default_executor.rb
```diff
@@ -0,0 +1,101 @@
...
+      # @return [Boolean] true when shutdown is complete
+      def shutdown(timeout = nil)
+        @mutex.synchronize do
+          return true if @state == SHUTDOWN
+
+          @state = SHUTTING_DOWN
+          @pool.size.times { @queue << :shutdown }
+        end
+
+        if timeout
```
alextwoods, line 62: Is there a reason this code isn't in the synchronize block?

Atleast the later @state = SHUTDOWN should probably be

gems/aws-sdk-s3/lib/aws-sdk-s3/transfer_manager.rb
```diff
@@ -2,27 +2,70 @@
...
+    # You can provide your own executor (e.g., `Concurrent::ThreadPoolExecutor`) for fine-grained control over thread
+    # pools and resource management. When using a custom executor, you are responsible for shutting it down
+    # when finished. The executor may be reused across multiple TransferManager operations.
+    #
+    # In addition, it must implement the same interface as DefaultExecutor. The following is a list
+    # of required methods:
+    #
+    #   * `post(*args, &block)` - Execute a task with given arguments and block
+    #   * `shutdown(timeout = nil)` - Gracefully shutdown the executor with optional timeout
+    #   * `kill` - Immediately terminate all running tasks
```
alextwoods: Our code doesn't actually rely on these methods right? these are more of a suggestion

alextwoods (approved):

gems/aws-sdk-s3/lib/aws-sdk-s3/transfer_manager.rb
jterapin: Yeah, I would say `shutdown` is technically a suggestion (since custom executors are managed by the customer). 

With the context of Directory Uploader/Downloader, we do rely on `kill` to stop ongoing tasks in a context when they have their failure policy as "abort all ongoing uploads/downloads in a moment of exception" - see [demo](https://github.com/aws/aws-sdk-ruby/blob/0cb35cd018319b024ff1e6522c53211b5a002bd2/gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb#L72-L77).

I'll update the docs to label `shutdown` as suggestion.

gems/aws-sdk-s3/lib/aws-sdk-s3/default_executor.rb
jterapin: Good callout, yes - it should be wrapped. Thanks!

PR #3304, opened 2025-10-13
S3 Directory Support

jterapin:
Adds directory upload/download to Transfer Manager.

### `upload_directory`
- Upload all files from a local directory to S3
-  Optional recursive traversal of subdirectories
- Symlink handling with circular reference detection
- S3 key prefix support
- Filter callback to selectively upload files
- Request callback to modify upload parameters per file
- Progress callback for transfer monitoring
- Configurable failure handling (fail-fast or continue on error)

### `download_directory`
- Download all objects from an S3 bucket/prefix to a local directory
- S3 prefix stripping for clean local paths
- Path traversal detection 
- Filter callback to selectively download objects
- Request callback to modify download parameters per object
- Progress callback for transfer monitoring
- Configurable failure handling (fail-fast or continue on error)
- Automatic directory creation for nested structures
 
---


gems/aws-sdk-s3/lib/aws-sdk-s3/transfer_manager.rb
```diff
@@ -70,6 +71,90 @@ def initialize(options = {})
...
+      #   It will be invoked with `transferred_bytes` and `transferred_files`.
+      #
+      # @raise [DirectoryDownloadError] Raised when download fails with `ignore_failure: false` (default)
+      #
+      # @return [Hash] Returns a hash with download statistics:
+      #
+      #   * `:completed_downloads` - Number of objects successfully downloaded
+      #   * `:failed_downloads` - Number of objects that failed to download
+      #   * `:errors` - Array of errors for failed downloads (only present when failures occur)
+      def download_directory(destination, bucket:, **options)
```
jterapin, line 156: I need to allow for setting a custom thread for executor here. Same goes for uploader.

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
```diff
@@ -0,0 +1,211 @@
...
+
+        FileUtils.mkdir_p(File.dirname(entry.path)) unless Dir.exist?(File.dirname(entry.path))
+        downloader.download(entry.path, entry.params)
+        progress&.call(File.size(entry.path))
+      rescue StandardError => e
+        @mutex.synchronize { errors << e }
+        request_abort unless opts[:ignore_failure]
+      end
+
+      def process_download_queue(producer, downloader, opts)
```
jterapin: In a situation where we need to raise, I decided against doing `@queue_executor.kill` since there might be work in-flight that hangs. It would be best to exit gracefully and the bubbles up to raise within `#build_results`

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_uploader.rb
```diff
@@ -0,0 +1,262 @@
...
+
+      def request_abort
+        @mutex.synchronize { @abort_requested = true }
+      end
+
+      def upload(source_directory, bucket, **opts)
+        raise ArgumentError, 'Invalid directory' unless Dir.exist?(source_directory)
+
+        uploader = FileUploader.new(
+          multipart_threshold: opts.delete(:multipart_threshold),
```
jterapin, line 39: Self-reminder to documentation this at `TransferManager#upload_directory`

I need to take a look at other params available on `FileUploader` and `FileDownloader` level but I'm concerned about overlapped params if it occurs.

gems/aws-sdk-s3/lib/aws-sdk-s3/transfer_manager.rb
richardwang1124, line 67: What if you added `DefaultExecutor` here? Like 
```
@executor = options[:executor] || DefaultExecutor.new
```

gems/aws-sdk-s3/spec/transfer_manager_spec_helper.rb
alextwoods, line 1: Nit - `manger` mispelled.

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
```diff
@@ -0,0 +1,212 @@
...
+            next if directory_marker?(o)
+            next unless include_object?(o.key)
+
+            @object_queue << build_object_entry(o.key)
+          end
+          stream_objects(continuation_token: resp.next_continuation_token) if resp.next_continuation_token
+        end
+
+        def validate_path(path, key)
+          segments = path.split('/')
```
alextwoods: Should this be `File::SEPERATOR` rather than `/` here?  I think the path at this point will come from `File.join` and so would have os seperator?   I might be wrong about that though.

gems/aws-sdk-s3/lib/aws-sdk-s3/customizations.rb
```diff
@@ -7,15 +7,23 @@ module S3
     autoload :Encryption, 'aws-sdk-s3/encryption'
     autoload :EncryptionV2, 'aws-sdk-s3/encryption_v2'
     autoload :EncryptionV3, 'aws-sdk-s3/encryption_v3'
+    autoload :LegacySigner, 'aws-sdk-s3/legacy_signer'
+
+    # transfer manager + multipart upload/download utilities
+    autoload :DefaultExecutor, 'aws-sdk-s3/default_executor'
```
alextwoods, line 13: Nit: it looks like default_executor is listed twice here.

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
```diff
@@ -0,0 +1,212 @@
...
+      def initialize(options = {})
+        @client = options[:client]
+        @executor = options[:executor]
+        @abort_requested = false
+        @mutex = Mutex.new
+      end
+
+      attr_reader :client, :executor
+
+      def abort_requested
```
alextwoods: This might be more clear with a `?` suffix (ie `abort_requested?`) to make it clear that its a status check rather than an action.

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
```diff
@@ -0,0 +1,212 @@
...
+        @mutex.synchronize { @abort_requested }
+      end
+
+      def request_abort
```
alextwoods: "request_abort" feels more verbose than just "abort".  I think the internal `@abort_requested` makes sense, but the method's name I think could just be `abort`.  I assume the `request_abort` is to make it clear that aborting is async, but I think thats implied with abort already.  What do you think?

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
```diff
@@ -0,0 +1,212 @@
...
+          FileUtils.mkdir_p(destination)
+        end
+
+        download_opts, producer_opts = build_opts(destination, bucket, options)
+        downloader = FileDownloader.new(client: @client, executor: @executor)
+        producer = ObjectProducer.new(producer_opts)
+        downloads, errors = process_download_queue(producer, downloader, download_opts)
+        build_result(downloads, errors)
+      ensure
+        @abort_requested = false
```
alextwoods: Why are we setting @abort_requested to false here?

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
```diff
@@ -0,0 +1,212 @@
...
+            prefix = @s3_prefix.end_with?('/') ? @s3_prefix : "#{@s3_prefix}/"
+            key = key.delete_prefix(prefix)
+          end
+          File::SEPARATOR == '/' ? key : key.tr('/', File::SEPARATOR)
+        end
+
+        def stream_objects(continuation_token: nil)
+          resp = @client.list_objects_v2(bucket: @bucket, prefix: @s3_prefix, continuation_token: continuation_token)
+          resp.contents&.each do |o|
+            break if @directory_downloader.abort_requested
```
alextwoods: This will block on the mutex - I'm not sure we always need to that here.  It is definitely safe to do so, but likely has a performance hit.  Ditto I think with the check on line 91: 
```
        begin
          producer.each do |object|
              break if abort_requested
```

Since we're using a SizedQueue for communicating between threads - maybe we could use clear/close on it rather than constantly blocking on the mutex?  If we used `close` It will cause threads waiting to raise ClosedQueueError which we could catch and handle to detect aborts?   I haven't fully thought that out, but I think it might simplify the code and avoid locking as much.

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
richardwang1124: Does this need `@mutex.synchronize`?

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_progress.rb
```diff
@@ -0,0 +1,24 @@
...
+        @progress_callback = progress_callback
+        @mutex = Mutex.new
+      end
+
+      def call(bytes_transferred)
+        @mutex.synchronize do
+          @transferred_bytes += bytes_transferred
+          @transferred_files += 1
+
+          @progress_callback.call(@transferred_bytes, @transferred_files)
```
alextwoods: I see why we're calling the progress_callback inside the `synchronize` block - it does ensure progress is linear for users.  However, depending on what they're doing in the callback (things like IO like printing/writing to file, ect) - this could end up being a small bottleneck.  I'm not sure how much that matters vs the fully ordered progress callbacks.

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_uploader.rb
```diff
@@ -0,0 +1,262 @@
...
+              upload_file(f, uploader, opts, progress, errors)
+            ensure
+              completion_queue << :done
+            end
+          end
+        rescue StandardError => e
+          @mutex.synchronize { errors << e }
+          request_abort
+        end
+        upload_attempts.times { completion_queue.pop }
```
alextwoods: Is there a chance that upload_attemps will ever be larger than the completion queue?

alextwoods (commented):
Nice - this is looking good!  Great test coverage and documentation. Good thread safety.

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_uploader.rb
```diff
@@ -0,0 +1,262 @@
...
+          http_chunk_size: opts.delete(:http_chunk_size),
+          client: @client,
+          executor: @executor
+        )
+        upload_opts, producer_opts = build_opts(source_directory, bucket, opts)
+        producer = FileProducer.new(producer_opts)
+        uploads, errors = process_upload_queue(producer, uploader, upload_opts)
+        build_result(uploads, errors)
+      ensure
+        @abort_requested = false
```
richardwang1124: Same as my previous comment

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
```diff
@@ -0,0 +1,212 @@
...
+              download_object(o, downloader, opts, progress, errors)
+            ensure
+              completion_queue << :done
+            end
+          end
+        rescue StandardError => e
+          @mutex.synchronize { errors << e }
+          request_abort
+        end
+        download_attempts.times { completion_queue.pop }
```
richardwang1124: Is it guaranteed that download_attempts will be less than or equal to completion_queue size?

richardwang1124 (commented):
Nice! Looks pretty good overall. Left a few comments.

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
jterapin: Yeah, I agree that `abort` is cleaner. I was trying to match the `@abort_requested` variable name, but there's no need to be that verbose for the method. I'll simplify it to what you suggested!

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
jterapin: Good catch. We don't need this anymore since we're cleaning up in-flight work and this is single-use.

jterapin: I'll removing it per above.

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
jterapin: Good idea. I'll give it a whirl to see how it pans out.

gems/aws-sdk-s3/lib/aws-sdk-s3/transfer_manager.rb
jterapin: When I initially implemented TransferManager, I considered this but decided against it. 

I need a way to gracefully shutdown executors when a custom one isn't provided, so I scoped the default executors to their own methods. The plan is: when a custom executor is given, we use it but don't shut it down after work finishes. 

For default executors, we treat them as single-use (like the internal classes themselves) and shut them down. They are also quite limited and shouldn't share resources across methods. I need to track this at the TM-level to differentiate between custom and default executors.

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
jterapin: I think I should reorder the normalization/validation here.  You are correct - `File.join` will always use `/` separator based on documentation. 

What I should actually do 
1. Validate that object key does not contain `.` or `..` path segments first
2. Remove s3-prefix if applies
3. `File.join` to combine destination directory and new key path
4. Update file separator if relevant

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
jterapin: Yes, I think the key here is the `ensure` bloc always pushes `:done` even on error

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_progress.rb
jterapin: Good callout. How does Java handle their directory progress?

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_uploader.rb
jterapin: Removing per above.

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_uploader.rb
jterapin: I believe `ensure` will push `:done` regardless what happens.

jterapin: I re-ordered the count to happen right after the work has been posted.

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
```diff
@@ -0,0 +1,221 @@
+# frozen_string_literal: true
+
+module Aws
+  module S3
+    # @api private
+    class DirectoryDownloader
+      def initialize(options = {})
+        @client = options[:client]
+        @executor = options[:executor]
```
alextwoods: I know this is API private, but should we validate that these required options are set? (I can't remember what we do for other similar cases....)

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
```diff
@@ -0,0 +1,221 @@
...
+        download_opts, producer_opts = build_opts(destination, bucket, options)
+        @producer = ObjectProducer.new(producer_opts)
+        downloader = FileDownloader.new(client: @client, executor: @executor)
+        downloads, errors = process_download_queue(downloader, download_opts)
+        build_result(downloads, errors)
+      end
+
+      private
+
+      def build_opts(destination, bucket, opts)
```
alextwoods: nit, feel free to ignore - the download and producer opts feel unrelated in this code - they aren't really sharing anything by being in the same method - I'd probably lean towards having them as seperate methods.

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
```diff
@@ -0,0 +1,221 @@
+# frozen_string_literal: true
+
+module Aws
+  module S3
+    # @api private
+    class DirectoryDownloader
```
alextwoods, line 18: The parallelization and logic of this class is fairly complex - The interaction of queue executor, producer, provided executor/file downloader, ect.  It might be worth documenting that here.

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
```diff
@@ -0,0 +1,221 @@
...
+        downloader.download(entry.path, entry.params)
+        progress&.call(File.size(entry.path))
+      rescue StandardError => e
+        @mutex.synchronize { errors << e }
+        abort unless opts[:ignore_failure]
+      end
+
+      def process_download_queue(downloader, opts)
+        progress = DirectoryProgress.new(opts[:progress_callback]) if opts[:progress_callback]
+        queue_executor = DefaultExecutor.new
```
alextwoods: What are the defaults on the DefaultExector? (ie, how many threads does this create?) Do we want to hard code that default here to make it more explicit and... should this executor be configurable?  Trying to think through cases where someone wants to leverage `async` more instead of using threads.

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_progress.rb
alextwoods: Java actually does not provide any directory upload/download level progress.  

I guess we can leave this as is for now - but its a potential optimization - maybe we should just add a comment about that here.

alextwoods (approved):
LGTM - just minor comments.

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
jterapin: Per our conversation, we lowered the number of the count to `2` to make this lightweight. However, we could consider adding an option in the future if needed.

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_progress.rb
jterapin: Per our conversation, we decided to remove directory progress callbacks due to performance concerns and unresolved design questions. We can revisit this later, potentially after discussing with other SDK teams.

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
jterapin: Agreed, I updated this and other class with further context.

gems/aws-sdk-s3/lib/aws-sdk-s3/directory_downloader.rb
```diff
@@ -83,7 +83,9 @@ def download_object(entry, downloader, errors, opts)
 
         FileUtils.mkdir_p(File.dirname(entry.path)) unless Dir.exist?(File.dirname(entry.path))
         downloader.download(entry.path, entry.params)
+        @logger&.debug("Downloaded #{entry.params[:key]} from #{entry.params[:bucket]} to #{entry.path}")
       rescue StandardError => e
+        @logger&.warn("Failed to download #{entry.params[:key]} from #{entry.params[:bucket]}")
```
alextwoods: add the message from exception?

richardwang1124 (approved):

PR #3305, opened 2025-10-16
Add CREDENTIALS_CODE for explicit profile

richardwang1124:
When resolving credentials in the default chain for a client explicitly configured with a profile, the `CREDENTIALS_CODE` metric should be emitted. This PR adds the `CREDENTIALS_CODE` metric to all `static_profile_` prefixed locations in the default chain.

---


build_tools/services.rb
```diff
@@ -9,10 +9,10 @@ class ServiceEnumerator
     MANIFEST_PATH = File.expand_path('../../services.json', __FILE__)
 
     # Minimum `aws-sdk-core` version for new gem builds
-    MINIMUM_CORE_VERSION = "3.231.0"
+    MINIMUM_CORE_VERSION = "3.234.0"
```
richardwang1124, line 12: Reminder to double check this before merging

gems/aws-sdk-core/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Add `CREDENTIALS_CODE` metric for `static_profile_` methods in default credential chain.
```
jterapin:
```suggestion
* Issue - Add `CREDENTIALS_CODE` metric for `static_profile` methods in default credential chain.
```
Is that a typo? Jw

gems/aws-sdk-core/lib/aws-sdk-core/credential_provider_chain.rb
```diff
@@ -91,7 +103,7 @@ def static_profile_process_credentials(options)
         process_provider = Aws.shared_config.credential_process(profile: options[:config].profile)
         if process_provider
           creds = ProcessCredentials.new([process_provider])
-          creds.metrics << 'CREDENTIALS_PROFILE_PROCESS'
+          creds.metrics << 'CREDENTIALS_PROFILE_PROCESS' << 'CREDENTIALS_CODE'
```
mullermp: Use concat

jterapin: Here's a better way to add to an array:
```
irb(main):001> arr = []
=> []
irb(main):002> arr += ['foo', 'bar']
=> ["foo", "bar"]
irb(main):003> arr
=> ["foo", "bar"]
```

gems/aws-sdk-core/lib/aws-sdk-core/credential_provider_chain.rb
```diff
@@ -55,31 +55,43 @@ def static_credentials(options)
 
     def static_profile_assume_role_web_identity_credentials(options)
       if Aws.shared_config.config_enabled? && options[:config] && options[:config].profile
```
mullermp: If you wanted to improve these methods you can guard clause this and credentials returning.

mullermp (approved):

gems/aws-sdk-core/lib/aws-sdk-core/credential_provider_chain.rb
```diff
@@ -55,31 +55,43 @@ def static_credentials(options)
...
+          creds
+        end
       end
     end
 
     def static_profile_credentials(options)
       if options[:config] && options[:config].profile
         creds = SharedCredentials.new(profile_name: options[:config].profile)
-        creds.metrics = ['CREDENTIALS_PROFILE']
+        creds.metrics << 'CREDENTIALS_PROFILE'
```
jterapin: Do we need to check if creds exists here? Same question for line 106

gems/aws-sdk-core/lib/aws-sdk-core/credential_provider_chain.rb
```diff
@@ -55,31 +55,43 @@ def static_credentials(options)
...
-          region: options[:config].region
-        )
+        with_metrics('CREDENTIALS_CODE') do
+          creds = Aws.shared_config.assume_role_web_identity_credentials_from_config(
+            profile: options[:config].profile,
+            region: options[:config].region
+          )
+          creds.metrics << 'CREDENTIALS_CODE' if creds
+          creds
+        end
```
jterapin: I'm just asking this for my understanding. Why do we also add this metric to `creds.metrics` + have a metrics wrapper with same metric?

jterapin (approved):

gems/aws-sdk-core/CHANGELOG.md
richardwang1124: All the methods start with `static_profile_` so I left the underscore. I can specify that they are `static_profile_` prefixed methods.

gems/aws-sdk-core/lib/aws-sdk-core/credential_provider_chain.rb
richardwang1124: Creds should exist here and for line 106 because we are calling the initializer for SharedCredentials/ProcessCredentials. We needed to check if creds exist for the previous methods because we aren't directly initializing credentials but calling other methods which may or may not return credentials.

gems/aws-sdk-core/lib/aws-sdk-core/credential_provider_chain.rb
richardwang1124: So the `with_metrics` call is there so that when we are resolving assume role web id credentials and call STS's `assume_role_with_web_identity` operation, the request headers have the `CREDENTIALS_CODE` metric. However, just wrapping this block of code with the `with_metrics` doesn't actually add the metric to the features array of the resolved credential itself, so we need to add it to `creds.metrics` so that when we make a real client operation (e.g. `s3.list_buckets`), that request will add all the metrics from `creds.metrics` to the request header.

PR #3323, opened 2025-12-06
Support S3 unsigned body

jterapin:
This PR intends to improve performance for S3's `PutObject` and `UploadPart` .

---


build_tools/services.rb
```diff
@@ -9,10 +9,10 @@ class ServiceEnumerator
     MANIFEST_PATH = File.expand_path('../../services.json', __FILE__)
 
     # Minimum `aws-sdk-core` version for new gem builds
-    MINIMUM_CORE_VERSION = "3.239.1"
+    MINIMUM_CORE_VERSION = "3.240.0"
 
     # Minimum `aws-sdk-core` version for new S3 gem builds
-    MINIMUM_CORE_VERSION_S3 = "3.234.0"
+    MINIMUM_CORE_VERSION_S3 = "3.240.0"
```
jterapin: Self-reminder to update the minimum core version when this is ready for release.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
```diff
@@ -466,52 +462,65 @@ def response_header_to_verify(headers, validation_list)
...
+          @current_chunk.clear
+          @eof = false
         end
 
-        def read(length, buf = nil)
-          # account for possible leftover bytes at the end, if we have trailer bytes, send them
-          if @trailer_io
-            return @trailer_io.read(length, buf)
-          end
+        def read(_length = nil, buf = nil)
```
jterapin: The length parameter is ignored since we always expect 16KB (default) or whatever is set by `@chunk_size`. This is an internal implementation accessed by either our patch or `Net::HTTP`.

I could make this more dynamic but that gets tricky... the `#size` method needs to know the chunk size ahead of time to calculate the overhead bytes plus actual IO size. Supporting variable-length reads would break that calculation.

gems/aws-sdk-core/lib/seahorse/client/net_http/patches.rb
```diff
@@ -6,28 +6,61 @@ module Seahorse
...
+        module RequestPatches
+          # For requests with bodies, Net::HTTP sets a default content type of:
+          #   'application/x-www-form-urlencoded'
+          # There are cases where we should not send content type at all.
+          # Even when no body is supplied, Net::HTTP uses a default empty body
+          # and sets it anyway. This patch disables the behavior when a Thread
+          # local variable is set.
+          # See: https://github.com/ruby/net-http/issues/205
           def supply_default_content_type
             return if Thread.current[:net_http_skip_default_content_type]
```
jterapin, line 25: I think we could eventually get rid of this - given that `net-http` no longer sets the default content type in their recent patch. We may want to be move this in later versions of Ruby when the default net-http gem version is the recent one.

gems/aws-sdk-s3/spec/client_spec.rb
```diff
@@ -167,7 +167,8 @@ module S3
           tmpfile.unlink
           s3 = Client.new(stub_responses: true)
           resp = s3.put_object(bucket: 'bucket', key: 'key', body: tmpfile)
-          expect(resp.context.http_request.body_contents).to eq(data)
+          # Need to discuss
+          expect(resp.context.http_request.body.instance_variable_get(:@io).read).to eq(data)
```
jterapin: I realized the body gets replaced with TrailerIO when we use the unsigned body path. I'm wondering if customers ever access this body object directly? If they do, TrailerIO doesn't have all the methods that StringIO  would have, which could break their code.

One option is to swap the body back to the original IO object after we finish sending the request. That way customers would see the original object if they need access later.

build_tools/customizations.rb
```diff
@@ -153,13 +153,18 @@ def dynamodb_example_deep_transform(subsegment, keys)
...
         # and preserve request uri as /
         if operation['http'] && operation['http']['requestUri']
           operation['http']['requestUri'].gsub!('/{Bucket}', '/')
           operation['http']['requestUri'].gsub!('//', '/')
         end
+
+        next unless %w[PutObject UploadPart].include?(key)
+
+        operation['authType'] = 'v4-unsigned-body'
+        operation['unsignedPayload'] = true
```
jterapin, line 167: I didn't add test cases for unsigned body since its already covered by our test cases here: https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/spec/aws/plugins/checksum_algorithm_spec.rb#L24

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
```diff
@@ -346,16 +343,16 @@ def calculate_request_checksum(context, checksum_properties)
 
         def apply_request_checksum(context, headers, checksum_properties)
           header_name = checksum_properties[:name]
-          body = context.http_request.body_contents
           headers[header_name] = calculate_checksum(
             checksum_properties[:algorithm],
-            body
+            context.http_request.body
```
richardwang1124, line 354: Was this change from `context.http_request.body_contents` to `context.http_request.body` intentional?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
richardwang1124: What would be the benefit of it being dynamic? If we're always expecting 16KB and this is an established expectation, I think it's fine that we don't use the length input. Not sure how feasible this is, but we could consider removing the parameter altogether in the future as to not cause confusion when using this method.

gems/aws-sdk-s3/spec/client_spec.rb
richardwang1124: If it's feasible, it may be better to swap the body back to the original IO object to minimize potentially breaking customers. However, I'm assuming most of these details should be API private?

richardwang1124 (approved):
Nice! I think it looks good overall to me. Added a few comments.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
jterapin: Well, I don't think we could remove the parameter altogether due to the interfaces... this IO-like object is being read by http-client (net-http in our case) and length is a required parameter. Removing the parameter will break.

> What would be the benefit of it being dynamic?

In the scenario where we don't use the patch - net-http could change the way their reads work - like increasing the chunk size and etc. We are bottlenecking the cap to 16KB here. We won't know if things have changed unless net-http checks the length on their end.

gems/aws-sdk-s3/spec/client_spec.rb
jterapin: Should be easy to swap back to the original - given that TrailerIO has access to the original io object. On that question... I would assume so but you never know!

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
jterapin: Great question. And yes! This was the root cause why the memory was high during testing. Calling `#body_contents` on the request loads the entire body into memory. It is unknown on why this have been done previously when I asked around so I think just having that body here makes sense (I do a rewind much later before we start calculating checksums).

alextwoods: Do we need to do a rewind after calculating as well?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
```diff
@@ -466,52 +462,65 @@ def response_header_to_verify(headers, validation_list)
...
-          @digest = ChecksumAlgorithm.digest_for_algorithm(algorithm)
-          @trailer_io = nil
+        def initialize(options = {})
+          @io = options.delete(:io)
+          @location_name = options.delete(:location_name)
+          @algorithm = options.delete(:algorithm)
+          @digest = ChecksumAlgorithm.digest_for_algorithm(@algorithm)
+          @chunk_size = Thread.current[:net_http_override_body_stream_chunk] || MIN_CHUNK_SIZE
+          @overhead_bytes = calculate_overhead(@chunk_size)
+          @max_chunk_size = @chunk_size - @overhead_bytes
```
alextwoods: I find `max_chunk_size` to be a little bit of a confusing name.  Why is it less than chunk size?  This is actually the chunk size that we're reading from the underlying io right?  so maybe like... `base_chunk_size` or `underlying_chunk_size` or something like that?

alextwoods (approved):

gems/aws-sdk-s3/spec/client_spec.rb
alextwoods: Fair to worry about breaking someone with this... but it doesn't seem like part of our API contract that we won't change/manipulate the request body during a request.  In fact, It feels like this should be expected and it seems like it would be better for the contents to match what we actually sent rather than a customer provided operation input.

jterapin: I'm going to leave it as it is. If a customer is accessing something after the call has been made - it should be the response body if anything.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
jterapin: Yeah we should - seems like this was never added so I'll add!

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
jterapin: Oof, naming is hard but I like the `base_chunk_size` so I'll go with that. Yup this is the chunk size of the io.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
```diff
@@ -331,6 +331,11 @@ def calculate_request_checksum(context, checksum_properties)
           if (algorithm_header = checksum_properties[:request_algorithm_header])
             headers[algorithm_header] = checksum_properties[:algorithm]
           end
+
+          # Trailer implementation within JRUBY environment is facing some
+          # network issues that will need further investigation
```
alextwoods: maybe link to the possible open issue?

gems/aws-sdk-s3/lib/aws-sdk-s3/transfer_manager.rb
```diff
@@ -208,6 +208,7 @@ def download_file(destination, bucket:, key:, **options)
       # @option option [Integer] :http_chunk_size (16384) Size in bytes for each chunk when streaming request bodies
       #   over HTTP. Controls the buffer size used when sending data to S3. Larger values may improve throughput by
       #   reducing the number of network writes, but use more memory. Custom values must be at least 16KB.
+      #   Only Ruby MRI is supported.
```
alextwoods, line 211: The value is still being set on the thread and the patch is being applied right? So the low level http chunk size would still be applied in both cruby and jruby?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
```diff
@@ -466,52 +467,87 @@ def response_header_to_verify(headers, validation_list)
...
+          @current_chunk = +''
+          @eof = false
         end
 
-        def read(length, buf = nil)
-          # account for possible leftover bytes at the end, if we have trailer bytes, send them
-          if @trailer_io
-            return @trailer_io.read(length, buf)
+        def read(length = nil, buf = nil)
+          return @io.read unless length
```
alextwoods: I'm a little confused about this - this means we won't use AWS chunked encoding when we're not given a length?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
```diff
@@ -466,52 +467,87 @@ def response_header_to_verify(headers, validation_list)
...
+        end
+
+        private
+
+        def calculate_overhead(chunk_size)
+          chunk_size.to_s(16).size + 4 # hex_length + "\r\n\r\n"
+        end
+
+        def fill_chunk
+          chunk = @io.read(@base_chunk_size)
```
alextwoods: What happens here is fewer (but non-zero) bytes than the base_chunk_size are returned? (I can't 100% remember Ruby, but I think IO objects are allowed to return <= the requested length when the rest of the bytes aren't yet available)

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
jterapin: I'm unsure if these two are related but I think they helped me to narrow down the issue to SSL so i'll link those two.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
jterapin: I think I was trying to make set of specific tests to pass. 

An actual fix will be if `length` is not given, I should resort to using the minimum chunk size. I also made some edits to fix the expectations on those tests as well.

gems/aws-sdk-s3/lib/aws-sdk-s3/transfer_manager.rb
jterapin: Yeah the value is still being set - I could check for JRUBY version here to ensure that it doesn't (unnecessary thread local variable). That being said, Checksum plugin checks at runtime to determine if the thread local need to be used or not.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
jterapin: Returning fewer bytes are allowed. From what I recall (and observed), returning `nil` signals that we are at EOL.

https://ruby-doc.org/3.4.1/IO.html#method-i-read
> Opens the stream, reads and returns some or all of its content, and closes the stream; returns nil if no bytes were read.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
```diff
@@ -466,52 +469,93 @@ def response_header_to_verify(headers, validation_list)
...
+
+          while !@eof && fill_data?(required_length)
+            chunk = @io.read(@base_chunk_size)
+            if chunk && !chunk.empty?
+              @digest.update(chunk)
+              @encoded_buffer << "#{chunk.bytesize.to_s(HEX_BASE)}\r\n#{chunk}\r\n"
+            else
+              @encoded_buffer << "0\r\n#{trailer_string}\r\n\r\n"
+              @eof = true
+              break
```
alextwoods: nit: since the while loop has a condition on @eof, you probably don't need the break here

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
```diff
@@ -5,6 +5,7 @@ module Plugins
     # @api private
     class ChecksumAlgorithm < Seahorse::Client::Plugin
       CHUNK_SIZE = 1 * 1024 * 1024 # one MB
+      MIN_CHUNK_SIZE = 16_384 # 16 KB
```
alextwoods: I find these constant names a bit confusing - I think we should be more specific about what they are for?
If I'm understanding/remembering correctly, `CHUNK_SIZE` is the size for checksum/digest computation - so maybe `CHECKSUM_CHUNK_SIZE` and `MIN_CHUNK_SIZE` is being used as the default trailer chunk size? so maybe `DEFAULT_TRAILER_CHUNK_SIZE`

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
```diff
@@ -466,52 +469,93 @@ def response_header_to_verify(headers, validation_list)
...
-          n_full_chunks = orig_body_size / CHUNK_SIZE
-          partial_bytes = orig_body_size % CHUNK_SIZE
-          chunked_body_size = n_full_chunks * (CHUNK_SIZE + 8)
-          chunked_body_size += partial_bytes.to_s(16).size + partial_bytes + 4 unless  partial_bytes.zero?
+          n_full_chunks = orig_body_size / @base_chunk_size
+          partial_bytes = orig_body_size % @base_chunk_size
+
+          full_chunk_overhead = @base_chunk_size.to_s(HEX_BASE).size + CHUNK_OVERHEAD
+          chunked_body_size = n_full_chunks * (@base_chunk_size + full_chunk_overhead)
+          chunked_body_size += partial_bytes.to_s(HEX_BASE).size + partial_bytes + 4 unless partial_bytes.zero?
```
alextwoods: nit, why hard code the `4` here? should it be `CHUNK_OVERHEAD`?

alextwoods (approved):

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
```diff
@@ -466,52 +469,93 @@ def response_header_to_verify(headers, validation_list)
       # Wrapper for request body that implements application-layer
       # chunking with Digest computed on chunks + added as a trailer
       class AwsChunkedTrailerDigestIO
```
alextwoods, line 470: Wasn't really sure where to leave this comment - but this class is super complex.  I know its indirectly tested through other things, but I think given the complexity we may want to add specs for it by itself to test different cases of length/chunk size

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
jterapin: I like those names better. I'll update.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
jterapin: Good catch! Meant to use the const.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb
jterapin: Added in this commit: https://github.com/aws/aws-sdk-ruby/pull/3323/commits/49e82f0538c7d8c7fdf02665dbaa2eb78dd2511a

alextwoods (approved):

PR #3336, opened 2026-01-02
Normalize endpoints in ClockSkew retry plugin to prevent memory leak

### LINKED ISSUE #3332: Memory leak in Aws::Plugins::Retries::ClockSkew for long-running applications
### Describe the bug

The `Aws::Plugins::Retries::ClockSkew` class causes unbounded memory growth in long-running applications due to storing every unique endpoint URL in hash tables that never get cleaned up.

The ClockSkew class maintains two hashes that grow indefinitely:

```ruby
@endpoint_clock_corrections = Hash.new(0)
@endpoint_estimated_skews = Hash.new(nil)
```

These hashes use the full endpoint URL as keys, including paths and query parameters. For applications making many API calls to different resources (e.g., different S3 objects), this creates a new hash entry for every unique URL, leading to memory exhaustion.

- **My application**: 40,000 S3 operations per 10 minutes with unique object keys
- **Result**: OOM crash within 38 hours due to hash growth
- **Scope**: Affects all AWS SDK clients (S3, DynamoDB, Lambda, etc.)

### Root Cause
Clock skew is a server-level property, not a URL-level property. AWS servers synchronize time at the host level, so tracking skew per `scheme://host:port` is sufficient and correct.

### Regression Issue

### Expected Behavior

- Clock skew should be tracked per server (host), not per individual API endpoint
- Memory usage should remain bounded regardless of API call diversity
- Hash size should be limited to the number of unique AWS hosts/regions accessed
- Applications should run indefinitely without memory leaks from clock skew tracking
- Example: All S3 operations should share the same entry: `https://mybucket.s3.amazonaws.com:443`

### Current Behavior

- Each unique endpoint URL (including path and query parameters) creates a permanent entry in `@endpoint_clock_corrections` and 
`@endpoint_estimated_skews` hashes
- Memory usage grows linearly with the number of unique API endpoints accessed
- Hash entries are never removed or cleaned up
- Long-running applications eventually crash with OOM errors
- Example: S3 operations on different objects create entries like:
 

  `https://mybucket.s3.amazonaws.com/path/to/object1.jpg`
  `https://mybucket.s3.amazonaws.com/path/to/object2.jpg`
  `https://mybucket.s3.amazonaws.com/path/to/object3.jpg`

### Reproduction Steps

```ruby
require 'aws-sdk-s3'

# Create S3 client
client = Aws::S3::Client.new(region: 'us-east-1', retry_mode: 'standard')

# Get reference to the ClockSkew instance
clock_skew = client.config.clock_skew

puts "Initial hash sizes:"
puts "Clock corrections: #{clock_skew.instance_variable_get(:@endpoint_clock_corrections).size}"
puts "Estimated skews  : #{clock_skew.instance_variable_get(:@endpoint_estimated_skews).size}"

# Simulate many different S3 object operations
1000.times do |i|
  # Each operation creates a unique endpoint URL
  begin
    client.head_object(bucket: 'test-bucket', key: "object-#{i}.mp3")
  rescue Aws::S3::Errors::ServiceError
    # Ignore errors
  end
end

puts "\nAfter 1000 operations:"
puts "Clock corrections: #{clock_skew.instance_variable_get(:@endpoint_clock_corrections).size}"
puts "Estimated skews  : #{clock_skew.instance_variable_get(:@endpoint_estimated_skews).size}"

# Show the actual keys being stored
skews = clock_skew.instance_variable_get(:@endpoint_estimated_skews)
puts "\nSample hash keys (first 3):"
skews.keys.first(3).each { |key| puts "  #{key}" }
```

### Possible Solution

Normalize endpoints to `scheme://host:port` before using as hash keys:

```ruby
def normalize_endpoint(endpoint)
  uri = endpoint.is_a?(URI::Generic) ? endpoint : URI(endpoint.to_s)
  "#{uri.scheme}://#{uri.host}:#{uri.port}"
end

def clock_correction(endpoint)
  @mutex.synchronize { @endpoint_clock_corrections[normalize_endpoint(endpoint)] }
end

def estimated_skew(endpoint)
  @mutex.synchronize { @endpoint_estimated_skews[normalize_endpoint(endpoint)] }
end

def update_estimated_skew(context)
  endpoint    = context.http_request.endpoint
  now_utc     = Time.current
  server_time = server_time(context.http_response)

  return unless server_time

  @mutex.synchronize do
    @endpoint_estimated_skews[normalize_endpoint(endpoint)] = server_time - now_utc
  end
end

def set_clock_correction(endpoint, correction)
  @mutex.synchronize { @endpoint_clock_corrections[normalize_endpoint(endpoint)] = correction }
end
```

### Benefits
- **Bounded memory usage**: Hash size limited to number of unique AWS hosts/regions
- **Maintains functionality**: Clock skew tracking remains accurate
- **Cross-service compatibility**: Works for all AWS services
- **Backward compatible**: No API changes required

I am an Amazon employee, although not in AWS, so please reach out if you like. I am more than happy to share my current implementation details that seem to work or provide a PR if you like me to.

aws-sdk-core

MRI Ruby 3.3, any OS any platform (arm & amd)

johvet:

## Fix memory leak in ClockSkew retry plugin

Fixes #3332

### Problem
Long-running applications with high request volume (1000+ ops/sec) experience memory exhaustion due to unlimited growth in clock skew correction hashes. Each unique URL path creates a separate hash entry instead of sharing corrections per server.

### Solution  
- Normalize endpoints to `scheme://host:port` format, removing paths/queries
- Add graceful handling for malformed URIs
- Maintain backward compatibility for edge cases

### Testing
- Comprehensive edge case coverage (malformed URIs, IPv6, case variations)
- Memory leak prevention verified through internal state inspection

## Further opportunities
While this solution normalizes the scheme provided by endpoint URLs, it does not do that for the host name. In other words, https://example.com and https://EXAMPLE.com would still be treated as 2 different endpoint servers. This could be easily tackled by using Ruby's `URI().normalize` which already honors [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986), but that comes at the cost of extra object allocations and is likely scope creep for this PR so I decided to leave that up to the team for a later decision.


richardwang1124 (approved):
Looks good! Thanks for your contribution!

PR #3340, opened 2026-01-06
Add Catapult PR GitHub workflow

richardwang1124:
Adds Catapult GitHub PR support. Workflow triggers automatically when a PR is approved, and `workflow_dispatch` allows the workflow to be trigger manually through GitHub Actions.

---


.github/workflows/pull-request-build.yaml
```diff
@@ -0,0 +1,54 @@
+name: Build SDK
+on:
+  pull_request_review:
+    types: [ submitted ]
+    branches: [ main ]
```
jterapin: I'm guessing the branch that we should mention should be `version-3` rather than `main`?

.github/workflows/pull-request-build.yaml
```diff
@@ -0,0 +1,54 @@
...
+  DOWNLOAD_FOLDER: '.build-scripts/'
+  SCRIPT_LOCATION: 'workflows/start-pull-request-build/pull-request-build-v1.sh'
+
+  # custom variables
+  IAM_ROLE_ARN: 'arn:aws:iam::664418967014:role/AwsSdkRubyGitHubRole' # dev account role for testing
+  ROLE_SESSION_DURATION_SECONDS: 7200
+
+jobs:
+  aws-sdk-pr-build:
+    if: github.event.review.state == 'approved'
```
jterapin: I think it will be fine to run when it is not in approve state yet - for testing purposes.

.github/workflows/pull-request-build.yaml
```diff
@@ -0,0 +1,54 @@
+name: Build SDK
+on:
+  pull_request_review:
+    types: [ submitted ]
```
jterapin: Might want to look over other implementations - they used different activity types and we might want to consider those as well.

richardwang1124: We could edit the workflow so that it runs for any open PR. My main concern for this would be that every time a new commit is pushed the workflow runs again, and these start up CodeBuild jobs which might be resource expensive. I feel like this workflow should only be run when the PR is in a pretty finalized state. For now I think I'll modify this to be like the CPP one so we can test, but afterwards we should consider how often this should run.

.github/workflows/pull-request-build.yaml
richardwang1124: Good catch, I'll update this.

jterapin (approved):

.github/workflows/pull-request-build.yaml
```diff
@@ -0,0 +1,55 @@
...
+    branches: [ version-3 ]
+  workflow_dispatch:
```
richardwang1124, line 6: Adding in this configuration should allow us to trigger this workflow manually once this PR is merged.

jterapin (approved):
Looks great! Thank you for working on this.

jterapin (approved):

PR #3341, opened 2026-01-06
Correctly tests US-ASCII encoding in S3 200 errors

jterapin:

### DIFF
--- a/gems/aws-sdk-s3/spec/plugins/http_200_errors_spec.rb
+++ b/gems/aws-sdk-s3/spec/plugins/http_200_errors_spec.rb
@@ -33,7 +33,7 @@ module Aws
             .to raise_error(Seahorse::Client::NetworkingError, /Empty or incomplete response body/)
         end
 
-        it 'gracefully handle non-UTF encoding' do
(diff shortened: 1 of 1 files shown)


gems/aws-sdk-s3/spec/plugins/http_200_errors_spec.rb
```diff
@@ -44,14 +44,16 @@ module Plugins
             </DeleteResult>
           XML
 
+          allow_any_instance_of(Seahorse::Client::Http::Response)
```
richardwang1124, line 47: Any reason why this is allow instead of expect?

richardwang1124 (approved):
Looks good, left a comment.

gems/aws-sdk-s3/spec/plugins/http_200_errors_spec.rb
jterapin: WebMock automatically normalizes all response bodies to UTF-8, which prevents us from testing encoding edge cases. I need to mock the HTTP response body at a lower level to simulate the problematic encoding scenarios we're trying to fix.

PR #3352, opened 2026-01-29
Improve error message for providers

### LINKED ISSUE #3350: Improve error message when ECS credentials endpoint returns HTTP 429 (rate limited)
### Describe the feature

When the ECS task metadata credentials endpoint returns HTTP 429 (rate limited), the Ruby SDK raises Aws::Errors::MissingCredentialsError. in my case the complete error was:

```
Error Class: Aws::Errors::MissingCredentialsError
Message: unable to sign request without credentials set

FULL STACK TRACE:

0: /usr/local/bundle/gems/aws-sdk-core-3.241.4/lib/aws-sdk-core/plugins/sign.rb:145:in `rescue in sign'
1: /usr/local/bundle/gems/aws-sdk-core-3.241.4/lib/aws-sdk-core/plugins/sign.rb:136:in `sign'
2: /usr/local/bundle/gems/aws-sdk-core-3.241.4/lib/aws-sdk-core/plugins/sign.rb:48:in `call'
3: /usr/local/bundle/gems/aws-sdk-core-3.241.4/lib/aws-sdk-core/plugins/transfer_encoding.rb:27:in `call'
4: /usr/local/bundle/gems/aws-sdk-core-3.241.4/lib/aws-sdk-core/plugins/helpful_socket_errors.rb:12:in `call'
5: /usr/local/bundle/gems/aws-sdk-core-3.241.4/lib/aws-sdk-core/plugins/retry_errors.rb:365:in `block in call'
6: /usr/local/bundle/gems/aws-sdk-core-3.241.4/lib/aws-sdk-core/plugins/user_agent.rb:92:in `metric'
7: /usr/local/bundle/gems/aws-sdk-core-3.241.4/lib/aws-sdk-core/plugins/retry_errors.rb:385:in `with_metric'
8: /usr/local/bundle/gems/aws-sdk-core-3.241.4/lib/aws-sdk-core/plugins/retry_errors.rb:365:in `call'
9: /usr/local/bundle/gems/aws-sdk-core-3.241.4/lib/aws-sdk-core/plugins/http_checksum.rb:20:in `call'
```

This is misleading because credentials are not missing—they are being rate-limited by the ECS agent. misleading my troubleshooting efforts

The Python SDK (boto3/botocore) provides a much clearer error message that includes the HTTP status code and the actual response from the endpoint.

Python SDK (for comparison):

```
botocore.exceptions.CredentialRetrievalError: Error when retrieving credentials from container-role: Error retrieving metadata: Received non 200 response 429 from container metadata: You have reached maximum request limit.

```

The ECS agent logs do not help either as there were no rate limiting errors in the agent logs. I had to run a tcp dump command below to troubleshoot the issue:

```
sudo tcpdump -i any -A port 80 and host 169.254.170.2 2>/dev/null | grep -E "HTTP/1\.[01] (200|429|503)"

```

### Use Case

When running applications on Amazon ECS with IAM task roles, the ECS agent enforces rate limiting on the credentials endpoint (http://169.254.170.2/v2/credentials/) via the ECS_TASK_METADATA_RPS_LIMIT configuration (default: 40 steady, 60 burst requests per second).

When this rate limit is exceeded, the ECS agent returns HTTP 429 with the body: You have reached maximum request limit.

The current Ruby SDK error message leads developers to debug credential provider chain configuration burning hours investigating the wrong problem When the actual issue is simply rate limiting, which can be resolved by Caching/reusing SDK clients instead of creating new ones per request or adjusting ECS_TASK_METADATA_RPS_LIMIT on the ECS agent.

References
[1] https://github.com/aws/amazon-ecs-agent/blob/master/README.md#:~:text=ECS_TASK_METADATA_RPS_LIMIT

### Proposed Solution

Include HTTP status code and response body in the error message 

### Other Information

_No response_

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### SDK version used

3.241.4

### Environment details (OS name and version, etc.)

amazonlinux 2023

richardwang1124:
Fixes https://github.com/aws/aws-sdk-ruby/issues/3350.

---

### DIFF
--- a/gems/aws-sdk-core/lib/aws-sdk-core/ecs_credentials.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/ecs_credentials.rb
@@ -22,7 +22,7 @@ module Aws
         @status_code = status_code
         @body = body
         msg = "HTTP #{status_code}"
-        msg += ": #{body[0..200]}" if body && !body.empty?
+        msg += ": #{body}" if body && !body.empty?
         super(msg)
       end
     end
--- a/gems/aws-sdk-core/lib/aws-sdk-core/ecs_credentials.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/ecs_credentials.rb
@@ -15,7 +15,17 @@ module Aws
     include RefreshingCredentials
 
     # @api private
-    class Non200Response < RuntimeError; end
+    class Non200Response < RuntimeError
+      attr_reader :status_code, :body
+
+      def initialize(status_code, body = nil)
+        @status_code = status_code
+        @body = body
+        msg = "HTTP #{status_code}"
+        msg += ": #{body[0..200]}" if body && !body.empty?
+        super(msg)
+      end
+    end
 
     # Raised when the token file cannot be read.
     class TokenFileReadError < RuntimeError; end
@@ -251,7 +261,7 @@ module Aws
       request = Net::HTTP::Get.new(path)
       set_authorization_token(request)
       response = connection.request(request)
-      raise Non200Response unless response.code.to_i == 200
+      raise Non200Response.new(response.code.to_i, response.body) unless response.code.to_i == 200
 
       response.body
     end
--- a/gems/aws-sdk-core/CHANGELOG.md
+++ b/gems/aws-sdk-core/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Feature - Include HTTP status code and body in errors whehn retrieving ECS credentials and Instance Profile credentials.
+
 3.241.4 (2026-01-16)
 ------------------
 
--- a/gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
+++ b/gems/aws-sdk-core/lib/aws-sdk-core/instance_profile_credentials.rb
@@ -26,7 +26,17 @@ module Aws
     include RefreshingCredentials
 
     # @api private
-    class Non200Response < RuntimeError; end
+    class Non200Response < RuntimeError
+      attr_reader :status_code, :body
+
+      def initialize(status_code, body = nil)
+        @status_code = status_code
+        @body = body
+        msg = "HTTP #{status_code}"
+        msg += ": #{body[0..200]}" if body && !body.empty?
+        super(msg)
+      end
+    end
 
     # @api private
     class TokenRetrivalError < RuntimeError; end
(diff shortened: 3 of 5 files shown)


gems/aws-sdk-core/lib/aws-sdk-core/ecs_credentials.rb
```diff
@@ -15,7 +15,17 @@ class ECSCredentials
...
     # @api private
-    class Non200Response < RuntimeError; end
+    class Non200Response < RuntimeError
+      attr_reader :status_code, :body
+
+      def initialize(status_code, body = nil)
+        @status_code = status_code
+        @body = body
+        msg = "HTTP #{status_code}"
+        msg += ": #{body[0..200]}" if body && !body.empty?
```
alextwoods: I assume the ..200 is to truncate large bodies/limit the size of the message?

If so, may want to use a constant for 200 - otherwise at a quick glance it looks like it matches the 200 response code.

I can't remember do we do truncation like this anywhere else?  We might want to make it clear when the body is truncated (something like "..." or "[truncated]") - otherwise, it might look like an incomplete response from the service was part of the problem?

gems/aws-sdk-core/lib/aws-sdk-core/ecs_credentials.rb
```diff
@@ -251,7 +261,7 @@ def http_get(connection, path)
       request = Net::HTTP::Get.new(path)
       set_authorization_token(request)
       response = connection.request(request)
-      raise Non200Response unless response.code.to_i == 200
+      raise Non200Response.new(response.code.to_i, response.body) unless response.code.to_i == 200
```
alextwoods, line 264: just to double check - `.body` will always give a string and not an IO right?

alextwoods (approved):

gems/aws-sdk-core/lib/aws-sdk-core/ecs_credentials.rb
richardwang1124: Yes, I believe it will always return a string. From https://ruby-doc.org/3.3.2/stdlibs/net/Net/HTTP.html regarding Net::HTTP.get: 
> Sends a GET request and returns the HTTP response body as a string.

gems/aws-sdk-core/lib/aws-sdk-core/ecs_credentials.rb
richardwang1124: Yeah I added it in case error messages got really long but I don't think the truncation is necessary. I'll remove it.

Issue #3359, opened 2026-03-05
aws-sdk-s3 1.214.0 autoloads aws-sdk-s3/directory_progress but file is missing


marckohlbrugge:
### Describe the bug

In `aws-sdk-s3` `1.214.0`, requiring the gem raises:

`LoadError: cannot load such file -- aws-sdk-s3/directory_progress`

`lib/aws-sdk-s3/customizations.rb` includes:

```ruby
autoload :DirectoryProgress, 'aws-sdk-s3/directory_progress'
```

but `lib/aws-sdk-s3/directory_progress.rb` is not present in the packaged gem.

This also breaks Rails production boot when using `aws-sdk-rails` eager loading.

### Reproduction

```bash
ruby -v
# ruby 3.4.8

gem install aws-sdk-s3 -v 1.214.0
ruby -e 'gem "aws-sdk-s3", "1.214.0"; require "aws-sdk-s3"'
```

Output:

```text
cannot load such file -- aws-sdk-s3/directory_progress (LoadError)
```

### Additional context

- Downgrading to `aws-sdk-s3` `1.213.0` resolves the issue.
- In `1.213.0`, `customizations.rb` does not reference `DirectoryProgress`.

jterapin:
Hey Marc! Thanks for the issue. This is a miss on our part. I'll send out a fix today.

jterapin:
This is fixed in aws-sdk-s3 v1.215.0: https://rubygems.org/gems/aws-sdk-s3/versions/1.215.0

Please confirm this resolves the issue for you.

marckohlbrugge:
@jterapin All good now!

jterapin:
Great, I'll close the issue now. Feel free to reach out if you need anything else.

PR #3360, opened 2026-03-05
Fix LoadError in TM directory operations

jterapin:
#3359 

### DIFF
--- a/build_tools/services.rb
+++ b/build_tools/services.rb
@@ -12,7 +12,7 @@ module BuildTools
     MINIMUM_CORE_VERSION = "3.241.4"
 
     # Minimum `aws-sdk-core` version for new S3 gem builds
-    MINIMUM_CORE_VERSION_S3 = "3.241.4"
+    MINIMUM_CORE_VERSION_S3 = "3.243.0"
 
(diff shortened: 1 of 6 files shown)


alextwoods (approved):

build_tools/services.rb
```diff
@@ -12,7 +12,7 @@ class ServiceEnumerator
     MINIMUM_CORE_VERSION = "3.241.4"
 
     # Minimum `aws-sdk-core` version for new S3 gem builds
-    MINIMUM_CORE_VERSION_S3 = "3.241.4"
+    MINIMUM_CORE_VERSION_S3 = "3.243.0"
```
richardwang1124, line 15: From the changelog it looks like this is an issue. The current core version is 3.242.0, should this be 3.242.1 instead?

richardwang1124 (approved):
Left one comment about the min core version.

build_tools/services.rb
jterapin: Good question. In the S3's side - yes it is an issue but I would consider the additional user agent tracking to be a feature. Based on past changelog entries for any changes we made regarding user agent plugin, they usually fell under feature.

PR #3362, opened 2026-03-17
Fix waiter error matcher to handle boolean acceptors

jterapin:
Some waiter models (Lambda `FunctionExists`, ACM-PCA `CertificateAuthorityCSRCreated`, `CertificateIssued`) define the error matcher's expected value as a boolean (`false`) instead of a string (`"false"`).

The `matches_error?` method in Poller only matched against string values in its case/when, so boolean expected values fell through to the else branch, causing the acceptor to silently never match. This resulted in unnecessary retries until max attempts.

This change adds boolean literals alongside the existing string variants in the when clauses so both types are handled correctly.

---

### DIFF
--- a/build_tools/services.rb
+++ b/build_tools/services.rb
@@ -9,10 +9,10 @@ module BuildTools
     MANIFEST_PATH = File.expand_path('../../services.json', __FILE__)
 
     # Minimum `aws-sdk-core` version for new gem builds
-    MINIMUM_CORE_VERSION = "3.241.4"
+    MINIMUM_CORE_VERSION = "3.243.1"
 
     # Minimum `aws-sdk-core` version for new S3 gem builds
-    MINIMUM_CORE_VERSION_S3 = "3.243.0"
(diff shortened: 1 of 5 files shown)


build_tools/services.rb
```diff
@@ -9,10 +9,10 @@ class ServiceEnumerator
     MANIFEST_PATH = File.expand_path('../../services.json', __FILE__)
 
     # Minimum `aws-sdk-core` version for new gem builds
-    MINIMUM_CORE_VERSION = "3.241.4"
+    MINIMUM_CORE_VERSION = "3.243.1"
 
     # Minimum `aws-sdk-core` version for new S3 gem builds
-    MINIMUM_CORE_VERSION_S3 = "3.243.0"
+    MINIMUM_CORE_VERSION_S3 = "3.243.1"
```
jterapin, line 15: I believe a min core bump is necessary for all services to take this fix.

richardwang1124 (approved):

PR #3364, opened 2026-03-27
Fix require_https_for_sse_cpk option not being honored

### LINKED ISSUE #2519: `require_https_for_sse_cpk` option does not work
  
**Describe the bug**
I noticed this issue when trying  SSE-C in my local environment using minio.
`Aws::S3::Client` constructor can receive `require_https_for_sse_cpk` optional argument as below, but I think this option does not work.

https://github.com/aws/aws-sdk-ruby/blob/02a9af737c0a10ad354f16c65658439e8e11599e/gems/aws-sdk-s3/lib/aws-sdk-s3/client.rb#L280-L283

I think `Aws::S3::Plugins::SseCpk#require_https` method needs to be modified.
https://github.com/aws/aws-sdk-ruby/blob/a82c8981c95a8296ffb6269c3c06a4f551d87f7d/gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/sse_cpk.rb#L43-L53

I think it would be better to modify as follows.

```ruby
          def require_https(context)
            return unless context.config.require_https_for_sse_cpk

            unless URI::HTTPS === context.config.endpoint
              msg = <<-MSG.strip.gsub("\n", ' ')
                Attempting to send customer-provided-keys for S3
                server-side-encryption over HTTP; Please configure a HTTPS
                endpoint. If you are attempting to use a test endpoint,
                you can disable this check via `:require_https_for_sse_cpk`
              MSG
              raise ArgumentError, msg
            end
          end
```

'aws-sdk-s3'

ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-darwin19]

Steps to reproduce the behavior (please share code or minimal repo)
```ruby
[1] pry(main)> client = Aws::S3::Client.new(region: 'ap-northeast-1', credentials: xxx, require_https_for_sse_cpk: false)
=> #<Aws::S3::Client>
[2] pry(main)> bucket = Aws::S3::Resource.new(client: client).bucket('bucket_name')
=> #<Aws::S3::Bucket:0x00007fa59a4a70d0
 @arn=nil,
 @client=#<Aws::S3::Client>,
 @data=nil,
 @name="bucket_name",
 @resolved_region="ap-northeast-1",
 @waiter_block_warned=false>
[3] pry(main)> object = bucket.object('object_name')
=> #<Aws::S3::Object:0x00007fa59a3fdcd8
 @bucket_name="bucket_name",
 @client=#<Aws::S3::Client>,
 @data=nil,
 @key="object_name",
 @waiter_block_warned=false>
[4] pry(main)> object.presigned_url(:put, sse_customer_algorithm: 'AES256', sse_customer_key: 'customer_key')
ArgumentError: Attempting to send customer-provided-keys for S3                 server-side-encryption over HTTP; Please configure a HTTPS                 endpoint. If you are attempting to use a test endpoint,                 you can disable this check via `:require_https_for_sse_cpk`
from /Users/nkuroda/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/aws-sdk-s3-1.94.0/lib/aws-sdk-s3/plugins/sse_cpk.rb:51:in `require_https'
```

**Expected behavior**
Successful issuance of SSE-C type presigned_url

**Screenshots**

**Additional context**

monkeyWzr:
Fix #2519 (not sure why it is closed).

The `require_https_for_sse_cpk` config option was documented as a way to disable the HTTPS enforcement for SSE-CPK operations, but the `require_https` handler method never checked it — the guard always raised regardless of the configured value.

Gate the HTTPS check on `context.config.require_https_for_sse_cpk` so setting it to `false` actually allows HTTP endpoints (e.g. for local development against services like MinIO).

Changes:
  - Fix `SseCpk` plugin to honor `require_https_for_sse_cpk` config
  - Add specs for HTTPS enforcement, opt-out behavior, and key encoding


gems/aws-sdk-s3/spec/plugins/sse_cpk_spec.rb
```diff
@@ -0,0 +1,138 @@
+# frozen_string_literal: true
```
richardwang1124, line 1: Can you move this test under aws-sdk-s3/spec/plugins instead?

gems/aws-sdk-s3/lib/aws-sdk-s3/plugins/sse_cpk.rb
```diff
@@ -41,7 +41,8 @@ def compute_key_md5(context)
           end
 
           def require_https(context)
-            unless URI::HTTPS === context.config.endpoint
+            if context.config.require_https_for_sse_cpk &&
```
richardwang1124: Nit: Can you use a guard clause instead? Maybe something like
```
return unless context.config.require_https_for_sse_cpk

unless URI::HTTPS === context.config.endpoint
  # ... raise ArgumentError
end
```

richardwang1124 (commented):
Thanks for opening a PR! Overall the changes look good, left a few comments that need to be addressed.

gems/aws-sdk-s3/spec/plugins/sse_cpk_spec.rb
monkeyWzr: Oops sorry didn't notice I messed up with the path 😅 Fixed!

richardwang1124 (approved):
Looks good! Thanks again for your contribution!

PR #3379, opened 2026-05-12
feat: Support new 2026 retries

richardwang1124:
This PR adds support for new 2026 retries behavior, including changing the default retry mode, updating retry quotas, and introducing new behavior for long-polling operations and retry backoff headers.

---


gems/aws-sdk-core/lib/aws-sdk-core/plugins/retry_errors.rb
```diff
@@ -147,10 +157,10 @@ class RetryErrors < Seahorse::Client::Plugin
         :correct_clock_skew,
         default: true,
         doc_type: 'Boolean',
-        docstring: <<-DOCS) do |cfg|
-Used only in `standard` and adaptive retry modes. Specifies whether to apply
-a clock skew correction and retry requests with skewed client clocks.
-      DOCS
+        docstring: <<~DOCS) do |cfg|
+          Used only in `standard` and adaptive retry modes. Specifies whether to apply
```
jterapin: Backticks to keep things consistent

build_tools/services.rb
```diff
@@ -9,10 +9,10 @@ class ServiceEnumerator
     MANIFEST_PATH = File.expand_path('../../services.json', __FILE__)
 
     # Minimum `aws-sdk-core` version for new gem builds
-    MINIMUM_CORE_VERSION = "3.247.0"
+    MINIMUM_CORE_VERSION = "3.248.0"
```
richardwang1124, line 12: Note to self to check this before merging.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/retry_errors.rb
```diff
@@ -169,31 +179,40 @@ def self.resolve_retry_mode(cfg)
...
+          return value
+        end
+
+        # TODO: Remove gate and keep only the new retries branch
+        if RetryErrors.new_retries?
+          service_id = cfg.api.metadata['serviceId'] if cfg.respond_to?(:api)
+          ['DynamoDB', 'DynamoDB Streams'].include?(service_id) ? 4 : 3
+        else
+          3
         end
```
jterapin: Could extract this out to its own helper function, like:
```
def default_max_attempts(cfg)
    return 3 unless new_retries?

    service_id = cfg.api.metadata['serviceId'] if cfg.respond_to?(:api)
    ['DynamoDB', 'DynamoDB Streams'].include?(service_id) ? 4 : 3
end
```

gems/aws-sdk-core/lib/aws-sdk-core/plugins/retry_errors.rb
```diff
@@ -311,9 +351,51 @@ def retryable?(context, response, error_inspector)
...
+          backoff_duration = [retry_after, exp_backoff].max
+          [backoff_duration, exp_backoff + 5].min
+        end
+
+        # TODO: Remove gate, remove default nil params, keep only new retries branch
+        def exponential_backoff(retries, error_inspector = nil, service_id = nil)
+          if RetryErrors.new_retries?
+            backoff_scalar = if error_inspector.throttling_error?
+                               1
+                             elsif ['DynamoDB', 'DynamoDB Streams'].include?(service_id)
```
jterapin: We can make a quick helper check since we do this multiple times - like `dynamodb_service?(cfg)`

gems/aws-sdk-core/lib/aws-sdk-core/plugins/retry_errors.rb
```diff
@@ -10,6 +10,18 @@ module Aws
...
     class RetryErrors < Seahorse::Client::Plugin
+      # TODO: Remove this gate and hardcode new retry behavior once
+      # AWS_NEW_RETRIES_2026 is enabled by default, which includes:
+      # - Default retry_mode to 'standard'
+      # - Default max_attempts to 4 for DynamoDB
+      # - Remove the old retries branch in Handler#call
+      # - Remove the old retries branch in #exponential_backoff
+      # - Remove LEGACY_RETRY_COST and TIMEOUT_RETRY_COST from RetryQuota
+      # @api private
+      def self.new_retries?
```
jterapin: We have a quite a bit class methods here - we might as well do `class << self` to group there.

gems/aws-sdk-core/CHANGELOG.md
```diff
@@ -1,7 +1,9 @@
 Unreleased Changes
 ------------------
+
 * Feature - Add YJIT & ZJIT tracking to user agent.
 * Issue - Fix error messaging in SSO OIDC.
+* Feature - Add new retry behavior behind `AWS_NEW_RETRIES_2026` environment variable. When enabled, defaults to `standard` retry mode, service-specific tuning for DynamoDB, long-polling backoff support, and `x-amz-retry-after` header support.
```
jterapin: This changelog entry is a dense read with internal details so maybe:

```
* Feature - Add `AWS_NEW_RETRIES_2026` environment variable to opt-in to updated
  `standard` retry mode with reduced backoff intervals.
```

We may want to consider having changelog entries in DyanmoDB/DynamoDBStreams - customers who uses these service gems may not necessarily look at the core gem updates. I think we do generate a default "take a look at core for generated changes" but we could have more meaningful change entry for these services

gems/aws-sdk-core/lib/aws-sdk-core/plugins/retry_errors.rb
```diff
@@ -228,6 +247,14 @@ class Handler < Seahorse::Client::Handler
         # Max backoff (in seconds)
         MAX_BACKOFF = 20
 
+        # Hard-coded combination of services and operations as having the
+        # longPoll trait. To be removed when trait is enabled.
+        LONG_POLLING_OPERATIONS = {
```
jterapin, line 258: I'm not comfortable with service-specific logic living in core. I recognize this pattern exists in V3, but we should be more intentional about this in V4.

gems/aws-sdk-core/spec/aws/plugins/retries/retry_quota_spec.rb
```diff
@@ -17,30 +17,66 @@ module Plugins
...
-          expect(checked_out_capacity).to eq(Retries::RetryQuota::TIMEOUT_RETRY_COST)
+          it 'returns 0 when there is insufficient capacity' do
+            retry_quota.instance_variable_set(:@available_capacity, 1)
+
+            expect(retry_quota.checkout_capacity(error)).to eq(0)
+          end
         end
+      end
+
+      # TODO: Remove this context when new retries become default
```
jterapin, line 51: yeah i like that we have these todos for quick removals when time is ready

jterapin (approved):
Nice - you should check with everyone else when these changes should be going out. Also is there a blog post or documentation on how new standard retries work?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/retry_errors.rb
richardwang1124: I can define a constant which contains the list of DynamoDB related services, I think the check is simple enough that we don't need any method overhead.

gems/aws-sdk-core/lib/aws-sdk-core/json/handler.rb
```diff
@@ -35,6 +35,9 @@ def build_body(context)
 
       def parse_response(response)
         response.data = parse_body(response.context)
+      rescue Json::ParseError => e
+        # make JSON parsing errors on 200-range responses retryable
+        response.error = Seahorse::Client::NetworkingError.new(e)
```
jterapin: This is a new addition from last review so I have some questions:
* Does other parser handlers need the same treatment? Such as xml? Is there a reason why JSON is being specifically targeted here? 
* If this is the final attempt, does this NetworkingErr gets surfaced to the user? Will this be helpful to the user?

gems/aws-sdk-core/lib/aws-sdk-core/plugins/retry_errors.rb
```diff
@@ -163,71 +161,105 @@ class RetryErrors < Seahorse::Client::Plugin
...
+          value == 'true'
         end
-        value == 'true'
       end
 
       class Handler < Seahorse::Client::Handler
         # Max backoff (in seconds)
         MAX_BACKOFF = 20
 
+        # Hard-coded combination of services and operations as having the
```
jterapin, line 256: Should tag here as `TODO:` so that we can easily surface it.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/retry_errors.rb
```diff
@@ -311,9 +360,51 @@ def retryable?(context, response, error_inspector)
...
+                             end
+            Kernel.rand * [backoff_scalar * 2**retries, MAX_BACKOFF].min
+          else
+            [Kernel.rand * 2**retries, MAX_BACKOFF].min
+          end
+        end
+
+        def parse_retry_after(context)
+          retry_after = context.http_response.headers['x-amz-retry-after']
+          return nil unless retry_after
```
jterapin: just `return` will be nil - no need to be explicit.

gems/aws-sdk-core/lib/aws-sdk-core/plugins/retry_errors.rb
```diff
@@ -311,9 +360,51 @@ def retryable?(context, response, error_inspector)
...
+
+          unless retry_after.match?(/\A\d+\z/)
+            context.config.logger&.debug(
+              "Failed to parse x-amz-retry-after header value: #{retry_after.inspect}"
+            )
+            return nil
```
jterapin: ditto comment.

gems/aws-sdk-core/CHANGELOG.md
```diff
@@ -1,9 +1,13 @@
 Unreleased Changes
 ------------------
 
+* Feature - Add `AWS_NEW_RETRIES_2026` environment variable to opt-in to updated `standard` retry mode with reduced backoff intervals.
```
jterapin, line 4: Nice, have we considered having a similar changelog entries for DynamoDB since they have specific retryable behavior or is that not needed? 

Now that I think about it - are service-specific behaviors documented anywhere?

jterapin: I also have a question on the users who ARE already on `standard` mode for their `retry_mode` - do they feel any changes from this update?

gems/aws-sdk-core/lib/aws-sdk-core/json/handler.rb
richardwang1124: The SCP tests didn't cover the other parser handlers, but I believe they should also be updated. I'll make the changes.

With these changes, the error message raised on the final attempt will be the NetworkingErr, but it wraps the parsing error message so the root cause will still be surfaced.

gems/aws-sdk-core/CHANGELOG.md
richardwang1124: Can I add a DynamoDB changelog entry even though only core was updated? We could extend the core changelog entry to mention that DynamoDB defaults are changed if new retry behavior is enabled. I don't believe any service specific behaviors are documented anywhere yet. Externally I think the blogpost should mention this new DynamoDB behavior, internally I could add more comments or documentation?

Customers who are already on standard and opt in to new retries will feel a difference. Due to the updated backoff timing, retries will be much faster. Throttling behavior will be the same, and due to the updated retry quota draining, customers will fail faster during sustained service errors, but this is intentional to help services recover faster.

gems/aws-sdk-core/lib/aws-sdk-core/json/handler.rb
jterapin: > With these changes, the error message raised on the final attempt will be the NetworkingErr, but it wraps the parsing error message so the root cause will still be surfaced.

Is this how other SDKs are handling this case? Labeling it as a service error. Just curious 🤔

gems/aws-sdk-core/CHANGELOG.md
jterapin: > DynamoDB 

I personally feel that blogpost is not enough for documentation. Not everyone will read blogpost releases nor `aws-sdk-core`'s CHANGELOG entries.


Rethinking this...
This might be a good use case where we should have service-specific plugins. With plugins, you can be specific about this behavior + documentation. Now that I think about this - we might need to do something about the autogenerated config. See:
> :max_attempts (Integer) — default: 3 — An integer representing the maximum number attempts that will be made for a single request, including the initial attempt. For example, setting this value to 5 will result in a request being retried up to 4 times. Used in standard and adaptive retry modes. 

Above is what I see when I run codegen. Let's talk offline.


> Customers who are already on standard ... will feel a difference. 

We should probably add a separate entry about this. The way I read the above entry is like: "Ok so if I don't use that env var, i'm still on old `standard` retries mechanism"

richardwang1124: Sure I can try adding a DynamoDB plugin instead for retries.

> We should probably add a separate entry about this. The way I read the above entry is like: "Ok so if I don't use that env var, i'm still on old standard retries mechanism"

Your original understanding is correct, customers who are already on standard **and opt in to new retries** will feel a difference. If they do not set the environment variable, there will not be any differences. New retry behavior is disabled by default and opt in only.

jterapin (approved):
LGTM!

jterapin (approved):

PR #3381, opened 2026-05-12
Fix Ruby 3.2+ compatibility in aws-sdk-neptunedata detailed_message overrides

ricardogpsf:
## Summary

In Ruby 3.2, `Exception#full_message` was updated to internally call `detailed_message(highlight:)` passing a keyword argument ([source](https://github.com/ruby/ruby/blob/v3_3_0/error.c#L1491)). All 34 error classes in `aws-sdk-neptunedata` override `detailed_message` to expose the API-level `detailed_message` field, but none accepted the `highlight:` keyword — causing an `ArgumentError` whenever Ruby or tooling calls `full_message` on these exceptions.

A common trigger is OpenTelemetry SDK's `span.record_exception(e)` ([code here](https://github.com/open-telemetry/opentelemetry-ruby/blob/e851ad8fe3acdd22dea55e6c31a57e3057c81370/sdk/lib/opentelemetry/sdk/trace/span.rb#L199)), which calls `e.full_message(highlight: false, order: :top)` ([code here](https://github.com/open-telemetry/opentelemetry-ruby/blob/e851ad8fe3acdd22dea55e6c31a57e3057c81370/sdk/lib/opentelemetry/sdk/trace/span.rb#L199)) to capture the stacktrace. This silently breaks tracing and any instrumentation that records exceptions on spans.

## Change

Added `highlight: true, **` to all 34 `detailed_message` overrides in `errors.rb` and updated the corresponding `.rbs` type signatures.

```ruby
# Before
def detailed_message
  @data[:detailed_message]
end

# After
def detailed_message(highlight: true, **)
  @data[:detailed_message]
end
```

## Backward Compatibility

Fully backward compatible. Keyword arguments with defaults are valid Ruby syntax since 2.0. In Ruby < 3.2, `full_message` never calls `detailed_message` with keyword arguments, so the added parameter is never passed — no behavior change.

## Note on Code Generation

`errors.rb` and `sig/errors.rbs` are generated files. The corresponding generator template will also need to be updated to preserve this fix on future regeneration.

### DIFF
--- a/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/error_list.rb
+++ b/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/error_list.rb
@@ -13,10 +13,12 @@ module AwsSdkCodeGenerator
         # excluding event shapes marked as error
         if error_struct?(shape)
           members = shape['members'].inject([]) do |arr, (k, v)|
+            member_name = Underscore.underscore(k)
             arr << {
-              name: Underscore.underscore(k),
+              name: member_name,
               type: Docstring.ucfirst(v['type'] ||'String'),
-              shared: k.downcase == 'message' || k.downcase == 'code'
+              shared: k.downcase == 'message' || k.downcase == 'code',
+              detailed_message: member_name == 'detailed_message'
             }
             arr
           end
--- a/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/rbs/error_list.rb
+++ b/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/rbs/error_list.rb
@@ -11,10 +11,14 @@ module AwsSdkCodeGenerator
         @errors = @api['shapes'].inject([]) do |es, (name, shape)|
           if error_struct?(shape)
             members = shape["members"].map do |member_name, member_body|
-              MethodSignature.new(
-                method_name: Underscore.underscore(member_name),
-                overloads: ["() -> #{Docstring.ucfirst(member_body['type'] ||'::String')}"]
-              )
+              method_name = Underscore.underscore(member_name)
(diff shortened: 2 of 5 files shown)


jterapin:
Hey, thanks for the PR! This is a tricky case where we have a name collision between an API member name and a Ruby built-in method. The fix as proposed will still always return the API member value; meaning when Ruby's `full_message` calls `detailed_message(highlight: true)` internally, it will render the API member value rather than Ruby's formatted error message output. 

We are currently looking into a few different solutions and will circle back with a resolution.

ricardogpsf:
> Hey, thanks for the PR! This is a tricky case where we have a name collision between an API member name and a Ruby built-in method. The fix as proposed will still always return the API member value; meaning when Ruby's full_message calls detailed_message(highlight: true) internally, it will render the API member value rather than Ruby's formatted error message output.
> 
> We are currently looking into a few different solutions and will circle back with a resolution.

NP, thanks for the feedback. Actually this PR should be a draft since I still have no good solution, and realized (as you said) doing those changes will not solve the issue properly but just silence a not ideal behavior (at least would not break the code). I'm patching it in my application since it's braking there.
Anyway, was good you already seen it and you're already working on it too. Thank you for it. I'll close this PR then. Do you want me opening an Issue just for tracking and other people would be able to know you're aware of it?

jterapin:
Yes please to closing the pr and creating an issue. That would be great. I will post updates in the issue once we come to a resolution.

ricardogpsf:
> Yes please to closing the pr and creating an issue. That would be great. I will post updates in the issue once we come to a resolution.

Issue created: https://github.com/aws/aws-sdk-ruby/issues/3383

thanks for paying attention to this. =)

ricardogpsf:
Closing this PR in favor of discussing the solution or getting info from aws team in this issue: https://github.com/aws/aws-sdk-ruby/issues/3383

PR #3382, opened 2026-05-13
feat: RBS size reduction

sichanyoo:

### Changes & Impact
- Reduces size of `aws-sdk-<service>/sig/` folder in generated service gems by factoring out repeatedly used nested input struct definitions into `sig/params.rbs` as typealiases that get reused in `sig/client.rbs`.
  - The type definitions already defined under `sig/types.rbs` cannot be reused because that would make methods no longer accept hash literals.
- The biggest beneficiary of this change is `aws-sdk-quicksight`. Its `sig/` folder at the time of this PR (`aws-sdk-quicksight v1.179.0`) was 35.5MB, but after this refactor it's reduced to 0.925MB, which is a reduction of about 97.3%. With other big services like EC2 and S3, reduction is much less due to those services primarily using flat input shape structure. EC2 gets about 5% reduction to its `sig/` (3.34MB => 3.17MB) and S3 gets about 1.5% reduction to its `sig/` (445KB => 439KB).
- With this change, method auto-complete suggestions for QuickSight methods actually render instead of being shortened as `...` in RubyMine.
- Gem build time doesn't seem to be impacted based on local, brief manual testing for QuickSight (running build commands and not seeing even a tick of difference AFAICT).
- `InputTypeAliasCollector` walks through shapes, caching names of shapes within operation input shapes that appear more than once & would end up taking more than 5 lines in rendered RBS if it were to be generated fully without typealias. It then returns topologically sorted shapes so that leaf shapes (shapes without dependencies) can get rendered first in RBS, allowing latter shapes to simply refer to leaf shapes in their own definitions.
  - Because current changes determine a shape needs aliasing based on how long it would be if it was fully expanded, there is a possibility of unnecessary typealias generation in a special case where a large shape actually becomes 1~2 lines long after child typealias substitutions so it doesn't need to be aliased. But customers would need to refer to `params.rbs` to look at child typealias definitions anyways & couple extra type alises don't hurt, so additional complexity isn't worth it.


build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/rbs/input_type_alias_collector.rb
```diff
@@ -0,0 +1,156 @@
...
+        when 'map'
+          walk_map(value_shape, ancestors)
+        end
+      end
+
+      def rendered_rbs_line_count_heuristic(shape_name, visited = Set.new)
+        return 0 if visited.include?(shape_name)
+
+        # Cache results to deduplicate calculation for nested structure types that get used multiple times in the model.
+        @size_cache ||= {}
```
jterapin: Appears to be redundant since it was done at class init?

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/rbs/keyword_argument_builder.rb
```diff
@@ -8,11 +8,13 @@ class KeywordArgumentBuilder
...
       attr_reader :newline
 
-      def initialize(api:, shape:, newline:, options: {})
+      def initialize(api:, shape:, newline:, options: {}, aliased_shapes: Set.new, alias_namespace: nil)
         @api = api
         @shape = shape
         @newline = newline
         @options = options
+        @aliased_shapes = aliased_shapes
+        @alias_namespace = alias_namespace
```
jterapin: The signature are getting long here. We can do `options = {}` hash and deconstruct the has inside. Something like:
```ruby
def initialize(options = {})
  @api = options[:api]
  ...
```
We have some examples on this pattern in our codebase.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/rbs/keyword_argument_builder.rb
```diff
@@ -69,10 +71,16 @@ def struct_member(struct, member_name, member_ref, i, visited, keyword:)
       def ref_value(ref, i, visited)
         if visited.include?(ref['shape'])
           return "untyped"
-        else
-          visited  = visited + [ref['shape']]
         end
```
jterapin: style: 
```
return "untyped" if visited.include?(ref['shape'])
```

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/rbs/params.rb
```diff
@@ -0,0 +1,43 @@
...
+            builder = AwsSdkCodeGenerator::RBS::KeywordArgumentBuilder.new(
+              api: @api,
+              shape: shape,
+              newline: true,
+              aliased_shapes: aliased_set - [shape_name],
+              alias_namespace: 'Params',
+            )
+            {
+              'name' => Underscore.underscore(shape_name),
+              'definition' => builder.struct(shape, '      ', []),
```
jterapin: I'm not crazy with calling `struct` directly here. Looking at the prior version of `KeywordArgumentBuilder`, every external caller uses `.format` so I think `.struct` meant to be an internal detail so we could do the following:

```ruby
def format_as_alias(indent: '')
  struct(@shape, indent, [])
end
```

Benefits of this:
* `struct`, `struct_members`, `ref_value`, etc. can all move under `private`
* Alias callers don't need to know about the visited array or indent format
* Clear two-method public API: `format` for method signatures, `format_as_alias for` type alias bodies

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/rbs/client_class.rb
```diff
@@ -60,6 +61,8 @@ def operations
                 api: @api,
                 shape: input_shape,
                 newline: true,
+                aliased_shapes: @aliased_shapes,
+                alias_namespace: 'Params',
```
jterapin: I might be off base here, but it looks like `alias_namespace` is only ever `'Params'` or `nil` so effectively functioning as a boolean flag. If so, could we do a simple boolean and default value will be `false`?

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/code_builder.rb
```diff
@@ -114,8 +114,17 @@ def rbs_files(options = {})
         prefix = options.fetch(:prefix, '')
         codegenerated_plugins = codegen_plugins(prefix)
         unless @service.h2_required_setting?
```
jterapin, line 125: These changes are done for regular Client but what about AsyncClient?

jterapin (commented):
Looking great so far. I'm curious to see the overall size reduction. Could you calculate the before vs after for all of our service gems?  I think this will be great to highlight once this gets to prod.

As for the CodeQL failure, will this failure persist after this merge? If so, we should find a way to suppress (if not useful) or find a fix for it.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/code_builder.rb
sichanyoo: Oh oops, good catch. I didn't know Ruby published separate client type for bidirectional streaming lol. Added to async clients as well as pointed out.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/rbs/client_class.rb
sichanyoo: Instead of substituting with a flag, I just removed that entirely and used string literal in the keyword argument builder when it swaps shape name with type alias.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/views/rbs/params.rb
sichanyoo: Good point, changed as suggested (put private keyword below `format_as_alias` and `format` so all other methods are now private).

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/rbs/keyword_argument_builder.rb
sichanyoo: Changed as recommended.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/rbs/keyword_argument_builder.rb
sichanyoo: Good point, I just merged the two new fields into existing options field.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/rbs/input_type_alias_collector.rb
sichanyoo: Good catch, forgot to remove it after moving it off to initializer. Removed.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/rbs/input_type_alias_collector.rb
```diff
@@ -0,0 +1,155 @@
...
+            rendered_rbs_line_count_heuristic(child['member']['shape'], visited) : 1
+        when 'map'
+          value_shape = @api['shapes'][child['value']['shape']]
+          # If it's a map with structure member, add 2 lines & recurse.
+          #   (one line each for 'Hash[::String, {' and '}]' in rendered RBS)
+          # If it's a map with primitive member, add 1 and return
+          #   (e.g., renders as 1-liner, like 'map: Hash[::String, ::String]')
+          value_shape['type'] == 'structure' ? 2 +
+            rendered_rbs_line_count_heuristic(child['value']['shape'], visited) : 1
+        else 1
```
jterapin, line 153: Needs to align with case/when blocks. For the files you touch, I'll recommend running `rubocop` ([docs](https://rubocop.org/)) to catch these. I.e. 
```
rubocop <file-path.rb>
```

If the suggested changes makes sense, you can do: `rubocop <file-path.rb> -a` to automatically apply the fixes.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/rbs/input_type_alias_collector.rb
```diff
@@ -0,0 +1,155 @@
...
+      # Leaf dependencies come first so aliases can reference each other.
+      # params.rbs uses aliases in this list to define RBS type aliases.
+      # client_class.rbs uses aliases in this list to deduplicate content.
+      def shapes_to_alias
+        count_shape_usage
+        aliased = @shape_usage_count.select do |shape_name, count|
+          shape = @api['shapes'][shape_name]
+          shape['type'] == 'structure' &&
+            count > 1 &&
+            rendered_rbs_line_count_heuristic(shape_name) > 5
```
jterapin: nit: Could turn this `5` into a const, i.e. `MIN_LINES_FOR_ALIAS` or add a small `# min lines for alias` 

i recognized that the `5` is mentioned in the pr description but we could self-document here.

jterapin (approved):
Looks great. I would recommend that you run `rake sdk:build` to see the codegenerated outputs are exactly you expected.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/rbs/input_type_alias_collector.rb
jterapin: Disregard! I mislooked.

build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/rbs/input_type_alias_collector.rb
sichanyoo: Changed as suggested; used a constant in place of a int literal.

jterapin (approved):

PR #3384, opened 2026-05-14
fix: Resolve Exception#detailed_message name collision

### LINKED ISSUE #3383: Name collision of `detailed_message` method between Exception classes in `aws-sdk-neptunedata` gem and Ruby 3.2+
### Describe the bug

In Ruby 3.2, `Exception#full_message` was updated to internally call `detailed_message(highlight:)` (a Ruby built-in method) passing a keyword argument ([code here](https://github.com/ruby/ruby/blob/v3_3_0/error.c#L1491)). All 34 error classes in `aws-sdk-neptunedata` define a `detailed_message` ([code here](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-neptunedata/lib/aws-sdk-neptunedata/errors.rb#L185-L187)) to expose the API-level `detailed_message` field (consequently, that method definition becomes a method override when using ruby 3.2+), but none accepted the `highlight:` keyword, causing an `ArgumentError` whenever Ruby or tooling calls `full_message` on these exceptions.

A real example of that error happening is in OpenTelemetry SDK's `span.record_exception(e)` ([code here](https://github.com/open-telemetry/opentelemetry-ruby/blob/e851ad8fe3acdd22dea55e6c31a57e3057c81370/sdk/lib/opentelemetry/sdk/trace/span.rb#L199)), which calls `e.full_message(highlight: false, order: :top)` ([code here](https://github.com/open-telemetry/opentelemetry-ruby/blob/e851ad8fe3acdd22dea55e6c31a57e3057c81370/sdk/lib/opentelemetry/sdk/trace/span.rb#L199)) to capture the stacktrace. This silently breaks tracing and any instrumentation that records exceptions on spans since raises the error.

As a temporary fix, I've overridden the `detailed_message` in all error classes in `aws-sdk-neptunedata` as monkey patch to also accept right parameters, so it avoids the exception being raised when calling the `detailed_message`, so my code can work properly.

### Regression Issue

### Expected Behavior

A way to make the exceptions classes from API work well with Ruby 3.2+ updates. 

### Current Behavior

There is a name collision of method `detailed_message` between  error classes in `aws-sdk-neptunedata` API and Ruby 3.2+ built-in base exception class, so it raises an ArgumentError exception when the method is called with parameters.

### Reproduction Steps

Instantiate an exception/error class from `aws-sdk-neptunedata`, call the method `full_message` (buil-in in ruby 3.2+), it'll raise an ArgumentError.

### Possible Solution

Change the method name in the API from `detailed_message` to any other. 🤷  It'll require a new major version of the gem. =/
At least it'll keep the API behavior consistent and keep the ruby exception behavior as it is.

Overrride explicit the `detailed_message` with a a specific behavior for the API. It'll keep backward compatibilities (a minor/patch bump), but will change the default behavior of the ruby exception message, anyway a good documentation could make it ok for users.

Well, looking forward to know how you folks will decide to resolve that issue. =)

_No response_

aws-sdk-neptunedata in any version with detailed_message method defined in error classes

Ruby 3.2+

jterapin:
Fixes #3383 

---

## Summary
- Fix name collision between generated `detailed_message` accessor on NeptuneData error classes and Ruby 3.2+ `Exception#detailed_message(highlight:, **)`
- Generated accessor now uses kwargs-based dispatch: no-arg calls return the API field, calls with keyword arguments (e.g., from `full_message`) delegate to `super`
- Adds `exception_method` flag to error codegen, separate from existing `shared` flag

## Context
NeptuneData is the only AWS service whose error shapes have a `detailedMessage` member. The generated zero-arg `def detailed_message` shadowed Ruby 3.2+'s `Exception#detailed_message(highlight:, **)`, causing `ArgumentError` when `full_message` is called (triggered by OpenTelemetry, logging frameworks, etc.).

## Testing
Tested locally with Ruby 3.0 and Ruby 4.0
Expected code-generated changes: https://github.com/aws/aws-sdk-ruby/commit/99ec98044bc41130278a48a8a2b250847e2d9b25

### DIFF
--- a/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/error_list.rb
+++ b/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/error_list.rb
@@ -3,6 +3,10 @@
 module AwsSdkCodeGenerator
   class ErrorList
 
+    EXCEPTION_KWARGS_METHODS = %w[
+      detailed_message
+    ].freeze
+
     include Enumerable
 
     def initialize(options)
@@ -13,10 +17,12 @@ module AwsSdkCodeGenerator
         # excluding event shapes marked as error
         if error_struct?(shape)
           members = shape['members'].inject([]) do |arr, (k, v)|
+            member_name = Underscore.underscore(k)
             arr << {
-              name: Underscore.underscore(k),
+              name: member_name,
               type: Docstring.ucfirst(v['type'] ||'String'),
-              shared: k.downcase == 'message' || k.downcase == 'code'
+              shared: k.downcase == 'message' || k.downcase == 'code',
+              exception_method: EXCEPTION_KWARGS_METHODS.include?(member_name)
             }
             arr
           end
--- a/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/rbs/error_list.rb
+++ b/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/rbs/error_list.rb
@@ -11,9 +11,18 @@ module AwsSdkCodeGenerator
         @errors = @api['shapes'].inject([]) do |es, (name, shape)|
           if error_struct?(shape)
             members = shape["members"].map do |member_name, member_body|
+              underscored = Underscore.underscore(member_name)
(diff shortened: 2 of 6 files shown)


sichanyoo (approved):
Nice!

richardwang1124 (approved):
LGTM!

PR #3406, opened 2026-07-08
Handle document types in json builder

richardwang1124:
Fixes BigDecimal values in document types being serialized as JSON strings (e.g., "0.425e2") instead of JSON numbers (42.5) for services using REST-JSON and JSON protocols.

---


build_tools/services.rb
```diff
@@ -9,10 +9,10 @@ class ServiceEnumerator
     MANIFEST_PATH = File.expand_path('../../services.json', __FILE__)
 
     # Minimum `aws-sdk-core` version for new gem builds
-    MINIMUM_CORE_VERSION = "3.248.0"
+    MINIMUM_CORE_VERSION = "3.253.0"
```
richardwang1124: Check this before merging

gems/aws-sdk-core/lib/aws-sdk-core/json/builder.rb
```diff
@@ -56,10 +56,26 @@ def format(ref, value)
         when TimestampShape then timestamp(ref, value)
         when BlobShape      then encode(value)
         when FloatShape     then Util.serialize_number(value)
+        when DocumentShape  then format_document(value)
         else value
         end
       end
 
+      def format_document(value)
```
jterapin: Nit: To follow other private methods, `document(value)` might make more sense here?

gems/aws-sdk-core/lib/aws-sdk-core/json/builder.rb
```diff
@@ -56,10 +56,26 @@ def format(ref, value)
...
+        case value
+        when Hash
+          value.transform_values { |v| format_document(v) }
+        when Array
+          value.map { |v| format_document(v) }
+        when BigDecimal
```
jterapin: `bigdecimal` is not required at the aws-sdk-core entrypoint so i would require it at the top like `base64`

jterapin: Should we handle `Infinity`/`NaN` the same way Float does? 🤔

gems/aws-sdk-core/spec/aws/json/builder_spec.rb
```diff
@@ -74,6 +74,41 @@ def json(params)
         expect(json(string: 'abc', integer: nil)).to eq('{"String":"abc"}')
       end
 
+      describe 'document types' do
```
jterapin, line 77: Slightly off-topic but will there be protocol test coverages for this? I'm guessing its planned but not done yet.

gems/aws-sdk-core/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Feature - Serialize `BigDecimal` values as JSON numbers in document types.
```
jterapin: We should mention that this is scoped specifically for JSON protocols.

jterapin (approved):
Nice, the overall approach makes sense. Once the feedbacks are addressed, this is ready to merge.

gems/aws-sdk-core/lib/aws-sdk-core/json/builder.rb
```diff
@@ -56,10 +56,26 @@ def format(ref, value)
...
         else value
         end
       end
 
+      def format_document(value)
+        case value
+        when Hash
+          value.transform_values { |v| format_document(v) }
+        when Array
+          value.map { |v| format_document(v) }
```
jterapin: What are the expected changes here for these cases? Now that we do extra work on Document types that falls under these types.

richardwang1124: The main change is that now for documents we'll walk through them and convert BigDecimals to Floats.

gems/aws-sdk-core/spec/aws/json/builder_spec.rb
richardwang1124: Yeah that's a good question, since Kevin said there will be a SEP later regarding BigDecimal I would think there might be protocol test coverage later.

jterapin (approved):
LGTM!

PR #3407, opened 2026-07-09
fix: Address unbounded memory growth in upload_stream when source outpaces upload

### LINKED ISSUE #3393: Object#upload_stream retains unbounded memory when the source outpaces the upload
### Version

aws-sdk-s3 1.225.0 (Ruby 3.4)

### Problem

`Object#upload_stream` holds memory proportional to the *whole* object when the block writes faster than parts upload: a fast local source feeding a slower S3 sink. That defeats the purpose of a streaming upload.

### Cause

`MultipartStreamUploader#upload_with_executor` reads each part into a `StringIO` (`read_to_part_body`) and submits it through `DefaultExecutor#post`, which appends to an unbounded `Queue` and returns immediately (non-blocking). The internal `IO.pipe` backpressures the *writer* against the *reader*, but the reader then offloads into that unbounded queue, so nothing bounds read-ahead to the `~max_threads` parts that are actually uploading. Unsent 5 MB `StringIO`s accumulate up to roughly object size minus bytes already uploaded.

`tempfile: true` masks it (parts spill to disk), but the default is in-memory and unbounded.

### Impact

Streaming a multi-GB on-disk file spikes RSS by GBs. Concretely, Active Storage's `S3Service` archives large files via `upload_stream { |out| IO.copy_stream(file, out) }`; a 12 GB file drove ~3.6 GB of RSS and OOM'd a 4 GB worker. `upload_file` (lazy `FilePart`, reads from disk on demand) is unaffected under the same conditions; that divergence is the tell.

### Suggested fix

Bound read-ahead to the worker count: a `SizedQueue`, or have `post` block while all workers are busy. (`DefaultExecutor` arrived with the #3302 executor refactor; #1824 previously tuned `upload_stream` memory.)

richardwang1124:
Fixes unbounded memory growth in `upload_stream` when the source produces data faster than parts can be uploaded. Previously every part read off the pipe was buffered with no limit. Now `DefaultExecutor` supports a bounded queue that applies backpressure and `upload_stream` sets a limit to `thread_count`. Also switches in memory read path from `copy_stream` to single right-sized read.

Fixes https://github.com/aws/aws-sdk-ruby/issues/3393.

---


gems/aws-sdk-s3/lib/aws-sdk-s3/default_executor.rb
```diff
@@ -12,7 +12,7 @@ class DefaultExecutor
       def initialize(options = {})
         @max_threads = options[:max_threads] || DEFAULT_MAX_THREADS
         @state = RUNNING
-        @queue = Queue.new
+        @queue = SizedQueue.new(@max_threads)
```
jterapin: I think its worthwhile to offer `@max_queue` as a separate option and pass in a reasonable default to `:max_queue` when creating an executor within the scope of the streaming uploader.

gems/aws-sdk-s3/lib/aws-sdk-s3/default_executor.rb
```diff
@@ -25,9 +25,9 @@ def post(*args, &block)
         @mutex.synchronize do
           raise 'Executor has been shutdown and is no longer accepting tasks' unless @state == RUNNING
 
-          @queue << [args, block]
           ensure_worker_available
         end
+        @queue << [args, block]
```
jterapin: You moved the enqueue out of the mutex to avoid holding the lock while parked but what happens `shutdown` or `kill` is called? There is also a gap between `@state` check and the push: what happens when `shutdown`/`kill` slips into that gap?

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_stream_uploader.rb
```diff
@@ -113,18 +113,21 @@ def complete_opts(options)
       def read_to_part_body(read_pipe)
         return if read_pipe.closed?
 
-        temp_io = @tempfile ? Tempfile.new('aws-sdk-s3-upload_stream') : StringIO.new(String.new)
-        temp_io.binmode
-        bytes_copied = IO.copy_stream(read_pipe, temp_io, @part_size)
-        temp_io.rewind
-        if bytes_copied.zero?
-          if temp_io.is_a?(Tempfile)
+        if @tempfile
```
jterapin, line 116: For my own understanding: What's the reason for the split here? The tempfile branch needs `copy_stream` (streams to disk, no heap copy), but the old code used `copy_stream` for the in-memory path too...was there a reason to switch it to read?

jterapin (commented):
Thanks for picking this up! this is a real gap and a good catch. I've left a few questions on the current implementation inline. We're also missing a `CHANGELOG` entry, so we'll want to add one before this merges.

gems/aws-sdk-s3/lib/aws-sdk-s3/default_executor.rb
richardwang1124: Good callout, I think I'll try closing the queue instead.

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_stream_uploader.rb
richardwang1124: I'll add a comment here in the code too to clarify.

`IO.copy_stream` doesn't know how big the part will be, so it grows `StringIO`'s string as it goes, doubling from 1 to 2 to 4 to 8. A 5MB part ends up holding an 8MB buffer and the 1, 2, and 4 MB buffers are discarded, fragmenting the heap. `read` asks for the exact size so it's one allocation and doesn't run into the fragmentation issue. I ran some tests with the existing `copy_stream` approach vs the `read` approach and you can see that the peak process RSS levels off higher.

|                          | 30 parts | 120 parts | 480 parts |
|--------------------------|----------|-----------|-----------|
| copy_stream + SizedQueue | 162 MB   | 381 MB    | 382 MB    |
| read + SizedQueue        | 47 MB    | 48 MB     | 52 MB     |

The tempfile branch keeps copy_stream on purpose since it streams to disk instead of memory and doesn't run into this issue.

gems/aws-sdk-s3/lib/aws-sdk-s3/default_executor.rb
```diff
@@ -11,8 +11,10 @@ class DefaultExecutor
 
       def initialize(options = {})
         @max_threads = options[:max_threads] || DEFAULT_MAX_THREADS
+        @max_queue = options[:max_queue] || 0
         @state = RUNNING
-        @queue = Queue.new
+        # A bounded queue applies backpressure to producers when full. 0 means unbounded.
```
jterapin: Could remove this comment. Seems self-explanatory.

gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
```diff
@@ -405,7 +407,10 @@ def public_url(options = {})
       # @see Client#upload_part
       def upload_stream(options = {}, &block)
         upload_opts = options.merge(bucket: bucket_name, key: key)
-        executor = DefaultExecutor.new(max_threads: upload_opts.delete(:thread_count))
+        thread_count = upload_opts.delete(:thread_count) || DefaultExecutor::DEFAULT_MAX_THREADS
+        # A bounded queue prevents the source from reading ahead without limit when it
+        # produces data faster than parts can be uploaded.
+        executor = DefaultExecutor.new(max_threads: thread_count, max_queue: thread_count)
```
jterapin, line 413: What's the rationale for keeping the queue size same as thread count?

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_stream_uploader.rb
```diff
@@ -113,18 +113,24 @@ def complete_opts(options)
...
+          temp_io = Tempfile.new('aws-sdk-s3-upload_stream')
```
jterapin, line 117: This just occurred to me. Is it important to make the file name unique or no?

gems/aws-sdk-s3/lib/aws-sdk-s3/default_executor.rb
```diff
@@ -25,10 +27,15 @@ def post(*args, &block)
         @mutex.synchronize do
           raise 'Executor has been shutdown and is no longer accepting tasks' unless @state == RUNNING
 
-          @queue << [args, block]
           ensure_worker_available
         end
+        # Pushed outside the mutex because a bounded queue blocks the caller when
+        # full and holding the lock while parked would deadlock #shutdown and #kill.
+        @queue.push([args, block])
```
jterapin, line 33: I have some thoughts on this but i think offline discussion is better for this. Will DM you.

jterapin (commented):
The overall approach makes sense. I would like to see a fuller PR description for historical purposes (aside from the issue link).

gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
mashhurs: @jterapin we have been using `Aws::S3::TransferManager#upload_file` in [`logstash-output-s3` plugin](https://github.com/logstash-plugins/logstash-integration-aws/blob/24f35166801a2ef85e4ed046ea501aee235bc33e/lib/logstash/outputs/s3/uploader.rb#L41) and recently what I found is, native memory increases when using the S3 output. I am not super clear if this change is related but when I capture the source path with async-profiler (`jemalloc` env w/o auto cleaning retained memory) I do see ` customizations/object.rb` path. Can you please help me understanding if my issue also relates to this?

```
Logstash S3 output uploader
  outputs/s3/uploader.rb  upload_async → upload
   └─ aws-sdk-s3-1.192.0  customizations/object.rb  upload_file
      └─ aws-sdk-s3-1.192.0  file_uploader.rb  upload → put_object → open_file   (RubyIO.open)
         └─ aws-sdk-s3-1.192.0  client.rb  put_object
            └─ aws-sdk-core-3.226.3  seahorse/client/request.rb  send_requestlinked
```

jterapin: Hi! This pr is for 'upload_stream' so at glance, it appears to be unrelated but I would suggest submitting an issue to us so that we can take a closer look. Also please include a stack trace if possible.

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_stream_uploader.rb
richardwang1124: `Tempfile.new` automatically appends date, pid, and some random characters to the end so it should be unique already.

gems/aws-sdk-s3/lib/aws-sdk-s3/customizations/object.rb
```diff
@@ -405,7 +407,10 @@ def public_url(options = {})
       # @see Client#upload_part
       def upload_stream(options = {}, &block)
         upload_opts = options.merge(bucket: bucket_name, key: key)
-        executor = DefaultExecutor.new(max_threads: upload_opts.delete(:thread_count))
+        thread_count = upload_opts.delete(:thread_count) || DefaultExecutor::DEFAULT_MAX_THREADS
```
jterapin, line 410: Could we expose the queue size as its own option instead of deriving it from thread_count? Right now max_queue: thread_count couples two separate concerns: how many uploads run in parallel, and how far ahead we're allowed to buffer. 

Or this could be an additional thing to have down the road. Your choice.

jterapin: Could we expose the queue size as its own option instead of deriving it from thread_count? Right now max_queue: thread_count couples two separate concerns: how many uploads run in parallel, and how far ahead we're allowed to buffer.

gems/aws-sdk-s3/lib/aws-sdk-s3/default_executor.rb
```diff
@@ -25,10 +26,15 @@ def post(*args, &block)
...
         true
+      rescue ClosedQueueError
+        # shutdown or kill happened while parked on a full queue
+        raise 'Executor has been shutdown and is no longer accepting tasks'
```
jterapin, line 37: We should define an actual error class for this. Something like, `RejectedExecutionError` which is what concurrent-ruby has.

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
```diff
@@ -149,21 +149,32 @@ def upload_with_executor(pending, completed, options)
...
+              abort_upload = true
+              errors << e
+            ensure
+              p[:body].close
+              Thread.current[:net_http_override_body_stream_chunk] = nil if @http_chunk_size
+              completion_queue << :done
+            end
+            # Count only successfully queued parts; a failed post never runs the
+            # block, so it never pushes :done and must not be waited on below.
+            upload_attempts += 1
```
jterapin, line 170: Could rename this var into `queued_parts` to be consistent and accurate

gems/aws-sdk-s3/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Bound memory usage in `upload_stream` when the source produces data faster than parts can be uploaded.
```
jterapin, line 4:
```suggestion
* Issue - Fix unbounded memory growth in `upload_stream` when the data source outpaces the upload (#3407).
```

gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb
```diff
@@ -149,21 +149,32 @@ def upload_with_executor(pending, completed, options)
...
 
-          upload_attempts += 1
-          @executor.post(part) do |p|
-            Thread.current[:net_http_override_body_stream_chunk] = @http_chunk_size if @http_chunk_size
-            update_progress(progress, p)
-            resp = @client.upload_part(p)
-            completed_part = { etag: resp.etag, part_number: p[:part_number] }
-            apply_part_checksum(resp, completed_part)
-            completed.push(completed_part)
+          begin
```
jterapin, line 152: I believe we will have to apply similar changes to FileDownloader but you could this in a fast-follow PR.

gems/aws-sdk-s3/lib/aws-sdk-s3/default_executor.rb
```diff
@@ -91,8 +100,6 @@ def ensure_worker_available
       def spawn_worker
         Thread.new do
           while (job = @queue.shift)
-            break if job == :shutdown
-
             args, block = job
             block.call(*args)
```
jterapin, line 104: Now that the queue is bounded, I think this needs a guard around block.call. If a task raises outside StandardError (OOM, a signal), the worker dies here, and with a full queue the producer stays parked on push with nothing to drain it and no way to spawn a replacement, so the upload freezes silently. 

Any thoughts on how to counter this?

